PR #2, opened 2021-09-02
CXXCBC-5 Handle no meter in http requests

davidkelly:
Motivation
==========
The diagnostics ping test is segfaulting.

Modification
============
Simply guarded against calling into it if null.   We can follow up later
and decide how to deal with it generally (like, use a noop_meter).

While there, fixed up some issues with building this inside another
project - CMAKE_SOURCE_DIR is the _outer_ source directory, so migrated
to PROJECT_SOURCE_DIR throughout.

Also a small issue with the test helpers.  Now, we can pass in the server version.  However,
I believe we may have an issue specifying DP, will follow up later on that.

Results
=======
passing tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ endif()
 
 include(cmake/VersionInfo.cmake)
 
-include_directories(${CMAKE_SOURCE_DIR}/couchbase)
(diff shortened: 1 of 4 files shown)


raycardillo (approved):
LGTM

PR #3, opened 2021-09-13
[CXXCBC-14] Update build instructions in preparation for test refactoring

raycardillo:
https://issues.couchbase.com/browse/CXXCBC-14

In preparation for more work, making this PR to improve build instructions:
- Updated CMake version reference
- Added more instructions/clarifications to the README


davidkelly (approved):
LGTM.  The bump in the cmake version seems ok - could be an issue with getting builds on jenkins but we should be able to update.  Can always bump down if necessary.

PR #5, opened 2021-09-15
Fix race in HTTP session

avsej:


couchbase/io/http_session.hxx
```diff
@@ -395,8 +398,16 @@ class http_session : public std::enable_shared_from_this<http_session>
                 spdlog::error("{} IO error while writing to the socket: {}", self->log_prefix_, ec.message());
                 return self->stop();
             }
-            self->writing_buffer_.clear();
-            if (!self->output_buffer_.empty()) {
+            {
+                std::scoped_lock inner_lock(self->writing_buffer_mutex_);
+                self->writing_buffer_.clear();
+            }
+            bool want_write = false;
```
davidkelly, line 405: seems like this is always false, and only used on line 410 below, so maybe can be eliminated and no if statement necessary.

davidkelly (changes requested):

couchbase/io/http_session.hxx
avsej: Yes, it makes sense although in this case clang-tidy complains about uninitialized variable.

davidkelly (approved):
LGTM

PR #27, opened 2021-11-03
fix build with openssl 1.0

avsej:


jrawsthorne (approved):
Built successfully with openssl 1.0

PR #35, opened 2021-11-06
represent CAS as type instead of std::uint64_t

avsej:


avsej:
Makes sense. I agree with it and will move it on the same level as document_id

avsej:
Oh, one thing for `protocol::cas` is that with top-level cas you cannot type something like

```

brett19 (dismissed):

brett19 (approved):

PR #36, opened 2021-11-08
CXXCBC-25: keep scope and collection separate in document_id

avsej:


brett19 (changes requested):
Why did you change document_id to use accessors rather than direct access with accessor only for things that are actually calculated?

avsej:
I decided to do it because I want to control construction, and hide fields from direct modification into "private" section. Also it helped during refactoring :) as compiler complained more.

brett19:
So document_id is immutable.  I see :)

brett19 (approved):

brett19 (approved):

brett19 (approved):

brett19 (approved):

PR #45, opened 2021-11-12
fix move semantics for handlers

avsej:
in particular, allow to use lambdas that capture move-only objects as
operation handlers


davidkelly (approved):
LGTM

PR #50, opened 2021-11-15
SDKQE-2487: Port N1QL query tests from lcb

jrawsthorne:


couchbase/operations/management/collections_manifest_get.hxx
```diff
@@ -37,6 +37,7 @@ struct collections_manifest_get_response {
 struct collections_manifest_get_request {
```
jrawsthorne, line 37: Should this be under management, it's an mcbp operation

avsej: yes, because the nature of the operation is rather related to management area.

avsej (approved):

PR #52, opened 2021-11-18
use mutex lock when using buckets map

avsej:


jrawsthorne (dismissed):
Also ping in cluster.cxx

avsej:
Thanks @jrawsthorne. Fixed now

brett19 (changes requested):
It appears that this works, but also means keeping the buckets lock for excessive periods of time (specifically, through the handler calls, or during operation execution).  I think this will unnecessarily impact performance, as we should only need to hold the lock when looking through the buckets list.

davidkelly (commented):
Agree with @brett19 - an iterator into a map remains valid as the map changes, unless the thing it points to is deleted elsewhere, I believe.  Locking around the find should be sufficient.

brett19:
> Agree with @brett19 - an iterator into a map remains valid as the map changes, unless the thing it points to is deleted elsewhere, I believe. Locking around the find should be sufficient.

The bucket object itself is a shared_ptr, so once we've pulled our own copy of that pointer, we shouldn't even need the iterator anymore.

avsej:
@davidkelly @brett19, so I extracted mutex-related code into `find_bucket_by_name` and `for_each_bucket`. The latter copies all pointers into temporary vector and then iterates that vector invoking the callbacks.

couchbase/cluster.hxx
```diff
@@ -99,14 +97,16 @@ class cluster
...
+        if (b) {
             return handler({});
         }
+
         std::vector<protocol::hello_feature> known_features;
         if (session_ && session_->has_config()) {
             known_features = session_->supported_features();
         }
-        auto b = std::make_shared<bucket>(id_, ctx_, tls_, tracer_, meter_, bucket_name, origin_, known_features);
+        b = std::make_shared<bucket>(id_, ctx_, tls_, tracer_, meter_, bucket_name, origin_, known_features);
```
brett19, line 109: b is being emplaced into buckets_ without a lock being held.

couchbase/cluster.hxx
```diff
@@ -119,9 +119,9 @@ class cluster
     template<typename Handler>
     void close_bucket(const std::string& bucket_name, Handler&& handler)
     {
-        auto ptr = buckets_.find(bucket_name);
-        if (ptr != buckets_.end()) {
-            ptr->second->close();
+        auto bucket = find_bucket_by_name(bucket_name);
+        if (bucket) {
+            bucket->close();
```
brett19, line 124: bucket->close only shuts down the bucket, but doesn't remove it from the buckets_ list.  This means that the actual object pointer would leak.

brett19 (changes requested):
There are a couple of comments in the review.

couchbase/cluster.hxx
avsej: right, thanks for noticing it

couchbase/cluster.hxx
avsej: thanks

couchbase/cluster.hxx
```diff
@@ -99,29 +97,34 @@ class cluster
     template<typename Handler>
     void open_bucket(const std::string& bucket_name, Handler&& handler)
     {
-        if (buckets_.find(bucket_name) != buckets_.end()) {
+        auto b = find_bucket_by_name(bucket_name);
```
brett19, line 100: We need to keep the lock between finding the bucket and adding it to the list.  Specifically the order that's probably ideal is to lock the lock, search for the bucket, if its not there, we should create a new bucket instance, then add it to the map, then unlock and finally bootstrap the connection. Otherwise there is a race where two of the same bucket could end up being bootstrapped.

couchbase/cluster.hxx
```diff
@@ -99,29 +97,34 @@ class cluster
...
         buckets_.emplace(bucket_name, b);
     }
 
     template<typename Handler>
     void close_bucket(const std::string& bucket_name, Handler&& handler)
     {
-        auto ptr = buckets_.find(bucket_name);
-        if (ptr != buckets_.end()) {
-            ptr->second->close();
+        auto bucket = find_bucket_by_name(bucket_name);
```
brett19, line 123: Similarly to the comment above, we need to acquire the lock, find the bucket, then remove it from the list and then unlock the lock, then finally close the bucket instance.  Otherwise there is a chance that the bucket gets erased and closed twice.

brett19 (changes requested):

couchbase/cluster.hxx
avsej: New version addresses this issue.

brett19 (approved):

PR #53, opened 2021-11-18
ensure MCBP session closure when topology changes

avsej:


davidkelly (approved):
LGTM

brett19:
I wonder if we should modify certain things like mcbp_session to be non-copyable?

couchbase/bucket.hxx
```diff
@@ -213,7 +214,7 @@ class bucket : public std::enable_shared_from_this<bucket>
         } else {
             new_session = std::make_shared<io::mcbp_session>(client_id_, ctx_, origin_, name_, known_features_);
         }
-        new_session->bootstrap([self = shared_from_this(), new_session, h = std::forward<Handler>(handler)](
+        new_session->bootstrap([self = shared_from_this(), new_session = std::move(new_session), h = std::forward<Handler>(handler)](
```
avsej, line 217: yes, it is not necessary here, but above it is the last usage of the pointer, so we can safely move it into the closure

avsej:
@brett19 I've updated the patch, and removed questionable `std::move`. Now the only move left in the place, where we know for sure the pointer is not used anymore

brett19 (approved):

PR #62, opened 2021-11-24
fix snappy compression of the value

avsej:
When the value was compressed, the body length and datatype fields were
not updated because of non-reference argument. The patch improves API a
bit, and modification of the packet done by the caller now.


davidkelly (approved):
LGTM

PR #66, opened 2021-11-29
tests: try to detect cluster version using REST API

avsej:
fallback to version from environment only if the cluster does not agree
on version or returns invalid version string

### DIFF
--- a/couchbase/operations/management/cluster_describe.hxx
+++ b/couchbase/operations/management/cluster_describe.hxx
@@ -1,0 +1,68 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2021 Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


couchbase/operations/management/cluster_describe.hxx
```diff
@@ -0,0 +1,70 @@
...
+    cluster_info info{};
+};
+
+struct cluster_describe_request {
+    using response_type = cluster_describe_response;
+    using encoded_request_type = io::http_request;
+    using encoded_response_type = io::http_response;
+    using error_context_type = error_context::http;
+
+    std::string name;
```
jrawsthorne, line 60: Should name be here?

avsej: thanks. copy-paste issue.

test/utils/server_version.hxx
```diff
@@ -45,38 +45,44 @@ struct server_version {
 
     [[nodiscard]] bool is_cheshire_cat() const
     {
-        // [7.0.0, inf)
-        return major >= 7;
+        // [7.0.0, 7.1.0)
+        return major >= 7 && minor < 1;
```
jrawsthorne, line 49: == 7 ?

avsej: let it me >= until we get next version defined

avsej: ah, right. sorry

jrawsthorne (approved):

PR #68, opened 2021-11-30
CXXCBC-29: add storage option for bucket manager

avsej:


couchbase/operations/management/bucket_settings.hxx
```diff
@@ -81,7 +82,20 @@ struct bucket_settings {
     bool flush_enabled{ false };
     eviction_policy eviction_policy{ eviction_policy::unknown };
     conflict_resolution_type conflict_resolution_type{ conflict_resolution_type::unknown };
+
+    /**
+     * UNCOMMITTED: This API may change in the future
+     */
+    storage_backend_type storage_backend{ storage_backend_type::unknown };
```
avsej, line 89: This patch only exposes storage type option, because @chvck is going to propose to remove magma-specific options and expose "raw option" API instead.

brett19 (approved):

PR #72, opened 2021-12-01
Fix live cluster version parse

jrawsthorne:
Cluster version appends -edition. Account for this and parse it so we can test community edition
Fix neo version to apply to all future versions

### DIFF
--- a/test/utils/server_version.hxx
+++ b/test/utils/server_version.hxx
@@ -22,12 +22,16 @@
 
 namespace test::utils
 {
+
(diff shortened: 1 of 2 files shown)


test/utils/server_version.hxx
```diff
@@ -22,12 +22,16 @@
 
 namespace test::utils
 {
+
+enum class server_edition { enterprise, community, unknown };
```
avsej, line 26: could you make `unknown` first? so that it will be default for uninitialized cases

PR #79, opened 2021-12-15
Updated compiler warnings to not apply outside client and fixed a few minor issues.

brett19:


raycardillo (approved):
Approving per discussion since @avsej and @davidkelly already reviewed.

PR #88, opened 2021-12-17
Skip tests that don't work in certain configurations

jrawsthorne:
Analytics is not supported with magma buckets
Analytics external link management is not supported with cert auth on versions < 7.1


avsej:
jenkins, test this please

avsej (approved):

PR #89, opened 2021-12-18
fix thread_sanitizer offences

avsej:


brett19 (changes requested):
I believe this implementation still secretly hides a bunch of race conditions.  The need to wrap `async_write` in a lock implies that it is being called from multiple places concurrently, the problem being that boost::async_write internally calls async_write_some potentially multiple times.  This can lead to network data being written interleaved.  I believe the proper solution to this problem is to use strands to ensure that writes are executed in an orderly fashion.  I suspect that most of the places here where a lock was added actually needs to use stranding instead to fully solve the problem.

avsej:
@brett19 we actually use stands in the current stream implementation to avoid interleaved requests. I've added locks here to protect socket as memory objects, because, for instance, the connection sequence might be running in one thread, while another thread would like to initiate "stop" sequence, which would need to invoke "close()" method on the stream and the socket.

avsej:
jenkins, test this please

avsej:
After a little more pondering on this problem, I think you right. While we use strands for the streams (asio sockets) themselves, which helps to order all the data operations, occasionally we try directly signal state changes from other threads. In this particular case, the "close" event is external, and what we should do is not to cancel/stop the asio objects from the initiating thread, but rather post task to be executed on the same executor as the strand is using. I will change this patch to make it clear.

bin/init-cluster
```diff
@@ -30,7 +30,7 @@ options = {
   username: ENV.fetch("CB_USERNAME", "Administrator"),
   password: ENV.fetch("CB_PASSWORD", "password"),
   bucket: ENV.fetch("CB_BUCKET", "default"),
-  server_quota: ENV.fetch("CB_SERVER_QUOTA", 512).to_i,
+  server_quota: ENV.fetch("CB_SERVER_QUOTA", 1024).to_i,
```
brett19, line 33: This doesn't seem related to thread_sanitizer

couchbase/io/streams.hxx
```diff
@@ -72,71 +76,83 @@ class stream_impl
...
     void reopen() override
     {
-        id_ = uuid::to_string(uuid::random());
-        asio::error_code ec{};
-        stream_.shutdown(asio::socket_base::shutdown_both, ec);
-        stream_.close(ec);
-        stream_ = asio::ip::tcp::socket(strand_);
+        return close([this](std::error_code) {
+            id_ = uuid::to_string(uuid::random());
+            stream_ = std::make_shared<asio::ip::tcp::socket>(strand_);
```
brett19, line 119: Should this not set open_ back to true?

couchbase/io/streams.hxx
```diff
@@ -72,71 +76,83 @@ class stream_impl
...
             return {};
         }
         return res;
     }
 
-    void close() override
+    void close(std::function<void(std::error_code)>&& handler) override
     {
-        stream_.close();
+        if (!open_ || !stream_) {
```
brett19, line 103: I think this use of open_ and stream_ is not thread safe since it's not occurring in the strand.

brett19 (changes requested):

couchbase/io/streams.hxx
avsej: it will set to open, when it will open the socket. here it is just constructor

bin/init-cluster
avsej: I will eject this into separate PR

couchbase/io/streams.hxx
avsej: open_ is atomic_bool, so it should be safe

brett19: If that's the case, then stream_ needs to be std::atomic<std::shared_ptr> in that case, and it's important that both open_ and stream_ are always checked together, since they are independently atomic.

brett19 (changes requested):

avsej:
test this please

couchbase/io/streams.hxx
avsej: I removed this check, as `stream_` cannot be null and close could be called safely several times.

avsej:
test this please

avsej:
@davidkelly this is the different one. Wasn't it addressed in https://github.com/couchbaselabs/couchbase-cxx-client/commit/e95ec78cc8e3d2e2c74eda182c324d8b0ac48fb0?

davidkelly:
@avsej that one fixed tracing's issue, but this is a metrics issue.  Looks like it just needs a mutex around the map of service_recorders.   I'll tweak this slightly and try it out, see if that's all.

davidkelly:
@avsej - That did it.  Just added a mutex to logging_meter, and used it when finding the service recorder.  Can't repro that issue now, and when I run I see no thread sanitizer issues.   The test has txn collisions for like 5 minutes when using cbdyncluster, so seems ok.   I didn't really look hard at whether or not there's still a race when actually recording a value -- it appears we use `hdr_record_value`, and there is also a `hdr_record_value_atomic` we could use.   I'd guess it is possible we are ok, but perhaps we would be better off using the atomic instead.

Adding the mutex when finding the recorder:
```
diff --git a/couchbase/metrics/logging_meter.cxx b/couchbase/metrics/logging_meter.cxx
index 1981704..4be8f62 100644
--- a/couchbase/metrics/logging_meter.cxx
+++ b/couchbase/metrics/logging_meter.cxx
@@ -171,7 +171,6 @@ logging_meter::get_value_recorder(const std::string& name, const std::map<std::s
     if (service == tags.end()) {
         return &noop_recorder;
     }
-    auto& service_recorders = recorders_[service->second];

     static std::string operation_tag = "db.operation";
     auto operation = tags.find(operation_tag);
@@ -179,6 +178,9 @@ logging_meter::get_value_recorder(const std::string& name, const std::map<std::s
         return &noop_recorder;
     }

+    std::lock_guard<std::mutex> lock(mutex_);
+    auto& service_recorders = recorders_[service->second];
+
     auto recorder = service_recorders.find(operation->second);
     if (recorder != service_recorders.end()) {
         return recorder->second.get();
diff --git a/couchbase/metrics/logging_meter.hxx b/couchbase/metrics/logging_meter.hxx
index b01c80d..eabeb6a 100644
--- a/couchbase/metrics/logging_meter.hxx
+++ b/couchbase/metrics/logging_meter.hxx
@@ -19,7 +19,7 @@

 #include <couchbase/metrics/logging_meter_options.hxx>
 #include <couchbase/metrics/meter.hxx>
-
+#include <thread>
 #include <asio/steady_timer.hpp>

 namespace couchbase::metrics
@@ -31,6 +31,7 @@ class logging_meter : public meter
   private:
     asio::steady_timer emit_report_;
     logging_meter_options options_;
+    std::mutex mutex_;
     std::map<std::string, std::map<std::string, std::shared_ptr<logging_value_recorder>>> recorders_{};

     void log_report() const;
```

davidkelly (changes requested):
Lets add a mutex to `logging_meter`, and use it in `logging_meter::get_value_recorder` to protect access to the `_recorders` map.   I _think_ it is worth it to then use `hdr_record_value_atomic` in `logging_value_recorder::record_value` as well.

avsej:
Jenkins, test this please

avsej:
test this please, jenkins

couchbase/io/http_session.hxx
```diff
@@ -121,8 +121,8 @@ class http_session : public std::enable_shared_from_this<http_session>
       , ctx_(ctx)
       , resolver_(ctx_)
       , stream_(std::make_unique<plain_stream_impl>(ctx_))
```
brett19, line 123: I think it would be better if the strand was owned by the http_session rather than the socket itself (and then having it shared outwards).  Though I don't think it's necessarily important enough to change right away.

couchbase/io/streams.hxx
```diff
@@ -72,71 +81,77 @@ class stream_impl
...
     void reopen() override
     {
-        id_ = uuid::to_string(uuid::random());
-        asio::error_code ec{};
-        stream_.shutdown(asio::socket_base::shutdown_both, ec);
-        stream_.close(ec);
-        stream_ = asio::ip::tcp::socket(strand_);
+        return close([this](std::error_code) {
+            id_ = uuid::to_string(uuid::random());
+            stream_ = std::make_shared<asio::ip::tcp::socket>(strand_);
```
brett19, line 118: There is likely still race conditions here as while the close handler might execute on the strand, there are a bunch of places where the the fields of the stream (like id_ and stream_) are being accessed off the strand.  The shared_ptr holding the stream is not safely accessible from off the strand since it can change on-strand.  id_ is similar.  That being said, again, I think it's worthwhile to merge this and address these issues later.

couchbase/metrics/logging_meter.cxx
```diff
@@ -171,14 +171,16 @@ logging_meter::get_value_recorder(const std::string& name, const std::map<std::s
...
     static std::string operation_tag = "db.operation";
     auto operation = tags.find(operation_tag);
     if (operation == tags.end()) {
         return &noop_recorder;
     }
 
+    std::scoped_lock lock(recorders_mutex_);
+    auto& service_recorders = recorders_[service->second];
+
     auto recorder = service_recorders.find(operation->second);
```
brett19, line 184: Looks reasonable, although I noticed below this section there is a bit of weirdness in the implementation which will likely have a performance impact.  Will file an issue for it.

brett19 (approved):

couchbase/io/http_session.hxx
avsej: https://issues.couchbase.com/browse/CXXCBC-73

couchbase/io/streams.hxx
avsej: https://issues.couchbase.com/browse/CXXCBC-74

couchbase/metrics/logging_meter.cxx
brett19: https://issues.couchbase.com/browse/CXXCBC-75

brett19 (approved):

couchbase/metrics/logging_meter.cxx
avsej: https://issues.couchbase.com/browse/CXXCBC-76

PR #90, opened 2021-12-20
CXXCBC-44: add ability to perform freeform requests to HTTP services

avsej:


couchbase/operations/management/freeform.hxx
```diff
@@ -0,0 +1,54 @@
...
+
+struct freeform_request {
+    using response_type = freeform_response;
+    using encoded_request_type = io::http_request;
+    using encoded_response_type = io::http_response;
+    using error_context_type = error_context::http;
+
+    service_type type;
+    std::string method;
+    std::string path;
```
jrawsthorne, line 43: Method could be an enum. Same applies to the whole SDK

jrawsthorne (dismissed):
Looks good to me once tests are fixed. Had one comment

couchbase/operations/management/freeform.hxx
avsej: Yes, it could be, but on the other hand, this is freeform request, and HTTP protocol does not limit number of HTTP methods.

jrawsthorne: That's true

jrawsthorne (approved):

davidkelly (approved):

PR #93, opened 2021-12-23
CXXCBC-62: handle rate/quota limit errors during MCBP bootstrap

avsej:


jrawsthorne (changes requested):
In my test I'm getting the rate limited error on the error map response because it is sent after auth. This is a separate issue as I think it should be pipelined after hello. However, after fixing that I then get rate limited on the get cluster config because it's a bucketless bootstrap so no select bucket. The same issue exists in LCB.

avsej:
@jrawsthorne, I moved the status check higher, so it will be applied to all bootstrap responses.

jrawsthorne (approved):

PR #105, opened 2022-01-12
Identify race when message added to pending buffer

jrawsthorne:


avsej:
The fix of this issue is included in #89

PR #116, opened 2022-01-25
CXXCBC-58: do not use nodes ejected during rebalance

avsej:


brett19 (approved):

jrawsthorne (changes requested):
See my comment on the ticket, still seeing errors after rebalance with this patch

brett19:
@jrawsthorne To be fair, I think this change makes sense generally speaking, even if we don't yet have a fix for the issue you were seeing.

PR #123, opened 2022-01-27
Fix metrics for analytics and query results

avsej:
depends on #122


couchbase/operations/document_query.cxx
```diff
@@ -212,10 +212,10 @@ query_request::make_response(error_context::query&& ctx, const encoded_response_
             meta_metrics.result_size = m->at("resultSize").get_unsigned();
             meta_metrics.elapsed_time = utils::parse_duration(m->at("elapsedTime").get_string());
             meta_metrics.execution_time = utils::parse_duration(m->at("executionTime").get_string());
-            meta_metrics.sort_count = m->template optional<std::uint64_t>("sortCount");
-            meta_metrics.mutation_count = m->template optional<std::uint64_t>("mutationCount");
-            meta_metrics.error_count = m->template optional<std::uint64_t>("errorCount");
-            meta_metrics.warning_count = m->template optional<std::uint64_t>("warningCount");
+            meta_metrics.sort_count = m->template optional<std::uint64_t>("sortCount").value_or(0);
```
jrawsthorne, line 215: Is there somewhere that defines which of these are optional? It's not in the RFC

avsej: Right, for analytics these are not optional, but query still writes them optionally (https://github.com/couchbase/query/blob/f0109b241dd084d8e52ad4f53bcdba8a339492c0/server/http/service_response.go#L510-L520)

avsej:
test this please

avsej:
test this please

couchbase/operations/document_analytics.cxx
```diff
@@ -173,38 +163,36 @@ analytics_request::make_response(error_context::analytics&& ctx, const encoded_r
...
-                            break;
-                        case 24006: /* Link [string] does not exist | Link [string] does not exist */
-                            link_not_found = true;
-                            break;
-                        default:
-                            if (error.code >= 24000 && error.code < 25000) {
-                                compilation_failure = true;
-                            }
-                    }
+            for (const auto& error : response.meta.errors) {
```
brett19, line 166: In the previous PR, you modify error handling to use the first_error_code and what not for error handling, but in this PR you loop over them.  I don't think a loop makes sense here since we can only return a single error code.

brett19 (changes requested):

couchbase/operations/document_analytics.cxx
avsej: it is not previous PR, it is subsequent, which will come after this one

brett19 (approved):

PR #135, opened 2022-02-07
CXXCBC-123: substitute $HOST in the configuration

avsej:


brett19 (changes requested):
I'm not sure it actually matters, but the actual server spec says that you must string-replace $HOST before doing JSON parsing (since it can show up in compound strings).

avsej:
I think it does not matter in this particular case, `ns_server` uses it only for nodes:
https://github.com/couchbase/ns_server/blob/e05110041c31bd990ed698cf6f0af78f4967eefe/src/bucket_info_cache.erl#L224

brett19:
I was mainly just concerned as later changes within their spec could break our code, but honestly, context-less arbitrary search replace on JSON data is silly anyways.  Fair enough.

brett19 (approved):

PR #144, opened 2022-02-09
CXXCBC-128: fix build on Windows

brett19:


avsej:
I fixed build script (restored old bash script, and renamed portable version to `.rb`). Your fixes pulled into #143

PR #147, opened 2022-02-11
use shared pointers for metrics collector

avsej:


couchbase/cluster.hxx
```diff
@@ -98,7 +98,7 @@ class cluster
             handler();
             work_.reset();
             tracer_.reset();
-            delete meter_;
+            meter_.reset();
             meter_ = nullptr;
```
jrawsthorne, line 102: Don't need the meter_ = nullptr;

jrawsthorne (approved):

PR #159, opened 2022-02-17
Allow to statically link MSVC runtime

avsej:


brett19 (approved):

brett19 (changes requested):
Need to use set property with TARGET to apply it only to project_options.

avsej:
I will leave this open until I will update ruby SDK to recent version of C++ SDK. Or until somebody else will fix it for us.

PR #161, opened 2022-02-17
CXXCBC-138: protect HTTP round-robin index with mutex

avsej:


couchbase/io/http_session_manager.hxx
```diff
@@ -64,6 +64,7 @@ class http_session_manager : public std::enable_shared_from_this<http_session_ma
     void set_configuration(const topology::configuration& config, const cluster_options& options)
     {
         options_ = options;
+        std::scoped_lock index_lock(next_index_mutex_);
```
davidkelly, line 67: Makes sense.   Only worry I have is nested mutexes - if the config_mutex_ is ever locked, _then_ the next_index_mutex_ is locked, we could get a deadlock.  As long as that never happens (like, maybe this is the only place we take both?), then cool.

davidkelly (approved):
My comment was just in case.   I'm sure you already thought of it -- if so LGTM.

couchbase/io/http_session_manager.hxx
avsej: I've updated pr, to use `std::scoped_lock` on both mutexes

davidkelly (approved):

PR #165, opened 2022-02-22
CXXCBC-124: remove status code validity assertion from protocol parser

avsej:
Unknown status codes will be handled using error map


jrawsthorne (approved):
Caves tests now pass

PR #169, opened 2022-02-22
Split management tests into separate files

jrawsthorne:


avsej:
I'm closing this PR at the moment as the change has been outdated, but it is a good idea, and we will get back to it later.

PR #170, opened 2022-02-28
SDKQE-2596: Search query testing for CB++

jrawsthorne:


avsej:
moved to #380 and #379. The search testing will be part of Public API pull request

PR #172, opened 2022-03-18
CXXCBC-95: use address from configuration for served_by_node in query

avsej:


couchbase/error_context/analytics.hxx
```diff
@@ -39,6 +39,8 @@ struct analytics {
     std::string path{};
     std::uint32_t http_status{};
     std::string http_body{};
+    std::string hostname{};
```
brett19, line 42: It seems like it would be better to have the hostname/port as a single string, as it is specified in the last_dispatched_to and served_by_node fields since error_context's are also user-facing.

brett19 (approved):

PR #176, opened 2022-04-07
CXXCBC-69: Add support for collections to index management

chvck:


couchbase/operations/management/query_index_build_deferred.cxx
```diff
@@ -26,14 +26,24 @@ namespace couchbase::operations::management
 std::error_code
 query_index_build_deferred_request::encode_to(encoded_request_type& encoded, http_context& /* context */) const
 {
+    std::string statement;
+    if (!scope_name.empty() && !collection_name.empty()) {
```
chvck, line 30: I suspect some of these should check if only one of these is set and error if so. However I don't have a clue how to do that...

avsej (approved):

PR #177, opened 2022-04-07
Add support for raw to view request

chvck:


couchbase/operations/document_view.cxx
```diff
@@ -95,6 +95,9 @@ document_view_request::encode_to(document_view_request::encoded_request_type& en
         }
         body["keys"] = keys_array;
     }
+    for (const auto& [name, value] : raw) {
+        body[name] = value;
+    }
```
avsej, line 100: Should be some equivalent for query string? I don't think that `ns_server` allows anything except `"keys"` in the body

PR #178, opened 2022-04-07
Add on error option to view request

chvck:


couchbase/operations/document_view.hxx
```diff
@@ -86,6 +86,8 @@ struct document_view_request {
     std::optional<std::uint32_t> group_level;
     bool debug{ false };
 
+    enum class on_error { _continue, stop };
```
chvck, line 89: Not really sure what the idiom would be for a enum option that is also a keyword.

avsej: I would rather use synonym here something like `keep_up`, `proceed`, `resume` or even `ignore`.

chvck: I agree that that would be nicer but it would mean that it isn't consistent with other SDKs and the RFC. It might just have to be a place where cxx SDK has to differ slightly though.

avsej: yes, and also Views API is going to be removed in the future :)

avsej (approved):

PR #179, opened 2022-04-07
CXXCBC-153: Added automatic inclusion of the Capella Root CA.

brett19:


chvck:
Have tested this with the PHP client and it looks good to me. An extra test from python could be good though, I'm making this up as I go along...

avsej (changes requested):

avsej (approved):

PR #189, opened 2022-05-11
Repro for CXXCBC-179

jrawsthorne:


avsej:
Fixed in #191. This testcase included there.

PR #190, opened 2022-05-12
use matrix strategy in workflows where it is possible

avsej:
Also updated 6.6.4 to 6.6.5 and added 7.1.0


jrawsthorne:
Looks good. For the failing magma test you can increase the ram quota to 1024 or reduce the minimum server side with a request to /internalSettings with magmaMinMemoryQuota=256.

jrawsthorne (approved):

PR #195, opened 2022-05-27
CXXCBC-173: complete streaming lexer even if pointer didn't match

avsej:


jrawsthorne:
Looks good except for the failing test on 6.6.5

PR #211, opened 2022-06-23
CXXCBC-94, CXXCBC-161: update ping implementation

avsej:


davidkelly (approved):
Nice. I like that it opens the bucket if it is specified.

PR #216, opened 2022-06-28
CXXCBC-156: use vector of bytes instead of string for binary data

avsej:


couchbase/document_id.cxx
```diff
@@ -86,4 +89,18 @@ document_id::has_default_collection() const
 {
     return !use_collections_ || collection_path_ == "_default._default";
 }
+
+std::vector<std::byte>
+make_protocol_key(const document_id& id)
+{
+    std::vector<std::byte> key{};
+    if (id.is_collection_resolved()) {
```
brett19, line 97: Probably a bit of an ancillary point, but the naming of this (is_collection_resolved) implies that this might incorrectly encode a document_id which is intended to have a collection_id, but which was not properly resolved yet?

couchbase/utils/json.cxx
```diff
@@ -68,4 +68,214 @@ generate(const tao::json::value& object)
 {
     return tao::json::to_string(object);
 }
+
+class to_byte_vector
```
brett19, line 72: If I understand what this is doing, I would argue that it is unnecessary.  The data that is output from tao is actually a UTF8 encoded string which is perfect for std::string, we should instead simply convert the std::string to the std::vector<std::byte> when its being included into a buffer when that needs to happen rather than having custom output logic to try and output directly to a vector of bytes.

brett19 (changes requested):

couchbase/document_id.cxx
avsej: maybe it is bad naming, but it is there to cover the legacy case where we don't need to inject collection id

couchbase/utils/json.cxx
avsej: it is still possible to do, but that would involve `reinterpret_cast`, right? this one is just an option to avoid casting and copying by writing directly as `std::byte`

brett19 (approved):

PR #219, opened 2022-07-12
Increase resolution of threshold tracer up to microseconds

avsej:
reporter is using microseconds to report outliers, but we store duration
with millisecond resolution, which basicaly zeroes last three digits


davidkelly (approved):
LGTM 👍

PR #220, opened 2022-07-12
generate documentation using doxygen

avsej:


davidkelly (commented):
LGTM 👍 

I like the split where there are 'public' documented things in the /api, and the internals elsewhere.

davidkelly (approved):

PR #228, opened 2022-07-19
CXXCBC-55: External Tracing and Metrics support

davidkelly:
Motivation
==========
Allow clients to plug in their own tracing and/or metrics solutions,
such as OpenTelemetry

Modifications
=============
Expose the tracer and meter in cluster_options.  Then, add a parent_span
to many (but for now, not all) requests.   The tests use a simple tracer
and meter implementation to facilitate testing.

*** Will update this with a header that wraps OpenTelemetry objects in
particular, which will not be referenced by anything in the client so we
will only require otel when it is used.

Results
=======
Tests pass

### DIFF
--- a/couchbase/cluster.hxx
+++ b/couchbase/cluster.hxx
@@ -70,15 +70,26 @@ class cluster : public std::enable_shared_from_this<cluster>
         }
 
         origin_ = origin;
-        if (origin_.options().enable_tracing) {
-            tracer_ = std::make_shared<tracing::threshold_logging_tracer>(ctx_, origin.options().tracing_options);
+        // ignore the enable_tracing flag if a tracer was passed in
+        if (nullptr != origin_.options().tracer) {
+            tracer_ = origin_.options().tracer;
         } else {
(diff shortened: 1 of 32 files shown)


avsej:
@davidkelly looks like Jenkins failures are legit and related to tracer and meter:

http://sdk.jenkins.couchbase.com/job/c-cpp/job/cxx/job/cxx-scripted-build-pipeline/1197/

```
/home/couchbase/jenkins/workspace/c-cpp/cxx/cxx-scripted-build-pipeline/ws_centos7/couchbase-cxx-client/test/test_integration_tracer.cxx:189: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```
```
/home/couchbase/jenkins/workspace/c-cpp/cxx/cxx-scripted-build-pipeline/ws_centos7/couchbase-cxx-client/test/test_integration_meter.cxx:120: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```

avsej (approved):

PR #229, opened 2022-07-21
fix division by zero issue in KV loader

avsej:


couchbase/io/http_session.hxx
```diff
@@ -250,9 +250,8 @@ class http_session : public std::enable_shared_from_this<http_session>
             }
         }
 
-        if (on_stop_handler_) {
-            on_stop_handler_();
-            on_stop_handler_ = nullptr;
+        if (auto handler = std::move(on_stop_handler_); handler) {
```
chvck, line 253: Not going to pretend that I fully understand this bit.

chvck (approved):

couchbase/io/http_session.hxx
avsej: this is just thing I noticed and decided to fix by the way.

just improve reentrancy, so that `on_stop_handler_` won't be called second time. So first, move it to local variable (and at the same time clear the instance property), and only then invoke handler.

PR #230, opened 2022-07-21
split private/public API and update namespaces

avsej:
`couchbase/` contains Public API
`core/` contains current implementation of the SDK, which right now should be considered "internal" or "implementation details".

Similar logic in C++ namespaces.

everything under `core/` directory is defined in `couchbase::core::*` namespace.
public API from `couchbase/` is using just `couchbase::` namespace.

Next steps will be gradually move operations/structures/functions that we can expose as public API into `couchbase/` directory and document them on the way.

The Transactions C++ library will be merged and live in `transactions/` directory, which public API going into `couchbase/`


avsej:
jenkins, test this please

davidkelly:
@avsej - I noticed that there are things like `query_scan_consistency` which moved into `couchbase::core`.   But -- it seems like this is/will be exposed to the public for use in the options, etc...   Do you plan to pull things out of `couchbase::core` as they are exposed (perhaps via a `query_options`) in the `couchbase::` namespace?

avsej:
these will be exposed on demand when the related API will be exposed. I'm doing it it this way to ensure that only public parts will be visible, and it also simplifying documenting task

davidkelly:
Ok -- that seems reasonable to me.   Anything that doesn't get exposed, stays in the `couchbase::core` namespace like we'd want.   The wrappers will just have to change their `#import ...` and namespace references as it changes.

davidkelly (approved):
LGTM - the incremental approach to exposing things in the `couchbase` namespace, keeping everything else in `couchbase::core` seems reasonable to me.   Every time the wrappers update the client, they will just have to do a bit of editing.

avsej:
Yes, and technically everything still exposed, but `::core::` in namespace and `#include <core/` in headers should be a flag, that the unite is using some of the "core" APIs

PR #233, opened 2022-07-26
CXXCBC-203: disable clustermap nofication by default

avsej:
In order to optimize freeze/thaw cycle clustermap notification should be
disabled. Otherwise the SDK will be required to process all stale
notifications after thaw.


avsej:
jenkins, test this please

davidkelly (approved):
👍

PR #234, opened 2022-07-26
CXXCBC-52: fix read replica behavior for missing keys 

avsej:


avsej:
Now it helps with custom transcoder for result:
Screenshot from 2022-07-27 10-18-20

davidkelly (approved):

thejcfactor:
Thanks for the changes @avsej!  I like the smuggling transcoder 😄

PR #235, opened 2022-07-28
record line, file, and function names into log message

avsej:
the default pattern does not include file location info to reduce
verbosity, but the tests enable it for easier debugging.


davidkelly (approved):
👍 LGTM!

PR #238, opened 2022-08-02
CXXCBC-198: implement build deferred indexes per RFC-54

avsej:


avsej:
Right now it might still give unwanted errors when trying to build scoped indexes, that are pulled by just bucket name. See comment https://github.com/couchbaselabs/sdk-rfcs/pull/112#issuecomment-1203165094

In that case it will return internal error code with the response body like this
```json
{
"requestID": "abb7ef29-5755-4315-b800-8917b3d53d77",
"clientContextID": "14bd2217-bc64-4f0a-5c5e-07e92fc434c1",
"signature": null,
"results": [
],
"errors": [{"code":5411,"msg":"index collections_index_112551505388924 is not found - cause: Index Not Found - cause: GSI index collections_index_112551505388924 not found."}],
"status": "errors",
"metrics": {"elapsedTime": "610.318µs","executionTime": "562.388µs","resultCount": 0,"resultSize": 0,"serviceLoad": 3,"errorCount": 1}
}
```

core/operations/management/query_index_get_all_deferred.cxx
```diff
@@ -0,0 +1,79 @@
...
+#include "core/utils/json.hxx"
+#include "error_utils.hxx"
+
+namespace couchbase::core::operations::management
+{
+std::error_code
+query_index_get_all_deferred_request::encode_to(encoded_request_type& encoded, couchbase::core::http_context& /* context */) const
+{
+    std::string where;
+    if (collection_name.empty() || collection_name == "_default") {
```
chvck, line 29: I'm not sure this is right? collection_name being _default doesn't mean that a scope name hasn't been set. "_default" is also a legitimate collection/scope name to limit indexes to I think?

core/operations/management/query_index_build.cxx
```diff
@@ -15,34 +15,26 @@
...
     if (!scope_name.empty() && !collection_name.empty()) {
         statement = fmt::format(
-          R"(BUILD INDEX ON `{}`.`{}`.`{}` ((SELECT RAW name FROM system:indexes WHERE bucket_id = "{}" AND scope_id = "{}" AND keyspace_id = "{}" AND state = "deferred")))",
-          bucket_name,
-          scope_name,
-          collection_name,
-          bucket_name,
-          scope_name,
-          collection_name);
+          R"(BUILD INDEX ON `{}`.`{}`.`{}` ({}))", bucket_name, scope_name, collection_name, utils::join_strings(index_names, ","));
```
chvck, line 35: We should probably escape each of the index names.

core/operations/management/query_index_get_all_deferred.cxx
avsej: yes, you right.

chvck (approved):

PR #240, opened 2022-08-09
CXXCBC-53: expose get and mutation operations to public API

avsej:
This is the part of CXXCBC-53, it does not include implementation of durability, but refactores several internal structures to expose them in public API section. In particular get() and upsert() operations. Next PR will depend on this one, and add the rest of mutation APIs including durability options.


core/operations/document_get_projected.cxx
```diff
@@ -173,21 +175,22 @@ get_projected_request::make_response(key_value_error_context&& ctx, const encode
     get_projected_response response{ std::move(ctx) };
     if (!response.ctx.ec()) {
         response.cas = encoded.cas();
-        if (with_expiry && !encoded.body().fields()[0].value.empty()) {
-            response.expiry = gsl::narrow_cast<std::uint32_t>(std::stoul(encoded.body().fields()[0].value));
+        response.flags = gsl::narrow_cast<std::uint32_t>(std::stoul(encoded.body().fields()[0].value));
```
thejcfactor, line 178: @avsej - Thanks for adding this update!

thejcfactor:
+1 -- The changes look good to me, but probably should have another pair of eyes to confirm due to the breadth of change.

davidkelly (commented):
LGTM too -- but there were a couple fails in the jenkins run.   I'll re-run that check.

davidkelly (approved):
search test issue being worked on separately, so LGTM

PR #243, opened 2022-08-16
add more checks to search management test

avsej:


avsej:
jenkins test this please

jrawsthorne (approved):
LGTM. Instead of that could we not just append the scope and collection to the index name if on server >= 7.2

PR #249, opened 2022-08-21
Fix transcoder options for get_replica API

avsej:
@thejcfactor this patch also removes "hacks" to access flags/content directly, but instead adds more handy API for `content_as`, just like for `get`/`get_and_touch`.

Your passthrough transcoder will look like this:
```c++
struct passthrough_transcoder {
    using document_type = couchbase::codec::encoded_value;

    static auto decode(const couchbase::codec::encoded_value& data) -> document_type
    {
        return data;
    }
};
```
```c++
auto encoded = result.content_as<passthrough_transcoder>();
encoded.data    // => std::vector<std::byte>
encoded.flags   // => std::uint32_t
```

Also I plan to add other types of the transcoders, described in the RFC

### DIFF
--- a/core/impl/get_all_replicas.cxx
+++ b/core/impl/get_all_replicas.cxx
@@ -74,7 +74,8 @@ initiate_get_all_replicas_operation(std::shared_ptr<cluster> core,
                               return;
                           }
                       } else {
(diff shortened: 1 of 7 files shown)


avsej:
jenkins test this please

davidkelly (approved):

PR #250, opened 2022-08-21
fix framing boundaries calculation in case of multiple frames

avsej:
When more then one framing extras is being used, the boundaries
calcualted incorrectly (old size is not taken into account), which leads
to write attempts beyond vector's buffer, and leads to memory
corruption.


avsej:
jenkins test this please

chvck (approved):

PR #252, opened 2022-08-24
CXXCBC-53: implement legacy durability for low-level API

avsej:


brett19 (changes requested):
Looks good to me.  I'm slightly curious why you opted to use a trait to implement the check for being a compound operation rather than using something that can be checked through the operation itself as we did with the existing execute() checks (via encoded_request_type)?  If we had some constexpr on with_legacy_durability for instance, it would save needing to explicitly define the trait for the types (and their typedefs).  It risks missing one which would cause a compilation failure.  Note that I've not used traits very much, so I genuinely don't know much about their benefits :)

avsej:
@brett19 with http commands it was easier because encoded request/response is the same there. Here each mutation has different type of type. If you know the trick that does not require to tag all types that should give negative result for check, I will be happy to consider it.

avsej:
jenkins test this please

avsej:
jenkins, test this please

avsej:
jenkins test this please

avsej:
jenkins test this please

brett19 (approved):

PR #254, opened 2022-08-26
Fixed some missing includes in newer files.

brett19:


thejcfactor (approved):
LGTM.  Although one more update would be ideal:

Add #include "query_index_get_all_deferred.hxx" to core/operations/management/query.hxx

PR #264, opened 2022-09-02
CXXCBC-219: compound operation for build deferred indexes

avsej:


core/cluster.hxx
```diff
@@ -226,7 +226,11 @@ class cluster : public std::enable_shared_from_this<cluster>
         if (stopped_) {
             return handler(request.make_response({ errc::network::cluster_closed }, response_type{}));
         }
-        return session_manager_->execute(request, std::forward<Handler>(handler), origin_.credentials());
+        if constexpr (operations::is_compound_operation_v<Request>) {
+            return request.execute(shared_from_this(), std::forward<Handler>(handler));
```
avsej, line 230: @thejcfactor HTTP execute() method didn't have branch for compound requests.

thejcfactor: Ahhh, yes -- Good catch!

test/test_integration_management.cxx
```diff
@@ -1115,12 +1115,20 @@ TEST_CASE("integration: query index management", "[integration]")
...
         }
 
+        SECTION("public API")
         {
             auto manager = couchbase::cluster(integration.cluster).query_indexes();
             auto ctx = manager.build_deferred_indexes(integration.ctx.bucket, {}).get();
             REQUIRE_SUCCESS(ctx.ec());
         }
 
+        SECTION("core API")
```
thejcfactor, line 1125: I think it might be better to just copy the build test and instead of using the manager to build use the request. The way this test is means there aren't any any indexes to build b/c they have previously been built, right?

thejcfactor (approved):
Thanks for making this change!  If you agree on the comment for the tests...I can create a new PR to add the changes if that works. 😄

test/test_integration_management.cxx
avsej: nested SECTION works like multiplication of the cases

https://github.com/catchorg/Catch2/blob/3644b4135d2f4ec37fe6858dfc03b6273b0948ef/docs/tutorial.md#test-cases-and-sections

for each SECTION the test restarted from the beginning

thejcfactor: Ahhh, okay!  Great 😄

PR #271, opened 2022-09-21
CXXCBC-226: ensure that moving from movable_function will clear the source

avsej:
Right now we just delegate move assignment and constructor to std::function, which does not clean the source, which in turn lets "operator bool()" return true after the function has been moved.


davidkelly (approved):
LGTM 👍

PR #273, opened 2022-09-22
CXXCBC-232 Update transactions api

davidkelly:
Move to core, adjust namespaces and includes accordingly.  Also beginnings of public transaction api.   More to come, but wanted to start on it here so we can see how things will be organized, more or less.


CMakeLists.txt
```diff
@@ -268,26 +268,25 @@ set(couchbase_cxx_client_FILES
...
+        core/transactions/utils.cxx
+        core/transactions/transaction_attempt.cxx
+        core/transactions/transaction_config.cxx
+        core/transactions/transaction_context.cxx
+        core/transactions/transaction_get_result.cxx
+        core/transactions/transaction_links.cxx
+        core/transactions/transaction_query_options.cxx
+        core/transactions/transactions.cxx
+        core/transactions/transactions_cleanup.cxx
+        core/transactions/uid_generator.cxx)
```
avsej, line 289: you can use `cmake-format` here

```
pip install --user cmake-format
```

will install `cmake-format` somewhere into `$HOME/Library/Python/3.10/bin`

Then format the file in place.
```
cmake-format -i CMakeLists.txt
```

davidkelly: K -- done!  Thx for noticing it and mentioning `cmake-format`.

avsej (approved):

PR #275, opened 2022-10-03
CXXCBC-172: refresh DNS SRV when cluster uncontactable

avsej:
When DNS SRV records are used to connect to the SDK it's possible for the underlying addresses to change, i.e. the cluster to move. We need to detect this happening and react accordingly so that the SDK can continue to work correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ set(couchbase_cxx_client_FILES
     core/utils/split_string.cxx
     core/utils/url_codec.cxx
     core/impl/analytics_error_category.cxx
+    core/impl/dns_srv_tracker.cxx
     core/impl/append.cxx
     core/impl/build_deferred_query_indexes.cxx
     core/impl/common_error_category.cxx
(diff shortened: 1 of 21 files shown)


avsej:
Testing

```
brew install coredns
```

Create `db.example.org` with the following content
```
$ORIGIN example.org.
@	3600 IN	SOA sns.dns.icann.org. noc.dns.icann.org. (
				2017042745 ; serial
				7200       ; refresh (2 hours)
				3600       ; retry (1 hour)
				1209600    ; expire (2 weeks)
				3600       ; minimum (1 hour)
				)

	3600 IN NS a.iana-servers.net.
	3600 IN NS b.iana-servers.net.

_couchbase._tcp 10800 IN SRV 10 10 11210 node1.example.org.


node1     IN A     127.0.0.1
node2     IN A     127.0.0.1
```

Create `Corefile` with the following content
```
example.org {
    file db.example.org
    log
}
```

Run coredns
```
$ coredns -dns.port=1053
example.org.:1053
CoreDNS-1.10.0
darwin/amd64, go1.19.1, 
```

Export the following environment variables

```
export TEST_DNS_NAMESERVER=localhost
export TEST_DNS_PORT=1053
export TEST_CONNECTION_STRING=couchbase://example.org
```

PR #276, opened 2022-10-05
CXXCBC-237: expose get_with_lock, unlock and exists to public API

avsej:


brett19:
Out of curiosity, wasn't it the plan that the public API in couchbase-cxx-client would use std::future or std::async?

PR #278, opened 2022-10-06
CXXCBC-246: convert not_store code to document_exists

avsej:
The server can return either NOT_STORED or EXISTS error when inserting an existing document (MB-50742).


programmatix (approved):
LGTM

PR #279, opened 2022-10-06
CXXCBC-243 and CXXCBC-244: Expose kv transactions as public api

davidkelly:
Initial exposure of public api for transactions.   Will need to add transaction configs when we have a public api for creating the cluster (so there can be a transactions_config, etc...).   Also need queries exposed - felt they needed public query api first, but I'll look again, to be sure.

### DIFF
--- a/core/impl/transaction_op_error_category.cxx
+++ b/core/impl/transaction_op_error_category.cxx
@@ -24,7 +24,7 @@ namespace couchbase::core::impl
 struct transaction_op_error_category : std::error_category {
     [[nodiscard]] const char* name() const noexcept override
     {
-        return "couchbase.transaction";
+        return "couchbase.transaction_op";
     }
 
     [[nodiscard]] std::string message(int ev) const noexcept override
@@ -75,7 +75,7 @@ struct transaction_op_error_category : std::error_category {
             case errc::transaction_op::transaction_already_committed:
                 return "transaction already committed (1321)";
         }
-        return "FIXME: unknown error code (recompile with newer library): couchbase.transaction." + std::to_string(ev);
+        return "FIXME: unknown error code (recompile with newer library): couchbase.transaction_op." + std::to_string(ev);
     }
 };
 
--- a/core/impl/transaction_op_error_category.cxx
+++ b/core/impl/transaction_op_error_category.cxx
@@ -1,0 +1,116 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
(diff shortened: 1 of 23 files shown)


core/impl/transaction_op_error_category.cxx
```diff
@@ -0,0 +1,90 @@
...
+#include <couchbase/error_codes.hxx>
+
+#include <string>
+namespace couchbase::core::impl
+{
+
+struct transaction_op_error_category : std::error_category {
+    [[nodiscard]] const char* name() const noexcept override
+    {
+        return "couchbase.transaction";
```
avsej, line 27: This should use unique name of the category. `"couchbase.transaction_op"` maybe

core/impl/transaction_op_error_category.cxx
```diff
@@ -0,0 +1,90 @@
...
+            case errc::transaction_op::commit_not_permitted:
+                return "commit not permitted (1318)";
+            case errc::transaction_op::rollback_not_permitted:
+                return "rollback not permitted (1319)";
+            case errc::transaction_op::transaction_already_aborted:
+                return "transaction already aborted (1320)";
+            case errc::transaction_op::transaction_already_committed:
+                return "transaction already committed (1321)";
+        }
+        return "FIXME: unknown error code (recompile with newer library): couchbase.transaction." + std::to_string(ev);
```
avsej, line 78: error category as above: `couchbase.transaction_op.`

avsej (changes requested):

core/transactions.hxx
```diff
@@ -188,6 +188,9 @@ class transactions : public couchbase::transactions::transactions
 
     void run(const couchbase::transactions::per_transaction_config& config, async_logic&& code, txn_complete_callback&& cb);
 
+    void run(couchbase::transactions::async_txn_logic&& code,
+             couchbase::transactions::async_txn_complete_logic&& complete_cb,
+             const couchbase::transactions::per_transaction_config& cfg = {}) override;
```
programmatix, line 193: In ExtSDKIntegration it's no longer per_transaction_config, it's transaction_options.  This and other feedback was raised on https://github.com/couchbaselabs/couchbase-cxx-client/pull/251, which was an expensive full API review, and one I can't afford to redo here  Am I safe in assuming that all feedback on that PR will be used and that the final API will conform to the specced ExtSDKIntegration API from https://hackmd.io/cHIcXWZSQOi23qpJJAa8OQ?

core/impl/transaction_op_error_category.cxx
davidkelly: Oh thanks -- copy/paste failure :)

avsej (approved):

PR #281, opened 2022-10-11
CXXCBC-235: load system CAs when the trust certificate is not provided

avsej:
When the user has not set any root ca provider but is using TLS then we should trust both the system store and the Capella root CA.

### DIFF
--- a/core/cluster.hxx
+++ b/core/cluster.hxx
@@ -361,9 +361,13 @@ class cluster : public std::enable_shared_from_this<cluster>
                     return handler(ec);
                 }
             } else {
-                // add the cappela Root CA if no other CA was specified.
-                std::error_code ec{};
                 LOG_DEBUG(R"([{}]: use default CA for TLS verify)", id_);
+                std::error_code ec{};
+                tls_.set_default_verify_paths(ec);
+                if (ec) {
+                    LOG_WARNING(R"([{}]: failed to load system CAs: {})", id_, ec.message());
+                }
+                // add the capella Root CA if no other CA was specified.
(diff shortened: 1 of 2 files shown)


davidkelly (approved):
Ok -- seems set_default_verify_paths gets the CAs for the OS, so this LGTM.

ingenthr:
Interesting commentary:

> Load a set of default “certification authority” (CA) certificates from a filesystem path defined when building the OpenSSL library. Unfortunately, there’s no easy way to know whether this method succeeds: no error is returned if no certificates are to be found. When the OpenSSL library is provided as part of the operating system, though, it is likely to be configured properly.

core/cluster.hxx
```diff
@@ -361,9 +361,13 @@ class cluster : public std::enable_shared_from_this<cluster>
...
             } else {
-                // add the cappela Root CA if no other CA was specified.
-                std::error_code ec{};
                 LOG_DEBUG(R"([{}]: use default CA for TLS verify)", id_);
+                std::error_code ec{};
+                tls_.set_default_verify_paths(ec);
+                if (ec) {
+                    LOG_WARNING(R"([{}]: failed to load system CAs: {})", id_, ec.message());
+                }
+                // add the capella Root CA if no other CA was specified.
```
ingenthr, line 370: I think this comment is now incorrect, isn't it?  It adds the capella CA in addition to adding the platform certs.

avsej: yes, you right I will adjust the comment. Thanks for the note

PR #283, opened 2022-10-13
CXXCBC-235, CXXCBC-248: do not fail if trust certificate is not specified

avsej:
Now, when we load system certificates automatically, we can relax requirement to pass trust certificate in the connection options.

Instead the library will issue log warning reminding to either pass server sertificate in the options, or register it in the system store.


core/cluster.hxx
```diff
@@ -352,28 +351,33 @@ class cluster : public std::enable_shared_from_this<cluster>
                     tls_.set_verify_mode(asio::ssl::verify_peer);
                     break;
             }
-            if (!origin_.options().trust_certificate.empty()) {
```
avsej, line 355: Here I've just swapped condition branches to avoid negation for readability

jrawsthorne (approved):

PR #290, opened 2022-10-17
CXXCBC-172: fix reporting timeout error during bootstrap

avsej:


avsej:
Jenkins failures are not related to this patch and maybe ignored for now.

DemetrisChr (approved):
Fixes the issue - tested using the Ruby wrapper

PR #292, opened 2022-10-18
Revert move from thread.detach for now

davidkelly:
Turns out I'm blocking in transactions, so we can't {{asio::post}} instead.   Until the blocking code in transactions is refactored, we need to stick with detached threads.


thejcfactor (approved):
LGTM 👍

PR #293, opened 2022-10-18
TXNCXX-265: ExtSDKIntegration

davidkelly:
Motivation
==========
Now that transactions is integrated into the sdk, we needed to update it to support the ExtSDKIntegration changes.

Modification
============
Configs have changed names, and the cleanup no longer opens every
bucket, and assumes the atrs are in the default collection.    Each time
we create an atr, we update the cleanup to clean that keyspace, if it is
a new one.

Also - changed a json file, and noticed our .clang-format doesn't like json, so added a section for that.


core/transactions/transactions_cleanup.cxx
```diff
@@ -265,17 +262,26 @@ transactions_cleanup::create_client_record(const std::string& bucket_name)
 
 const client_record_details
 transactions_cleanup::get_active_clients(const std::string& bucket_name, const std::string& uuid)
+{
+    if (config_.metadata_collection()) {
+        return get_active_clients(config_.metadata_collection().value(), uuid);
+    }
+    return get_active_clients({ bucket_name, "_default", "_default" }, uuid);
```
avsej, line 269: could you use `couchbase::collection::default_name` and `couchbase::scope::default_name` constants here?

davidkelly: ooh good point - was being lazy -- will do.

PR #295, opened 2022-10-20
TXNCXX-266: Transaction config/option rework

davidkelly:
* naming consistent with ExtSDKIntegration
* builder pattern for all public config/options
* chainable mutators


couchbase/transactions/transactions_config.hxx
```diff
@@ -110,54 +116,99 @@ class transactions_config
...
         metadata_collection_ = keyspace;
+        return *this;
     }
 
-    void metadata_collection(const std::string& bucket, const std::string& scope, const std::string& collection)
+    /*void metadata_collection(const std::string& bucket, const std::string& scope, const std::string& collection)
     {
         metadata_collection_.emplace(bucket, scope, collection);
-    }
+    }*/
```
avsej, line 143: I guess, this meant to be removed? Or uncomment as it might be handy overload.

couchbase/transactions/transactions_config.hxx
```diff
@@ -175,7 +226,34 @@ class transactions_config
...
+        built(const built&);
+        built(const couchbase::durability_level,
+              const std::chrono::nanoseconds,
+              const std::optional<std::chrono::milliseconds>,
+              const std::unique_ptr<couchbase::core::transactions::attempt_context_testing_hooks>&,
+              const std::unique_ptr<couchbase::core::transactions::cleanup_testing_hooks>&,
+              const std::optional<transaction_keyspace>&,
+              const transactions_query_config::built&,
+              const transactions_cleanup_config::built&);
+        built& operator=(const built&);
```
avsej, line 242: the constructors and destructors might be removed it. Let compiler to generate defaults.

The idea is that this "struct built" should be simple, without any methods, just plain data. And it is the "build()" method below, that initializes everything.

couchbase/transactions/transactions_config.hxx
davidkelly: It was unused, so I was going to remove it.   I guess I forgot - though it could be handy...

couchbase/transactions/transactions_config.hxx
davidkelly: I had compile issues with the unique_ptr to the testing hooks (which are forward declared), and added this stuff...  But perhaps I need to look at that again.

davidkelly: After some thought, I realized a shared_ptr is just fine.  That eliminates the extra stuff in built.

avsej (approved):

PR #303, opened 2022-10-27
CXXCBC-142:  Update SRV resolution for Windows

thejcfactor:


thejcfactor:
ugh -- I will try to workout the Windows build.  Wasn't using VS2019 (was using VS2022).

avsej (approved):

core/io/CMakeLists.txt
```diff
@@ -1,12 +1,24 @@
...
+else()
+    target_link_libraries(
+            couchbase_io
+            PRIVATE project_options
+            project_warnings
+            snappy
+            http_parser
+            fmt::fmt
+            spdlog::spdlog)
+endif()
```
avsej, line 23: Also looks like you can use multiple `target_link_libraries` calls, and guard only `lphlpapi`.

avsej (changes requested):
Right, code looks good, just failing to build on other windows platform.

core/io/CMakeLists.txt
thejcfactor: I didn't know it was additive -- cool!  Updates to guard only `iphlpapi` 👍

avsej (approved):

PR #307, opened 2022-11-10
TXNCXX-254 Bugfixes from first pass through new fit performer.

davidkelly:
Needed for a crash-free run through the tests.

* Crash bug if atr for blocking document already cleaned
* Couple crash bugs from not moving (or moving then using) a callback.
* Update forward compat to declare support for ExtSDKIntegration


core/transactions/attempt_context_impl.cxx
```diff
@@ -635,8 +643,9 @@ attempt_context_impl::remove_staged_insert(const core::document_id& id, VoidCall
     };
     debug("removing staged insert {}", id);
 
-    if (auto err = hooks_.before_remove_staged_insert(this, id.key()); err) {
-        error_handler(*err, "before_remove_staged_insert hook returned error");
+    auto err = hooks_.before_remove_staged_insert(this, id.key());
+    if (err) {
+        error_handler(*err, "before_remove_staged_insert hook returned error", std::move(cb));
```
avsej, line 648: why the scope for `err` was expanded here? the old way seems to be okay

avsej (dismissed):

core/transactions/attempt_context_impl.cxx
davidkelly: Oh -- ya it was -- actually I liked it better that way.   I'll change it back!

avsej (approved):

PR #308, opened 2022-11-11
WIP: CXXCBC-242: range scan implementation

avsej:


avsej:
I will re-open refactoring PR using new branch on [couchbaselabs/couchbase-cxx-client](https://github.com/couchbaselabs/couchbase-cxx-client). This way following PR for range scan can use it as a target

PR #310, opened 2022-11-14
CXXCBC-242: core API for KV range scan

avsej:


core/crud_component.cxx
```diff
@@ -36,8 +36,137 @@
...
 
+#include <tl/expected.hpp>
+
 namespace couchbase::core
 {
+std::pair<std::vector<std::byte>, std::error_code>
+serialize_range_scan_create_options(const range_scan_create_options& options)
+{
+    tao::json::value body{};
+    if (options.without_content) {
```
programmatix, line 47: n.b. in latest meeting this was changed to ids_only.  It hasn't hit the RFC as a formal revision yet but I'm assuming it'll go through.

core/crud_component.cxx
```diff
@@ -36,8 +36,137 @@
...
+    } else {
+        return { {}, errc::common::invalid_argument };
+    }
+
+    if (options.snapshot_requirements) {
+        const auto& snapshot = options.snapshot_requirements.value();
+        tao::json::value requirements = {
+            { "vb_uuid", std::to_string(snapshot.vbucket_uuid) },
+            { "seqno", snapshot.sequence_number },
+            { "timeout_ms", options.timeout.count() },
```
programmatix, line 76: I couldn't see any mention of kvScanTimeout?

core/crud_component.cxx
avsej: fixed

core/crud_component.cxx
avsej: fixed

avsej:
jenkins, test this please

davidkelly (commented):
LGTM -- but lets see if @brett19 and @thejcfactor see anything that makes using in the wrappers problematic.

avsej:
Third patch will include manager class, which will encapsulate logic of tracking all vbucket-stream pairs with optional sorting.

brett19 (approved):
Core API here looks good, seems to essentially map exactly to how Go handles this (using callbacks for items rather than streams).  Note that after discussing with @chvck , having the item callback return batches seems like it's fine.  There is no compelling reason we _need_ to unbatch the items (as there might have been if you had used a channel here).

test/test_integration_range_scan.cxx
```diff
@@ -0,0 +1,699 @@
...
+          vbucket_id,
+          {},
+          [&items_callback_invoked](auto) { items_callback_invoked = true; },
+          [barrier](auto res, auto ec) {
+              barrier->set_value({ std::move(res), ec });
+          });
+        EXPECT_SUCCESS(op);
+
+        auto [resp, ec] = f.get();
+        REQUIRE(ec == couchbase::errc::key_value::document_not_found);
```
thejcfactor, line 447: Should this be a range scan cancelled error code?

test/test_integration_range_scan.cxx
```diff
@@ -0,0 +1,699 @@
...
+              barrier->set_value({ std::move(res), ec });
+          });
+        EXPECT_SUCCESS(op);
+
+        auto [res, ec] = f.get();
+        if (iteration == 1) {
+            REQUIRE(res.complete == false);
+            REQUIRE_SUCCESS(ec);
+        } else {
+            REQUIRE(ec == couchbase::errc::key_value::document_not_found); // scan has been cancelled
```
thejcfactor, line 551: Should this be a range scan cancelled error code?

thejcfactor:
> Third PR will include manager class, which will encapsulate logic of tracking all vbucket-stream pairs with optional sorting.

I was curious about this.  Good to know it will be coming.

thejcfactor (approved):
I did have comments on the type of error code returned when issuing a continue request after a cancel request, but I imagine there will be some iterations.  LGTM so far 👍

test/test_integration_range_scan.cxx
avsej: This is what the server gives us. I guess it does not differentiate "scan_uuid does not exist" from "scan_uuid has been canceled". Or does it, but the test cannot catch it.

PR #316, opened 2022-11-17
remove pipefail option from test shell script

avsej:
with this option, we don't get chance to inspect coredump files in case of test failures


avsej:
jenkins, test this please

avsej:
jenkins, test this please

davidkelly (approved):

PR #325, opened 2022-12-08
CXXCBC-242: Range scan orchestrator

avsej:


davidkelly (commented):
This all looks good to me.   I believe the scan_result_iterator returning a future is fine for Python, not sure if Node will be able to use that, but I'm sure Jared will have looked at it and I'll defer to his judgement.

core/range_scan_orchestrator.cxx
```diff
@@ -0,0 +1,404 @@
...
+    auto next() -> std::future<std::optional<range_scan_item>> override
+    {
+        auto barrier = std::make_shared<std::promise<std::optional<range_scan_item>>>();
+        if (options_.sort == scan_sort::none) {
+            next_item(streams_.begin(), [barrier](std::optional<range_scan_item> item) { barrier->set_value(std::move(item)); });
+        } else {
+            next_item_sorted({}, streams_.begin(), [barrier](std::optional<range_scan_item> item) { barrier->set_value(std::move(item)); });
+        }
+        return barrier->get_future();
+    }
```
thejcfactor, line 314: Instead of returning a future, can we allow wrappers/users of the API to pass in a callback?  As is, this won't work for Node.js.

thejcfactor (changes requested):
I think this looks good so far. 👍  

I think if the wrappers could pass in a callback to the next() method they should be able to consume the API (speaking for Python and Node.js that is 😄).

core/range_scan_orchestrator.cxx
avsej: okay, sure

thejcfactor (dismissed):
LGTM 👍

core/range_scan_options.hxx
```diff
@@ -53,11 +59,11 @@ struct range_snapshot_requirements {
 };
 
 struct range_scan_create_options {
-    std::chrono::milliseconds timeout;
     std::string collection_name;
     std::string scope_name;
-    std::uint32_t collection_id{ 0 };
     std::variant<range_scan, sampling_scan> scan_type;
```
thejcfactor, line 64: Could the scan_type be std::variant<std::monostate, range_scan, sampling_scan>?  This would allow the range_scan_create_options to be default-constructible as well, right?

The node autogen relies objects being default-constructible.

core/range_scan_options.hxx
```diff
@@ -37,8 +37,14 @@ struct scan_term {
 };
 
 struct range_scan {
```
thejcfactor, line 39: Would it be possible to allow the range_scan struct be default-constructible?

core/range_scan_orchestrator.cxx
```diff
@@ -0,0 +1,420 @@
...
+
+    void next(utils::movable_function<void(range_scan_item, std::error_code)> callback) override
+    {
+        auto handler = [callback = std::move(callback)](std::optional<range_scan_item> item) mutable {
+            if (item) {
+                callback(std::move(item.value()), {});
+            } else {
+                callback({}, errc::key_value::document_not_found);
```
thejcfactor, line 322: Is there a better way to signal that the streaming is complete than returning a document_not_found error?

avsej: I will create another code here, as it still internal object. Just for easier handling other errors.

davidkelly (approved):
LGTM

PR #326, opened 2022-12-13
CXXCBC-242: range scan fixes

avsej:


davidkelly (approved):

test/test_integration_range_scan.cxx
```diff
@@ -924,3 +924,76 @@ TEST_CASE("integration: manager scan range with content", "[integration]")
...
+        REQUIRE(entry->body);
+        REQUIRE_FALSE(entry->body->cas.empty());
+        REQUIRE(entry->body->value == value);
+        REQUIRE(entry->body->expiry_time() > now);
+
+        auto [_, inserted] = entry_ids.insert(entry->key);
+        REQUIRE(inserted);
+    } while (true);
+
+    REQUIRE(ids.size() >= 10);
```
davidkelly, line 994: Should we be testing the entry_ids here?   Also - should this be exactly 10, not >=?

davidkelly (changes requested):
Just that one question...

test/test_integration_range_scan.cxx
davidkelly: Ok this will be part of the next fix (for canceling the scans, etc...)

PR #328, opened 2022-12-15
fix ctest label for transaction tests on CI

avsej:
CXXCBC-277 - Update transaction tests to use existing integration test framework.


davidkelly:
I think the solution to the slow tests (when running `ctest ...`) is for me to just transition to the existing integration guard in the `test/utils`.   I'll open a ticket for that conversion, and add that work in here.

davidkelly (approved):

PR #332, opened 2023-01-10
CXXCBC-197: Add support for ExtInsertExisting

davidkelly:
Updates transactions to give an ignorable exception (in the core api) when a transactional insert discovers the document already exists (as long as it isn't in a transaction itself).  The public api just returns an error, and in either case that error doesn't stop the transaction.


thejcfactor (approved):
LGTM 👍

avsej (approved):

PR #333, opened 2023-01-10
TXNCXX-260 - ExtBinaryMetadata bug

davidkelly:
The atrs and client-records had `"u0000"` as the body, instead of a single `0` byte.   Thus they would be indexed by the index service, and other problematic things.

Simple fix, and added a test to catch any regressions in the future.


thejcfactor (approved):
LGTM 👍

avsej (approved):

PR #334, opened 2023-01-11
CXXCBC-259 Transactions public api expects bucket already open

davidkelly:
That's inconvienent - since you have to perform a kv op on it before the transaction.  So - added a function to insure it is open, on the public api calls.

While there, noticed an issue with cache_error_async, where we move the callback yet use it in the function as well.  For now, not using movable_function, allowing a copy, until I rework that function.


thejcfactor (approved):
LGTM 👍

avsej (approved):

avsej (approved):

PR #335, opened 2023-01-11
CXXCBC-188: Implement ChangePassword

SecureCake:
[CXXCBC-188](https://issues.couchbase.com/browse/CXXCBC-188)

Motivation
----------
Implement change password across SDKs

Changes
-------
*Initial commit


test/test_integration_management.cxx
```diff
@@ -884,6 +884,66 @@ TEST_CASE("integration: user management", "[integration]")
...
+        new_user.roles = {
+            couchbase::core::management::rbac::role{ "admin" },
+        };
+        auto [cluster, ec] = couchbase::cluster::connect(io, integration.ctx.connection_string, options_original).get();
+        auto coreCluster = cluster.core();
+        couchbase::core::operations::management::user_upsert_request upsertReq{};
+        upsertReq.user = new_user;
+        auto upsertResp = test::utils::execute(coreCluster, upsertReq);
+        REQUIRE_SUCCESS(upsertResp.ctx.ec);
+        usleep(100000); // Sleep so server can process
```
davidkelly, line 913: usleep isn't supported in windows.   Try `std::this_thread::sleep_for(std::chrono::seconds(10)` instead.   It would be nice not to sleep for an arbitrary amount of time, though, if possible.   I wonder if we can ping the server, looking for the user, until they are there?   

That may be worth doing here -- perhaps worth exploring?

test/test_integration_management.cxx
```diff
@@ -884,6 +884,66 @@ TEST_CASE("integration: user management", "[integration]")
...
+        asio::io_context io2;
+        auto guard2 = asio::make_work_guard(io2);
+        std::thread io_thread2([&io2]() { io2.run(); });
+        auto [cluster_new, ec_new] = couchbase::cluster::connect(io2, integration.ctx.connection_string, options_outdated).get();
+        auto coreCluster_new = cluster_new.core();
+        couchbase::core::operations::management::change_password_request changePasswordReq{};
+        changePasswordReq.newPassword = "newPassword";
+        auto changePasswordResp = test::utils::execute(coreCluster_new, changePasswordReq);
+        REQUIRE_SUCCESS(changePasswordResp.ctx.ec);
+        usleep(100000);
```
davidkelly, line 928: Same comment on the `usleep`.   Here, though, I don't suppose we can ping the server until the change happens -- I can't imagine we send the password back.   It's unfortunate there isn't a `mutation_token` equivalent that these sort of management calls would return, to allow us a deterministic way of knowing...

davidkelly (changes requested):
For sure we need to eliminate the `usleep`.   Since the password change probably can't be monitored until the change has happened, it may not be worth trying to eliminate the upsert user sleep.  Your call!

test/test_integration_management.cxx
SecureCake: Thanks David! Added a "wait until user present" to remove the arbitrary wait time.

test/test_integration_management.cxx
SecureCake: In the same vein as waiting for the user upsert we could "_wait until cluster connects_" by attempting (and retrying) a new connection with the cluster using the new credentials? Or even do that inside the test:
_while_ (cluster doesn't connect using new password) _do_ (initiate connection with new password) with timeout.

Seems a bit heavy for this use but it would be less arbitrary than sleep, what do you think?

davidkelly: Yea -- I do like that approach - good idea.

test/utils/wait_until.cxx
```diff
@@ -56,4 +56,20 @@ wait_until_collection_manifest_propagated(std::shared_ptr<couchbase::core::clust
...
+wait_until_user_present(std::shared_ptr<couchbase::core::cluster> cluster, const std::string& username)
+{
+    auto present = test::utils::wait_until([cluster, username]() {
+        couchbase::core::operations::management::user_get_request req{};
+        req.username = username;
+        auto resp = test::utils::execute(cluster, req);
+        return resp.user.username >= username;
+    });
+    if (present) {
+        // FIXME: The above check does not wait for all nodes to be up to date
```
davidkelly, line 70: True - this is always an issue with the management calls.   Any calls after this that assume the cluster completely agrees could fail.   The above call to `wait_until_cluster_manifest_propogated`, same thing.  We can live with this, no prob.

davidkelly: But -- why the >= username?  Does the user_get_response just have an error (like `couchbase::errc::management::user_not_found`)?   I didn't look, but I'm fairly sure it should.

davidkelly (changes requested):
Just a small change with the wait_until_user_present.   I like the approach of looping until the cluster doesn't connect with old credentials, then looping until it does connect with newest credentials a lot.

davidkelly (approved):
LGTM 👍

PR #337, opened 2023-01-12
CXXCBC-275 Update query_context handling in query management calls

davidkelly:
The decision we have made is to always send a `query_context`, and live with the errors we see from 6.x and earlier for the management calls.  We could follow up with an option that allowed to user to specify this isn't put in the calls, if really needed.

### DIFF
--- a/core/operations/management/query_index_drop.cxx
+++ b/core/operations/management/query_index_drop.cxx
@@ -29,31 +29,27 @@ query_index_drop_request::encode_to(encoded_request_type& encoded, http_context&
         return errc::common::invalid_argument;
     }
     encoded.headers["content-type"] = "application/json";
-    std::string keyspace = fmt::format("`{}`", bucket_name);
-    std::string query_context = keyspace;
+    std::string query_context = fmt::format("default:{}", bucket_name);
+    auto keyspace = query_context;
     if (!scope_name.empty()) {
-        keyspace += ".`" + scope_name + "`";
         query_context += ".`" + scope_name + "`";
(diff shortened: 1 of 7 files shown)


core/operations/management/query_index_drop.cxx
```diff
@@ -29,31 +29,27 @@ query_index_drop_request::encode_to(encoded_request_type& encoded, http_context&
         return errc::common::invalid_argument;
     }
     encoded.headers["content-type"] = "application/json";
-    std::string keyspace = fmt::format("`{}`", bucket_name);
-    std::string query_context = keyspace;
+    std::string query_context = fmt::format("default:{}", bucket_name);
```
thejcfactor, line 32: Would it be better to use the namespace_id here?  Same as _query_index_create_?

thejcfactor (dismissed):
I wonder if adding some default variables would be beneficial?  Could then replace the '_default' strings.  Just curious.  

static constexpr auto default_collection_name = "_default";
static constexpr auto default_collection_name = "_default";

LGTM 👍

core/operations/management/query_index_drop.cxx
davidkelly: ooh good catch - I'll fix it and make sure I didn't do that elsewhere.   Someday, we may have other namespaces.

davidkelly: Also those do exist -- the _default strings, somewhere... I'll look.

avsej (approved):

PR #340, opened 2023-01-20
CXXCBC-205: perf, use thread-local UUID generator

avsej:
before:
image

after:
image


davidkelly (approved):
Ya -- only create the random_device and generator once per-thread.  👍

PR #341, opened 2023-01-20
allow kv_loader to exit after certain number ops executed

avsej:
Small improvement, that makes it easier compare
perf output between fixes.


davidkelly (approved):
👍

PR #342, opened 2023-01-20
log DNS resolver on debug level

avsej:


avsej:
we are logging this info already. See https://github.com/couchbaselabs/couchbase-cxx-client/blob/4fe8ec748f58a811a2eef37ff4717f8bc6a1164a/core/io/dns_config.cxx#L127

PR #347, opened 2023-01-27
Eliminate looping transform in mcbp_parser::next

davidkelly:
I found that in mcbp_session_impl::do_read, 39% of the time was in mcbp::parser::next, with the vast majority of that in the std::transform
call after uncompressing the data.   By using vector::insert instead,
that dropped to only 11% of the time in do:read.

Probably the upshot here is to look for everywhere we are filling containers with a back_inserter and considering a way to modify that to not loop and instead copy in blocks.

before: 
image
after:
image


davidkelly:
After doing same in the `mcbp_parser::feed` call:
image

avsej:
cool, good catch!

davidkelly:
For upserts, etc... there was this same issue with `couchbase::core::utils::json::generate_binary`.   

Before: 
image

After:
image

avsej (approved):

PR #350, opened 2023-01-29
CXXCBC-286: decode value of user_agent_extra argument

avsej:


brett19:
Why are you only doing specifically for the user agent, but all connection string query strings should be URL decoded before use?

avsej:
in other cases there are nothing to encode anyway. those are either literals or numbers.

davidkelly (approved):

PR #359, opened 2023-02-03
CXXCBC-289 Refactor txn logging to match rest of cxx client

davidkelly:
Also, we were not rearming the logging meter, so we only got the metrics at shutdown.  When doing that, we _must_ do an io.close() after closing the cluster, or we wait around until the metrics timer goes off (10 min by default).   Small bit of code cleanup in a few places, added fmt::formatters for a number of classes/structs.


avsej:
I thought it might be something to do with MSVC preprocessor, but this patch does not look like helping: https://github.com/couchbaselabs/couchbase-cxx-client/commit/93a51d350dbc0d121c5b534ff97de6a147324855

davidkelly:
Thanks!  I'll look around a bit more....   But perhaps I can rewrite the macros to avoid this.   Will see later today/tomorrow....

avsej:
Dear Jenkins, test this please.

avsej (approved):

PR #360, opened 2023-02-04
CXXCBC-291: allow to disable mutation tokens

avsej:


davidkelly (approved):
Looks like we check the token in the "crud on default collection" test, so we know it is enabled by default.   So... LGTM!

PR #366, opened 2023-02-07
add cbc tool

avsej:
Adds cbc tool similar to what libcouchbase has. It is using the same name, because the goal is to eventually replace old tool. At the moment it implements:

* version: dumps version and build info
* get: retrieve document(s) from the collection
* query: perform N1QL query(s)

See built-in documentation (--help switch) for more info.


avsej:
Jenkins, test this please

davidkelly (approved):

PR #368, opened 2023-02-14
CXXCBC-296: force PLAIN SASL auth if TLS enabled

avsej:


brett19 (changes requested):
I think allowed_sasl_mechanisms needs to be optional, and if it's blank it uses PLAIN or SCRAM.  If it's explicitly specified, we need to use what's explicit.  Since, for instance, LDAP on non-TLS still needs be able to set just PLAIN.  If it's explicitly specified, we need to use what's explicit.  Since, for instance, LDAP on non-TLS still needs be able to set just PLAIN

avsej:
@brett19 now it will force PLAIN for TLS connections, and falls back to SCRAM if the user didn't specify mechanisms. And on the high-level the user does not select mechamisms directly, but rather indirectly using "ldapCompatible" constructor, which will set array of allowed mechanisms to "PLAIN" only.

brett19 (changes requested):
What I meant is that we should still support users who are using TLS, but also want to force the use of SCRAM by specifying an explicit set of authentication mechanisms.  The logic defined by https://issues.couchbase.com/browse/CCBC-1307 is such that if the user has specified no mechanisms, it should default to {PLAIN} for TLS, and {SCRAM-512, SCRAM-256, SCRAM-SHA1} for Non-TLS.  And if the user has specified a list of mechanisms explicitly, it should use those instead of any automatically determined defaults.

avsej:
it would be kind of fragile, IMHO. Now users should be aware of TLS and different SASL modes

brett19:
By default, the users don't need to care about the SASL mechanisms / TLS behaviour, etc...  Since the default auto-selects.  It's only if the user wants to be explicit that it becomes more fragile, but that's "expected" with advanced use cases.

avsej:
but can we to the other change separately? the logic you propose will require updating all wrappers to "NOT setting SASL mechanisms".

avsej:
Right, then I will do it here and open appropriate tickets on Jira.

brett19 (approved):

PR #369, opened 2023-02-14
CXXCBC-275 Update how we handle query_context.

davidkelly:
And, completely flesh out the query index managers.


tools/query.cxx
```diff
@@ -122,9 +122,6 @@ print_result_json_line(const std::optional<scope_with_bucket>& scope_id,
         meta["bucket_name"] = scope_id->bucket_name;
         meta["scope_name"] = scope_id->scope_name;
     }
-    if (built_options.scope_qualifier && !built_options.scope_qualifier->empty()) {
-        meta["scope_qualifier"] = built_options.scope_qualifier.value();
-    }
```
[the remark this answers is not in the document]
davidkelly: ooh good catch!

thejcfactor:
With the new changes coming, I think the query used in _query_index_get_all_deferred_request::encode_to_ needs to be updated to reflect the special case on when the collection == "_default" since the CollectionQueryIndexManager will always send the collection_name (at least I think that is the goal)?  I suppose this would be good to have regardless of the new changes, to cover the case where a user is passing the collection_name w/in the options when creating a query index.

Instead of:

`SELECT RAW name
FROM system:indexes
WHERE (bucket_id = 'default' AND scope_id = '_default' AND keyspace_id = '_default' )
    AND state = 'deferred'
    AND `using` = 'gsi'
ORDER BY is_primary DESC, name ASC`

I think it should update the WHERE clause to add " OR (bucket_id IS MISSING AND keyspace_id = 'default')" as referenced in the [RFC here](https://github.com/couchbaselabs/sdk-rfcs/blob/master/rfc/0054-sdk3-management-apis.md#sql).  So the query would become:

`SELECT RAW name FROM system:indexes
WHERE ((bucket_id = 'default' AND scope_id = '_default' AND keyspace_id = '_default')
OR (bucket_id IS MISSING AND keyspace_id = 'default'))
AND state = 'deferred' AND `using` = 'gsi'
ORDER BY is_primary DESC, name ASC`

davidkelly:
@thejcfactor - I'll take a look at that - but I should have removed the scope and collection from all the options in the public api - so that the only time a scope is in the request is when it comes from someone using the core api.   I'll look and see about this -- good catch.

avsej:
jenkins test this please

core/operations/management/query_index_drop.cxx
```diff
@@ -18,38 +18,33 @@
...
-        query_context += fmt::format(".`{}`", couchbase::scope::default_name);
-    }
-    if (!collection_name.empty()) {
-        keyspace += ".`" + collection_name + "`";
-    }
+    auto keyspace = utils::build_keyspace(*this);
     std::string drop_index_stmt;
     if (is_primary && index_name.empty()) {
-        drop_index_stmt = fmt::format(R"(DROP PRIMARY INDEX ON {} USING GSI)", keyspace);
+        drop_index_stmt = fmt::format(R"(DROP PRIMARY INDEX ON {})", keyspace);
```
thejcfactor, line 36: Should we keep the `USING GSI` portion of the query statement (same w/ other queries below)?

couchbase/collection_query_index_manager.hxx
```diff
@@ -0,0 +1,220 @@
...
+    {
+        auto barrier = std::make_shared<std::promise<manager_error_context>>();
+        auto future = barrier->get_future();
+        create_primary_index(options, [barrier](auto ctx) { barrier->set_value(ctx); });
+        return future;
+    }
+    /**
+     * Drop primary index on a collection.
+     *
+     * @param bucket_name name of the bucket in which to drop the primary index
```
thejcfactor, line 121: No `bucket_name` param for the `CollectionQueryIndexManager`

couchbase/collection_query_index_manager.hxx
```diff
@@ -0,0 +1,220 @@
...
+        auto barrier = std::make_shared<std::promise<manager_error_context>>();
+        auto future = barrier->get_future();
+        build_deferred_indexes(options, [barrier](auto ctx) { barrier->set_value(std::move(ctx)); });
+        return future;
+    }
+
+    /**
+     * Polls the state of a set of indexes, until they all are online.
+     *
+     * @param bucket_name name of the bucket in which to look for the indexes
```
thejcfactor, line 181: No `bucket_name` param for the `CollectionQueryIndexManager`

thejcfactor (changes requested):
LGTM!  Just some minor changes. I don't know if the drop index query missing `USING GSI` is important or not, but probably better to explicitly state.

core/operations/management/query_index_drop.cxx
davidkelly: Oh -- ya turns out that is really optional.   But...  I should keep it throughout.   I'll fix.

thejcfactor (approved):
LGTM 👍

avsej (approved):

PR #372, opened 2023-02-18
fix build with gcc-13

avsej:
gcc 13 does not provide `cstdint` implicitly.


avsej:
jenkins, test this please

davidkelly (approved):

PR #374, opened 2023-02-28
CXXCBC-194 Support ExtThreadSafe transaction extension.

davidkelly:
While at it, updated the public version of transaction_get_result to not strip out the metadata and links, as they are needed when staging replace/remove operations.

Also, the fit tests would sometimes run into issues with the naive retry
strategy when there was a FAIL_AMBIGUOUS while setting the ATR.   Redid
the transaction_context's utility for delaying to no longer be
recursive, avoiding the issue.   There are other places where this
strategy should be used in the future, in the utilities that do the
various forms of delay and exponential backoff, which I'll create a
ticket to improve later.


avsej (approved):

avsej:
jenkins test this please

PR #385, opened 2023-03-17
CXXCBC-316: Core txn get_optional after query issue

davidkelly:
When we have removed a doc in this transaction, and then call
get_optional, we expect to just get an empty optional.   However, the
handling was raising a transaction_operation_failed.  Simple fix - the
get_doc call can return no error, and no document.  When that happens
and we are in a get_optional, we should just return the empty optional.

While at it, added some simple `get` and `get_optional` tests that were missing.   And
added a test reproducing the issue Sergey saw.


avsej:
jenkins, test this please

avsej (approved):

PR #389, opened 2023-03-21
CXXCBC-310 LostTxnCleanup wasn't stopping right away

davidkelly:
When shutting down, for instance, the lost attempt loop would not stop sometimes, leading to issues with tests sometimes.  Now, I'm only accessing the running_ bool under the lock, so it is safe to use it in the wait calls as well.  Seems to have been an issue, simple fix.

Also - I was creating a client record doc if it was missing, when deleting, which does nothing.

### DIFF
--- a/core/transactions/transactions_cleanup.cxx
+++ b/core/transactions/transactions_cleanup.cxx
@@ -46,9 +46,8 @@ transactions_cleanup::transactions_cleanup(std::shared_ptr<core::cluster> cluste
   : cluster_(cluster)
   , config_(config)
   , client_uuid_(uid_generator::next())
-  , running_(false)
+  , running_(config.cleanup_config.cleanup_client_attempts || config.cleanup_config.cleanup_lost_attempts)
 {
-    running_ = config.cleanup_config.cleanup_client_attempts || config.cleanup_config.cleanup_lost_attempts;
     if (config.cleanup_config.cleanup_client_attempts) {
         cleanup_thr_ = std::thread(std::bind(&transactions_cleanup::attempts_loop, this));
     }
@@ -66,9 +65,9 @@ byteswap64(uint64_t val)
 {
     uint64_t ret = 0;
     for (std::size_t ii = 0; ii < sizeof(uint64_t); ii++) {
-        ret <<= 8ull;
-        ret |= val & 0xffull;
-        val >>= 8ull;
+        ret <<= 8ULL;
+        ret |= val & 0xffULL;
+        val >>= 8ULL;
     }
     return ret;
 }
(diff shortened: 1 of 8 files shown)


core/transactions/transactions_cleanup.cxx
```diff
@@ -526,7 +517,7 @@ transactions_cleanup::attempts_loop()
 void
 transactions_cleanup::add_attempt(attempt_context& ctx)
 {
-    auto& ctx_impl = static_cast<attempt_context_impl&>(ctx);
+    auto& ctx_impl = dynamic_cast<attempt_context_impl&>(ctx);
```
avsej, line 520: What is the reason if this change? I think it is exactly one kind of derived class here, right? I wonder if we really need this dynamic_cast, which can do extra checks in runtime and might return `nullptr`.

https://en.cppreference.com/w/cpp/language/dynamic_cast
> If the cast is successful, dynamic_cast returns a value of type new-type. If the cast fails and new-type is a pointer type, it returns a null pointer of that type. If the cast fails and new-type is a reference type, it throws an exception that matches a handler of type [std::bad_cast](https://en.cppreference.com/w/cpp/types/bad_cast).

core/transactions/transactions_cleanup.cxx
```diff
@@ -273,135 +274,127 @@ transactions_cleanup::create_client_record(const couchbase::transactions::transa
...
+                    for (const auto& [override, param] : value.get_object()) {
+                        if (override == FIELD_OVERRIDE_ENABLED) {
+                            details.override_enabled = param.get_boolean();
+                        } else if (override == FIELD_OVERRIDE_EXPIRES) {
+                            details.override_expires = param.as<std::uint64_t>();
+                        }
+                    }
+                } else if (key == FIELD_CLIENTS_ONLY) {
+                    for (const auto& [other_client_uuid, cl] : value.get_object()) {
+                        uint64_t heartbeat_ms = parse_mutation_cas(cl.at(FIELD_HEARTBEAT).get_string());
```
avsej, line 318: could you use the namespace for `uint64_t` and `int64_t` below to ensure that stdlib versions will be used

davidkelly: ooh - good catch.   I'll sweep through and make it consistent.

core/transactions/transactions_cleanup.cxx
davidkelly: Oh -- IIRC CLion had suggested it.  I never check the result anyways, as this has to succeed given the class hierarchy.   Ya back to static_cast.

avsej (approved):

PR #396, opened 2023-04-06
CXXCBC-227 Transaction public api docs

davidkelly:
While at it, renamed the transaction tests to not have that extra transactions in them.   Added a transaction test for examples.


avsej:
jenkins, test this please

avsej (approved):

PR #404, opened 2023-05-08
allow to run tests with gocaves

avsej:


test/test_integration_crud.cxx
```diff
@@ -35,6 +35,10 @@ TEST_CASE("integration: switching off mutation token", "[integration]")
     opts.enable_mutation_tokens = false;
     test::utils::integration_test_guard integration(opts);
 
+    if (integration.cluster_version().is_mock()) {
+        SKIP("GOCAVES does not allow to switching off mutation tokens. See https://github.com/couchbaselabs/gocaves/issues/100");
```
programmatix, line 39: I'm confused by is_mock() but also GOCAVES.  Does is_mock() always mean Gocaves, could it not mean the classic mock?  If it's always gocaves and the logging is explicitly referring to that, should it be is_gocaves()?

programmatix (approved):
Looks mostly good to me on a skim, but you might want someone who knows Gocaves to review also.

test/test_integration_crud.cxx
avsej: yes, it is always gocaves, the capital GOCAVES here is more like a marker to grep the logs and code. I hope eventually it will be removed, while is_mock() will remain, because we unlikely will implement search/analytics/query for the general tests in the gocaves.

PR #406, opened 2023-05-10
CXXCBC-328: fix restarting MCBP sessions

avsej:
Instead of letting session restarting itself, only allow it to removed itself from the bucket. And in the end, just notify the bucket to check configuration and establish connections to the endpoints that don't have it yet.


thejcfactor (approved):
LGTM 👍

PR #412, opened 2023-05-25
Skip one test for elixir deployment type

Couch-Matt:


test/test_integration_crud.cxx
```diff
@@ -328,7 +333,7 @@ TEST_CASE("integration: touch with zero expiry resets expiry", "[integration]")
     // set expiry with touch
     {
         couchbase::core::operations::touch_request req{ id };
-        req.expiry = 1;
+        req.expiry = 600;
```
avsej, line 336:
```suggestion
        
        if (integration.ctx.deployment == test::utils::deployment_type::elixir) {
            req.expiry = 600;
        } else {
            req.expiry = 1;
        }
```

avsej: not relevant, as later we indend to sleep for `req.expiry + 1`, which cannot be 601 seconds

avsej (approved):

PR #414, opened 2023-05-25
CXXCBC-331: fix FTS index provisioning script and improve test stability

avsej:
it contains improved setup scripts, as well as some test stability improvements, 
* better waiting for collections to be propagated (still not perfect, but closer), 
* cleanup of the collections (it behaves worse when tests leave a lot of the scopes and collections)
* skipping travel-sample for Capella
* avoid using non-portable index options for FTS
* improved test case naming in transactions


avsej:
jenkins, test this please

brett19 (approved):

PR #416, opened 2023-06-07
CXXCBC-333: fix parsing resolv.conf on Linux

avsej:
It also adds more details to the log entries


avsej:
This patch will be easier to review with hidden whitespace changes, like this

image

PR #424, opened 2023-06-29
Add `is_cancelled()` method to `scan_result`

DemetrisChr:
Added `is_cancelled()` method which is useful for some wrappers (e.g. Python). Included a test to verify that it works correctly

### DIFF
--- a/core/range_scan_orchestrator.cxx
+++ b/core/range_scan_orchestrator.cxx
@@ -442,6 +442,11 @@ class range_scan_orchestrator_impl
         }
     }
 
+    bool is_cancelled() override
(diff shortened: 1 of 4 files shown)


test/Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jun 26 12:14 BST
+----------------------------------------------------------
+End testing: Jun 26 12:14 BST
```
avsej, line 3: did you run `cmake` in the root?

avsej (approved):

PR #426, opened 2023-07-05
Return invalid_argument error code when concurrency is not positive in range scan

DemetrisChr:
* `errc::common::invalid_argument` when concurrency is non-positive
* Some tidying up in the tests - created a `get_vbucket_map()` function to make the code less repetitive

### DIFF
--- a/core/range_scan_orchestrator.cxx
+++ b/core/range_scan_orchestrator.cxx
@@ -391,13 +391,16 @@ class range_scan_orchestrator_impl
     {
 
         if (std::holds_alternative<sampling_scan>(scan_type_)) {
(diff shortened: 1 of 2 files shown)


avsej (approved):

avsej:
Jenkins, test this again

programmatix:
Optional but if you get a moment, would be awesome to add a FIT test for this so that other SDKs benefit.

DemetrisChr:
> 



> Optional but if you get a moment, would be awesome to add a FIT test for this so that other SDKs benefit.

I've added a few tests here (both for valid values of concurrency and 0) https://review.couchbase.org/c/transactions-fit-performer/+/193436

PR #430, opened 2023-07-12
CXXCBC-349: allow to pass trust certificate by value

avsej:


core/cluster.hxx
```diff
@@ -422,11 +424,22 @@ class cluster : public std::enable_shared_from_this<cluster>
...
+                        CB_LOG_WARNING("[{}]: unable to load CA passed via options object: {}", id_, ec.message());
+                    }
+                }
+                if (!origin_.options().trust_certificate.empty()) {
+                    CB_LOG_DEBUG(R"([{}]: use TLS verify file: "{}")", id_, origin_.options().trust_certificate);
+                    tls_.load_verify_file(origin_.options().trust_certificate, ec);
+                    if (ec) {
+                        CB_LOG_ERROR("[{}]: unable to load verify file \"{}\": {}", id_, origin_.options().trust_certificate, ec.message());
+                        return close([ec, handler = std::forward<Handler>(handler)]() mutable { return handler(ec); });
+                    }
```
avsej, line 442: I don't know if there should be some precedence between certificate passed by value and certificated passed via connection string.

I would imagine the case, where application has a certificate compiled in the binary, but the QE environment might require override it via connection string (environment variable).

So right now it just adds both certificate (if they are presented).

DemetrisChr (approved):

PR #446, opened 2023-08-31
CXXCBC-367 & CXXCBC-370 support bucket settings for bucket no dedup feature & collection management

Matt-Woz:
Changes:

- Implement collection management
- Implement update_collection
- Implement bucket deduplication

### DIFF
--- a/core/operations/management/collection_update.cxx
+++ b/core/operations/management/collection_update.cxx
@@ -25,7 +25,7 @@
 namespace couchbase::core::operations::management
 {
 std::error_code
-collection_update_request::encode_to(encoded_request_type& encoded, http_context& context ) const
+collection_update_request::encode_to(encoded_request_type& encoded, http_context& /* context */) const
 {
     encoded.method = "PATCH";
     encoded.path = fmt::format("/pools/default/buckets/{}/scopes/{}/collections/{}", bucket_name, scope_name, collection_name);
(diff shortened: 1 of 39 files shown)


core/operations/management/collection_update.cxx
```diff
@@ -0,0 +1,85 @@
...
+std::error_code
+collection_update_request::encode_to(encoded_request_type& encoded, http_context& /* context */) const
+{
+    encoded.method = "PATCH";
+    encoded.path = fmt::format("/pools/default/buckets/{}/scopes/{}/collections/{}", bucket_name, scope_name, collection_name);
+    encoded.headers["content-type"] = "application/x-www-form-urlencoded";
+    if (max_expiry.has_value()) {
+        encoded.body.append(fmt::format("maxTTL={}", max_expiry.value()));
+        if (history.has_value()) {
+            encoded.body.append(fmt::format("&history={}", history.value()));
```
Matt-Woz, line 36: Is there a way to get around this? Unlike create collection, update doesn't have a first default name attribute in the body, and it doesn't like &maxTTL or &history by itself

core/operations/management/collection_update.cxx
```diff
@@ -0,0 +1,85 @@
...
+
+collection_update_response
+collection_update_request::make_response(error_context::http&& ctx, const encoded_response_type& encoded) const
+{
+    collection_update_response response{ std::move(ctx) };
+    if (!response.ctx.ec) {
+        switch (encoded.status_code) {
+            case 400: {
+                if (encoded.body.data().find("Not allowed on this version of cluster") != std::string::npos ||
+                    encoded.body.data().find("Bucket must have storage_mode=magma") != std::string::npos) {
```
Matt-Woz, line 54: This is also not ideal, the history should be guarded by a 'nonDedupedHistory' bucket capability check, and return a feature_not_available before sending the request, but I tried adding it with the http_context but it never seemed to update the bucket capabilities, unless I e.g. ran an upsert just before

avsej: makes sense

DemetrisChr:
We need some feature macros for FIT purposes so we can exclude relevant code in the performer for older SDK versions. I think we should have separate feature macros for collection management public API and bucket no dedup

PR #451, opened 2023-09-14
CXXCBC-376 Create and Update bucket sending unnecessary fields to server

Matt-Woz:
Changes: 

- Change BucketSettings unrequired fields to be optional
- Update request encoding logic so the SDK doesn't send the fields unless they are specified

### DIFF
--- a/core/operations/management/bucket_create.cxx
+++ b/core/operations/management/bucket_create.cxx
@@ -168,15 +168,9 @@ bucket_create_request::make_response(error_context::http&& ctx, const encoded_re
                 auto* errors = payload.find("errors");
                 if (errors != nullptr) {
                     std::vector<std::string> error_list{};
-                    bool error_overridden = false;
                     for (const auto& [code, message] : errors->get_object()) {
                         if (message.get_string().find("Bucket with given name already exists") != std::string::npos) {
                             response.ctx.ec = errc::management::bucket_exists;
-                            error_overridden = true;
-                        }
-                        if (message.get_string().find("History Retention can only used with Magma") != std::string::npos &&
-                            !error_overridden) {
-                            response.ctx.ec = errc::common::feature_not_available;
                         }
                         error_list.emplace_back(message.get_string());
                     }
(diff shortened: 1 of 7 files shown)


core/operations/management/bucket_create.cxx
```diff
@@ -48,27 +48,35 @@ bucket_create_request::encode_to(encoded_request_type& encoded, http_context& /*
         case couchbase::core::management::cluster::bucket_type::unknown:
             break;
     }
-    encoded.body.append(fmt::format("&ramQuotaMB={}", bucket.ram_quota_mb));
-    if (bucket.bucket_type != couchbase::core::management::cluster::bucket_type::memcached) {
-        encoded.body.append(fmt::format("&replicaNumber={}", bucket.num_replicas));
+    if (bucket.ram_quota_mb == 0) {
+        encoded.body.append(fmt::format("&ramQuotaMB={}", 100)); // If not explicitly set, set to prior default value of 100
```
thejcfactor, line 52: I have asked the team if we possibly want some logic here to set the default to 1024 if the storage backend is set and set to magma.  As is, 100 MB is insufficient for a magma bucket and I am quite sure the server will return an error.

couchbase/management/bucket_settings.hxx
```diff
@@ -99,17 +99,17 @@ struct bucket_settings {
     std::string name;
     cluster::bucket_type bucket_type{ cluster::bucket_type::unknown };
     std::uint64_t ram_quota_mb{ 100 };
```
thejcfactor, line 101: I think we want this to be 0 MB, right?  For create I don't think it matters, but for update_bucket this might be a logic error if the user does not intend to update the buckets RAM quota.

core/operations/management/bucket_create.cxx
avsej: My cbdyncluster says minumum is 256, and it seems this setting might be dynamic.

https://github.com/couchbase/ns_server/blob/96072ad51c6739ec75f80737756f4db7bb7af13d/src/menelaus_web_buckets.erl#L1256-L1261

image

avsej: 256MB is for serverless

https://github.com/couchbase/ns_server/blob/96072ad51c6739ec75f80737756f4db7bb7af13d/etc/serverless_profile#L11

couchbase/management/bucket_settings.hxx
Matt-Woz: Yep, I forgot to update it - fixed now

avsej (approved):

PR #453, opened 2023-09-25
CXXCBC-382: Fix raw_binary_transcoder so that gets on binary data are possible.

zgroza:
In get_result.hxx:79, it is required that decode() method be templated. raw_binary_transcoder, unlike other transcoders, does not comply with this - and thus, it's not possible to perform any gets on binary data.


avsej (approved):

avsej:
fixed in https://github.com/couchbaselabs/couchbase-cxx-client/pull/459

PR #457, opened 2023-10-04
CXXCBC-377: Implement ExtParallelUnstaging in transactions

DemetrisChr:
## Motivation

Implement the ExtParallelUnstaging extension. Unstaging in parallel will be a major performance improvement, especially for larger transactions.

## Changes

* Each mutation unstaging is posted to the cluster's IO context to achieve parallelism - modified `commit_doc`, `remove_doc`, `rollback_insert` and `rollback_remove_or_replace` to achieve this, making use of the KV operations' handlers
* Added helpers for error handling which either set the exception on the promise or trigger a delayed retry
* Add `unstaging_state` that is used to limit the number of mutations being unstaged in parallel and prevents any further unstaging if an error has occurred
* Add `async_exp_delay` and `async_constant_delay` which use an asio timer to perform an asynchronous wait and then  invoke a callback (delay calculated in the same way as `exp_delay` and `constant_delay`) - used for retries during unstaging

## Results

* SDK tests pass
* `ExtParallelUnstagingTest` in FIT passes


core/transactions/staged_mutation.cxx
```diff
@@ -25,6 +25,23 @@
 namespace couchbase::core::transactions
 {
 
+void
+unstaging_state::wait_until_unstage_possible()
+{
+    std::unique_lock lock(mutex_);
+    cv_.wait(lock, [this]{ return in_flight_count_ < MAX_PARALLELISM; });
```
programmatix, line 32: If something goes wrong and the predicate never succeeds, do we wait here forever?

core/transactions/staged_mutation.cxx
```diff
@@ -165,14 +184,21 @@ staged_mutation_queue::commit(attempt_context_impl* ctx)
                 commit_doc(ctx, item);
                 break;
         }
-    }
+        state.notify_unstage_complete();
```
programmatix, line 187: If I'm reading commit_doc() right it can throw.   So is try-finally style logic needed?

Also what is going to be your policy on handling errors from commit_doc and remove_doc.  It's crucial that we don't have lingering background ops continuing after we're done with staged_mutation_queue::commit/rollback.

Please see the 2 rules [here](https://couchbase.slack.com/archives/C014WB8U2MQ/p1695209631970039) for more details.

core/transactions/staged_mutation.cxx
```diff
@@ -150,12 +167,14 @@ staged_mutation_queue::iterate(std::function<void(staged_mutation&)> op)
...
+std::future<void>
+staged_mutation_queue::commit_mutation(attempt_context_impl* ctx, staged_mutation& item, unstaging_state& state)
 {
-    CB_ATTEMPT_CTX_LOG_TRACE(ctx, "staged mutations committing...");
-    std::lock_guard<std::mutex> lock(mutex_);
-    for (auto& item : queue_) {
+    auto barrier = std::make_shared<std::promise<void>>();
+    auto future = barrier->get_future();
+
+    std::async(std::launch::async, [&]{
```
programmatix, line 176: Could you give me a brief description of how this (_gestures vaguely at std::async_) works in terms of threads, pools, etc.

core/transactions/staged_mutation.cxx
```diff
@@ -25,6 +25,23 @@
 namespace couchbase::core::transactions
 {
 
+void
+unstaging_state::wait_until_unstage_possible()
+{
+    std::unique_lock lock(mutex_);
```
programmatix, line 31: Is mutex_ reentrant?  Because we've already locked it at the top of staged_mutation_queue::commit/rollback.  Just checking that this isn't an accidental no-op.

DemetrisChr: That's the `mutex_` of `unstaging_state` which is different from `mutex_` of the `staged_mutation_queue` that's locked at commit/rollback

core/transactions/staged_mutation.cxx
DemetrisChr: According to the [docs](https://en.cppreference.com/w/cpp/thread/async) if you specify the `std::launch::async` launch policy to `std::async`, it executes the function in a new thread equivalent to using `std::thread`. Then `std::async` returns a future (so I shouldn't have used the `barrier` promise there and should've actually stored the future that it returns - the updated code I'll push shortly should address this).

Another issue with this piece of code is that if I have N mutations I'm creating N threads immediately and a lot of them will be blocked until they can do the commit/rollback if N > 1000. I will update it to have the main thread (or a thread responsible for starting the threads for each mutation) block until the in flight commits/rollbacks are less than 1000, so only that one thread will be blocked

core/transactions/staged_mutation.cxx
DemetrisChr: At the moment yes, so this is something I should fix. Theoretically, since the mutation operations have timeouts this should not happen, but since we are using concurrency I can't be 100% certain that things can't go wrong so I'll probably add a timeout to `wait` - but I'm not sure what timeout I should use

programmatix: How about what's left of the transaction timeout?

core/transactions/staged_mutation.cxx
DemetrisChr: Now using the cluster's IO context instead

core/transactions/staged_mutation.cxx
```diff
@@ -22,9 +22,46 @@
...
+    std::lock_guard lock(mutex_);
+    in_flight_count_--;
+    cv_.notify_one();
+}
+
+void
+unstaging_state::notify_unstage_error()
+{
+    std::lock_guard lock(mutex_);
+    abort_ = true;
```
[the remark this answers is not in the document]
DemetrisChr: I don't think is necessary at the moment. If `abort_` is set, the  `(in_flight_count_ < MAX_PARALLELISM) || abort_` condition will always evaluate to true and then the unstaging of any further mutations gets cancelled, so the value of `in_flight_count_` doesn't matter. But I will reduce it just in case there are any changes in the future and it gets missed

core/transactions/staged_mutation.cxx
```diff
@@ -153,78 +190,495 @@ staged_mutation_queue::iterate(std::function<void(staged_mutation&)> op)
...
+                    }
+                });
                 break;
         }
     }
+
+    std::exception_ptr exc{};
+    for (auto& future : futures) {
+        try {
+            future.get();
```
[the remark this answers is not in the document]
DemetrisChr: The try-catch block is inside the loop that iterates over all the futures. If a future throws an exception, I am storing the pointer to it, and then continuing to iterate over the futures of the rest of the in-flight operations and waiting for them to be completed. The exception is then re-thrown at the very end after all futures have completed, so I think we don't return while still having in-flight operations

core/transactions/staged_mutation.cxx
```diff
@@ -153,78 +190,495 @@ staged_mutation_queue::iterate(std::function<void(staged_mutation&)> op)
...
+
     for (auto& item : queue_) {
+        aborted = !state.wait_until_unstage_possible();
+        if (aborted) {
+            // Do not commit any more mutations
+            break;
+        }
+
+        auto barrier = std::make_shared<std::promise<void>>();
+        futures.push_back(barrier->get_future());
```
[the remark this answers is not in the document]
DemetrisChr: Yes, I don't think that can currently happen - but yes, it wouldn't hurt to have a try-catch block here. However, if an exception was to be thrown the `future.get()` code won't be reached, so it won't block - but weird things might happen if we don't wait for the in-flight operations. So I will add a try-catch that stores the exception pointer, then skips ahead to waiting for all the futures, and rethrows.

Should this catch block catch _any_ exception and then throw a `transaction_operation_failed`? I could set `aborted = true` in that case so the generic exception at the bottom gets thrown

core/transactions/staged_mutation.cxx
```diff
@@ -153,78 +190,495 @@ staged_mutation_queue::iterate(std::function<void(staged_mutation&)> op)
...
+
+        auto timer = std::make_shared<asio::steady_timer>(ctx->cluster_ref()->io_context());
+        async_constant_delay delay(timer);
```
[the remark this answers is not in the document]
DemetrisChr: `async_constant_delay` uses the asio timer to asynchronously wait for a certain amount of time and then execute an operation, used for retries when a `retry_operation` exception gets thrown. The behaviour is equivalent to  `retry_op` of the current implementation but using asynchronous waits and callbacks instead

core/transactions/staged_mutation.cxx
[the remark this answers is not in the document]
DemetrisChr: Oh now I see what you meant. Even if I do what I said above, I should not store the future until I'm sure that no exception was raised by `remove_doc`/`commit_doc`/etc.

core/transactions/internal/utils.hxx
```diff
@@ -300,6 +307,90 @@ struct constant_delay {
...
+          std::chrono::duration_cast<std::chrono::microseconds>(initial_delay * (jitter() * pow(2, static_cast<double>(retries++))));
+        if (delay > max_delay) {
+            delay = max_delay;
+        }
+        timer->expires_after(delay);
+        timer->async_wait([callback = std::move(callback)](std::error_code ec) mutable {
+            if (ec == asio::error::operation_aborted) {
+                callback(std::make_exception_ptr(retry_operation_retries_exhausted("retry aborted")));
+                return;
+            }
```
DemetrisChr, line 351: I'm not sure if this is the right approach here, but it seems like something very unlikely to happen since the timer is not used by anything else?

avsej (approved):

PR #459, opened 2023-10-09
CXXCBC-382: Fix raw_binary_transcoder so that gets on binary data are possible

avsej:
The issue was reported in https://github.com/couchbaselabs/couchbase-cxx-client/pull/453 initially.


thejcfactor (approved):
LGTM 👍

PR #460, opened 2023-10-12
Migrate dependencies from git submodules to CPM.cmake

avsej:
Use cmake package manager instead of git modules for spdlog/fmt. Also update the versions of these libraries.


avsej:
Jenkins, test this please once more time

thejcfactor (approved):
LGTM 👍

PR #463, opened 2023-10-24
Announce SSL library version in user agent string

avsej:
It is going to be always short but useful addition like "bssl/1010107f", which we can track on the server side to help debugging TLS-related issues.


thejcfactor (approved):
LGTM 👍

PR #464, opened 2023-10-25
CXXCBC-387: fix inconsitency with cached addresses with and without port

avsej:


thejcfactor (approved):
Test w/ the Node.js SDK and `CouchbaseMock.jar` look good now 👍

PR #469, opened 2023-10-30
CXXCBC-393: Treat third party dependences as system libraries

avsej:


thejcfactor (approved):
LGTM 👍 

Also confirmed CI builds w/ Python and Node.js.

PR #474, opened 2023-11-06
Patch BoringSSL for Windows MinGW and detect Apple ARM

avsej:
Apple ARM should provide support for all three features below, otherwise BoringSSL will fail to compile:

* __ARM_NEON
* __ARM_FEATURE_AES
* __ARM_FEATURE_SHA2

https://github.com/google/boringssl/blob/ad57528d2c978543106f9b115bd0eb658f3ebdd2/crypto/cpu_aarch64_apple.c#L49-L54

It turns out, that at this time all three versions of gcc from homebrew (11, 12, 13) only expose `__ARM_NEON`

```
$ echo | g++-13 -dM -E -x c++ - | grep '__ARM_NEON\|__ARM_FEATURE_AES\|__ARM_FEATURE_SHA2'
#define __ARM_NEON 1

$ echo | g++-12 -dM -E -x c++ - | grep '__ARM_NEON\|__ARM_FEATURE_AES\|__ARM_FEATURE_SHA2'
#define __ARM_NEON 1

$ echo | g++-11 -dM -E -x c++ - | grep '__ARM_NEON\|__ARM_FEATURE_AES\|__ARM_FEATURE_SHA2'
#define __ARM_NEON 1
```

while system compiler which is Apple Clang, exposes all of them
```
$ echo | g++ -dM -E -x c++ - | grep '__ARM_NEON\|__ARM_FEATURE_AES\|__ARM_FEATURE_SHA2'
#define __ARM_FEATURE_AES 1
#define __ARM_FEATURE_SHA2 1
#define __ARM_NEON 1
#define __ARM_NEON_FP 0xE
#define __ARM_NEON__ 1
```

So as a solution, we try to detect dysfunctional compilers for Apple ARM platform, and stop the build. Because BoringSSL is not default, and must be explicitly selected, we cannot just fallback to OpenSSL.

The second part of the commit just applies patch for MinGW to allow building BoringSSL, the patch is in the source tree already, we just missed the part that applies it during porting process.

### DIFF
--- a/cmake/0001-fix-build-for-mingw-w64-ucrt-x86_64-toolchain.patch
+++ b/cmake/0001-fix-build-for-mingw-w64-ucrt-x86_64-toolchain.patch
@@ -1,21 +1,21 @@
-From 51e44ffee037f4528407c81087fa18b60bf7fb27 Mon Sep 17 00:00:00 2001
+From 6bdc93aac66f2282823dceee802b60e2447c4ba9 Mon Sep 17 00:00:00 2001
 From: Sergey Avseyev <sergey.avseyev@gmail.com>
 Date: Fri, 20 Oct 2023 20:48:33 -0700
 Subject: [PATCH] fix build for mingw-w64-ucrt-x86_64-toolchain
 
 ---
- crypto/CMakeLists.txt        | 4 ++++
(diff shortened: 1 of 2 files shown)


thejcfactor (approved):
LGTM 👍

PR #477, opened 2023-11-08
update BlackDuck manifest

avsej:
* fix YAML formatting to use two spaces indentation
* synchronize BoringSSL version to what we declared in CPM
* use project names we use in CPM as the keys, as BD automatically uses bd-id to locate the project.


thejcfactor (approved):
LGTM 👍

PR #478, opened 2023-11-08
relax check for support on MacOS/aarch64

avsej:
we need further investigation of the issue with BoringSSL and gcc on ARM architectures for MacOS. For some reason, gcc that shipped by Homebrew cannot compile our BoringSSL


thejcfactor (dismissed):
LGTM 👍

thejcfactor (approved):

PR #482, opened 2023-11-22
CXXCBC-391: Fix transactions API bugs

DemetrisChr:
## Changes

* In `transactions_config`, `transactions_options` and the command line arguments of tools:
    * Remove `kv_timeout`
    * Rename `expiration_time` to timeout
* Set the value `"kvTimeoutMs"` in the transactions query requests to the `key_value_durable_timeout` instead of `key_value_timeout`

### DIFF
--- a/core/transactions/attempt_context_impl.cxx
+++ b/core/transactions/attempt_context_impl.cxx
@@ -340,8 +340,7 @@ attempt_context_impl::replace_raw(const transaction_get_result& document, const 
                             }
                             if (existing_sm != nullptr && existing_sm->type() == staged_mutation_type::INSERT) {
                                 CB_ATTEMPT_CTX_LOG_DEBUG(this, "found existing INSERT of {} while replacing", document);
-                                exp_delay delay(
-                                  std::chrono::milliseconds(5), std::chrono::milliseconds(300), overall_.config().expiration_time);
+                                exp_delay delay(std::chrono::milliseconds(5), std::chrono::milliseconds(300), overall_.config().timeout);
                                 create_staged_insert(document.id(), content, existing_sm->doc().cas().value(), delay, op_id, std::move(cb));
                                 return;
                             }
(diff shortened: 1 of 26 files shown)


core/transactions/attempt_context_impl.cxx
```diff
@@ -764,8 +763,7 @@ attempt_context_impl::query_begin_work(std::optional<std::string> query_context,
     txdata["state"] = tao::json::empty_object;
     txdata["state"]["timeLeftMs"] = overall_.remaining().count() / 1000000;
     txdata["config"] = tao::json::empty_object;
-    txdata["config"]["kvTimeoutMs"] =
-      overall_.config().kv_timeout ? overall_.config().kv_timeout->count() : core::timeout_defaults::key_value_timeout.count();
+    txdata["config"]["kvTimeoutMs"] = core::timeout_defaults::key_value_durable_timeout.count();
```
DemetrisChr, line 766: Here should we be fetching the `key_value_durable_timeout` set in the cluster options instead of always setting the default?

avsej: Do we have access to cluster options somewhere here?

Looking at Java implementation, it seems like they also don't have it on overall config, and rather pull it from global options (which still modifiable, kind of equivalent of our cluster options)

https://github.com/couchbase/couchbase-jvm-clients/blob/7007f9ce455ec88a8266d17a7c117dff6c4dc3a1/core-io/src/main/java/com/couchbase/client/core/transaction/CoreTransactionAttemptContext.java#L274-L275

avsej (dismissed):

core/transactions/attempt_context_impl.cxx
DemetrisChr: Will do this on a separate PR so this can get merged (since my change doesn't really change the existing behaviour)

Matt-Woz (approved):

core/transactions/attempt_context_impl.cxx
DemetrisChr: It was quite simple so I've added it to this PR

avsej (approved):

PR #484, opened 2023-11-28
CXXCBC-308: Add analytics index manager in the public API

DemetrisChr:
## Changes

* Add `analytics_index_manager` in the public API and all associated options classes
* Add `management::analytics_index`, `management::analytics_dataset` and `management::analytics_link` (with its three concrete implementations as per the RFC) in the public API
* Add relevant tests
* Fix the assertions for `get_pending_mutations` in the Core API test

### DIFF
--- a/couchbase/analytics_index_manager.hxx
+++ b/couchbase/analytics_index_manager.hxx
@@ -1,0 +1,463 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2023-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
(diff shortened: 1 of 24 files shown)


couchbase/analytics_index_manager.hxx
```diff
@@ -0,0 +1,463 @@
...
+     * Fetches the existing analytics remote links.
+     *
+     * @param options optional parameters
+     * @return future object that carries the result of the operation
+     *
+     * @since 1.0.0
+     * @committed
+     */
+    [[nodiscard]] auto get_links(const get_links_analytics_options& options) const
+      -> std::future<std::pair<manager_error_context, std::vector<std::unique_ptr<management::analytics_link>>>>;
```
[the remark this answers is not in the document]
avsej: Yes, I think it makes sense

test/test_integration_management.cxx
```diff
@@ -2866,7 +2867,10 @@ TEST_CASE("integration: analytics index management", "[integration]")
             couchbase::core::operations::management::analytics_get_pending_mutations_request req{};
             auto resp = test::utils::execute(integration.cluster, req);
             REQUIRE_SUCCESS(resp.ctx.ec);
-            REQUIRE(resp.stats[index_name] == 0);
+            // In the Core API the key has the `dataverse.dataset` format
+            auto key = fmt::format("{}.{}", dataverse_name, dataset_name);
+            REQUIRE(resp.stats.count(key) == 1);
```
DemetrisChr, line 2872: This assertion is failing against 6.6 because the server returns an empty response back for some reason. It seems that a similar issue with server versions < 7.0 was observed in the JVM SDKs as well: [JVMCBC-1075](https://issues.couchbase.com/browse/JVMCBC-1075). The same failure happens with the equivalent Public API test.

(Previously this test was passing because the incorrect key was used so `resp.stats[index_name]` was returning the default value which is 0.)

avsej: @DemetrisChr lets mask this test and skip it for <7.0

avsej (approved):

PR #486, opened 2023-11-29
CXXCBC-100: Add ability to set timeout for ping

DemetrisChr:
## Changes

* Add option for timeout in ping operations
* Set the ping status to `timeout` if the error code was a timeout

### DIFF
--- a/test/test_integration_diagnostics.cxx
+++ b/test/test_integration_diagnostics.cxx
@@ -359,8 +359,9 @@ TEST_CASE("integration: ping", "[integration]")
     {
         auto barrier = std::make_shared<std::promise<couchbase::core::diag::ping_result>>();
         auto f = barrier->get_future();
-        integration.cluster.ping(
-          "my_report_id", {}, {}, [barrier](couchbase::core::diag::ping_result&& resp) mutable { barrier->set_value(std::move(resp)); });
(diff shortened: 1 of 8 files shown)


test/test_integration_diagnostics.cxx
```diff
@@ -441,3 +444,59 @@ TEST_CASE("integration: ping allows to select bucket and opens it automatically"
         REQUIRE(res.services[couchbase::core::service_type::key_value][0].bucket.value() == integration.ctx.bucket);
     }
 }
+
+TEST_CASE("integration: ping allows setting timeout", "[integration]")
```
DemetrisChr, line 448: I think this test might have to be removed - it checks that with 1 millisecond timeout the ping for all services times out, which might not necessarily be the case. Do we have a good way to test that timeouts work?

avsej: No, I think it makes sense. Do think it would be source of unstable tests when sometimes the ping will not timeout in one millisecond?

DemetrisChr: Yes, potentially, if it manages to ping before the timeout. So far that does not seem to be the case either on localhost or CI so we could just keep an eye on it and remove it if that's an issue

avsej: OK

avsej (approved):

PR #490, opened 2023-12-11
CXXCBC-368: Subscribe to clustermap notifications to speedup failover

avsej:
Implement only deduplication flag from the RFC, and enable clustermap
change notifications. I this patch, we only want to use full
notifications, and do not turn off polling. So the SDK will be able to
receive new configuration sooner, but still be chatty as older versions.

Also this patch introduces tool cbc-beam that can send requests to specific vbucket or node id

### DIFF
--- a/core/bucket.cxx
+++ b/core/bucket.cxx
@@ -557,6 +557,7 @@ class bucket_impl
             } else {
                 added = config.nodes;
             }
+            config_.reset();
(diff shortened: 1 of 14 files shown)


thejcfactor (approved):
LGTM 👍 

Node testing looks good as well (no errors during rebalance).

PR #492, opened 2023-12-13
CXXCBC-409: Add handling for 'index does not exist' query error

DemetrisChr:
## Changes

* Handle `"Index does not exist"` query error
* Add handling for the 5000 error code in `document_query` (copied the one from the query management operations)
* Handle the possibility of `"status"` not being included in the query response - an example where this happens is with the query `drop index notfound on default;`

## Results

All query FIT tests pass

### DIFF
--- a/core/operations/document_query.cxx
+++ b/core/operations/document_query.cxx
@@ -26,6 +26,7 @@
 #include <couchbase/error_codes.hxx>
 
 #include <gsl/assert>
+#include <regex>
 
 namespace couchbase::core::operations
 {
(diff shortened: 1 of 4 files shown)


DemetrisChr:
TODO: I should also make sure that we match on `exist` and not just `exists` as described in the RFC & review the checks on the Python side to ensure that they are all handled within the C++ SDK instead.
> Note: the uppercase index for 5000 is not a mistake (also only match on exist not exists because there is a typo somewhere in query engine which might either print exist or exists depending on the codepath)
. Code 5000 AND message contains Index .+ already exist
. Code 4300 AND message contains index .+ already exist

avsej (approved):

PR #495, opened 2023-12-18
Revert "CXXCBC-405 Return type for touch is 'result' instead of 'mutation_result'"

avsej:
Reverts couchbaselabs/couchbase-cxx-client#493

The Touch operation does not return mutation token, so it is not really a Mutation.

I see that DCP has opcode for Expiration replication: https://github.com/couchbase/kv_engine/blob/master/docs/dcp/documentation/commands/expiration.md

Also @Matt-Woz said that protostellar returns MutationResult for Touch, should not we expect something special for GetAndTouch operation too?

@brett19, what do you think? Shall we revert this change?

The spec is not consistent, it mentions both IMutationResult and IResult at the same time:
https://github.com/couchbaselabs/sdk-rfcs/blob/c84c821cc738e1244d9c97f6a0e3f7814bd1f599/rfc/0053-sdk3-crud.md?plain=1#L719-L735

### DIFF
--- a/core/impl/collection.cxx
+++ b/core/impl/collection.cxx
@@ -144,7 +144,7 @@ class collection_impl : public std::enable_shared_from_this<collection_impl>
             options.timeout,
             { options.retry_strategy },
           },
-          [handler = std::move(handler)](auto resp) mutable { return handler(std::move(resp.ctx), mutation_result{ resp.cas }); });
+          [handler = std::move(handler)](auto resp) mutable { return handler(std::move(resp.ctx), result{ resp.cas }); });
     }
 
     void get_any_replica(std::string document_key,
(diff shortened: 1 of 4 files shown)


avsej:
I think, I'd rather revert it and raise question on the next team meeting.

brett19:
Interestingly, looking at CNG.  It appears that Protostellar exposes a 'MutationToken' field in our TouchResponse, but it's never actually specified as gocbcorex's TouchResult object does not contain a MutationToken.  Additionally, while Touch is indeed a 'Mutating Operation' on the surface, MutationToken's are used for the purposes of indexing consistency, and it could be argued that Touch does not change the contents of the document.  I don't think there is any technical reason that the server could not produce a MutationToken (since it ultimately will have DCP event associated with it, and thus a sequence number), but it probably does not do that today considering the confusion in the first place.

avsej:
Okay, then lets keep this PR open. I've added item to agenda.

brett19 (approved):

PR #496, opened 2023-12-18
CXXCBC-419: ensure that MCBP protocol parser starts with clean state

avsej:


thejcfactor (approved):
LGTM 👍

brett19:
@avsej Can you explain where I've made an error in the following thinking for understanding why this reset call is neccessary?  The `mcbp_parser` instance (`_parser`) is itself actually part of the mcbp_session object.  That `mcbp_session` appears to be one-time use, and you are calling `reset()` in `on_connect`, which is inherently invoked only once, and always prior to the `on_read`'s which ultimately feed data to the parser and pull packets off.  Additionally, `reset()` doesn't appear to have any useful behaviour except clearing the buffer inside the parser, which would already be empty since it's a new object I would think.

avsej:
no, it is not a one time object, and we reuse it if it cannot bootstrap, we just reopen the underlying TCP socket. Yes it is rather a design flaw, and this bug proves the weakness of reusing the session during retrying bootstrap.

PR #498, opened 2023-12-20
CXXCBC-401: Add ping & diagnostics to public API

DemetrisChr:
## Motivation

The `ping` and `diagnostics` operations currently only exist in the Core API, they also need to be exposed in the Public API

## Changes

* Add `ping()` to public `Cluster` and `Bucket` APIs
* Add `diagnostics() to the public `Cluster` API
* Add any relevant options & results to Public API (`diagnostics_options`, `ping_options`, `diagnostics_result`, `ping_result`, `endpoint_diagnostics`, `endpoint_ping_report`, `service_type`, `ping_state`, `endpoint_state`)
* Add tests (the public API equivalents of the existing core diagnostics tests)
* Moved the existing core diagnostics unit tests to a file separate from the integration tests

## Results

All tests pass

### DIFF
--- a/couchbase/cluster.hxx
+++ b/couchbase/cluster.hxx
@@ -22,6 +22,8 @@
 #include <couchbase/bucket.hxx>
 #include <couchbase/bucket_manager.hxx>
 #include <couchbase/cluster_options.hxx>
+#include <couchbase/diagnostics_options.hxx>
+#include <couchbase/ping_options.hxx>
 #include <couchbase/query_index_manager.hxx>
 #include <couchbase/query_options.hxx>
 #include <couchbase/search_index_manager.hxx>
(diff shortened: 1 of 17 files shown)


couchbase/endpoint_diagnostics.hxx
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Returns the namespace of this endpoint (likely the bucket name if present).
+     *
+     * @return the namespace.
+     *
+     * @since 1.0.0
+     * @committed
+     */
+    [[nodiscard]] auto endpoint_namespace() const -> std::optional<std::string>
```
DemetrisChr, line 163: Since we can't call this `namespace` that the RFC wants, I called it `endpoint_namespace` to work around the fact that it's a reserved keyword in C++. In the core it's called `bucket` which is what it mostly is at the moment, but could it be possible that in the future the namespace is not just the bucket name?

couchbase/cluster.hxx
```diff
@@ -207,6 +209,66 @@ class cluster
...
+     * @note This operation performs active I/O against services and endpoints to assess their health. If you do not
+     * wish to performs I/O, consider using @ref diagnostics() instead.
+     *
+     * @param options custom options to change the default behavior.
+     * @return future object that carries result of the operation.
+     *
+     * @since 1.0.0
+     * @committed
+     */
+    [[nodiscard]] auto ping(const ping_options& options = {}) const -> std::future<ping_result>;
```
DemetrisChr, line 238: Would we ever need to return an error code or an error context in the result here? At the moment, we do not need to, since if there are any errors, they are part of the ping result itself. In both ping & diagnostics the _current_ core implementation always returns a ping/diagnostics result.

avsej: I believe we don't need error context here

avsej (approved):

PR #504, opened 2024-01-20
CXXCBC-284: Use round-robin for config polling

avsej:
Instead of letting each MCBP session to poll for configuration automatically, make bucket object responsible for the polling orchestration and iterate over the list of known session to use only one socket to fetch configuration.

### DIFF
--- a/core/bucket.cxx
+++ b/core/bucket.cxx
@@ -28,6 +28,7 @@
 #include "mcbp/queue_response.hxx"
 #include "origin.hxx"
 #include "ping_collector.hxx"
+#include "protocol/cmd_get_cluster_config.hxx"
 #include "retry_orchestrator.hxx"
 
 #include <couchbase/metrics/meter.hxx>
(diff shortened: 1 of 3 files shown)


mikereiche (approved):

mikereiche (approved):

ingenthr:
Doesn't the RFC say to have a random list and then iterate through it so we don't have a stampeding herd if there's a network issue?

core/bucket.cxx
```diff
@@ -476,12 +482,52 @@ class bucket_impl
...
+        std::optional<io::mcbp_session> session{};
+        {
+            std::scoped_lock lock(sessions_mutex_);
+
+            std::size_t start = heartbeat_next_index_.fetch_add(1);
+            std::size_t i = start;
+            do {
+                auto ptr = sessions_.find(i % sessions_.size());
+                if (ptr != sessions_.end() && ptr->second.supports_gcccp()) {
+                    session = ptr->second;
```
mikereiche, line 503: should this break out of the while loop?

mikereiche:
the "SHOULD" leaves a little wiggle room.   But I looked again at the code, and it looks like the do/while is missing an an exit condition (other than checking all the possible  sessions).  It looks like it will always choose the last one it checks (?). 

"The SDK MUST perform these connections sequentially and not in parallel, and SHOULD perform this in a consistently random order (the SDK should pre-shuffle the list, and then use the pre-shuffled order consistently). "

ingenthr:
Thanks for checking, sounds like we should file anew or reopen that issue @mikereiche?

PR #505, opened 2024-01-23
CXXCBC-431: Fix history retention bucket capability check

DemetrisChr:
## Motivation
The `http_context` appears to be at the cluster level, so it wasn't clear which bucket's capabilities were being checked. This resulted in unpredictability, as seen in FIT and in the Python SDK's test suite.

## Changes
Add a new `execute_with_bucket_capability` method that after ensuring that the bucket is open, it checks for the presence of the capability using `with_bucket_configuration` and then proceeds to do the operation if successful. This is now used for `create_collection` and `update_collection`.

## Results
Tests pass (including FIT)

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -333,6 +333,38 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
         }
     }
 
+    template<class Request,
+             class Handler,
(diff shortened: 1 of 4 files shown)


avsej (approved):

thejcfactor (approved):
Thanks!! LGTM 👍

PR #508, opened 2024-01-26
CXXCBC-446: log more details on connection failures

avsej:


thejcfactor (approved):
This is good!! Useful for debugging for sure.

PR #516, opened 2024-02-03
CXXCBC-447: Use addresses from the config to bootstrap bucket

avsej:
Use hostnames from the cluster configuration to bootstrap bucket connections instead of original list from the connection string.

It is possible that connection string does not have any node that has KV service, and in this case the SDK will fail to open bucket.


thejcfactor (dismissed):
Changes LGTM.  Also, I verified locally running the same test/setup that I logged in CXXCBC-447. 👍

thejcfactor (approved):
LGTM 👍

PR #518, opened 2024-02-07
Change encoded search request showrequest param from string to boolean.

thejcfactor:


DemetrisChr (approved):
It looks like a reasonable change to me and the RFC agrees that it should be a boolean as well:
> The SDK will now send `"showrequest": false` in the top-level JSON, for all queries (vector or not) sent via the new `cluster/scope.search()` interface.

I have also verified this with Ruby.

avsej (approved):

PR #519, opened 2024-02-08
CXXCBC-191: Create Index Key Encoding

DemetrisChr:
## Motivation

Make the SDK consistent with the management RFC revision/clarification in couchbaselabs/sdk-rfcs#116

## Changes
* Rename `fields` parameter to `keys` in the Public API's `create_index()`
* Encode each index key provided to `create_index()` by surrounding them with backticks
* Add unit test for the encoding & update integration test to reflect these changes


thejcfactor (approved):
LGTM 👍

PR #522, opened 2024-02-16
CXXCBC-452: detect new search capabilities

avsej:
Fail fast if the cluster does not support new search features.

The patch also updates list of known capabilities and provides utilities to help with asserting their presence in more general way.

### DIFF
--- a/core/topology/configuration_json.hxx
+++ b/core/topology/configuration_json.hxx
@@ -213,33 +213,47 @@ struct traits<couchbase::core::topology::configuration> {
             for (const auto& entry : m->get_array()) {
                 const auto& name = entry.get_string();
                 if (name == "couchapi") {
-                    result.bucket_capabilities.insert(couchbase::core::bucket_capability::couchapi);
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::couchapi);
                 } else if (name == "collections") {
-                    result.bucket_capabilities.insert(couchbase::core::bucket_capability::collections);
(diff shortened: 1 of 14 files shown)


core/topology/configuration_json.hxx
```diff
@@ -213,51 +213,76 @@ struct traits<couchbase::core::topology::configuration> {
...
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::subdoc_revive_document);
+                } else if (name == "dcp.IgnorePurgedTombstones") {
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::dcp_ignore_purged_tombstones);
+                } else if (name == "preserve_expiry") {
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::preserve_expiry);
+                } else if (name == "querySystemCollection") {
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::query_system_collection);
+                } else if (name == "mobileSystemCollection") {
+                    result.capabilities.bucket.insert(couchbase::core::bucket_capability::mobile_system_collection);
+                } else if (name == "subdoc.ReplicaRead") {
```
Matt-Woz, line 257: Duplicated `subdoc.ReplicaRead` check on L239 (causing build failure on alpine)

avsej: Thank you for notice. I will update it, I think better leave the newer version of the enum value `subdoc_replica_read` as it reflects the underlying capability better (otherwise this error would not happen)

Matt-Woz (approved):

PR #525, opened 2024-02-20
CXXCBC-345: Range scan improvements & resolve concurrency issues

DemetrisChr:
## Motivation

A number of issues were observed with the existing range scan implementation, such as items missing & the scan getting stuck indefinitely when more than one IO threads are used in the SDK. This change simplifies the implementation of range scan and resolves issues with concurrency and IO threads.

## Changes

* Create a `range_scan_load_balancer` class to contain all of the logic around selecting which vbucket to scan.
* Instead of using one concurrent channel for each vbucket scan when sending items to the caller (i.e. having 1024 channels), only use a single channel shared by all vbucket scans. This significantly simplifies the logic of `next_item()`. Items are sent through this channel, as well as signals that a vbucket scan has been completed (either successfully or with a fatal error), which allows the orchestrator to keep track of whether there are any outstanding vbucket scans.
* Use `asio::post` when triggering KV operations within a KV callback (e.g. when doing a `range_scan_continue` from the callback of `range_scan_create`)
* Add unit tests for the load balancing logic
* Fix an issue with the "orchestrator prefix scan, get 10 items and cancel" test where it was fetching documents created in a different test.
* For sampling scans, if a seed is provided it is now used for shuffling the vbuckets in the load balancer to ensure that a given seed always yields the same sample when concurrency 1 is used. We cannot have the same guarantee for concurrency > 1, as the order the vbuckets are scanned depends on factors such as how long each scan takes and how many scans are taking place on each node at any given time.

## Results

All tests pass, including FIT


Matt-Woz (commented):
LGTM but maybe want another pair of eyes to confirm - also verified with PHP and it seems to have solved the previous issues w/concurrency

thejcfactor (commented):
Apart from my 1 comment, FIT looks good w/ the Python and Node.js performers.  I can consistently get all tests to pass (wasn't the case previously).  I will look more closely at the changes tomorrow if @avsej cannot get to them first. 👍

core/range_scan_load_balancer.cxx
```diff
@@ -0,0 +1,145 @@
...
+range_scan_load_balancer::range_scan_load_balancer(const topology::configuration::vbucket_map& vbucket_map,
+                                                   std::optional<std::uint64_t> seed)
+  : seed_{ seed }
+{
+    std::map<std::int16_t, std::queue<std::uint16_t>> node_to_vbucket_map{};
+    for (std::uint16_t vbucket_id = 0; vbucket_id < vbucket_map.size(); vbucket_id++) {
+        auto node_id = vbucket_map[vbucket_id][0];
+        if (node_to_vbucket_map.count(node_id) == 0) {
+            node_to_vbucket_map[node_id] = {};
+        }
```
avsej, line 85: no need to initialize the queue, it is default-constructible, and `map::operator[]` is non-const, it access-or-insert operator.

godbolt: https://godbolt.org/z/Eo3b8G89f
reference: https://en.cppreference.com/w/cpp/container/map/operator_at
> Returns a reference to the value that is mapped to a key equivalent to key or x respectively, performing an insertion if such key does not already exist.

core/range_scan_load_balancer.cxx
```diff
@@ -0,0 +1,145 @@
...
+    std::iota(iterators.begin(), iterators.end(), nodes_.begin());
+    std::mt19937_64 gen{ std::random_device{}() };
+    if (seed_.has_value()) {
+        gen.seed(seed_.value());
+    }
+    std::shuffle(iterators.begin(), iterators.end(), gen);
+
+    for (auto it : iterators) {
+        auto node_id = it->first;
+        auto* node_status = &it->second;
```
avsej, line 118: It is sugar, but I think you can do structured binding here

```c++
        auto &[node_id, node_status] = *it;
```

avsej (changes requested):
Looks good in general, couple of style notes

core/range_scan_load_balancer.cxx
DemetrisChr: Done

core/range_scan_load_balancer.cxx
DemetrisChr: Thanks for the tip, done

avsej (approved):

PR #529, opened 2024-02-29
Improvements in cbc-pillowfight

avsej:
* record latencies and display histogram in the end of the run
* fix termination condition when using multiple threads
* use running average for runtime rate display
* allow to configure separate batch size for queries
* fixed links in the reference pages


thejcfactor (approved):
LGTM 👍

thejcfactor (approved):
LGTM 👍

PR #532, opened 2024-03-04
CXXCBC-462: Fix hanging when specifying a custom metadata collection via the public API & expose errors

DemetrisChr:
## Motivation

Specifying a custom metadata collection when calling `cluster::connect` results in the SDK to hang as the constructor of `core::transactions::transactions`, which is called in the callback of `core.open` (running on an IO thread), attempts to open the bucket of the custom metadata (requires an available IO thread) and blocks until that is completed (which means that it does not release its IO thread to be used by `open_bucket`). In addition, the constructor throws exceptions when an error occurs in the constructor which are not being handled by either the Public API or the wrappers.

## Changes

* Add a `core::transactions::transactions::create` method to construct the core transactions object, which takes a callback and can be chained with other asynchronous operations. It also returns an appropriate error code via the callback when an error occurs.
* In `cluster::connect`, handle any errors that occur during `transactions::create` by closing the cluster and surfacing the error via the `cluster_connect_handler`
* Added test to ensure that `cluster::connect` returns `bucket_not_found` when the specified metadata collection's bucket does not exist

## Results

* Test passes
* FIT tests which require a custom metadata collection (e.g. `CleanUpInMultipleBucketsTest`) no longer hang when the performer runs with 1 IO thread


core/transactions.hxx
```diff
@@ -127,21 +129,49 @@ using txn_complete_callback =
 class transactions : public couchbase::transactions::transactions
 {
   public:
+    /**
+     * @internal
+     * Called internally
+     */
+    transactions(core::cluster cluster, const couchbase::transactions::transactions_config::built& config);
```
DemetrisChr, line 136: Should I make the constructor private?

core/transactions.hxx
```diff
@@ -127,21 +129,49 @@ using txn_complete_callback =
...
      *
      * @param cluster The cluster to use for the transactions.
      * @param config The configuration parameters to use for the transactions.
+     * @param cb Called when the transactions object has been created or an error has occurred during the creation.
      */
-    transactions(core::cluster cluster, const couchbase::transactions::transactions_config::built& config);
-    transactions(core::cluster cluster, const couchbase::transactions::transactions_config& config);
+    static void create(core::cluster cluster,
+                       const couchbase::transactions::transactions_config::built&,
+                       utils::movable_function<void(std::error_code, std::shared_ptr<transactions>)>&& cb);
```
DemetrisChr, line 155: @thejcfactor Does returning a shared pointer here work for Python/Node?

core/transactions.hxx
avsej: yes, it makes sense. Everything will go through `create` method then.

core/transactions.hxx
avsej: the wrapper could always use future/promise to wait for completion

avsej (dismissed):

core/transactions.hxx
thejcfactor: I agree -- I think with the changes, we need to force the use of the `transactions::create()`.  I suppose it is not a problem if not using the metadata collection, but I think going the route of not providing any "foot-guns" is the better path.

core/transactions.hxx
thejcfactor: I have tested out the changes w/o any problems.  When we create the `transactions` obj we do it in a blocking fashion so I think using one of the overloads that returns the `std::future<std::pair<>>` is the path we would take. Either way, we can get to correctly raising an exception is there is an issue w/ opening the metadata bucket. 👍

thejcfactor:
This question is not related to the changes specific to this PR, but it is related. Why do we "implicitly" create the transactions object when we create a cluster object in the public API?  I don't get the sense that is what we need to do in order to support `ExtSDKIntegration` ([ref Transaction instantiation](https://hackmd.io/cHIcXWZSQOi23qpJJAa8OQ#Transaction-instantiation)).  To me, it would seem better to lazily create the transactions object when users call `cluster::transactions()`.  Is there something I am missing?

DemetrisChr:
> This question is not related to the changes specific to this PR, but it is related. Why do we "implicitly" create the transactions object when we create a cluster object in the public API? I don't get the sense that is what we need to do in order to support `ExtSDKIntegration` ([ref Transaction instantiation](https://hackmd.io/cHIcXWZSQOi23qpJJAa8OQ#Transaction-instantiation)). To me, it would seem better to lazily create the transactions object when users call `cluster::transactions()`. Is there something I am missing?

The spec doesn't specify this behaviour. It only says "The life-cycle of that Transactions object is now bound to the Cluster it's created from." and it appears to leave the details up to the implementation. However, I agree that it would be better to lazily create the transactions object to avoid spinning up all the transactions infrastructure for users who are not using transactions. A problem here is that `transactions::create` currently returns an error because of this `bucket_open` which would be inconvenient if it was returned via `transactions()` I think (since we return errors rather than throwing in the Public API). 

I'll need to investigate further if we should actually be doing this `bucket_open` here or if we can just do the transactions create asynchronously and surface any errors elsewhere (e.g. on txn.run). If we end up doing this, we should be able to easily move the creation to `transactions()`.

core/transactions.hxx
DemetrisChr: It appears that you can't use `std::make_shared<transactions>` if the constructor is private, and the workarounds are slightly ugly. I think I'm leaning towards leaving them public as it's a core API, though happy to do otherwise.

avsej:
> To me, it would seem better to lazily create the transactions object when users call cluster::transactions(). Is there something I am missing?

I agree, this is what we do with buckets, scopes and collections. Lets capture this as a ticket and discuss it separately. Maybe it should go one step further, and not just lazily create transaction object, but defer running background tasks and opening a bucket until first transaction will be scheduled.

avsej (approved):

DemetrisChr:
> > To me, it would seem better to lazily create the transactions object when users call cluster::transactions(). Is there something I am missing?
> 
> I agree, this is what we do with buckets, scopes and collections. Lets capture this as a ticket and discuss it separately. Maybe it should go one step further, and not just lazily create transaction object, but defer running background tasks and opening a bucket until first transaction will be scheduled.

I've filed a ticket for this: [CXXCBC-474](https://issues.couchbase.com/browse/CXXCBC-474)

PR #533, opened 2024-03-04
CXXCBC-336: Do not fallback to 8.8.8.8, if we cannot obtain system DNS server

avsej:
Avoid assuming Google DNS and rather report DNS-SRV lookup failure. The details still logged and user still can specify their DNS server in options.

### DIFF
--- a/core/io/dns_config.hxx
+++ b/core/io/dns_config.hxx
@@ -40,7 +40,7 @@ class dns_config
     [[nodiscard]] std::chrono::milliseconds timeout() const;
 
   private:
-    std::string nameserver_{ default_nameserver };
+    std::string nameserver_{};
     std::uint16_t port_{ default_port };
     std::chrono::milliseconds timeout_{ timeout_defaults::dns_srv_timeout };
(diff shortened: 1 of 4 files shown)


core/io/dns_config.hxx
```diff
@@ -40,7 +40,7 @@ class dns_config
     [[nodiscard]] std::chrono::milliseconds timeout() const;
 
   private:
-    std::string nameserver_{ default_nameserver };
+    std::string nameserver_{};
```
thejcfactor, line 43: Do we need to keep the `static constexpr auto default_nameserver = "8.8.8.8";` line?  Seems like it might be better to not even hint at a potential hard-coded default value.  I think it is used in the tests, but that is easily moved to only reside w/in the test suite.

avsej: It is a constant, let it be there tests are using it.

thejcfactor (approved):
Fair enough -- LGTM, but I still don't like that constant hanging around 🙂

PR #536, opened 2024-03-05
CXXCBC-448: Expose hooks for fork() scenarious

avsej:
In order to gracefully survive fork() syscall, the library should be able to move itself into stable and predictable state. In particular ASIO should not have any pending work, and all IO thread should be stopped. After the fork() the parent process retains control over all file descriptors, but the child have to reconnect them (ASIO does it in its hooks), and also the SDK should restart all the IO workers.

At the moment, the only worker thread that the library controls is the one that in charge of the transactions cleanup, but later once the public API will hide ASIO context, it will become in charge of the restarting IO threads.

In case of the wrappers, that use core API directly, they must be able to stop all IO threads before fork(), wait for work completion, then perform the syscall and restart IO threads and contexts in both child and parent later. See example in test/test_integration_examples.cxx

### DIFF
--- a/core/impl/cluster.cxx
+++ b/core/impl/cluster.cxx
@@ -234,6 +234,13 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
                              });
     }
 
+    void notify_fork(fork_event event)
+    {
+        if (transactions_) {
+            transactions_->notify_fork(event);
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


avsej:
jenkins, test this please

thejcfactor (approved):
LGTM 👍

PR #542, opened 2024-03-11
CXXCBC-482: Range scan orchestrator should use best effort retry strategy by default

DemetrisChr:
The retry strategy is unset instead of being set to best-effort by default.

This has been found via the custom transcoder FIT tests which create a collection and immediately perform a range scan on them, where the SDK fast-fails instead of retrying the get_collection_id operation (which uses the retry strategy specified in the orchestrator).


avsej (approved):

avsej:
```
$ grep 'data from Mozilla' cacert.pem
## Certificate data from Mozilla as of: Mon Mar 11 15:25:27 2024 GMT
```
looks like mozilla just updated certificate bundles, and  we were unlucky to pickup some stale caches. I believe it should be fixed when restarted

PR #543, opened 2024-03-14
Update dependencies

avsej:


couchbase-sdk-cxx-black-duck-manifest.yaml
```diff
@@ -2,13 +2,13 @@
 components:
   boringssl:
     bd-id: 1e303b15-b82b-4a4d-8cd6-a2048fb2c73d
-    versions: [ 20200921 ]
+    versions: [ 20240201 ]
```
avsej, line 5: Here I'm not sure. The old one `20200921` did not reflect the actual version being used, but rather some rather recent version existing in BlackDuck database. Maybe we should rather ask them to add this one. It is being used in current grpc repo:

https://github.com/grpc/grpc/blob/db51a3f69ad6de2d4cce0f7642b829e0ebf6909e/bazel/grpc_deps.bzl#L246-L247

tools/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 if(NOT TARGET CLI11::CLI11)
+  # https://github.com/CLIUtils/CLI11/releases
   cpmaddpackage(
     NAME
     cli11
     VERSION
-    2.3.2
+    2.4.1
```
avsej, line 7: With GA we are going to package tools too, so CLI parser also have to be tracked, but we don't have BD component yet.

couchbase-sdk-cxx-black-duck-manifest.yaml
thejcfactor, line 5: I don't think we want to change this value.  See CBD-5677 for details outlining why we chose the previous value (20200921).  It has to do w/ what is in the blackduck knowledge base.  I think if we want to try and change we need to make sure the BD scans are okay.

thejcfactor (commented):
I would be fine w/ submitting this...checking the BD scans and making updates accordingly.  We could also check w/ the build team to see if there are any updates for BoringSSL.  I don't think there has been any progress on my request to get info back from BD on updating their knowledge base.

couchbase-sdk-cxx-black-duck-manifest.yaml
thejcfactor: Oh -- sorry!  I didn't see this comment.  I think we either need to follow-up w/ the build team (re: CBD-5677) or maybe we can merge the changes and see how the BD scan goes?  We can just be sure to update the yaml prior to the next release.

avsej: I have extracted BoringSSL update into separate PR: https://github.com/couchbaselabs/couchbase-cxx-client/pull/544

avsej: I have updated ticket to point to https://github.com/couchbaselabs/couchbase-cxx-client/pull/544

avsej:
@thejcfactor BD KB has all versions except BoringSSL. So I extracted it into separate PR.

thejcfactor (approved):
LGTM 👍

PR #544, opened 2024-03-18
update BoringSSL

avsej:


avsej:
Not clear what would the BoringSSL upgrade process look like. Turning the PR to Draft

PR #548, opened 2024-03-27
CXXCBC-489: Support scoped eventing functions in the core API

DemetrisChr:
Adding optional fields `bucket_name` and `scope_name` to all eventing management requests in the core API to support the new `function_scope` property in eventing.

These fields are used to add the corresponding URI parameters to the request where applicable and to add the `function_scope` field in the body of a function upsert request. The `function_scope` itself is an internal field but it is used for filtering the results of `get_all_functions` and `get_status`, as the Eventing Service does not provide scoped endpoints for those.


core/operations/management/eventing_get_status.cxx
```diff
@@ -48,7 +49,33 @@ eventing_get_status_request::make_response(error_context::http&& ctx, const enco
...
+                include = function.internal.bucket_name.has_value() && function.internal.scope_name.has_value() &&
+                          function.internal.bucket_name.value() == bucket_name.value() &&
+                          function.internal.scope_name.value() == scope_name.value();
+            } else {
+                include = (!function.internal.bucket_name.has_value() && !function.internal.scope_name.has_value()) ||
+                          (function.internal.bucket_name.has_value() && function.internal.scope_name.has_value() &&
+                           function.internal.bucket_name.value() == "*" && function.internal.scope_name.value() == "*");
+            }
+            if (include) {
+                CB_LOG_CRITICAL("Including function {} at {} / {}",
```
avsej, line 66: Lets use CB_LOG_TRACE, or don't use logging here at all. After all we are logging the thing which is accessible in the response. If necessary, the caller can log it for us

test/CMakeLists.txt
```diff
@@ -17,6 +16,8 @@ integration_test(meter)
 integration_test(transcoders)
 integration_test(range_scan)
 integration_test(search)
+integration_test(management)
+integration_test(management_eventing)
```
avsej, line 20: 👍 for starting split process

avsej (changes requested):
With logging fixed, looks good.

core/operations/management/eventing_get_status.cxx
DemetrisChr: Oops, yes, this was for debugging only, it wasn't supposed to stay

DemetrisChr: Removed 👍

avsej (approved):

PR #557, opened 2024-04-11
CXXCBC-499 Use couchbase::error for Query, Search and Analytics operations

Matt-Woz:
Changes:

- Convert query, search & analytics to use couchbase::error
- Add assign & as() methods for JSON conversion for analytics, search, query (and http) error contexts
- Add make_error util methods
- Add as() method to error context

### DIFF
--- a/core/impl/cluster.cxx
+++ b/core/impl/cluster.cxx
@@ -360,6 +360,26 @@ cluster::diagnostics(const couchbase::diagnostics_options& options) const -> std
 }
 
 void
+cluster::search_query(std::string index_name,
+                      const class search_query& query,
+                      const search_options& options,
+                      search_handler&& handler) const
+{
+    return impl_->search_query(std::move(index_name), query, options.build(), std::move(handler));
+}
+
+auto
+cluster::search_query(std::string index_name, const class search_query& query, const search_options& options) const
+  -> std::future<std::pair<error, search_result>>
+{
+    auto barrier = std::make_shared<std::promise<std::pair<error, search_result>>>();
+    search_query(std::move(index_name), query, options, [barrier](auto ctx, auto result) mutable {
+        barrier->set_value(std::make_pair(std::move(ctx), std::move(result)));
+    });
+    return barrier->get_future();
+}
+
+void
 cluster::search(std::string index_name, search_request request, const search_options& options, search_handler&& handler) const
 {
     return impl_->search(std::move(index_name), std::move(request), options.build(), std::move(handler));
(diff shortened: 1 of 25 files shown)


couchbase/retry_reason.hxx
```diff
@@ -99,6 +101,9 @@ enum class retry_reason {
     views_no_active_partition,
 };
 
+auto
+retry_reason_to_enum(const std::string& reason) -> couchbase::retry_reason;
```
DemetrisChr, line 105: Do we need to expose this in the Public API?

core/impl/error.hxx
```diff
@@ -0,0 +1,43 @@
...
+make_error(const core::error_context::search& core_ctx);
+
+error
+make_error(const core::error_context::analytics& core_ctx);
+
+error
+make_error(const core::error_context::http& core_ctx);
+
+error
+make_error(core::error_context::http core_ctx, std::optional<std::error_code>);
```
DemetrisChr, line 42: Is this needed? Could `make_error(const core::error_context::http& core_ctx)` be used instead?

couchbase/query_options.hxx
```diff
@@ -567,5 +568,5 @@ struct query_options : public common_options<query_options> {
  * @since 1.0.0
  * @uncommitted
  */
-using query_handler = std::function<void(couchbase::query_error_context, query_result)>;
+using query_handler = std::function<void(couchbase::error, query_result)>;
```
DemetrisChr, line 571: NIT: I think we could just use `error` without the full namespace since we're already inside the `couchbase` namespace.

core/impl/error.hxx
Matt-Woz: It's only for watch_indexes where we overwrite the error code, but we can move that logic up to keep this cleaner

couchbase/retry_reason.hxx
Matt-Woz: no

core/impl/cluster.cxx
```diff
@@ -297,11 +300,13 @@ cluster::query(std::string statement, const query_options& options, query_handle
...
 
 auto
-cluster::query(std::string statement, const query_options& options) const -> std::future<std::pair<query_error_context, query_result>>
+cluster::query(std::string statement, const query_options& options) const -> std::future<std::pair<error, query_result>>
 {
-    auto barrier = std::make_shared<std::promise<std::pair<query_error_context, query_result>>>();
+    auto barrier = std::make_shared<std::promise<std::pair<error, query_result>>>();
     auto future = barrier->get_future();
-    query(std::move(statement), options, [barrier](auto ctx, auto result) { barrier->set_value({ std::move(ctx), std::move(result) }); });
+    query(std::move(statement), options, [barrier](auto ctx, auto result) {
```
DemetrisChr, line 307: It would be good to rename any references to `ctx` where it is a `couchbase::error` to `err`

DemetrisChr (approved):

PR #560, opened 2024-04-15
CXXCBC-498 Use couchbase:error for management operations

Matt-Woz:
Changes:
- Convert public API management operations to use couchbase::error


couchbase/analytics_index_manager.hxx
```diff
@@ -343,7 +342,7 @@ class analytics_index_manager
      * @committed
      */
     [[nodiscard]] auto get_pending_mutations(const get_pending_mutations_analytics_options& options) const
-      -> std::future<std::pair<couchbase::manager_error_context, std::map<std::string, std::map<std::string, std::int64_t>>>>;
+      -> std::future<std::pair<couchbase::error, std::map<std::string, std::map<std::string, std::int64_t>>>>;
```
DemetrisChr, line 345:
```suggestion
      -> std::future<std::pair<error, std::map<std::string, std::map<std::string, std::int64_t>>>>;
```

core/impl/collection_manager.cxx
```diff
@@ -23,6 +23,8 @@
 #include "core/operations/management/scope_create.hxx"
 #include "core/operations/management/scope_drop.hxx"
 #include "core/operations/management/scope_get_all.hxx"
+
+#include "core/impl/error.hxx"
 #include "internal_manager_error_context.hxx"
```
DemetrisChr, line 28: Is this include needed?

core/impl/query_index_manager.cxx
```diff
@@ -135,7 +117,7 @@ class watch_context : public std::enable_shared_from_this<watch_context>
         return timeout_ - std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start_time_);
     }
 
-    bool check(const couchbase::core::operations::management::query_index_get_all_response& resp)
+    bool check(couchbase::core::operations::management::query_index_get_all_response resp)
```
DemetrisChr, line 120: Why is this not a const reference anymore?

core/impl/analytics_index_manager.cxx
```diff
@@ -682,12 +711,14 @@ analytics_index_manager::get_pending_mutations(const get_pending_mutations_analy
 
 auto
 analytics_index_manager::get_pending_mutations(const get_pending_mutations_analytics_options& options) const
-  -> std::future<std::pair<couchbase::manager_error_context, std::map<std::string, std::map<std::string, std::int64_t>>>>
+  -> std::future<std::pair<couchbase::error, std::map<std::string, std::map<std::string, std::int64_t>>>>
```
DemetrisChr, line 714:
```suggestion
  -> std::future<std::pair<error, std::map<std::string, std::map<std::string, std::int64_t>>>>
```

core/impl/analytics_index_manager.cxx
```diff
@@ -682,12 +711,14 @@ analytics_index_manager::get_pending_mutations(const get_pending_mutations_analy
...
 {
-    auto barrier = std::make_shared<
-      std::promise<std::pair<couchbase::manager_error_context, std::map<std::string, std::map<std::string, std::int64_t>>>>>();
+    auto barrier =
+      std::make_shared<std::promise<std::pair<couchbase::error, std::map<std::string, std::map<std::string, std::int64_t>>>>>();
```
DemetrisChr, line 717:
```suggestion
      std::make_shared<std::promise<std::pair<error, std::map<std::string, std::map<std::string, std::int64_t>>>>>();
```

core/impl/collection_manager.cxx
DemetrisChr: Should we be able to completely remove `internal_manager_error_context` now?

core/impl/query_index_manager.cxx
Matt-Woz: Because finish() takes the resp as a reference, and we are modifying its error code there

core/impl/collection_manager.cxx
Matt-Woz: I think the include is needed for the core::impl::make_error methods.

Yes, let's remove the internal_manager_error_context and the underlying manager_error_context too. We should do this for the other operations which aren't used in the core too, but we can do that outside of this PR.

DemetrisChr: Yes, makes sense. I was referring to the "internal_manager_error_context.hxx" include above

core/impl/query_index_manager.cxx
DemetrisChr: Could we make this a non-const reference instead, to avoid making a copy of the response when `check()` is called?

DemetrisChr (approved):

PR #564, opened 2024-04-26
CXXCBC-500 Use couchbase::error for transactions

Matt-Woz:
Changes:
- Changed transactions APIs to use couchbase::error
- Made couchbase::error polymorphic & extended by transaction_operation_failed

### DIFF
--- a/core/error_context/query_public_json.hxx
+++ b/core/error_context/query_public_json.hxx
@@ -1,0 +1,87 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2024. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+#include <couchbase/fmt/retry_reason.hxx>
+#include <couchbase/query_error_context.hxx>
+
+#include <fmt/format.h>
+#include <tao/json/forward.hpp>
+
+namespace tao::json
+{
+template<>
+struct traits<couchbase::query_error_context> {
+    template<template<typename...> class Traits>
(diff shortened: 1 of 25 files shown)


test/test_transaction_public_blocking_api.cxx
```diff
@@ -499,7 +503,6 @@ TEST_CASE("transactions public blocking API: some query errors are seen immediat
       [](couchbase::transactions::attempt_context& ctx) {
           auto [e, res] = ctx.query("I am not a valid n1ql query");
           CHECK(e.ec());
-          CHECK(std::holds_alternative<couchbase::query_error_context>(e.cause()));
```
DemetrisChr, line 502: Instead of removing this entirely, we should replace this with a check that `e.cause().has_value()` and/or `e.cause().value() == errc::common::parsing_failure`

core/error_context/query_public_json.hxx
```diff
@@ -0,0 +1,87 @@
...
+        if (ctx.retry_attempts()) {
+            v["retry_attempts"] = ctx.retry_attempts();
+        }
+        if (!ctx.retry_reasons().empty()) {
+            std::vector<tao::json::basic_value<Traits>> reasons{};
+            for (couchbase::retry_reason r : ctx.retry_reasons()) {
+                reasons.emplace_back(fmt::format("{}", r));
+            }
+            v["retry_reasons"] = reasons;
+        }
```
DemetrisChr, line 48: Could it be better to always include the retry attempts & reasons fields in the JSON even when they're zero/empty?

avsej:
@Matt-Woz, @DemetrisChr is this PR still in progress?

core/transactions/attempt_context_impl.cxx
```diff
@@ -1127,22 +1130,22 @@ attempt_context_impl::do_core_query(const std::string& statement,
...
         auto result = do_core_query(statement, opts, query_context);
-        return core::impl::build_transaction_query_result(result);
+        auto [ctx, res] = core::impl::build_transaction_query_result(result);
+        return std::make_pair(core::impl::make_error(ctx), res);
     } catch (const transaction_operation_failed& e) {
-        return { e.get_error_ctx(), {} };
+        return { core::impl::make_error(e), {} };
     } catch (const op_exception& qe) {
-        return { qe.ctx(), {} };
+        return { core::impl::make_error(qe.ctx()), {} };
```
Matt-Woz, line 1145: Not sure if maybe we should create a `make_error` for the op_exception (which contain the transaction_op_error_context and a cause), as right now just using the err context loses the `cause` field.

DemetrisChr: It looks like that might not provide any additional information to what's already in the `transaction_op_error_context`, so I think it's fine as is. Considering we were only returning `transaction_op_error_context` in the Public API previously, this shouldn't lose any information that we used to expose.

DemetrisChr (approved):

PR #565, opened 2024-05-08
CXXCBC-511: Prevent use of HTTP session if idle timer has expired

thejcfactor:


avsej (approved):

core/io/http_session_manager.hxx
```diff
@@ -153,8 +153,12 @@ class http_session_manager
                     session->start();
                     session->on_stop([type, id = session->id(), self = this->shared_from_this()]() {
                         std::scoped_lock lock(self->sessions_mutex_);
-                        self->busy_sessions_[type].remove_if([&id](const auto& s) { return !s || s->id() == id; });
-                        self->idle_sessions_[type].remove_if([&id](const auto& s) { return !s || s->id() == id; });
+                        self->busy_sessions_[type].remove_if([&id](const auto& s) {
```
brett19, line 156: It looks like this is a code-formatter difference between @avsej and @thejcfactor ?  Did anything behavioural change here?

core/io/http_session.hxx
```diff
@@ -280,6 +280,15 @@ class http_session : public std::enable_shared_from_this<http_session>
         return stopped_;
     }
 
+    bool idle_timer_expired() const
+    {
+        auto expired = idle_timer_.expiry() <= asio::steady_timer::clock_type::now();
```
brett19, line 285: Are you sure this is sufficient to prevent the race?  I imagine there is a case where this check passes, but the idle-timer still ends up being invoked 'on the next line'.  Rather than relying on timing here, why not mark the session atomically with whether its in-use, or cancel the idle timer after a session is selected, where failure to cancel causes it to retry selecting?

core/io/http_session_manager.hxx
thejcfactor: Should be resolved.  Although need to see why the formatting is different.

core/io/http_session.hxx
thejcfactor: Have updated the logic to retry selecting a session if cancelling the idle timer is not successful.

core/io/http_session.hxx
```diff
@@ -349,13 +349,14 @@ class http_session : public std::enable_shared_from_this<http_session>
             if (ec == asio::error::operation_aborted) {
                 return;
             }
+            CB_LOG_DEBUG("{} idle timeout expired, stopping session: \"{}:{}\"", self->info_.log_prefix(), self->hostname_, self->service_);
             self->stop();
         });
     }
 
-    void reset_idle()
+    std::size_t reset_idle()
```
brett19, line 357: Since http_session internally manages the idle_timer, you should probably do the conversion from 'arbitrary number' to boolean here, where the context for that number is known.

core/io/http_session_manager.hxx
```diff
@@ -204,32 +204,42 @@ class http_session_manager
...
         idle_sessions_[type].remove_if([](const auto& s) { return !s; });
         busy_sessions_[type].remove_if([](const auto& s) { return !s; });
-        if (idle_sessions_[type].empty()) {
+        std::shared_ptr<http_session> session{};
+        while (!idle_sessions_[type].empty()) {
+            if (preferred_node.empty()) {
+                session = idle_sessions_[type].front();
+                idle_sessions_[type].pop_front();
+                if (session->reset_idle() != 0) {
+                    break;
```
brett19, line 213: I think this should be continue instead so that it goes back and tries again?

core/io/http_session.hxx
thejcfactor: Fair -- I will make that change.

core/io/http_session_manager.hxx
thejcfactor: Hmmm, if `reset_idle()` is returning 0 that would mean `cancel()` was unable to cancel any operations.  That is the "failure" condition in which we want to try and pick up another session, right?

thejcfactor: I think this resolved now.

avsej (approved):

brett19 (approved):

PR #566, opened 2024-05-09
CXXCBC-509: Zone-Aware replica reads

avsej:


avsej:
jenkins test this please

avsej:
jenkins test this please

couchbase/read_preference.hxx
```diff
@@ -0,0 +1,27 @@
...
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+namespace couchbase
+{
+enum class read_preference {
+    no_preference,
+    selected_server_group,
```
Matt-Woz, line 24: Should we add a codedoc here (or maybe in the respective option setters) about where the selected server group can be configured? i.e.`cluster_options::network()::preferred_server_group()` (or `server_group_affinity()` going off the discussion in the RFC?)

avsej: I will add docs with examples.

core/cluster.cxx
thejcfactor, line 1: Should we also change the `execute()` method to open the bucket for `lookup_in_all_replicas_request` and `lookup_in_any_replica_request`?

thejcfactor (commented):
Just had a small question...looks good otherwise.

core/cluster.cxx
avsej: Those operation handle it inside execute() method. It was added there before, I just found that regular get operations are sensitive to it and fixed it here without refactoring working code for lookup_in*.

I've opened ticket to track the unification, so it won't get lost: https://issues.couchbase.com/browse/CXXCBC-515

thejcfactor: Ahh -- fair.  I was seeing the `core->with_bucket_configuration` and not the `core->open_bucket` prior in the `lookup_in*` requests.  The CXXCBC to log the effort works.  👍

thejcfactor (approved):

PR #568, opened 2024-05-21
CXXCBC-445: return request_canceled on IO error in HTTP session

avsej:


core/io/http_session.hxx
```diff
@@ -441,7 +444,9 @@ class http_session : public std::enable_shared_from_this<http_session>
             return;
         }
         if (deadline_timer_.expiry() <= asio::steady_timer::clock_type::now()) {
-            stream_->close([](std::error_code) {});
+            cancel_current_response(couchbase::errc::common::ambiguous_timeout);
```
DemetrisChr, line 447: Should this be an `unambiguous_timeout` instead? My understanding is that the `deadline_timer_` is only used for the connect timeout, so that would mean that if we ended up here we can be certain that no data was modified on the server.

Though I'm a little confused by `check_deadline`. Isn't it only called when the `deadline_timer_` expires which would mean that it is always the case that `deadline_timer_.expiry() <= asio::steady_timer::clock_type::now()` ?

avsej: @DemetrisChr, thank you. You are right. I updated this handler, moved it closer to the place where we start connection attempt, fixed the error code and removed extra check.

PR #575, opened 2024-05-29
CXXCBC-514 Support for base64 encoded vector types

Matt-Woz:
Changes

- Added support for base64 encoded vector query
- Bumped stability up to uncommitted


avsej (changes requested):
Looks good, but looks like you forgot to reference the snippet you tagged in the test code.

avsej (approved):

PR #589, opened 2024-06-13
CXXCBC-517:  Add HTTP session retries

thejcfactor:
Motivation
==========
The client does not currently attempt to retry if the HTTP session fails to resolve or connect.  Adding retries will help to improve user experience.

Changes
=======
* Add retries if resolve or connect fails.
* If connect fails, and resolve provided multiple endpoints, attempt to connect to next endpoint in list of provided endpoints.
* If ping() fails due to resolve/connect issue, list "attempted" hostname and port in `remote_address` property of `endpoint_ping_info`

### DIFF
--- a/core/io/http_command.hxx
+++ b/core/io/http_command.hxx
@@ -127,15 +127,25 @@ struct http_command : public std::enable_shared_from_this<http_command<Request>>
     deadline.cancel();
   }
 
-  void send_to(std::shared_ptr<io::http_session> session)
+  void send_to()
   {
     if (!handler_) {
       return;
     }
-    session_ = std::move(session);
     if (span_->uses_tags())
       span_->add_tag(tracing::attributes::local_id, session_->id());
     send();
+  }
+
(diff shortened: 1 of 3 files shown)


thejcfactor:
I will update CXXCBC-517 tomorrow w/ details from my testing.

core/io/http_command.hxx
```diff
@@ -127,17 +129,27 @@ struct http_command : public std::enable_shared_from_this<http_command<Request>>
...
 
+  void set_command_session(std::shared_ptr<io::http_session> session)
+  {
+    session_.reset();
+    session_ = std::move(session);
+  }
+
+  [[nodiscard]] auto deadline_expired() const -> bool
+  {
+    return deadline_expired_;
```
avsej, line 150: will it work with just checking the time?
```cpp
return deadline.expiry() < std::chrono::steady_clock::now();
```

thejcfactor: Agreed.  Done in latest patch. 👍

PR #590, opened 2024-06-13
CXXCBC-381: use shared_ptr for transactions_context

avsej:


DemetrisChr:
We might need to define a feature macro for this so we can keep compatibility with previous versions in the FIT performer

PR #591, opened 2024-06-13
CXXCBC-81: fix upsert for locked key

avsej:


avsej:
Closing as invalid bug

PR #594, opened 2024-06-17
CXXCBC-526: Return couchbase::error from txn lambda

DemetrisChr:


examples/minimal.cxx
```diff
@@ -62,6 +62,31 @@ main()
...
+  //          });
+  //        });
+  //      },
+  //      [barrier](auto err, auto res) {
+  //        barrier->set_value({err, res});
+  //      },
+  //      {}
+  //      );
+  //    barrier->get_future();
+  //  }
```
avsej, line 89: Was this temporary change? Static analyzer complains

image

DemetrisChr: Oops, removed

avsej (approved):

PR #616, opened 2024-07-03
CXXCBC-547: Add initial columnar query core implementation

DemetrisChr:
## Motivation

Add core implementation for columnar queries

## Changes

* Add a columnar agent, providing access to all columnar-related operations. The columnar agent is the equivalent of "cluster agent" in the columnar side of things.
* Add a columnar query component, that supports executing queries and streaming the results.
* Add a high level row_streamer interface, that can be reused for different operations that stream rows (e.g. FTS, query, analytics). The row streamer allows reading the body header before returning from the operation i.e. the part of the body before the rows JSON array.
* Add an http_streaming_response that can wrap the http_session and provide access to the body using pull-based streaming. It also makes it possible to stop the HTTP session mid-stream. Add support in http_session for returning the http_streaming_response.

## Results

All tests pass

Further work is required to properly handle errors, return richer error information & add support for retries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,6 +371,7 @@ set(couchbase_cxx_client_FILES
     core/range_scan_options.cxx
     core/range_scan_orchestrator.cxx
     core/retry_orchestrator.cxx
+    core/row_streamer.cxx
     core/scan_result.cxx
     core/search_query_options.cxx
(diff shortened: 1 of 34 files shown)


avsej (approved):
Looks good to me. Lets merge it and continue

PR #620, opened 2024-07-06
Clang tidy fixes

avsej:


couchbase/analytics_index_manager.hxx
```diff
@@ -91,7 +91,7 @@ public:
    * @since 1.0.0
    * @committed
    */
-  void drop_dataverse(std::string dataverse_name,
+  void drop_dataverse(const std::string& dataverse_name,
```
DemetrisChr, line 94: Is turning `std::string` to `const std::string&` a breaking change?

avsej: Technically it is, as the compiled code will not be able to find the signature as it is using different type. But I think I'd rather make it early and right. Right now we do not ship binaries anyway yet. And the user would need to recompile it anyway.

couchbase/query_index_manager.hxx
```diff
@@ -88,7 +88,7 @@ public:
    * @since 1.0.0
    * @committed
    */
-  void create_index(std::string bucket_name,
+  void create_index(const std::string& bucket_name,
                     std::string index_name,
                     std::vector<std::string> keys,
```
DemetrisChr, line 93: Should the rest of the parameters here and in the methods below also become const references?

core/transactions/transactions_config.cxx
```diff
@@ -25,8 +25,7 @@ namespace couchbase::transactions
 {
 
 transactions_config::transactions_config()
-  : level_(couchbase::durability_level::majority)
-  , timeout_(std::chrono::seconds(15))
+  : timeout_(std::chrono::seconds(15))
```
DemetrisChr, line 28: Should this also be removed, similar to `level_`? It's being initialized to 15 seconds by default.

```c++
  std::chrono::nanoseconds timeout_{ std::chrono::seconds(15) };
```

couchbase/query_index_manager.hxx
avsej: Probably they should. While taking by value and moving it later looks more efficient, we still transform those keys into JSON at some point, and treat them as temporary values anyway, so it is kind of waste of the memory. 

So in this context, taking it by reference would be more reasonable, as we do not want to actually use this chunk of memory for a long time (as a part of request), but rather convert it into some internal representation in some front-end.

core/impl/fail_fast_retry_strategy.cxx
```diff
@@ -15,9 +15,15 @@
  *   limitations under the License.
  */
 
+#include "couchbase/retry_action.hxx"
+#include "couchbase/retry_reason.hxx"
+#include "couchbase/retry_request.hxx"
```
DemetrisChr, line 20: Should `<>` be used for these includes?

avsej: yes, in general I try to follow it as a style rule with <> for our "public API" and double quotes for the other parts, thanks for spotting it.

avsej:
jenkins, test this please

PR #621, opened 2024-07-09
CXXCBC-525: Columnar - open cluster connection in background

thejcfactor:
Motivation
==========
Columnar SDKs takes a different approach to connect. So that wrapper SDKs can still utilize the majority of the I/O functionality in the C++ client, the Columnar connect changes should be added.

Changes
=======
* Add path to allow SDKs to open and bootstrap a cluster in the background
* Add "infinite" retries when bootstrapping
* Add config_tracker
* Add functionality to track bootstrap failure/success
* Add deferred queue for HTTP requests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,11 @@ if(NOT DEFINED COUCHBASE_CXX_CLIENT_MASTER_PROJECT)
   endif()
 endif()
 
+option(COUCHBASE_CXX_CLIENT_COLUMNAR "Build with Columnar additions" FALSE)
+if(COUCHBASE_CXX_CLIENT_COLUMNAR)
+  message(STATUS "COUCHBASE_CXX_CLIENT_COLUMNAR=${COUCHBASE_CXX_CLIENT_COLUMNAR} building with Columnar additions.")
+endif()
+
 project(
   couchbase_cxx_client
   VERSION "1.0.0"
@@ -226,6 +231,7 @@ set(couchbase_cxx_client_FILES
     core/io/mcbp_message.cxx
     core/io/mcbp_parser.cxx
     core/io/mcbp_session.cxx
+    core/io/config_tracker.cxx
     core/key_value_config.cxx
     core/management/analytics_link_azure_blob_external.cxx
     core/management/analytics_link_couchbase_remote.cxx
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -15,7 +15,7 @@
  *   limitations under the License.
  */
 
-#include "couchbase/build_config.hxx"
+#include <couchbase/build_config.hxx>
 
 #include "cluster.hxx"
 
@@ -24,6 +24,9 @@
 #include "core/impl/get_replica.hxx"
 #include "core/impl/lookup_in_replica.hxx"
 #include "core/impl/observe_seqno.hxx"
+#ifdef COUCHBASE_CXX_CLIENT_COLUMNAR
+#include "core/io/config_tracker.hxx"
+#endif
 #include "core/io/http_command.hxx"
 #include "core/io/http_session_manager.hxx"
 #include "core/io/mcbp_command.hxx"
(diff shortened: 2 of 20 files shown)


CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ if(NOT DEFINED COUCHBASE_CXX_CLIENT_MASTER_PROJECT)
   endif()
 endif()
 
+option(COUCHBASE_CXX_CLIENT_COLUMNAR "Build with Columnar additions" FALSE)
```
avsej, line 17: We need CI build with this option to track regressions during future changes.

avsej (commented):
In general look. I would merge it if tests pass and CI had some verification for `COUCHBASE_CXX_CLIENT_COLUMNAR=ON` in the matrix

CMakeLists.txt
thejcfactor: Agreed -- added a path to build Columnar 'things' in the latest patchset.

core/cluster.cxx
```diff
@@ -276,6 +289,64 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+    } else {
+      if (origin_.options().enable_metrics) {
+        meter_ = std::make_shared<metrics::logging_meter>(ctx_, origin_.options().metrics_options);
+      } else {
+        meter_ = std::make_shared<metrics::noop_meter>();
+      }
+    }
+    meter_->start();
+    session_manager_->set_tracer(tracer_);
+    session_manager_->set_dispatch_timeout(origin_.options().dispatch_timeout);
```
DemetrisChr, line 338: We might need to also set the dispatch timeout in the standard `open()` as well. Right now, if you build the SDK with `COUCHBASE_CXX_CLIENT_COLUMNAR` but use `open` instead of `open_in_background` all operations time out because the dispatch timeout is 0.

core/http_component.cxx
```diff
@@ -182,20 +258,91 @@ class http_component_impl
...
 
+    if (!session->is_connected()) {
+      session_manager->connect_then_send_pending_op(
+        session,
+        {},
+#ifdef COUCHBASE_CXX_CLIENT_COLUMNAR
+        op->dispatch_deadline_expiry(),
+#endif
+        op->deadline_expiry(),
+        [op = std::move(op)](std::error_code ec, std::shared_ptr<io::http_session> http_session) {
```
DemetrisChr, line 276: `op` shouldn't be moved here, since we are later returning it from `do_http_request` - we can just copy it instead

core/http_component.cxx
```diff
@@ -182,20 +258,91 @@ class http_component_impl
...
+    if (session_manager->last_bootstrap_error().has_value()) {
+      return tl::unexpected(session_manager->last_bootstrap_error().value().ec);
+    }
+    pending_op->start([callback = std::move(callback)](auto resp, auto ec) mutable {
+      callback(std::move(resp), ec);
+    });
+    // TODO:  client_context_id?
+    CB_LOG_DEBUG(R"(Adding pending HTTP operation to deferred queue: {})",
+                 pending_op->request().service);
+    session_manager->add_to_deferred_queue([op = std::move(pending_op),
```
DemetrisChr, line 305: Similar here, `pending_op` shouldn't be moved here as it's being returned from this method.

core/cluster.cxx
thejcfactor: Hmmm, interesting.  Why would someone build w/ `COUCHBASE_CXX_CLIENT_COLUMNAR` and _not_ use `open_in_background()`?  That would violate (seems like maybe a harsh word 😆) the RFC right?

core/http_component.cxx
thejcfactor: Ahh, thanks!  Will update 👍

core/http_component.cxx
thejcfactor: Ahh, thanks!  Will update 👍

core/cluster.cxx
DemetrisChr: In my opinion, the code hidden behind `COUCHBASE_CXX_CLIENT_COLUMNAR` should be an extension to the existing core, so building with it ideally should not break existing non-columnar behaviour. I think if we don't want `open` to be used with `COUCHBASE_CXX_CLIENT_COLUMNAR` maybe instead of having a separate `open_in_background` we should just have two versions of `open` with `#ifdef #else #endif`?

The reason I raised this is because our testing setup tries to do `cluster_describe` request in order to get the server version etc., and that fails with a timeout because of this at the moment.

avsej: Lets think about it as "experimental" code for early preview. Now we have CI checking that both versions are buildable, and after merge we can iterate on the test coverage with the final goal of fusing all code paths into single version.

avsej (approved):

PR #622, opened 2024-07-09
configure cppcheck

avsej:


core/transactions/forward_compat.cxx
```diff
@@ -97,13 +100,14 @@ struct forward_compat_protocol_requirement : public forward_compat_requirement {
 struct forward_compat_extension_requirement : public forward_compat_requirement {
   std::string extension_id;
 
-  forward_compat_extension_requirement(forward_compat_behavior_full b, const std::string ext)
+  forward_compat_extension_requirement(forward_compat_behavior_full b, std::string ext)
     : forward_compat_requirement(b)
-    , extension_id(ext)
+    , extension_id(std::move(ext))
```
avsej, line 105: @DemetrisChr on the other hand, in case like this I see reason to allow accepting parameter by value. We are going to use it verbatim later, and copying it into the class instance.

avsej:
jenkins, test this please

PR #624, opened 2024-07-11
Fix behaviour when reading is complete before returning HTTP streaming resp

DemetrisChr:
## Motivation

If the entire HTTP body has been read in the initial read before returning the HTTP streaming response, any subsequent attempt to read from the TCP socket will never return.

## Changes

* Set `reading_complete` when initializing `http_streaming_response_body` with the value of `http_streaming_parser`'s `complete` field.
* Call the `stream_end_handler` when the entire HTTP message has been read successfully (not only on cancellation).

### DIFF
--- a/core/io/http_session.cxx
+++ b/core/io/http_session.cxx
@@ -681,7 +681,7 @@ http_session::do_read()
             std::scoped_lock lock(self->current_response_mutex_);
             std::swap(self->current_streaming_response_, ctx);
           } else {
-            if (auto handler = std::move(ctx.stream_end_handler)) {
+            if (auto handler = std::move(ctx.stream_end_handler); handler) {
               handler();
             }
           }
--- a/core/io/http_session.cxx
+++ b/core/io/http_session.cxx
@@ -501,6 +501,9 @@ http_session::read_some(
           std::scoped_lock lock(self->current_response_mutex_);
           std::swap(self->current_streaming_response_, ctx);
         }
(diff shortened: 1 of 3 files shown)


core/io/http_session.cxx
```diff
@@ -501,6 +501,9 @@ http_session::read_some(
           std::scoped_lock lock(self->current_response_mutex_);
           std::swap(self->current_streaming_response_, ctx);
         }
+        if (ctx.stream_end_handler) {
```
avsej, line 504: could you use the move idiom here?

core/io/http_session.cxx
```diff
@@ -677,6 +680,10 @@ http_session::do_read()
           if (!res.complete) {
             std::scoped_lock lock(self->current_response_mutex_);
             std::swap(self->current_streaming_response_, ctx);
+          } else {
+            if (auto handler = std::move(ctx.stream_end_handler)) {
```
avsej, line 684:
```suggestion
            if (auto handler = std::move(ctx.stream_end_handler); handler) {
```

core/io/http_session.cxx
DemetrisChr: The reason I didn't use it is that we're already moving the whole of `ctx` with `std::swap` and then we just let it get out of scope, it's not used later. Is it still necessary?

avsej: makes sense, you right

PR #634, opened 2024-07-17
Improve logging of DNS client

avsej:


core/io/dns_client.cxx
```diff
@@ -155,7 +156,6 @@ class dns_srv_command : public std::enable_shared_from_this<dns_srv_command>
                    self->address_.to_string(),
                    self->port_);
       self->udp_.cancel();
-      return self->retry_with_tcp();
```
avsej, line 158: Removed this retry as it would retry in the UDP read handler above

thejcfactor (approved):
LGTM 👍

PR #641, opened 2024-08-13
CXXCBC-569: resolve cycle in shared pointers for transaction_context

avsej:
The attempt object does not own the transaction context, so it should use weak_ptr instead.


avsej:
rerun tests

avsej:
rerun tests

DemetrisChr (approved):

PR #648, opened 2024-08-23
CXXCBC-576: Ensure all HTTP sessions are stopped when closing the core cluster

DemetrisChr:
## Motivation

When `cluster.close()` is called, all in-progress HTTP operations should be cancelled.

## Changes

* Add `pending_sessions_` in the HTTP session manager to ensure that it has a reference to sessions created by it at all stages of their lifetime. Pending operations are the sessions that are currently connecting (i.e. after they have been removed from `idle_sessions_` or created and before being added to `busy_sessions_`)
* Fix issue where the callback was not called in the case of a timeout (dispatch timeout or overall timeout) while a session is connecting.
* In an HTTP session that is connecting, call the connect callback if it is stopped before it has connected.
* Add the `couchbase::core::columnar::client_errc` error codes to represent errors that are not the result of a client-server interaction. These are the errors that should be exposed in a platform-idiomatic way in the wrappers, not as 
`ColumnarError`s - one example is cancellation which happens on the client-side.
* In `pending_query_operation` replace the use of `columnar::errc::generic` for cancellations with the new `columnar::client_errc::canceled`
* When `http_session_manager.close()` is called, force all the sessions owned by the manager to stop, not just discarding the pointers.
* Fix seg fault when accessing `retry_info_` for a `pending_query_operation` where the session wrapped by it is `nullptr` (e.g. in the case of a dispatch timeout)
* Add tests for closing the cluster before columnar `execute_query` has returned & for closing the cluster while rows are being iterated
* Add test for the dispatch timeout with a columnar query

## Results

All tests pass

### DIFF
--- a/core/columnar/error_codes.cxx
+++ b/core/columnar/error_codes.cxx
@@ -27,7 +27,7 @@ namespace couchbase::core::columnar
 struct columnar_error_category : std::error_category {
   [[nodiscard]] auto name() const noexcept -> const char* override
   {
-    return "couchbase.core.columnar";
+    return "couchbase.core.columnar.errc";
   }
 
   [[nodiscard]] auto message(int ev) const noexcept -> std::string override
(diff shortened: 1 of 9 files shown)


core/columnar/error_codes.cxx
```diff
@@ -17,6 +17,7 @@
 
 #include <couchbase/error_codes.hxx>
 
+#include "core/logger/logger.hxx"
```
thejcfactor, line 20: Is this include needed?

core/columnar/query_component.cxx
```diff
@@ -85,24 +83,19 @@ class pending_query_operation
           const std::scoped_lock lock{ self->pending_op_mutex_ };
           std::swap(op, self->pending_op_);
         }
+        if (ec) {
```
thejcfactor, line 86: Thanks for making this change 😄

test/test_integration_columnar_query.cxx
```diff
@@ -423,7 +451,7 @@ TEST_CASE("integration: columnar query collection does not exist")
   couchbase::core::columnar::agent agent{ integration.io, { { integration.cluster } } };
 
   couchbase::core::columnar::query_options options{ "SELECT * FROM `does-not-exist`" };
-  options.timeout = std::chrono::seconds(20);
+  options.timeout = std::chrono::seconds(60);
```
thejcfactor, line 454: Just curious: does the added seconds have an impact? At first I thought it might be so that it is greater than the dispatch timeout, but that shouldn't have to come into play here I don't think.

thejcfactor (commented):
I intend to play w/ this change a bit more in the Python Columnar client, but I think all looks well.  I just had a couple of comments.

test/test_integration_columnar_query.cxx
DemetrisChr: There wasn't really a reason for this. I've reverted it back to 20 seconds so it's the same with other tests. We might just need to keep an eye on them if it's slower on CI (but 20 seconds should be enough)

core/columnar/error_codes.cxx
DemetrisChr: No, it's not needed, I've removed it

thejcfactor (approved):
LGTM 👍 

Did further testing w/ the Python Columnar client.  Thanks for making these changes!

avsej (approved):

PR #650, opened 2024-08-28
CXXCBC-580: Report last error when timing out on columnar query retries

DemetrisChr:
## Motivation

The Error/Retry handling RFC Revision 8 requires that we list the last error code & message that was retried when reporting a timeout

## Changes

* Include a `last_errors` field in the error context that lists the set of server errors that were retried, when the error is a timeout resulting from retries.
* FIX: Timeout should be initialized before the HTTP request in `pending_query_operation`

## Results

Tested with [FIT](https://review.couchbase.org/c/transactions-fit-performer/+/215117) (via Python) - test passed

### DIFF
--- a/core/columnar/query_component.cxx
+++ b/core/columnar/query_component.cxx
@@ -41,6 +41,7 @@
 #include <cstdint>
 #include <memory>
 #include <mutex>
+#include <string>
 #include <system_error>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


avsej:
Interesting, why it started complaining about missing `<string>` header in `core\platform\string_hex.h`?

```
    24>D:\a\couchbase-cxx-client\couchbase-cxx-client\core\platform\string_hex.cc(95,18): error C2039: 'to_hex': is not a member of 'couchbase::core' [D:\a\couchbase-cxx-client\couchbase-cxx-client\cmake-build-tests\core\platform\couchbase_platform.vcxproj]
             D:\a\couchbase-cxx-client\couchbase-cxx-client\core\platform\string_hex.h(22,11):
             see declaration of 'couchbase::core'      
```

@DemetrisChr, could you check it? and open separate PR to fix it if necessary.

It is indeed using `std::string` without `#include <string>` and so far it looks like whenever the header was included the `<string>` header was already there.

https://github.com/couchbase/couchbase-cxx-client/blob/9c445aecb739227032e6368e3ef0f9f85054973e/core/platform/string_hex.h#L35-L36

DemetrisChr:
Opened #652

thejcfactor (approved):
LGTM 👍

PR #651, opened 2024-08-29
CXXCBC-552: cleanup network selection options

avsej:
Deprecates behavior_options#network() in favor to network_options#preferred_network()

### DIFF
--- a/core/impl/cluster.cxx
+++ b/core/impl/cluster.cxx
@@ -140,7 +140,6 @@ options_to_origin(const std::string& connection_string,
   user_options.enable_mutation_tokens = opts.behavior.enable_mutation_tokens;
   user_options.enable_unordered_execution = opts.behavior.enable_unordered_execution;
   user_options.user_agent_extra = opts.behavior.user_agent_extra;
-  user_options.network = opts.behavior.network;
 
   user_options.server_group = opts.network.server_group;
   user_options.enable_tcp_keep_alive = opts.network.enable_tcp_keep_alive;
@@ -150,8 +149,10 @@ options_to_origin(const std::string& connection_string,
   if (opts.network.max_http_connections) {
     user_options.max_http_connections = opts.network.max_http_connections.value();
   }
-  if (!opts.network.network.empty()) {
-    user_options.network = opts.network.network;
+  user_options.network = opts.network.network;
+  if (user_options.network.empty()) {
+    // this is deprecated option, but use it in case main options is empty
+    user_options.network = opts.behavior.network;
   }
   switch (opts.network.ip_protocol) {
     case ip_protocol::any:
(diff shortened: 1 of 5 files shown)


couchbase/behavior_options.hxx
```diff
@@ -60,19 +60,10 @@ public:
...
-   *
-   * @see
-   * https://docs.couchbase.com/server/current/learn/clusters-and-availability/connectivity.html#alternate-addresses
-   * @see https://docs.couchbase.com/server/current/rest-api/rest-set-up-alternate-address.html
-   */
+  [[deprecated("Use couchbase::network_options#preferred_network() instead.")]]
   auto network(std::string name) -> behavior_options&
   {
-    network_ = std::move(name);
+    [[maybe_unused]] auto name_ = std::move(name);
```
DemetrisChr, line 66: Even though it's deprecated, shouldn't we still use this if it's set? (but give priority to `network_options.network` if both of them are set)

avsej: I think, you right. I will adjust the change

avsej:
@DemetrisChr I've update the patch according to your comment

core/impl/cluster.cxx
```diff
@@ -150,8 +149,10 @@ options_to_origin(const std::string& connection_string,
   if (opts.network.max_http_connections) {
     user_options.max_http_connections = opts.network.max_http_connections.value();
   }
-  if (!opts.network.network.empty()) {
-    user_options.network = opts.network.network;
+  user_options.network = opts.network.network;
+  if (user_options.network.empty()) {
```
DemetrisChr, line 153: In the Public API's `couchbase::network_options`, we initialize `network_` with "auto" by default. If someone who uses the deprecated behavior.network option, doesn't also explicitly set the network.network option to the empty string, it will get ignored. The core API `network` option is also initialized with "auto" so `user_options.network` will not be empty here even without the above.

 I think we can remove the initialization of network with "auto" in the public API (as that is already the default value in the core API) and then do:
```cpp
if (!opts.network.network.empty()) {
  user_options.network = opts.network.network
} else if (!opts.behavior.network.empty()) {
  // this is deprecated option, but use it in case main options is empty
  user_options.network = opts.behavior.network
}
```

avsej: I rather leave public API keep initializing safe value, and expect people to pass empty value if they want old setting.

DemetrisChr: That's ok, but I think we should still change the condition to something other than `user_options.network.empty()`, since `user_options.network` is initialized to "auto" in the core as well, so it will never be empty.

I think we need something similar to the if/else-if statement from my comment above.

DemetrisChr: The comment above is not actually true, since `user_options.network = opts.network.network` happens before that condition, so it will be empty if `opts.network.network` is empty

DemetrisChr (approved):

PR #653, opened 2024-08-30
Update code samples in API reference

avsej:


couchbase/collection.hxx
```diff
@@ -348,6 +348,13 @@ public:
    * @param options the custom options
    * @return the future object with first available result, might be the active or a replica.
    *
+   * Select preferred server group in connection options:
```
DemetrisChr, line 351: Should this section also be added for the callback-based overload of `get_any_replica` above?

couchbase/collection.hxx
```diff
@@ -368,6 +375,13 @@ public:
    * @note Individual errors are ignored, so you can think of this API as a best effort
    * approach which explicitly emphasises availability over consistency.
    *
+   * Select preferred server group in connection options:
+   * @snippet{trimleft} test_integration_read_replica.cxx select-preferred_server_group
+   *
+   * Restrict get_any_replica to reach nodes that belong to selected server
+   * group only:
+   * @snippet{trimleft} test_integration_read_replica.cxx get_any_replica-with-read_preference
```
DemetrisChr, line 383:
```suggestion
   * Restrict get_all_replicas to reach nodes that belong to selected server
   * group only:
   * @snippet{trimleft} test_integration_read_replica.cxx get_all_replicas-with-read_preference
```

couchbase/transactions/attempt_context.hxx
```diff
@@ -50,6 +50,16 @@ public:
    * @param id The unique id of the document.
    * @return The result of the operation, which is an @ref error and a @ref
    * transaction_get_result.
+   *
+   * The example below shows how to use custom transcoder with transactions.
+   *
+   * The type should have regular transcoder to be defined. The only important
```
DemetrisChr, line 56: > The type should have regular transcoder to be defined.

It's not very clear to me what this means. Could we phrase it differently?

avsej: I will try to rephrase it. What I meant is that the transactions are using the same transcoders as general API, just with only exception that they should either encode types as JSON or Binary with returning corresponding flags, encoding with different set of flags is not defined.

couchbase/collection.hxx
avsej: We do not have ready example for that case yet.

couchbase/transactions/attempt_context.hxx
avsej: Updated.

DemetrisChr (approved):

PR #654, opened 2024-09-05
CXXCBC-588: Update timeout sent to server on each columnar query retry

DemetrisChr:
## Changes

* Update the timeout written in the query request payload when retrying, according to the time remaining until the deadline.
* Fix a bug where not setting the timeout in the options (i.e. relying on the cluster-wide default) resulted in the query timing out immediately.
  * This is because the HTTP component's timeout would not be set, which resulted in the epoch being passed on to the http_session_manager's `connect_then_send_pending_op` as the deadline.
  * The timeout is now always sent to the HTTP component.
* Remove the dispatch timeout test which was succeeding because of the bug above. Dispatching appears to succeed very quickly, so even with a timeout 1ms the test was not consistently passing - we probably need to find another way to test it.
* Use the default timeouts in tests that don't need a specific timeout value (this was previously hiding the bug)

### DIFF
--- a/core/columnar/query_component.cxx
+++ b/core/columnar/query_component.cxx
@@ -33,6 +33,7 @@
 
 #include <asio/io_context.hpp>
 #include <asio/steady_timer.hpp>
+#include <fmt/chrono.h>
 #include <fmt/format.h>
 #include <gsl/util>
 #include <tao/json/value.hpp>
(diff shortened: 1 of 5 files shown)


core/columnar/query_component.cxx
```diff
@@ -59,6 +60,7 @@ class pending_query_operation
                           std::chrono::milliseconds default_timeout)
     : client_context_id_{ uuid::to_string(uuid::random()) }
     , timeout_{ options.timeout.value_or(default_timeout) }
+    , payload_(build_query_payload(options))
```
DemetrisChr, line 63: Using the initializer list syntax for `tao::json::value` resulted in the `error: no class template named 'default_key'` error on Linux that we've seen before - unsure why

avsej (approved):

thejcfactor (approved):
LGTM 👍

PR #656, opened 2024-09-16
CXXCBC-596: do not include tao_json_serializer.hxx by default

avsej:
@thejcfactor, @DemetrisChr, @Matt-Woz this patch will likely require depending projects (like FIT, wrappers etc) to explicitly require `tao/json.hpp` or `couchbase/codec/tao_json_serializer.hxx`

But on the other hand, it allows users to use public API without tao/json with some other serializer, leaving tao/json as only dependency of the C++SDK itself.

This patch is necessary step in producing better packages.


thejcfactor (approved):
LGTM 👍 

The removal of `rows_as_json()` in `analytics_result`/`query_result` is unfortunate, but understandable w.r.t. the change.  Arguably a worthwhile "sacrifice".

Lets just make sure we mention the change in release notes.

PR #659, opened 2024-09-16
CXXCBC-303: build jsonsl with the library

avsej:
At this stage, we rather own jsonsl implementation, and it would be easier to compile and maintain it as a part of the main source set.

The changes in the library is just result of tidy and formatter scripts, nothing really changed in the library except the build process


avsej:
I think I'd rather defer this change for later

PR #666, opened 2024-09-30
CXXCBC-598: Columnar - propagate bootstrap errors to HTTP operations

thejcfactor:
Changes
=======
* Create and error_union type of std::variant<std::monostate, std::error_code, impl::bootstrap_error> that can be passed around to Columnar callbacks
* Allow the http_session_manager drain queue to accept an error_union
* Update error message in certain contexts to indicate a bootstrap error is the reason the HTTP operation failed.
* Confirm columnar testing passes


core/columnar/query_component.cxx
```diff
@@ -127,7 +137,23 @@ class pending_query_operation
       pending_op_ = op.value();
       return {};
     }
+    retry_timer_.cancel();
```
thejcfactor, line 140: An interesting situation (you can see it easily when there is a bootstrap error, but I think it applies to any situation to where creating the pending_op failed).  The situation is that http operation's callback is still "alive" at this point.  If we don't stop the timer, you see the op "timeout" later on in logs even though we never really got the query operation "off the ground".  

Also, when testing w/ the Node.js wrapper, I noticed that even if we cancel the timers, something is still holding onto the callback b/c you can see the logging continue infinitely even after the application has exited.  This infinite logging goes away with something like `invoke_callback({}, return_error);`.

Is it bad to both invoke the callback and return the same error?

avsej: Looks good to me

avsej (approved):

DemetrisChr (approved):
LGTM

I think eventually we want to use a non-columnar equivalent of `couchbase::core::columnar::error` in the HTTP component & when draining the deferred queue in the session manager (and in the core API more broadly). That might allow us to remove the error union from the HTTP operation signature in the agent, which could be cleaner. But that's something to consider later on as it would be a bigger change.

thejcfactor:
Yes!  I agree. At first I tried to work through a more generic approach, but realized how pervasive the change would need to be so I reverted to something I thought would be more pointed and less evasive.

PR #667, opened 2024-10-03
CXXCBC-602: Columnar - report first non-retriable code/message in query error

DemetrisChr:
## Motivation

Columnar Error/Retry RFC revision 9:
> If the list of errors in a query response contains at least one non-retriable error, the first non-retriable error in the array is selected. Otherwise, the first retriable error in the array is selected. 

## Changes

* Report the first non-retriable error in the columnar query error properties, if it exists. Otherwise, report the first error.

## Results

FIT tests pass (Tested via the Python SDK)


thejcfactor (approved):
LGTM 👍

PR #673, opened 2024-10-09
CXXCBC-606: Fix detection of disfunctional node

avsej:
* remove `reopen()` API of the stream, the close is asynchronous anyway, so it is not correct to assume that the socket will be ready immediately.
* remove `open_` flag for stream, and instead proxy it directly to the socket to avoid "operation is already in progress" type of errors when trying to open the same socket twice.
* re-arm `connection_deadline_` when the socket is ready for data, use KV timeout for whole handshake with the single node. Re-initiate bootstrap using next bootstrap address in case of single node timeout.
* reset `reading_` flag in case of error. Ensure that the flag is reset if the session is not reading.
* handle `socket_closed_while_in_flight` error, and if it happens during bootstrap, try next node in the list, otherwise preserve old behavior (close the connection, and let bucket to reopen it).

### DIFF
--- a/core/io/mcbp_session.cxx
+++ b/core/io/mcbp_session.cxx
@@ -1084,20 +1084,21 @@ public:
       return;
     }
     bootstrapped_ = false;
-    if (bootstrap_handler_) {
-      last_bootstrap_error_ = std::move(bootstrap_handler_)->last_bootstrap_error();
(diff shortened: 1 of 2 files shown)


thejcfactor (approved):
LGTM 👍 

Tested changes in Node.js client as well.

PR #677, opened 2024-10-15
CXXCBC-582: Add cluster labels, keyspace & outcome in metrics

DemetrisChr:
## Motivation

The Extended Observability RFC revisions 24 & 25 require us to record additional metrics tags: bucket, scope, collection, outcome, and cluster labels (name & UUID)

## Changes

* Update config parsing to include `clusterName` & `clusterUUID`
* Create a new `core::metrics::meter_wrapper` class that watches for config updates and keeps track of cluster labels 
* Replace all meter references in the core with the meter wrapper
* The meter wrapper is responsible for encoding the metric attributes & inserts tags for the cluster labels, if those are supported by the cluster
* Add some tests for KV

## Results

Tests pass

## Known issues

Non-KV operation metrics don't currently report collection, bucket and scope names.

### DIFF
--- a/core/metrics/meter_wrapper.cxx
+++ b/core/metrics/meter_wrapper.cxx
@@ -1,0 +1,188 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2024. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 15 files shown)


core/metrics/meter_wrapper.cxx
```diff
@@ -0,0 +1,184 @@
...
+{
+  const std::string::size_type pos = ec.message().find(' ');
+  if (pos != std::string::npos) {
+    return ec.message().substr(0, pos);
+  }
+  return ec.message();
+}
+
+auto
+service_to_string(service_type s) -> std::string
```
avsej, line 57: it will create new string every time. It should return `const std::string&`, `const char *` or at least `std::string_view`

DemetrisChr: Won't we have to eventually create a copy of the string anyway, because we have to insert it in the tags map that meters expect as `std::map<std::string, std::string>`?

PR #681, opened 2024-10-16
CXXCBC-616: Columnar - report retry information when receiving a timeout from the HTTP component

DemetrisChr:
We have client-side timeouts both for the overall query and for each individual HTTP operation, and it appears possible that the individual HTTP timeout gets triggered before the overall query one and we lose all the error context retry information (last errors, number of retry attempts).

We should handle a timeout error code coming from the HTTP component, and report the same retry fields in the error context that we would otherwise do.


thejcfactor (approved):
LGTM 👍

PR #683, opened 2024-10-17
CXXCBC-618: fix memory leak and data race

avsej:
using `std::function` along with `std::shared_ptr` for promise might affect the lifetime of the cluster object and result in memory leak and data races during destruction


avsej:
#783 does better job fixing the sanitizers

PR #686, opened 2024-10-24
CXXCBC-615: Expose insert_raw & replace_raw in core txns attempt ctx

DemetrisChr:
* Expose `insert_raw` and `replace_raw` in the core txns attempt context APIs, which have now been renamed to `insert` and `replace`.
* Remove the variants of `insert` and `replace` that use transcoders from the core API, as well as the overloads of `core::transactions::transaction_get_result::content()` that use transcoders
   * The C++ Public API, as well as the wrappers, have their own transcoder support, which makes these redundant.


thejcfactor (approved):
LGTM 👍 

Also tested w/ the PYCBC-1595 changes.  Thanks for making the updates!

avsej (approved):

PR #690, opened 2024-10-29
CXXCBC-609: Expose `parent_span` in Public API options

DemetrisChr:
## Changes

* Add a `parent_span` setter to all Public API options classes (by adding it to the `common_options` base class) – at 'volatile' stability
* Map the option to the corresponding Core API field where it is available

### DIFF
--- a/couchbase/common_options.hxx
+++ b/couchbase/common_options.hxx
@@ -76,9 +76,9 @@ public:
    * @since 1.0.4
    * @volatile
    */
-  auto parent_span(const std::shared_ptr<tracing::request_span>& span) -> derived_class&
+  auto parent_span(std::shared_ptr<tracing::request_span> span) -> derived_class&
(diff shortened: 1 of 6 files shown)


couchbase/common_options.hxx
```diff
@@ -69,6 +70,18 @@ public:
...
 
+  /**
+   * @param span
+   *
+   * @since 1.0.4
+   * @volatile
+   */
+  auto parent_span(const std::shared_ptr<tracing::request_span>& span) -> derived_class&
+  {
+    parent_span_ = span;
```
avsej, line 81:
```suggestion
  auto parent_span(std::shared_ptr<tracing::request_span> span) -> derived_class&
  {
    parent_span_ = std::move(span);
```

avsej (approved):

PR #692, opened 2024-10-30
CXXCBC-624: Fix user agent ID generation

avsej:
SDK-RFC-47 expects space after version and all extra information must be surrounded with parentheses.

Additionally this patch allows wrappers to override the ID during compile time, so that C++SDK version will be moved to the extras section of the identifier.

* standalone C++SDK build ("cbc" is runtime extra part)

      "cxx/1.0.3+5.1.074ce40c (Darwin/arm64;ssl/0x30400000;cbc)"

* cmake -DCOUCHBASE_CXX_CLIENT_WRAPPER_UNIFIED_ID=ruby/4.5.3

      "ruby/4.5.3 (cxx/1.0.3+5.1.074ce40c;Darwin/arm64;ssl/0x30400000;cbc)"


thejcfactor (approved):
LGTM 👍

PR #699, opened 2024-11-22
CXXCBC-632: Do not assume Analytics will always send meta fields in response

avsej:
"requestID", "clientContextID" and "metrics" might be missing.

For example, the following payload might be sent with status 503:

    {
      "errors": [
        {
          "code": 23000,
          "msg": "Analytics Service is temporarily unavailable"
        }
      ],
      "status": "fatal"
    }


thejcfactor (approved):
LGTM 👍

PR #703, opened 2024-12-09
CXXCBC-633: ensure DNS-SRV resolver reports timeouts

avsej:
* catch `asio::error::operation_aborted` in every handler, and return immediately giving chance to deadline_ handler to report timeout error code

* make sure that UDP round will always fallback to TCP in case of UDP timeout.


thejcfactor (approved):
Nice!  Checked out in the Node.js client.  Much better now. 👍

PR #704, opened 2025-01-07
CXXCBC-636: Expose preferred server group replica reads in transaction_context

Matt-Woz:
- Exposed `get_replica_from_preferred_server_group` in `transaction_context`


thejcfactor (approved):
LGTM 👍

PR #708, opened 2025-01-14
CXXCBC-640: do not force debug symbols for release builds

### LINKED ISSUE #702: Remove debugging information in Release builds when using CMake
Hi,

In the [Apache NiFi - MiNiFi C++](https://github.com/lordgamez/nifi-minifi-cpp) project we created a Couchbase extension using the Couchbase C++ Client. In this extension the Couchbase client is linked statically to the extension library. We are building the Couchbase client using CMake. What we discovered is that even when using the Release build type the library is more than 400MB in size. It seems that the cmake/Backtrace.cmake file is always included in the main CMakeLists.txt which adds the -ggdb3 compile option if the execinfo.h and dlfcn.h are found on the system. Without the Backtrace.cmake include the library is only 25MB.

Is this by design, or could this be changed to only add these compile options if the build type is Debug or RelWithDebInfo, or add it as a separate CMake option to include it or not?

Regards,
Gabor

avsej:
Fixes #702


thejcfactor (approved):
Nice!  LGTM 👍

PR #713, opened 2025-02-14
CXXCBC-646: Do not copy configuration when the operation depends on it

avsej:
For performance reasons, store bucket configuration as shared pointer, and copy it instead. The SDK can use shared pointer here, as configuration is readonly, and only access to the pointer have to be protected by the mutex


thejcfactor (approved):
LGTM 👍

PR #715, opened 2025-02-14
CXXCBC-646: Do not copy configuration when the operation depends on it

avsej:
For performance reasons, store bucket configuration as shared pointer, and copy it instead. The SDK can use shared pointer here, as configuration is readonly, and only access to the pointer have to be protected by the mutex

### DIFF
--- a/core/impl/observe_poll.cxx
+++ b/core/impl/observe_poll.cxx
@@ -96,24 +96,24 @@ number_of_replica_nodes_required(couchbase::replicate_to replicate_to) -> std::u
 }
 
 auto
-validate_replicas(const topology::configuration& config,
+validate_replicas(const std::shared_ptr<topology::configuration>& config,
                   couchbase::persist_to persist_to,
                   couchbase::replicate_to replicate_to) -> std::pair<std::error_code, std::uint32_t>
 {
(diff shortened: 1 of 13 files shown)


core/impl/observe_poll.cxx
```diff
@@ -96,24 +96,24 @@ number_of_replica_nodes_required(couchbase::replicate_to replicate_to) -> std::u
...
   }
 
   if (touches_replica(persist_to, replicate_to)) {
-    if (!config.num_replicas) {
+    auto number_of_replicas = config->num_replicas;
+    if (!number_of_replicas.has_value()) {
       return { errc::key_value::durability_impossible, {} };
     }
-    auto number_of_replicas = config.num_replicas.value();
     if (number_of_replica_nodes_required(persist_to) > number_of_replicas ||
```
thejcfactor, line 112: Need to access the value of the optional (`number_of_replicas`) here right?

avsej: `std::optional` would proxy all methods like `operator >()` to the underlying value.

Here is the example: https://godbolt.org/z/Po4n4j45q

thejcfactor (approved):
LGTM 👍

PR #721, opened 2025-02-21
[backport] CXXCBC-646: add feature flag for FIT performer

avsej:
FIT performer need it to compile, as it must support both old and new version of the core

Backport of #720


thejcfactor (approved):
LGTM 👍

PR #722, opened 2025-02-25
CXXCBC-660: Fix potential race condition in the logger

avsej:
All code paths must use update_file_logger to ensure thread safety. The create_file_logger just updated the logger pointer without incrementing version.


thejcfactor (approved):
LGTM 👍

PR #743, opened 2025-03-31
CXXCBC-605: Add support for user-defined log consumption 

Matt-Woz:
Motivation
-----------
Users should be able to redirect SDK logs to their custom logger implementations for easier integration with their logging infrastructure

Changes
---------
- Add `register_log_callback` & `unregister_log_callback` functions to allow registering of custom log handlers for SDK logs


core/logger/logger.cxx
```diff
@@ -72,6 +75,31 @@ get_file_logger() -> std::shared_ptr<spdlog::logger>
   return logger;
 }
 
+auto
+get_custom_callback() -> std::shared_ptr<couchbase::core::logger::log_callback>
+{
+  const std::scoped_lock lock(log_callback_mutex);
+  return log_callback;
```
avsej, line 82: This implementation will require to grab the mutex every time we want to log, which is slow. Previously we had similar implementation of the logger.

How about different approach, where we use lock only to write, and in addition to the shared pointer, use monotonically increasing version that will be updated every time the logger will be registered/unregistered? 

https://github.com/couchbase/couchbase-cxx-client/blob/a80817c6254fb6d06ef15ad3e1b492c47357b462/core/logger/logger.cxx#L62-L73

https://github.com/couchbase/couchbase-cxx-client/blob/a80817c6254fb6d06ef15ad3e1b492c47357b462/core/logger/logger.cxx#L75-L86

The pointer to the logger could be thread-local, and also it would have thread-local copy of the version. And because now we only need to compare two atomics, it should be faster than acquiring mutex lock for every log statement.

https://github.com/couchbase/couchbase-cxx-client/blob/a80817c6254fb6d06ef15ad3e1b492c47357b462/core/logger/logger.cxx#L58-L60

avsej (changes requested):

avsej (approved):

PR #759, opened 2025-05-01
CXXCBC-684: Allow setting both named and positional parameters for queries

DemetrisChr:
## Motivation

The query and analytics services allow queries that have named and positional parameters in the same request. We should allow users to make queries with both types of parameters.

## Changes

**Public API:**
* Update `query_options` & `analytics_options` so that setting named parameters doesn't clear any existing positional parameters and vice versa.
* Document the fact that setting positional parameters overrides existing positional parameters, and setting named parameters overrides existing named parameters.

**Core API:**
* Update encoding of query/analytics_query requests so both named and positional parameters are included in the payload if both have been set.

### DIFF
--- a/core/meta/features.hxx
+++ b/core/meta/features.hxx
@@ -223,3 +223,9 @@
  * bucket_settings in both the public and core management APIs have a num_vbuckets attribute
  */
 #define COUCHBASE_CXX_CLIENT_HAS_BUCKET_SETTINGS_NUM_VBUCKETS 1
+
+/**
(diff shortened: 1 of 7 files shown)


avsej (approved):

DemetrisChr:
I've proposed revisions to the relevant RFCs here: https://github.com/couchbaselabs/sdk-rfcs/pull/148

Once/if that's accepted and merged we can also merge this.

avsej (approved):

PR #765, opened 2025-05-22
CXXCBC-688: Don't convert Public API TOF from lambda to Core API's TOF, rely on internal state

DemetrisChr:
As part of the GetMulti changes, logic was added that converts the Public API TOF error, to a core exception and throws it. As we lose some of the context in the Public API, this was being done with the FAIL_OTHER error class. This caused many FIT failures, as for some error classes we expect different behaviour, like retries.

It’s not necessary to raise the TOF exceptions if returned from the lambda, as the core must always cache TOF errors internally with the full context around the error and the correct error class.

### DIFF
--- a/core/transactions/transactions.cxx
+++ b/core/transactions/transactions.cxx
@@ -29,6 +29,7 @@
 #include "internal/utils.hxx"
 
 #include <couchbase/error.hxx>
+#include <couchbase/fmt/error.hxx>
 
 #include <system_error>
 #include <utility>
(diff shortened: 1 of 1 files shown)


core/transactions/transactions.cxx
```diff
@@ -174,19 +174,11 @@ wrap_public_api_run(transactions& txns,
...
-      } else {
-        if (err.ec().category() == core::impl::transaction_op_category()) {
-          throw op_exception(err);
-        }
-        throw std::runtime_error(err.ec().message());
+    if (err && err.ec() != errc::transaction_op::transaction_op_failed) {
+      if (err.ec().category() == core::impl::transaction_op_category()) {
+        throw op_exception(err);
       }
+      throw std::runtime_error(err.ec().message());
```
avsej, line 181: how about throwing `std::system_error(err.ec())` it is still generic, but its constructor accepts `std::error_code`
https://en.cppreference.com/w/cpp/error/system_error/system_error
```
system_error( std::error_code ec );
```

DemetrisChr: Done

avsej (approved):

PR #777, opened 2025-06-16
CXXCBC-699: Allow to randomize list of bootstrap nodes

avsej:
By default the SDK would shuffle node list (including list received via DNS-SRV request). To disable this behavior use `preserve_bootstrap_nodes_order` option.


thejcfactor (approved):
LGTM 👍

PR #791, opened 2025-07-07
CXXCBC-567: Add support for field-level encryption library

DemetrisChr:
## Motivation

In order to support a separate Field Level Encryption library we need to provide a way to specify a pointer to a "crypto manager" which is then passed on to special transcoders which can handle encryption/decryption. The definition of the crypto transcoder class will be provided by the separate FLE library.

## Changes

* Add forward declaration for `couchbase::crypto::manager`
* Add crypto manager in cluster options & pass that on to the cluster/bucket/scope/collection classes
* Add `couchbase::codec::is_crypto_transcoder` transcoder trait
* For transcoders that satisfy `couchbase::codec::is_crypto_transcoder`, call their encode/decode methods with an additional parameter, the crypto manager pointer.
* Add an internal (not intended for use by any users other than the FLE library) header file with helpers for aead_aes_256_cbc_hmac_sha512 encryption/decryption and generating secure random initialization vectors. These utilize the relevant BoringSSL/OpenSSL functions internally, and allow the FLE library to use them without adding a direct dependency to BoringSSL or OpenSSL


test/test_unit_crypto.cxx
```diff
@@ -0,0 +1,112 @@
...
+                               0x6f,
+                               0xfd,
+                               0xd8,
+                               0x66,
+                               0x94,
+                               0x09,
+                               0x23,
+                               0x41,
+                               0xbc,
+                               0x04 });
```
avsej, line 69: trailing comma would force clang-format to avoid this column-like thing

PR #809, opened 2025-07-26
CXXCBC-707: update network selection heuristic

avsej:
* check all nodes, not just node marked with 'thisNode=true' property when selecting the network

* only check default and "external" networks


thejcfactor (approved):
I updated the tickets w/ my testing.  The changes look to be good (can connect to GCP w/o `network=external` and not regression on AWS).

PR #820, opened 2025-08-28
Add labels to all tests

DemetrisChr:
Some of our tests didn't have an 'integration' or 'unit' label. On GitHub Actions we use these labels to filter tests, which means that tests without a label weren't running on GHA.

Also added a script to verify that all tests have one of the known labels, to prevent any tests without a label from being added in the future.


thejcfactor (approved):
LGTM 👍

PR #821, opened 2025-09-02
CXXCBC-653: Support "access_deleted" for Replica Reads

thejcfactor:
Changes
=======
* Add subdoc access deleted to bucket capabilities
* Enable transactions to utilize subdoc access deleted if capability exists
* Feature not enabled for public API


core/operations/document_lookup_in_all_replicas.hxx
```diff
@@ -141,59 +150,64 @@ struct lookup_in_all_replicas_request {
             };
 
             auto ctx = std::make_shared<replica_context>(std::move(h), nodes.size());
+            auto access_deleted = false;
+            if (enable_access_deleted) {
+              access_deleted = config->capabilities.supports_subdoc_access_deleted();
+            }
```
DemetrisChr, line 156: In my opinion, it would be better for the core replica lookup_in requests to simply forward the value for `access_deleted` that the caller set, and let the server decide if it's valid or not. In that case, users of the core API can decide what to do when the feature isn't supported. We can implement the cap check in the transactions library instead.

thejcfactor: Good point, have made changes accordingly 👍

core/transactions/attempt_context_impl.cxx
```diff
@@ -3533,10 +3533,20 @@ attempt_context_impl::get_doc(const core::document_id& id,
...
-      core::operations::lookup_in_any_replica_request req{ id };
-      req.read_preference = couchbase::read_preference::selected_server_group_or_all_available;
-      req.specs = specs;
-      execute_lookup(this, req, cb);
+      cluster_ref().with_bucket_configuration(
+        id.bucket(),
+        [this, id, specs, cb = std::move(cb)](
+          std::error_code ec, const std::shared_ptr<topology::configuration>& config) {
+          if (ec) {
+            cb(FAIL_OTHER, std::nullopt, ec.message(), std::nullopt);
```
DemetrisChr, line 3541: NIT: Maybe we could add something to the error message so it's clearer that we failed here:


```suggestion
            cb(FAIL_OTHER, std::nullopt, "failed to check whether access_deleted is supported: " + ec.message(), std::nullopt);
```

DemetrisChr (approved):

DemetrisChr (approved):

avsej (approved):

PR #822, opened 2025-09-04
CXXCBC-718: Use cbdinocluster for GitHub Actions

DemetrisChr:
With this change our GHA workflows will use cbdinocluster to create Couchbase clusters, allowing to test against multi-node clusters, for which we currently rely on Jenkins.

In the future we can configure the workflows to test against pre-release versions, but for that we need to configure an access token in the repository's secrets.

### DIFF
--- a/.github/actions/create-cluster/action.yml
+++ b/.github/actions/create-cluster/action.yml
@@ -1,0 +1,77 @@
+name: create-cluster
+description: Sets up a Couchbase cluster using cbdinocluster
+inputs:
+  version:
+    description: 'Couchbase Server version to use'
+    required: true
+  suite:
+    description: 'Test suite the cluster is being created for'
+    required: true
+outputs:
+  ip:
+    description: 'The IP address of a node of the cluster'
+    value: ${{ steps.start-cluster.outputs.ip }}
+  connstr:
+    description: 'Connection string to connect to the cluster'
+    value: ${{ steps.start-cluster.outputs.connstr }}
+  tls-connstr:
+    description: 'Connection string to connect to the cluster over TLS'
+    value: ${{ steps.start-cluster.outputs.tls-connstr }}
+  id:
+    description: 'The cbdinocluster ID of the created cluster'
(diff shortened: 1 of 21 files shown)


.github/workflows/sanitizers.yml
```diff
@@ -48,9 +40,15 @@ jobs:
...
+      - name: Create Couchbase cluster with cbdinocluster
+        id: cbdino
+        uses: ./.github/actions/create-cluster
+        with:
+          version: ${{ matrix.server }}
       - name: Initialize couchbase
         env:
           CB_TRAVEL_SAMPLE: yes
+          CB_HOST: ${{ steps.cbdino.outputs.ip }}
         run: ./bin/init-cluster
```
avsej, line 52: Maybe we do not really need `bin/init-cluster` anymore?

Ruby SDK works without it just fine:
https://github.com/couchbase/couchbase-ruby-client/blob/8574a1a1d2c7a2687188af5dea7d29c522c4c37e/.github/workflows/tests.yml#L648-L672

It can create buckets, collections, etc.
```
          cbdinocluster -v buckets add ${CLUSTER_ID} default --ram-quota-mb=100 --flush-enabled=true
          cbdinocluster -v buckets load-sample ${CLUSTER_ID} travel-sample
```

.github/actions/create-cluster/action.yml
```diff
@@ -0,0 +1,56 @@
...
+    description: 'Connection string to connect to the cluster over TLS'
+    value: ${{ steps.start-cluster.outputs.tls-connstr }}
+runs:
+  using: composite
+  steps:
+    - name: Install cbdinocluster
+      shell: bash
+      run: |
+        mkdir -p "$HOME/bin"
+        wget -nv -O $HOME/bin/cbdinocluster https://github.com/couchbaselabs/cbdinocluster/releases/download/v0.0.41/cbdinocluster-linux-amd64
```
thejcfactor, line 24: Tiny suggestion to update the version of cbdino we are on v0.0.86 now 😄

PR #863, opened 2025-11-04
CXXCBC-745: Lazy connections

avsej:


Matt-Woz (approved):
Looks good to me, but may be good to have another pair of eyes on this too

thejcfactor (approved):

PR #902, opened 2026-03-12
allow usage of system snappy if -DSnappy_DIR is provided

Thomas-Barbier-1A:
Remove FORCE on CPM_USE_LOCAL_PACKAGES to allow users to build with local packages
Note on the intent: We don't want to build snappy as part of the this
build (we want to use our own snappy, same for other dependencies)
We want to link against our own packages


eitlane:
Hi, 

Any update on this?

Thanks

PR #930, opened 2026-03-27
[WIP] CXXCBC-810: Fix transactions layer and add unit tests

avsej:
**Tracking / umbrella PR for CXXCBC-810.** This is not meant to merge as-is — the transactions-layer
defects and the surrounding SDK / fit_performer fixes uncovered while running the Java FIT
transaction, thread-safety, and get-multi suites against the C++ SDK have been (and are being) split
into small, individually reviewable PRs, each against its own ticket. This description tracks that
slicing.

## Wave 1 — self-contained fixes (all MERGED)

| Ticket | PR | Change |
|---|---|---|
| CXXCBC-862 | #998 | Re-resolve collection id on `unknown_collection` on the `queue_request` KV path |
| CXXCBC-863 | #999 | Scan ATRs concurrently in lost-attempts cleanup |
| CXXCBC-864 | #1000 | Close thread-safety races in KV op routing and finalization |
| CXXCBC-865 | #1001 | Fix get_multi read-skew detection and resolution |
| CXXCBC-866 | #1002 | Remove the redundant `transactions_config` move constructor |
| CXXCBC-867 | #1003 | Return `string_view` from `staged_mutation::type_as_string` |
| CXXCBC-868 | #1004 | Harden `atr_entry::has_expired` against missing timestamps and overflow |
| CXXCBC-872 | #1009 | Fix replica-read leak of `mcbp_command` on io_context teardown |
| CXXCBC-873 | #1010 | Reject transaction KV query result rows without a usable CAS |
| CXXCBC-874 | #1011 | Handle per-document get_multi fetch errors and bound the reads |

## Wave 2 — final slice (OPEN)

After rebasing this branch onto the merged wave-1 work, the remaining commits were extracted into
standalone, single-commit PRs (each branches off `main` independently):

| Ticket | PR | Change |
|---|---|---|
| CXXCBC-875 | #1015 | fit_performer: forward named and positional query parameters independently ⚠ *behavioural — reverses named/positional mutual exclusion; confirm FIT contract* |
| CXXCBC-876 | #1016 | Detect and tear down dead idle HTTP connections promptly |
| CXXCBC-877 | #1017 | Regression guard for auto-rollback error surfacing (+ log-format fix) — *behaviour already correct on main; this is a guard, not a fix* |
| CXXCBC-878 | #1018 | fit_performer: fix transaction-stream completion hang |
| CXXCBC-879 | #1019 | fit_performer: propagate parallelize batch sub-command errors |
| CXXCBC-880 | #1020 | fit_performer: harden content decoding and range-scan error reporting |
| CXXCBC-881 | #1021 | fit_performer: serialize concurrent cleanup-hook RPCs (fix double-free) |

## Wave 2 — held back (not yet opened)

Two slices carry pre-existing real Jira ids and are prepared/verified locally but intentionally not
pushed yet:

| Ticket | Change |
|---|---|
| CXXCBC-810 | Classify non-transacted documents correctly in transactional `lookup_in` (the residual, genuinely behavioural core of this branch; branch churn dropped) |
| CXXCBC-861 | Harden transactions cleanup-hooks access with a no-op fallback (trimmed: dead setter + ineffective mutex removed) |

## Status

- Wave 1 (10 PRs): **merged**.
- Wave 2 (7 PRs, #1015–#1021): **open**, awaiting review/merge.
- Wave 2 held back (CXXCBC-810, CXXCBC-861): local, to be opened after the above.
- The four fit_performer PRs (#1018–#1021) also unblock the deferred get_multi FIT coverage promised
  on #1011 (ExtGetMulti permutations can run once the performer no longer crashes/hangs).

Once every slice lands, this umbrella PR carries nothing unique and can be **closed** rather than
merged.


DemetrisChr:
This PR is quite large and has some changes that appear distinct, touching a few separate components (Performer, Transactions, HTTP core, a new type of transcoder in the Public API, changes to KV retry handling). Would it be possible to split it up? I think that would make the review process easier.

avsej:
@DemetrisChr my plan was to merge it without squashing, so each commit will be landed separately in the git history.

tools/fit_performer/src/commands/common.hxx
```diff
@@ -171,30 +172,36 @@ result_to_content(const Result& result,
   std::visit(
     overloaded{
       [&](std::monostate) {
+        // No transcoder specified: the reference SDK uses a bare JsonSerializer here, which does
+        // not validate the document's common flags when reading (see ExtBinarySupport spec). Use
+        // the lenient JSON transcoder so e.g. JSON content stored with binary flags still reads.
+        using lenient = couchbase::codec::default_lenient_json_transcoder;
```
DemetrisChr: We shouldn't be setting the transcoder in the `std::monostate` case, as that would possibly obscure legitimate SDK bugs. This should be testing the default SDK behaviour. If we have any test failures here, we should address them in the SDK.

tools/fit_performer/src/commands/key_value.cxx
```diff
@@ -2040,11 +2040,24 @@ execute_streaming_command(const protocol::sdk::kv::rangescan::Scan& cmd, const c
...
-                            transcoder,
-                            content_as,
-                            proto_res.mutable_sdk()->mutable_range_scan_result());
+      try {
+        from_scan_result_item(stream_id,
+                              item.value(),
+                              transcoder,
+                              content_as,
+                              proto_res.mutable_sdk()->mutable_range_scan_result());
+      } catch (const std::exception& e) {
```
DemetrisChr, line 2054: I think we should narrow this down to the `performer_exception` so we can handle it separately from the decode exception that can come from the SDK.

tools/fit_performer/src/commands/key_value.cxx
```diff
@@ -2040,11 +2040,24 @@ execute_streaming_command(const protocol::sdk::kv::rangescan::Scan& cmd, const c
...
+        // Converting a scanned item into the protocol result can throw: result_to_content() raises
+        // performer_exception for an unsupported transcoder/content_as combination, and
+        // content_as<T>() can throw a decode error. Report it as an error result rather than
+        // letting the exception escape the producer. An escaping exception unwinds the stream
+        // writer task (stream.hxx) without sending a terminal stream message, leaving the driver
+        // blocked until its preemptive test timeout.
+        spdlog::debug("next_fn failed converting scan item for stream {}: {}", stream_id, e.what());
+        auto err_res = common::create_new_result();
+        err_res.mutable_sdk()->mutable_exception()->mutable_other()->set_name(e.what());
```
DemetrisChr, line 2063: Performer exceptions are meant to be surfaced as a non-success gRPC status. This is returning it as an SDK-error. This is to make it more obvious in the test result that the issue is likely to be an invalid request from the driver, rather than an SDK issue.

We do this in a few places, for example: https://github.com/couchbase/couchbase-cxx-client/blob/580ce35e33b5de386ec390619f6823c55f6b39da/tools/fit_performer/src/service.cxx#L2360-L2361

The decode error should continue to be returned as an SDK error.

core/codec/lenient_json_transcoder.hxx
```diff
@@ -0,0 +1,62 @@
...
+ * validate the document's common flags when decoding.
+ *
+ * This mirrors the behaviour of using a bare JsonSerializer (rather than a Transcoder) when reading
+ * in the transactions feature: per the ExtBinarySupport spec the SDK's transcoders "ignore [the
+ * flags] when reading", so e.g. a document written with binary common flags but holding valid JSON
+ * can still be read with the default JSON serializer. It is intended for the transactions read path
+ * only; ordinary KV reads continue to use the strict @ref json_transcoder.
+ */
+template<typename Serializer>
+class lenient_json_transcoder
```
DemetrisChr, line 46: I think this should not be in the Public API, since it's not intended to be used by users.

tools/fit_performer/src/service.cxx
```diff
@@ -1507,6 +1568,7 @@ TxnService::performerCapsFetch(grpc::ServerContext* /*context*/,
   response->add_performer_caps(protocol::performer::Caps::OBSERVABILITY_1);
 #endif
 
+  response->add_sdk_implementation_caps(protocol::sdk::Caps::WAIT_UNTIL_READY);
```
DemetrisChr, line 1567: I don't think we should be implementing WaitUntilReady in the performer, if the SDK does not truly support it. If there are any issues arising from the SDK not supporting WaitUntilReady, the driver is meant to find a way to work around that (currently it sleeps for a few seconds instead).

tools/fit_performer/src/commands/query.cxx
```diff
@@ -64,10 +64,10 @@ to_query_options(const protocol::sdk::query::Command& cmd, observability::span_o
...
     options.readonly(cmd.options().readonly());
   }
-  // Named and positional parameters are mutually exclusive; named takes precedence (matches the
-  // reference Java performer, which uses positional only when there are no named parameters).
-  if (cmd.options().parameters_named_size() == 0 &&
-      cmd.options().parameters_positional_size() > 0) {
+  // Named and positional parameters are independent: a statement may use both at once (named
+  // values become $name body fields, positional values populate the args array), so each is
+  // forwarded whenever present.
+  if (cmd.options().parameters_positional_size() > 0) {
```
DemetrisChr: We can just remove this if statement entirely. If the number of parameters is zero then the loop will just be a no-op. Same for the named parameters below.

tools/fit_performer/src/service.cxx
```diff
@@ -1196,6 +1203,9 @@ TxnService::execute_command(ConnectionPtr conn,
         }
         lock.lock();
         in_flight--;
+        if (op_err && !batch_error) {
+          batch_error = op_err;
```
DemetrisChr, line 1266: `batch_error` can be written to by multiple threads. This might need some locking or an atomic variable.

tools/fit_performer/src/service.cxx
```diff
@@ -1403,6 +1415,18 @@ TxnService::transactionStream(
...
+
+          // The transaction is finished, but the read loop below is still parked in a blocking
+          // stream->Read() waiting for the next driver message. The driver (RunningTransaction in
+          // the FIT test-driver) only stops blocking when it observes the stream close - receiving
+          // the final result message alone does not release blockOnResult(), and the driver never
+          // half-closes its outbound side. With nothing left to exchange, proactively end the RPC
+          // so the stream closes and the driver unblocks. TryCancel() is the only way to wake a
+          // synchronous gRPC reader from another thread; the final result was already written
+          // above, and gRPC delivers it before the cancellation, so the driver sees the result then
+          // treats the close as completion. This mirrors the reference .NET performer, whose
```
DemetrisChr, line 1484: I think it would be cleaner if we stopped the read loop instead. The driver should be using the fact that the `final_result` is populated as a signal that the performer will neither send nor accept any more messages.

tools/fit_performer/src/service.cxx
avsej: I know that, but this is an intermediate step to test drive the implementation and then move it into the SDK

avsej: I suppose I can move it into the SDK rather. It really helps with the tests, because the fallback sleep in test driver is not reliable and still generate errors. So there is no alternative at all

tools/fit_performer/src/service.cxx
avsej: but there is lock around it already

avsej:
@DemetrisChr the patch set is rebased. Lenient transcoder pushed to the core, wait until ready feature exposed in the public API now

tools/fit_performer/src/service.cxx
DemetrisChr: Yes, I missed that

DemetrisChr:
> @DemetrisChr the patch set is rebased. Lenient transcoder pushed to the core, wait until ready feature exposed in the public API now

Should the wait_until_ready feature be a separate CXXCBC & PR, since it is a new feature? It seems distinct enough from the rest of the changes here that it can be pulled out easily.

avsej:
> Should the wait_until_ready feature be a separate CXXCBC & PR, since it is a new feature? It seems distinct enough from the rest of the changes here that it can be pulled out easily.

I will create a ticket for it, but I want to keep the commit here in this PR. As I said previously, I will do merge without squashing everyting into the single commit.

core/transactions/internal/utils.hxx
```diff
@@ -160,6 +162,74 @@ error_class_from_response(const Resp& resp)
...
+ * misses these per-field failures.  This specialization additionally checks
+ * @c ctx.first_error_index() and maps the corresponding @c fields[i].ec to the appropriate
+ * @c error_class value (@c FAIL_PATH_NOT_FOUND, @c FAIL_PATH_ALREADY_EXISTS, or @c FAIL_OTHER).
+ *
+ * @param resp The @c lookup_in_response to classify.
+ * @return The @c error_class that best represents the failure, or @c std::nullopt on success.
+ */
+template<>
+inline std::optional<error_class>
+error_class_from_response(const core::operations::lookup_in_response& resp)
```
DemetrisChr: I think it is possible to achieve this without duplicating the error handling logic. The generic `error_class_from_response` already calls `error_class_from_response_extras` which is intended for the same purpose as described in the comment above, for subdocument errors.

`error_class_from_response_extras` is noop for all operations except mutate_in. I think the only change needed would be to also implement that for lookup_in in the same way as mutate_in.

core/transactions/attempt_context_impl.cxx
```diff
@@ -3660,7 +3666,9 @@ attempt_context_impl::create_staged_insert_error_handler(const core::document_id
               std::optional<external_exception> /* cause */,
               std::optional<std::string> err_message,
               std::optional<transaction_get_result> doc) mutable {
-              if (!ec3) {
+              // FAIL_PATH_NOT_FOUND with a valid doc means the document exists but has no
+              // transaction XATTRs — treat it like a successful fetch of a non-transacted doc.
+              if (!ec3 || (ec3 == FAIL_PATH_NOT_FOUND && doc.has_value())) {
```
DemetrisChr: I think this might now be inconsistent with the [spec](https://github.com/couchbaselabs/couchbase-transactions-specs/blob/361feb4614dba1d3e4be3c610d7dcd18466ff982/transactions-stages.md#doc-already-exists-when-staging-insert). It suggests that `FAIL_PATH_NOT_FOUND` cannot happen:
> ~FAIL_PATH_NOT_FOUND -> This is what will happen if the doc has been deleted, or if it's simply a normal tombstone. We will get the tombstone and it will be missing the "txn" xattr metadata. Error(ec, retry=true)~ This logic is redundant - it can never trigger - and this line will be removed soon.

It also says that we want to fast-fail if the document exists as a non-transacted doc:
> Most likely, the document simply existed before, as a regular doc. In which case we want to fast fail.

avsej:
converted to draft. I will extract commits from here into smaller PRs

tools/fit_performer/src/service.cxx
avsej: Resolved by CXXCBC-845 (#980): wait_until_ready is now a real SDK feature on both cluster and bucket — the move into the SDK discussed above. The standalone performer implementation was dropped when this branch was rebased onto main, so this cap now advertises genuine SDK support rather than a performer-only shim.

core/transactions/internal/utils.hxx
avsej: Addressed: the dedicated error_class_from_response specialization for lookup_in has been removed. A lookup_in that fetches a doc plus its txn XATTRs is no longer classified as a document-level error — subdoc_multi_path_failure surfaces as success at the document level (empty ctx.ec()) with the missing paths recorded per-field, so the generic template returns success and the caller decides via the parsed links, matching the reference SDKs and the spec. See the explanatory comment now in utils.hxx.

core/transactions/attempt_context_impl.cxx
avsej: Addressed: the FAIL_PATH_NOT_FOUND branch has been removed. create_staged_insert now keys off a successful lookup (doc present), inspects the parsed links, and fast-fails a non-transacted existing document with document_exists — consistent with the linked spec section (no reliance on a document-level FAIL_PATH_NOT_FOUND, which cannot trigger on this path).

avsej:
all changes has been extracted, refined and merged

PR #935, opened 2026-04-18
CXXCBC-821: Propagate node_id through KV operations and add collection::node_id_for

avsej:
Thread node identity through the entire KV operation path so callers can attribute results and errors to the serving cluster node, and add a client-side collection::node_id_for() helper that resolves a key to its owning node from the vBucket map with no network round-trip.

Core plumbing:

- base_error_context and key_value_error_context now carry node_id. For KV requests, make_key_value_error_context builds the node_id from the mcbp session's nodeUUID, canonical hostname, and KV port (TLS or plain, matching the session's transport). A node_id is only built once the session has actually been bound to a concrete node — either a non-empty UUID, or a non-empty canonical hostname together with a non-zero canonical KV port — so the fallback hash of ("", "", 0) never produces a truthy-but-meaningless identifier and the contract that a default-constructed node_id means "unknown" is preserved.
- subdocument_error_context gains a node_id-aware constructor and make_subdocument_error_context forwards last_dispatched_to_node_id from the source key_value_error_context so subdoc results/errors carry the node_id end-to-end.
- make_error for KV contexts always preserves node_id on both success and error paths so callers can identify the serving node regardless of outcome.

Public API:

- couchbase::result gains a node_id() accessor/setter; node_id is propagated onto every KV operation result at a single impl-layer point (core::impl::invoke_with_node_id) rather than in a wrapping std::function layer, avoiding an extra heap allocation that clang-static-analyzer previously flagged as a potential leak on each KV call.
- couchbase::error gains a node_id() accessor for failure attribution.
- configuration::node::effective_node_id(is_tls) exposes the canonical way to obtain a node_id from topology configuration. The is_tls flag lets request-side callers select the same KV port the session used on the response side, so the fallback hash (used on pre-8.0 servers that do not supply a nodeUUID) agrees on both sides.
- collection::node_id_for(document_id, options) resolves a key to its owning node using the client-side vBucket map. The API honors common_options (timeout, parent_span) by arming an asio::steady_timer guarded by an atomic "done" flag so exactly one of the timer or the bucket-configuration callback fires the user handler — with_bucket_configuration cannot be cancelled, so the timer is the only way to guarantee the handler completes when a config never arrives. A missing/empty vBucket map or an unresolvable server_index is surfaced as errc::network::configuration_not_available (not request_canceled), reserving bucket_not_found for the actual "bucket does not exist" case already surfaced via the ec parameter.
- get_all_replicas_response::entry and lookup_in_all_replicas_response::entry now carry dispatched_to_node_id, populated from each sub-operation's last_dispatched_to_node_id. The public get_replica_result and lookup_in_replica_result entries returned from get_all_replicas() and lookup_in_all_replicas() expose this per-entry node_id so callers can tell which node served each replica read.


thejcfactor (approved):
LGTM 👍

PR #936, opened 2026-04-21
[Backport] CXXCBC-832: HTTP session check_in not removing sessions from busy/pending session maps

thejcfactor:
Changes
-------
* Update check_in() to remove session from pending sessions map if session is not connected
* Update check_in() to always attempt to remove from busy/pending session maps (if the session exists and is connected)

### DIFF
--- a/core/io/http_session_manager.hxx
+++ b/core/io/http_session_manager.hxx
@@ -251,45 +251,45 @@ public:
             app_telemetry_meter_,
             options_.default_timeout_for(request.type));
 #endif
-          cmd->start([start = std::chrono::steady_clock::now(),
-                      self = shared_from_this(),
-                      type,
-                      cmd,
-                      handler =
-                        collector->build_reporter()](operations::http_noop_response&& resp) {
-            diag::ping_state state = diag::ping_state::ok;
-            std::optional<std::string> error{};
-            if (auto ec = resp.ctx.ec; ec) {
-              if (ec == errc::common::unambiguous_timeout ||
-                  ec == errc::common::ambiguous_timeout) {
-                state = diag::ping_state::timeout;
-              } else {
-                state = diag::ping_state::error;
+          cmd->start(
+            [start = std::chrono::steady_clock::now(),
+             self = shared_from_this(),
+             type,
+             cmd,
+             handler = collector->build_reporter()](operations::http_noop_response&& resp) {
+              diag::ping_state state = diag::ping_state::ok;
+              std::optional<std::string> error{};
+              if (auto ec = resp.ctx.ec; ec) {
+                if (ec == errc::common::unambiguous_timeout ||
+                    ec == errc::common::ambiguous_timeout) {
+                  state = diag::ping_state::timeout;
+                } else {
+                  state = diag::ping_state::error;
+                }
+                error.emplace(fmt::format("code={}, message={}, http_code={}",
+                                          ec.value(),
+                                          ec.message(),
+                                          resp.ctx.http_status));
               }
-              error.emplace(fmt::format("code={}, message={}, http_code={}",
-                                        ec.value(),
-                                        ec.message(),
-                                        resp.ctx.http_status));
-            }
-            auto remote_address = cmd->session_->remote_address();
-            // If not connected, the remote address will be empty.  Better to
-            // give the user some context on the "attempted" remote address.
-            if (remote_address.empty()) {
-              remote_address =
-                fmt::format("{}:{}", cmd->session_->hostname(), cmd->session_->port());
-            }
-            handler->report(
-              diag::endpoint_ping_info{ type,
-                                        cmd->session_->id(),
-                                        std::chrono::duration_cast<std::chrono::microseconds>(
-                                          std::chrono::steady_clock::now() - start),
-                                        remote_address,
-                                        cmd->session_->local_address(),
-                                        state,
-                                        {},
-                                        error });
-            self->check_in(type, cmd->session_);
-          });
+              auto remote_address = cmd->session_->remote_address();
+              // If not connected, the remote address will be empty.  Better to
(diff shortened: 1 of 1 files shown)


avsej:
## Integration-test timeout diagnosis

### Summary of CI state

- **build / unit / sanitizer / macOS / windows:** all pass.
- **integration (`<version>`, integration)** matrix — **all 8 legs hit the 40-minute step timeout.**
- **integration (`<version>`, transaction)** on the same matrix finishes in 2–6 min, so the failure is specific to the main integration suite.
- Baseline `release-1.3` without this commit runs the same matrix in ≈ 36 min.

Each failing leg hangs on a single ctest case (per-test limit = 1500 s = 25 min), continues to the next, hangs there too, and the job step is killed at 40 min. The specific hung test differs by server version, e.g.:

| server | hung test | ctest time |
|---|---|---|
| 6.6.6 | `integration: analytics external link management with core API` (#155) | 1500.10 s |
| 8.0.1-4642 | `integration: collection management create collection with max expiry` (#145) | 1500.09 s |

### Log signature

Every hanging test has the same tail pattern on the session that was most recently created:

```
[debu] ... attempt to establish HTTP connection
[trac] ... resolved "<host>:<port>" to 1 endpoint(s)
[debu] ... connecting to <host>:<port>, timeout=2000ms
[debu] ... connected to <host>:<port>
--- 23–25 minutes of silence on this session ---
```

The TCP connect completed (so `http_session::on_connect` ran and called `invoke_connect_callback()` at `core/io/http_session.cxx:677`), but **`http_command::send()` never logs its "HTTP request: …" line** (would appear at `core/io/http_command.hxx:261`). The only meaningful work between those two points is the connect-callback body at `core/io/http_session_manager.hxx:719-723`, which takes `sessions_mutex_` and then calls `cmd->send_to()`.

### Root-cause hypothesis

The new `check_in()` calls `pending_sessions_[type].remove_if(...)` and `busy_sessions_[type].remove_if(...)` **while holding `sessions_mutex_`** (lines 408-412, 436-447). When `remove_if` erases a matching element, the `shared_ptr<http_session>` copy stored in the list is destroyed in place. If that was the last strong reference, `~http_session()` runs synchronously — which calls `stop()` (`core/io/http_session.cxx:159`) — which calls `invoke_connect_callback()` synchronously inside `stop()` (`core/io/http_session.cxx:358`), **still on the same thread that holds `sessions_mutex_`**.

For any session that went through `connect_then_send`, the stored connect callback is the lambda at `http_session_manager.hxx:680-724`. Its "connected" branch does:

```cpp
std::scoped_lock inner_lock(self->sessions_mutex_);   // line 720
self->busy_sessions_[...].push_back(session);
cmd->send_to();                                        // line 722
```

Re-acquiring a non-recursive `std::mutex` on the same thread is UB and hangs in practice. The connecting session's strand is then parked forever waiting on `sessions_mutex_`. This matches the log exactly: `connected_ = true` was set (we see "connected to …"), but we never reach `cmd->send_to()`.

A secondary, less certain concern: `session->set_idle(idle_timeout)` at line 431 is called **outside** `sessions_mutex_`, before the push into `idle_sessions_` at 444, and the `if (can_pool && !session->is_stopped())` is re-evaluated under the lock. A session that becomes stopped between the two checks gets `set_idle` armed (which captures `shared_from_this()`), is not pooled, and then is posted a `stop()`. `stop()` will cancel the idle timer, so this is only a theoretical leak — but it exemplifies the "act on the session while not-under-lock, then re-check" pattern that's easy to get wrong.

### Suggested fix (sketch)

Never destroy a `shared_ptr<http_session>` while `sessions_mutex_` is held. Move any potential last-ref-drop outside the critical section — for example:

```cpp
void check_in(service_type type, std::shared_ptr<http_session> session) {
  if (!session) return;
  if (!session->is_connected()) {
    std::shared_ptr<http_session> dropped;
    {
      std::scoped_lock lock(sessions_mutex_);
      auto& pend = pending_sessions_[type];
      auto it = std::find_if(pend.begin(), pend.end(),
        [id = session->id()](const auto& s){ return s && s->id() == id; });
      if (it != pend.end()) { dropped = std::move(*it); pend.erase(it); }
    }
    CB_LOG_DEBUG("{} HTTP session never connected. Removed from pending.",
                 session->log_prefix());
    // `dropped` is released here, OUTSIDE the mutex
    return;
  }
  // ... same trick for the busy/pending removes in the main path:
  // collect victims into a local vector<shared_ptr>, release mutex, let them fall out of scope.
}
```

`asio::post(session->get_executor(), [session]{ session->stop(); })` in the non-poolable branch already defers `stop()` off-thread, but the *implicit* drop that happens whenever `remove_if` matches is what needs the same treatment.

### Other review notes

- The `pending_sessions_[type].remove_if` in the new not-connected branch (lines 410-412) does **not** call `session->stop()`, unlike every other non-pool path. That matches the old `return session.reset();` semantics, but combined with the removal from `pending_sessions_`, `session_manager::close()` at lines 486-491 will also **not** stop it at shutdown — the session was previously recovered there. If `async_connect` is still in flight via `shared_from_this`, shutdown ends up waiting for the io_context to drain it. Adding an explicit `asio::post(executor, [session]{ session->stop(); })` in that branch is probably safer and more consistent.
- Commit message mentions "search sessions" but PR title/body says "sessions" generically — worth reconciling before merge.
- Once the destroy-under-lock is fixed, re-run the integration matrix; the hang should go away.

### Recommendation

**Block merge until the CI hang is addressed.** Requested changes:

1. Restructure `check_in` so no `shared_ptr<http_session>` destruction happens while `sessions_mutex_` is held.
2. Explicitly `stop()` the session in the `!is_connected()` early-return so shutdown semantics match what used to happen via `session_manager::close()`.

avsej (approved):

PR #942, opened 2026-05-08
CXXCBC-823: HTTP session check_in not removing sessions from busy/pending session maps

thejcfactor:
Changes
--------
* Update `check_in()` to remove session from pending sessions map if session is not connected
* Update `check_in()` to always attempt to remove from busy/pending session maps (if the session exists and is connected)
* Add comment explaining that the declaration of container of sessions is declared outside lock so that destructor(s) run after mutex is released (gives context to cppcheck-suppress)

### DIFF
--- a/core/io/http_session_manager.hxx
+++ b/core/io/http_session_manager.hxx
@@ -251,45 +251,45 @@ public:
             app_telemetry_meter_,
             options_.default_timeout_for(request.type));
 #endif
-          cmd->start([start = std::chrono::steady_clock::now(),
-                      self = shared_from_this(),
-                      type,
-                      cmd,
-                      handler =
-                        collector->build_reporter()](operations::http_noop_response&& resp) {
-            diag::ping_state state = diag::ping_state::ok;
-            std::optional<std::string> error{};
-            if (auto ec = resp.ctx.ec; ec) {
-              if (ec == errc::common::unambiguous_timeout ||
-                  ec == errc::common::ambiguous_timeout) {
-                state = diag::ping_state::timeout;
-              } else {
-                state = diag::ping_state::error;
+          cmd->start(
+            [start = std::chrono::steady_clock::now(),
+             self = shared_from_this(),
+             type,
+             cmd,
+             handler = collector->build_reporter()](operations::http_noop_response&& resp) {
+              diag::ping_state state = diag::ping_state::ok;
+              std::optional<std::string> error{};
+              if (auto ec = resp.ctx.ec; ec) {
+                if (ec == errc::common::unambiguous_timeout ||
+                    ec == errc::common::ambiguous_timeout) {
+                  state = diag::ping_state::timeout;
+                } else {
+                  state = diag::ping_state::error;
+                }
+                error.emplace(fmt::format("code={}, message={}, http_code={}",
+                                          ec.value(),
+                                          ec.message(),
+                                          resp.ctx.http_status));
               }
-              error.emplace(fmt::format("code={}, message={}, http_code={}",
-                                        ec.value(),
-                                        ec.message(),
-                                        resp.ctx.http_status));
-            }
-            auto remote_address = cmd->session_->remote_address();
-            // If not connected, the remote address will be empty.  Better to
-            // give the user some context on the "attempted" remote address.
-            if (remote_address.empty()) {
-              remote_address =
-                fmt::format("{}:{}", cmd->session_->hostname(), cmd->session_->port());
-            }
-            handler->report(
-              diag::endpoint_ping_info{ type,
-                                        cmd->session_->id(),
-                                        std::chrono::duration_cast<std::chrono::microseconds>(
-                                          std::chrono::steady_clock::now() - start),
-                                        remote_address,
-                                        cmd->session_->local_address(),
-                                        state,
-                                        {},
-                                        error });
-            self->check_in(type, cmd->session_);
-          });
+              auto remote_address = cmd->session_->remote_address();
+              // If not connected, the remote address will be empty.  Better to
+              // give the user some context on the "attempted" remote address.
+              if (remote_address.empty()) {
+                remote_address =
+                  fmt::format("{}:{}", cmd->session_->hostname(), cmd->session_->port());
+              }
+              handler->report(
+                diag::endpoint_ping_info{ type,
+                                          cmd->session_->id(),
+                                          std::chrono::duration_cast<std::chrono::microseconds>(
+                                            std::chrono::steady_clock::now() - start),
+                                          remote_address,
+                                          cmd->session_->local_address(),
+                                          state,
+                                          {},
+                                          error });
+              self->check_in(type, cmd->session_);
+            });
 
           cmd->set_command_session(session);
           if (!session->is_connected()) {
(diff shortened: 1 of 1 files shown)


core/io/http_session_manager.hxx
```diff
@@ -402,23 +402,45 @@ public:
...
+        auto& pend = pending_sessions_[type];
+        auto it = std::find_if(pend.begin(), pend.end(), [id = session->id()](const auto& s) {
+          return s && s->id() == id;
+        });
+        if (it != pend.end()) {
+          dropped = std::move(*it);
+          pend.erase(it);
+        }
+      }
+      CB_LOG_DEBUG("{} HTTP session never connected.  Removed session from pending.",
```
avsej: Misleading log line. The block above only erases when `find_if` actually locates the session in `pending_sessions_[type]`; if the session has already transitioned to `busy_sessions_[type]` (or was already cleaned up by an `on_stop` racing this path), nothing was removed but we still log `"Removed session from pending."`.

Either move the log inside the `if (it != pend.end()) { ... }` branch, or soften the wording so it's true in both cases:

```suggestion
      CB_LOG_DEBUG("{} HTTP session never connected.  Ensured session is not in pending.",
                   session->log_prefix());
```

core/io/http_session_manager.hxx
```diff
@@ -402,23 +402,45 @@ public:
       return;
     }
     if (!session->is_connected()) {
-      CB_LOG_DEBUG("{} HTTP session never connected.  Skipping check-in", session->log_prefix());
-      return session.reset();
+      // Declared outside the lock so the destructor runs after sessions_mutex_ is released.
+      // cppcheck-suppress variableScope
+      std::shared_ptr<http_session> dropped;
```
avsej: Question: is `dropped` actually doing anything on this path?

The rationale comment ("declared outside the lock so the destructor runs after `sessions_mutex_` is released") is correct in the abstract, but here the *caller* of `check_in()` still holds a `std::shared_ptr<http_session>` (the `session` parameter), so the list element being moved out can't drop the refcount to zero — the local `dropped` extends nothing the function exit didn't already extend. Compare with the `on_stop` lambda further down (line 768 in the new file), where `dropped` *is* meaningful because no caller holds the reference.

Not wrong, just adds a `cppcheck-suppress` annotation for an idiom that isn't load-bearing on this branch. Either drop the local on this path, or add a one-liner explaining you're keeping it for symmetry with `on_stop`.

core/io/http_session_manager.hxx
```diff
@@ -402,23 +402,45 @@ public:
...
       }
     }
+    if (should_stop) {
+      return asio::post(session->get_executor(), [session]() {
+        session->stop();
+      });
+    }
     if (!session->is_stopped()) {
-      session->set_idle(options_.idle_http_connection_timeout);
+      session->set_idle(idle_timeout);
```
avsej, line 440: Worth a sentence in a comment here, or a small reorder. `set_idle()` runs **before** the `sessions_mutex_` block that adds to `idle_sessions_` and removes from `busy_/pending_`. If `set_idle()` is purely async (just schedules a deadline timer for `idle_timeout`), the order is fine — the timer can't fire until well after we've finished the maintenance under the lock. If for some reason `set_idle()` could trigger `stop()` synchronously (e.g. zero/negative timeout, or a path I'm missing), `on_stop` would block on `sessions_mutex_` until we release, then promptly remove the session from `idle_sessions_` *after* we just added it.

Probably benign even in the worst case (you'd just leak nothing into `idle_sessions_`), but a one-line comment confirming `set_idle` is non-firing-synchronous would save the next reader the same derivation. Or do the map maintenance first and call `set_idle` after releasing the lock — then the relative order is unambiguous.

core/io/http_session_manager.hxx
```diff
@@ -741,13 +763,26 @@ private:
...
-        return !s || s->id() == id;
-      });
+      // Declared outside the lock so destructors run after sessions_mutex_ is released.
+      // cppcheck-suppress variableScope
+      std::vector<std::shared_ptr<http_session>> dropped;
+      {
+        const std::scoped_lock inner_lock(self->sessions_mutex_);
+        for (auto* map :
+             { &self->busy_sessions_, &self->idle_sessions_, &self->pending_sessions_ }) {
+          auto& list = (*map)[type];
```
avsej: Tiny nit: `(*map)[type]` uses `std::map::operator[]`, which default-inserts an empty `std::list` if `type` isn't present. The pre-existing `check_in()` code at line 447–452 does the same so this isn't a regression — but since this region is being rewritten anyway, a `find()` early-skip would avoid littering the maps with empty lists for service types that never had a session of that kind.

avsej (commented):
## Review summary

The fix correctly addresses CXXCBC-823 — the FTS leak (sessions where `keep_alive()` always returns `false` get stuck in `busy_sessions_` / `pending_sessions_` forever). Two real changes here:

1. The connected-and-going-to-idle path in `check_in()` now removes the session from **both** `busy_sessions_` and `pending_sessions_` under a single lock — that's the actual leak fix.
2. The `on_stop` callback now also iterates `pending_sessions_` (previously only `busy_` and `idle_`), closing a second leak path the Jira description doesn't call out explicitly.

Deferred destruction (`std::shared_ptr<...> dropped` outside the lock with `cppcheck-suppress variableScope`) matches the idiom CXXCBC-832 documented for this file (#939, `3a4d30b1`). Lock ordering looks correct: `config_mutex_` is taken just for the `keep_alive` / `has_node` decision and released before the `asio::post(stop)` or the second `sessions_mutex_` acquisition.

A few things to address before merge — see inline comments. The biggest one isn't actually code: it's the title.

### ⚠ Title references the wrong ticket

The PR title says **CXXCBC-832**, but:
- the branch is `CXXCBC-823`,
- the Jira ticket about "HTTP session check_in not removing sessions from busy/pending session maps" is **CXXCBC-823**,
- **CXXCBC-832** is the unrelated, already-merged *Document deferred-destruction idiom in http_session_manager* (PR #939, commit `3a4d30b1`).

This is almost certainly a copy-paste from #936's title (`[Backport] CXXCBC-832: HTTP session check_in not removing sessions from`). The merge commit will inherit this title and pollute `git log` searches for CXXCBC-823. Please retitle to `CXXCBC-823: HTTP session check_in not removing sessions from busy/pending session maps`. (And while we're at it: #936's commit on `release-1.3` carries the same wrong tag — worth checking with the author whether that's the same fix as this one and whether the Jira numbering needs untangling.)

### Test coverage

`test/test_integration_http_session_manager.cxx` has a single analytics happy-path test. Nothing exercises the FTS-keep_alive=false path that the Jira ticket describes, so this entire bug had no automated detection. I'd strongly recommend adding a regression test in this PR.

I didn't get to *run* a test because I don't have a Couchbase cluster up locally (`cbdinocluster ps` came back empty), but the existing test uses `integration.cluster.http_session_manager()` directly and `session_manager::export_diag_info(diag::diagnostics_result&)` is already public, so the shape is straightforward — something like (untested, sketch only):

```cpp
TEST_CASE("integration: FTS check_in does not leak busy/idle sessions", "[integration]")
{
  test::utils::integration_test_guard integration;
  if (!integration.has_search_service()) {
    SKIP("Requires search service");
  }

  auto [mgr_ec, session_mgr] = integration.cluster.http_session_manager();
  REQUIRE_SUCCESS(mgr_ec);

  auto search_sessions = [&]() {
    couchbase::core::diag::diagnostics_result res{};
    session_mgr->export_diag_info(res);
    if (auto it = res.services.find(couchbase::core::service_type::search);
        it != res.services.end()) {
      return it->second.size();
    }
    return std::size_t{ 0 };
  };

  const auto baseline = search_sessions();

  for (int i = 0; i < 8; ++i) {
    auto [ec, session] =
      session_mgr->check_out(couchbase::core::service_type::search, "");
    REQUIRE_SUCCESS(ec);
    session_mgr->check_in(couchbase::core::service_type::search, session);
  }

  // Allow async session->stop() / on_stop to run.
  std::this_thread::sleep_for(std::chrono::milliseconds(200));

  // Pre-fix: each FTS check_in leaks the session in busy_sessions_, so this
  // count grows by 8.  Post-fix: bounded by the number of search nodes.
  REQUIRE(search_sessions() <= baseline + integration.number_of_search_nodes());
}
```

If the deadline assumptions or `number_of_search_nodes()` helper don't match what's in the test utils, the shape still holds — the key invariant is *"after N FTS round-trips the diagnostic session count for `service_type::search` is bounded, not growing linearly with N."* That's the assertion the original bug failed at runtime.

Nice patch otherwise — once the title is fixed and one of the two log/ordering nits below is addressed, I'm a +1.

avsej (approved):

PR #943, opened 2026-05-08
[Backport] CXXCBC-828: Use std::string::find() when building errors based on server response text

thejcfactor:
Changes
--------
* Replace inline `std::regex` construction w/ `std::string::find()` in collection/scope mgmt operations where possible
* Make `std::regex` objects `static const` in query_index_create/drop and error_utils where case-insensitive matching is required

### DIFF
--- a/core/operations/management/collection_create.cxx
+++ b/core/operations/management/collection_create.cxx
@@ -24,13 +24,11 @@
 #include <spdlog/fmt/bundled/core.h>
 #include <tao/json/value.hpp>
 
-#include <regex>
-
 namespace couchbase::core::operations::management
 {
 auto
-collection_create_request::encode_to(encoded_request_type& encoded,
-                                     http_context& /*context*/) const -> std::error_code
+collection_create_request::encode_to(encoded_request_type& encoded, http_context& /*context*/) const
+  -> std::error_code
 {
   encoded.method = "POST";
   encoded.path = fmt::format("/pools/default/buckets/{}/scopes/{}/collections",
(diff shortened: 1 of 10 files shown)


core/operations/management/collection_create.cxx
```diff
@@ -60,16 +58,14 @@ collection_create_request::make_response(error_context::http&& ctx,
...
+        if (encoded.body.data().find("already exists") != std::string::npos) {
           response.ctx.ec = errc::management::collection_exists;
         } else {
           response.ctx.ec = errc::common::invalid_argument;
         }
       } break;
       case 404: {
-        const std::regex scope_not_found("Scope with name .+ is not found");
-        if (std::regex_search(encoded.body.data(), scope_not_found)) {
+        if (encoded.body.data().find("is not found") != std::string::npos) {
```
avsej: Mirror of the high-impact issue called out on #944 — same line, same fix.

The original anchored regex `Scope with name .+ is not found` would not match `"Bucket with name X is not found"`. The new `find("is not found")` does, so a missing-bucket 404 becomes `errc::common::scope_not_found` instead of `errc::common::bucket_not_found`. Reproducible with the unit test linked from the top-level review.

```suggestion
        if (encoded.body.data().find("Scope with name") != std::string::npos) {
```

The analogous fix on `scope_drop.cxx:51` is also needed.

avsej (commented):
This patch is **byte-identical** to #944 (the main-branch counterpart): `git diff origin/main...pr-944` and `git diff origin/release-1.3...pr-943` produce 255 lines each and `diff` between the two reports no differences.

Substantive review — including a focused unit test that **fails 2/15 assertions on this PR's source** because of two loosened 404 substring matches — lives on the main PR: https://github.com/couchbase/couchbase-cxx-client/pull/944#pullrequestreview-4254959492.

The specific fixes called out there on `collection_create.cxx:68` and `scope_drop.cxx:51` apply **identically here on the same lines** (see the inline comment for the high-impact case). After applying the suggested tightenings the test goes from 13/15 to 15/15 passing.

### Recommendation for the dual-PR flow

Land #944 first, then `git cherry-pick` (or rebase) onto `release-1.3` so the two PRs stay in lockstep. Separately re-applying the same patch on each branch is exactly the maintenance burden this dual-PR pattern is meant to avoid — and a backport that quietly diverges from main is hard to spot later.

No backport-specific concerns from my side; once #944 is corrected, just bring the corrected patch over verbatim.

avsej:
could you fix clang-tidy note?
```
/home/runner/work/couchbase-cxx-client/couchbase-cxx-client/core/operations/management/error_utils.cxx:151:9: error: do not use 'else' after 'return' [llvm-else-after-return,readability-else-after-return,-warnings-as-errors]
  151 |       } else if (message.find("Index does not exist") != std::string::npos ||
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  152 |                  std::regex_search(message, index_not_found_re)) {
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  153 |         return errc::common::index_not_found;
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  154 |       } else if (message.find("Bucket Not Found") != std::string::npos) {
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  155 |         return errc::common::bucket_not_found;
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  156 |       }
      |       ~
```

or maybe just sync patch with `main` branch

avsej (approved):

PR #944, opened 2026-05-08
CXXCBC-828: Use std::string::find() when building errors based on server response text

thejcfactor:
Changes
--------
* Replace inline `std::regex` construction w/ `std::string::find()` in collection/scope mgmt operations where possible
* Make `std::regex` objects `static const` in query_index_create/drop and error_utils where case-insensitive matching is required

### DIFF
--- a/core/operations/management/collection_create.cxx
+++ b/core/operations/management/collection_create.cxx
@@ -24,13 +24,11 @@
 #include <spdlog/fmt/bundled/core.h>
 #include <tao/json/value.hpp>
 
-#include <regex>
-
 namespace couchbase::core::operations::management
 {
 auto
-collection_create_request::encode_to(encoded_request_type& encoded,
-                                     http_context& /*context*/) const -> std::error_code
+collection_create_request::encode_to(encoded_request_type& encoded, http_context& /*context*/) const
+  -> std::error_code
 {
   encoded.method = "POST";
   encoded.path = fmt::format("/pools/default/buckets/{}/scopes/{}/collections",
@@ -58,18 +56,20 @@ collection_create_request::make_response(error_context::http&& ctx,
 {
   collection_create_response response{ std::move(ctx) };
   if (!response.ctx.ec) {
-    switch (encoded.status_code) {
+    switch (const auto& body = encoded.body.data(); encoded.status_code) {
       case 400: {
-        const std::regex collection_exists("Collection with name .+ already exists");
-        if (std::regex_search(encoded.body.data(), collection_exists)) {
+        const auto prefix_pos = body.find("Collection with name ");
+        if (prefix_pos != std::string_view::npos &&
+            body.find(" already exists", prefix_pos) != std::string_view::npos) {
           response.ctx.ec = errc::management::collection_exists;
         } else {
           response.ctx.ec = errc::common::invalid_argument;
         }
       } break;
       case 404: {
-        const std::regex scope_not_found("Scope with name .+ is not found");
-        if (std::regex_search(encoded.body.data(), scope_not_found)) {
+        const auto prefix_pos = body.find("Scope with name ");
+        if (prefix_pos != std::string_view::npos &&
+            body.find(" is not found", prefix_pos) != std::string_view::npos) {
           response.ctx.ec = errc::common::scope_not_found;
         } else {
           response.ctx.ec = errc::common::bucket_not_found;
@@ -78,7 +78,7 @@ collection_create_request::make_response(error_context::http&& ctx,
       case 200: {
         tao::json::value payload{};
         try {
-          payload = utils::json::parse(encoded.body.data());
+          payload = utils::json::parse(body);
         } catch (const tao::pegtl::parse_error&) {
           response.ctx.ec = errc::common::parsing_failure;
           return response;
@@ -86,7 +86,7 @@ collection_create_request::make_response(error_context::http&& ctx,
         response.uid = std::stoull(payload.at("uid").get_string(), nullptr, 16);
       } break;
       default:
-        response.ctx.ec = extract_common_error_code(encoded.status_code, encoded.body.data());
+        response.ctx.ec = extract_common_error_code(encoded.status_code, body);
         break;
     }
   }
--- a/core/operations/management/error_utils.cxx
+++ b/core/operations/management/error_utils.cxx
@@ -143,15 +143,18 @@ translate_query_error_code(std::uint64_t error, const std::string& message, std:
 {
   switch (error) {
     case 5000: /* IKey: "Internal Error" */
-      if (std::regex_search(message, std::regex{ ".*[iI]ndex .*already exist.*" })) {
+    {
+      static const std::regex index_already_exists_re{ ".*[iI]ndex .*already exist.*" };
+      static const std::regex index_not_found_re{ ".*[iI]ndex .*[nN]ot [fF]ound.*" };
+      if (std::regex_search(message, index_already_exists_re)) {
         return errc::common::index_exists;
       } else if (message.find("Index does not exist") != std::string::npos ||
-                 std::regex_search(message, std::regex{ ".*[iI]ndex .*[nN]ot [fF]ound.*" })) {
+                 std::regex_search(message, index_not_found_re)) {
         return errc::common::index_not_found;
       } else if (message.find("Bucket Not Found") != std::string::npos) {
         return errc::common::bucket_not_found;
       }
-      break;
+    } break;
 
     case 12003: /* IKey: "datastore.couchbase.keyspace_not_found" */
       return errc::common::bucket_not_found;
--- a/core/operations/management/scope_create.cxx
+++ b/core/operations/management/scope_create.cxx
@@ -24,13 +24,11 @@
 #include <spdlog/fmt/bundled/core.h>
 #include <tao/json/value.hpp>
 
-#include <regex>
-
 namespace couchbase::core::operations::management
 {
 auto
-scope_create_request::encode_to(encoded_request_type& encoded,
-                                http_context& /* context */) const -> std::error_code
+scope_create_request::encode_to(encoded_request_type& encoded, http_context& /* context */) const
+  -> std::error_code
 {
   encoded.method = "POST";
   encoded.path = fmt::format("/pools/default/buckets/{}/scopes",
@@ -41,18 +39,19 @@ scope_create_request::encode_to(encoded_request_type& encoded,
 }
 
 auto
-scope_create_request::make_response(error_context::http&& ctx, const encoded_response_type& encoded)
-  const -> scope_create_response
+scope_create_request::make_response(error_context::http&& ctx,
+                                    const encoded_response_type& encoded) const
+  -> scope_create_response
 {
   scope_create_response response{ std::move(ctx) };
   if (!response.ctx.ec) {
-    switch (encoded.status_code) {
+    switch (const auto& body = encoded.body.data(); encoded.status_code) {
       case 400: {
-        const std::regex scope_exists("Scope with name .+ already exists");
-        if (std::regex_search(encoded.body.data(), scope_exists)) {
+        const auto prefix_pos = body.find("Scope with name ");
+        if (prefix_pos != std::string_view::npos &&
+            body.find(" already exists", prefix_pos) != std::string_view::npos) {
           response.ctx.ec = errc::management::scope_exists;
-        } else if (encoded.body.data().find("Not allowed on this version of cluster") !=
-                   std::string::npos) {
+        } else if (body.find("Not allowed on this version of cluster") != std::string::npos) {
           response.ctx.ec = errc::common::feature_not_available;
         } else {
           response.ctx.ec = errc::common::invalid_argument;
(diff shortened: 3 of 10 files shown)


core/operations/management/collection_create.cxx
```diff
@@ -60,16 +58,14 @@ collection_create_request::make_response(error_context::http&& ctx,
...
+        if (encoded.body.data().find("already exists") != std::string::npos) {
           response.ctx.ec = errc::management::collection_exists;
         } else {
           response.ctx.ec = errc::common::invalid_argument;
         }
       } break;
       case 404: {
-        const std::regex scope_not_found("Scope with name .+ is not found");
-        if (std::regex_search(encoded.body.data(), scope_not_found)) {
+        if (encoded.body.data().find("is not found") != std::string::npos) {
```
avsej: **Behavioural loosening — needs tightening.**

The original anchored regex was `Scope with name .+ is not found` — it would *not* match a body like `"Bucket with name X is not found"`. The new `find("is not found")` *does* match that body, so a 404 caused by a missing bucket now resolves to `errc::common::scope_not_found` instead of `errc::common::bucket_not_found`. Reproduced locally:

```
test_unit_management_collection.cxx:54: failed: response.ctx.ec == bucket_not_found for: couchbase.common:16 == couchbase.common:10
```

Suggest matching the discriminative prefix — same approach this PR already uses in `collection_drop.cxx` / `collection_update.cxx`:

```suggestion
        if (encoded.body.data().find("Scope with name") != std::string::npos) {
```

With this change applied (plus the analogous one in `scope_drop.cxx:51`), the suggested unit test in the top-level review comment goes from 13/15 → 15/15 passing.

avsej (commented):
## Review summary

Right-shaped fix for the libstdc++ ABI/heap-corruption bug — removing the inline `std::regex` constructions from these hot paths is exactly what CXXCBC-828 calls for, and keeping the case-insensitive ones as `static const` is the right compromise (one-time, thread-safe init under C++11). The mechanical change is fine in 5 / 8 files.

Two of the substring substitutions, however, are **strictly looser than the original anchored-regex matches** and will **misclassify a missing-bucket 404 as `scope_not_found`** (`collection_create.cxx` case 404, `scope_drop.cxx` case 404). Confirmed locally with a focused unit test built against this PR's HEAD (`b2b1e11`):

```
RNG seed: 2370362099
test_unit_management_collection.cxx:54: failed: response.ctx.ec == bucket_not_found for: couchbase.common:16 == couchbase.common:10
test_unit_management_collection.cxx:119: failed: response.ctx.ec == bucket_not_found for: couchbase.common:16 == couchbase.common:10
test cases:  5 |  3 passed | 2 failed
assertions: 15 | 13 passed | 2 failed
```

Those two failures correspond to the inline comments on `collection_create.cxx:68` and `scope_drop.cxx:51`. Replacing the two `find("is not found")` calls with `find("Scope with name")` (matching the pattern this PR already uses in `collection_drop.cxx` and `collection_update.cxx`) makes all 15 assertions pass:

```
All tests passed (15 assertions in 5 test cases)
```

### Suggested regression test

I'd suggest landing this alongside the fix — it's exactly the kind of behavioural delta that drifts back unnoticed. Drop into `test/test_unit_management_collection.cxx` and register as `unit_test(management_collection)` in `test/CMakeLists.txt`:

```cpp
#include "test_helper.hxx"

#include "core/error_context/http.hxx"
#include "core/io/http_message.hxx"
#include "core/operations/management/collection_create.hxx"
#include "core/operations/management/collection_drop.hxx"
#include "core/operations/management/collection_update.hxx"
#include "core/operations/management/scope_create.hxx"
#include "core/operations/management/scope_drop.hxx"

#include <couchbase/error_codes.hxx>

namespace {
auto make_encoded(std::uint32_t status, std::string body)
  -> couchbase::core::io::http_response
{
  couchbase::core::io::http_response encoded{};
  encoded.status_code = status;
  encoded.body.append(body);
  return encoded;
}
}

TEST_CASE("unit: collection_create::make_response error mapping", "[unit]")
{
  using couchbase::core::operations::management::collection_create_request;
  collection_create_request req{ "default", "_default", "c1" };

  SECTION("404 'Scope with name X is not found' -> scope_not_found") {
    auto encoded = make_encoded(404, "Scope with name `myscope` is not found in bucket `default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::scope_not_found);
  }
  SECTION("404 'Bucket ... is not found' -> bucket_not_found (NOT scope_not_found)") {
    auto encoded = make_encoded(404, "Bucket with name `nope` is not found");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::bucket_not_found);
  }
  SECTION("400 'Collection with name X already exists' -> collection_exists") {
    auto encoded = make_encoded(400, "Collection with name `c1` already exists in scope `_default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::management::collection_exists);
  }
  SECTION("400 unrelated body -> invalid_argument") {
    auto encoded = make_encoded(400, "name - max length (251) exceeded");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::invalid_argument);
  }
}

TEST_CASE("unit: scope_drop::make_response error mapping", "[unit]")
{
  using couchbase::core::operations::management::scope_drop_request;
  scope_drop_request req{ "default", "myscope" };

  SECTION("404 'Scope with name X is not found' -> scope_not_found") {
    auto encoded = make_encoded(404, "Scope with name `myscope` is not found in bucket `default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::scope_not_found);
  }
  SECTION("404 'Bucket ... is not found' -> bucket_not_found (NOT scope_not_found)") {
    auto encoded = make_encoded(404, "Bucket with name `nope` is not found");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::bucket_not_found);
  }
}

TEST_CASE("unit: scope_create::make_response error mapping", "[unit]")
{
  using couchbase::core::operations::management::scope_create_request;
  scope_create_request req{ "default", "myscope" };

  SECTION("400 'Scope with name X already exists' -> scope_exists") {
    auto encoded = make_encoded(400, "Scope with name `myscope` already exists in bucket `default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::management::scope_exists);
  }
  SECTION("400 'Not allowed on this version of cluster' -> feature_not_available") {
    auto encoded = make_encoded(400, "Not allowed on this version of cluster");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::feature_not_available);
  }
  SECTION("400 unrelated body -> invalid_argument") {
    auto encoded = make_encoded(400, "name - max length (251) exceeded");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::invalid_argument);
  }
}

TEST_CASE("unit: collection_drop::make_response error mapping", "[unit]")
{
  using couchbase::core::operations::management::collection_drop_request;
  collection_drop_request req{ "default", "_default", "c1" };

  SECTION("404 'Collection with name' -> collection_not_found") {
    auto encoded = make_encoded(404, "Collection with name `c1` is not found in scope `_default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::collection_not_found);
  }
  SECTION("404 'Scope with name' -> scope_not_found") {
    auto encoded = make_encoded(404, "Scope with name `nope` is not found in bucket `default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::scope_not_found);
  }
  SECTION("404 'Bucket ...' -> bucket_not_found") {
    auto encoded = make_encoded(404, "Bucket with name `default` is not found");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::bucket_not_found);
  }
}

TEST_CASE("unit: collection_update::make_response error mapping", "[unit]")
{
  using couchbase::core::operations::management::collection_update_request;
  collection_update_request req{ "default", "_default", "c1" };

  SECTION("404 'Collection with name' -> collection_not_found") {
    auto encoded = make_encoded(404, "Collection with name `c1` is not found in scope `_default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::collection_not_found);
  }
  SECTION("404 'Scope with name' -> scope_not_found") {
    auto encoded = make_encoded(404, "Scope with name `nope` is not found in bucket `default`");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::scope_not_found);
  }
  SECTION("404 'Bucket ...' -> bucket_not_found") {
    auto encoded = make_encoded(404, "Bucket with name `default` is not found");
    CHECK(req.make_response({}, encoded).ctx.ec == couchbase::errc::common::bucket_not_found);
  }
}
```

### Cross-PR note

The patch is **byte-identical** to PR #943 (release-1.3 backport): `git diff` against the respective bases produces 255 lines in each, and `diff` between the two patches reports no differences. Recommend landing this PR first, then cherry-picking onto `release-1.3` so the two stay in lockstep.

Nothing else blocking from my side — once the four substring substitutions are tightened (or the two 404 ones at minimum), this is good to go.

core/operations/management/scope_drop.cxx
```diff
@@ -50,8 +48,7 @@ scope_drop_request::make_response(error_context::http&& ctx,
         response.ctx.ec = errc::common::unsupported_operation;
         break;
       case 404: {
-        const std::regex scope_not_found("Scope with name .+ is not found");
-        if (std::regex_search(encoded.body.data(), scope_not_found)) {
+        if (encoded.body.data().find("is not found") != std::string::npos) {
```
avsej: **Behavioural loosening — needs tightening.**

Same issue as `collection_create.cxx:68`. Original anchored regex was `Scope with name .+ is not found`. The new `find("is not found")` will also match `"Bucket with name X is not found"`, so a missing-bucket 404 will now resolve to `errc::common::scope_not_found` instead of `errc::common::bucket_not_found`. Reproduced:

```
test_unit_management_collection.cxx:119: failed: response.ctx.ec == bucket_not_found for: couchbase.common:16 == couchbase.common:10
```

```suggestion
        if (encoded.body.data().find("Scope with name") != std::string::npos) {
```

core/operations/management/collection_create.cxx
```diff
@@ -60,16 +58,14 @@ collection_create_request::make_response(error_context::http&& ctx,
   if (!response.ctx.ec) {
     switch (encoded.status_code) {
       case 400: {
-        const std::regex collection_exists("Collection with name .+ already exists");
-        if (std::regex_search(encoded.body.data(), collection_exists)) {
+        if (encoded.body.data().find("already exists") != std::string::npos) {
```
avsej: Lower-priority but in the same family as the case-404 issue: the original anchored regex `Collection with name .+ already exists` would not match an unrelated 400 body that happens to contain `"already exists"`. Hard to think of a realistic such body for *this* endpoint, but `find("Collection with name")` is just as cheap and removes the question. Same applies to `scope_create.cxx:50`.

```suggestion
        if (encoded.body.data().find("Collection with name") != std::string::npos) {
```

core/operations/management/scope_create.cxx
```diff
@@ -41,15 +39,15 @@ scope_create_request::encode_to(encoded_request_type& encoded,
...
+                                    const encoded_response_type& encoded) const
+  -> scope_create_response
 {
   scope_create_response response{ std::move(ctx) };
   if (!response.ctx.ec) {
     switch (encoded.status_code) {
       case 400: {
-        const std::regex scope_exists("Scope with name .+ already exists");
-        if (std::regex_search(encoded.body.data(), scope_exists)) {
+        if (encoded.body.data().find("already exists") != std::string::npos) {
```
avsej: Mirror of `collection_create.cxx:61` — same concern, same fix:

```suggestion
        if (encoded.body.data().find("Scope with name") != std::string::npos) {
```

core/operations/management/error_utils.cxx
```diff
@@ -143,15 +143,18 @@ translate_query_error_code(std::uint64_t error, const std::string& message, std:
 {
   switch (error) {
     case 5000: /* IKey: "Internal Error" */
-      if (std::regex_search(message, std::regex{ ".*[iI]ndex .*already exist.*" })) {
+    {
+      static const std::regex index_already_exists{ ".*[iI]ndex .*already exist.*" };
+      static const std::regex index_not_found_re{ ".*[iI]ndex .*[nN]ot [fF]ound.*" };
```
avsej, line 148: Tiny naming nit: the variable above (`index_already_exists`) drops the `_re` suffix while this one (`index_not_found_re`) keeps it. `query_index_create.cxx` and `query_index_drop.cxx` both use `_re`. Worth normalising while you're here.

core/operations/management/scope_create.cxx
avsej: I agree with AI's point, that it is loosening the check, but I do not like its solution either. "Scope with name" might also mean something else. What we can do here is chained `find()` like this.

```suggestion
        const auto prefix_pos = body.find("Scope with name ");
        if (prefix_pos != std::string_view::npos &&
            body.find(" already exists", prefix_pos) != std::string_view::npos) {
```

avsej (changes requested):
Use chained find() in case of regexp with `.+` in the middle. Also it makes sense to add some unit tests similar to proposed in the previous comment.

avsej (approved):

avsej (approved):

PR #951, opened 2026-05-18
CXXCBC-820: Add couchbase::node_id public type for cluster node identification

avsej:
Introduce node_id as a value type that uniquely identifies a cluster node. On Server 8.0.1+ the identifier is the server-assigned nodeUUID; on older releases a stable CRC32-based hash is derived from the node's hostname and KV port. The type is equality-comparable, ordered, and hashable (std::hash specialization provided).

An internal_node_id factory provides the only construction path from core-level data, keeping the public constructor private.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,7 @@ set(couchbase_cxx_client_FILES
     core/impl/doc_id_query.cxx
     core/impl/error.cxx
     core/impl/error_context.cxx
+    core/impl/node_id.cxx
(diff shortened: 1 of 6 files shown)


thejcfactor (approved):
LGTM.  Should we cherry pick the review feedback onto release-1.3 to keep the branches consistent?

avsej:
`release-1.3` already got initial implementation and has been released. All feedback and improvements stay on `main` unless it is a critical fix.

PR #953, opened 2026-05-20
CXXCBC-829: Client should use static const for std::regex strings when building query errors

thejcfactor:
Changes
-------
* Update std::regex strings used in document_query.cxx to be static const
* Aligns remainder of client w/ CXXCBC-828 changes

### DIFF
--- a/core/operations/document_query.cxx
+++ b/core/operations/document_query.cxx
@@ -33,8 +33,8 @@
 namespace couchbase::core::operations
 {
 auto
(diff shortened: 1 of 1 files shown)


thejcfactor:
Going to update CXXCBC-829 for a change that will allow the removal of `std::regex`.  The customer issue that lead to this investigation shows that simply moving the regex strings to use `static const` does not address the ABI mismatch we see when building the Node.js prebuilds on an older toolchain (e.g. GCC-9 via devtoolset-9).

PR #961, opened 2026-06-06
CXXCBC-838: verify TLS server hostname against the certificate (CWE-297)

avsej:
Outgoing TLS connections drove the OpenSSL/asio handshake without ever verifying that the server certificate identifies the host being contacted, and without sending SNI. asio::ssl::verify_peer (the default for couchbases://) only validates that the certificate chains to a trusted CA; it does NOT check the certificate CN/SAN. Combined with the default trust store (system CAs + bundled Capella CA + full Mozilla bundle), this let an on-path attacker present any CA-issued certificate for any name and complete the handshake — a man-in-the-middle on the secure-by-default path (CWE-297). For Capella (trust_only_capella), any Capella-CA-signed node certificate was accepted for any Capella hostname, enabling cross-tenant interception.

Fix:
- Add core::io::configure_tls_handshake(stream, hostname): sets SNI (SSL_set_tlsext_host_name) and installs asio::ssl::host_name_verification so the leaf certificate's SAN/CN must match the requested hostname.
- Thread the canonical hostname through stream_impl::async_connect (the per-attempt connect, which is the correct granularity: the mcbp bootstrap session reuses one stream across rotating bootstrap addresses). Callers pass the intended name: mcbp_session (bootstrap_hostname_), http_session (hostname_), app_telemetry_reporter (address_.hostname). plain streams ignore the parameter.
- The verifier is installed unconditionally; in verify_none mode OpenSSL ignores the callback result, so behaviour there is unchanged.

Also harden the tls_verify=none kill switch (was applied silently):
- Log a prominent warning when certificate verification is disabled.
- Refuse to disable verification for *.cloud.couchbase.com hosts; enforce verify_peer instead.

Test: test_unit_tls_hostname_verification drives a real loopback TLS handshake with a certificate whose only SAN is correct.example and asserts the matching hostname is accepted while a mismatched hostname is rejected. With the verification absent the mismatch case fails (reproducing the bug), and passes once the verifier is installed.

### DIFF
--- a/test/test_unit_tls_hostname_verification.cxx
+++ b/test/test_unit_tls_hostname_verification.cxx
@@ -1,0 +1,211 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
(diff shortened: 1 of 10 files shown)


test/test_unit_tls_hostname_verification.cxx
thejcfactor, line 1: Do you think adding some more units tests would be useful?

- Confirm `tls_verify=none` works as expected?
- Confirm `has_capella_host()` works as expected?

thejcfactor (approved):
LGTM 👍

PR #967, opened 2026-06-13
CXXCBC-826: Add collection::node_ids() to enumerate KV-serving cluster nodes

avsej:
Expose the set of cluster nodes that currently serve key-value traffic for a collection's bucket as a public API drawn from the client-side topology snapshot, with no network round-trip. This closes the node_id surface introduced by CXXCBC-820 (the value type) and CXXCBC-821 (node_id_for and result/error attribution): it is the piece application-side state machines keyed on couchbase::node_id need in order to retire tracking entries when a node leaves the cluster topology.

Public API:

- couchbase::collection gains node_ids(options, handler) and a std::future-returning overload, both marked `@since 1.3.2` and `@uncommitted` to mirror node_id_for.
- couchbase::node_ids_options carries the standard common_options (timeout, parent_span, retry_strategy) and builds an immutable node_ids_options::built for the impl layer.
- Each entry in the returned vector is the same node_id the SDK reports on KV results and errors that touched that node, so the result is directly comparable to the keys of any unordered_map<node_id, ...> the application keeps (e.g. an external circuit-breaker registry).

Semantics:

- Nodes that do not expose a key-value port for the configured transport (TLS or plain) are filtered out: keeping them would derive the pre-8.0 fallback hash from a meaningless port=0 that could collide across KV-less nodes.
- Order is topology order with no deduplication; topology nodes are unique by construction. Callers wanting set semantics can hash the entries themselves (std::hash<couchbase::node_id> is specialized).
- An empty effective node set surfaces errc::network::configuration_not_available rather than a successful empty vector, so the mid-rebalance "topology momentarily exposes no KV nodes" transient is a hard error, matching node_id_for()'s contract on an empty vbmap.
- Other errors mirror node_id_for: unambiguous_timeout when the bucket configuration does not arrive within the request timeout, bucket_not_found / configuration_not_available when it is absent, and any error returned from core::cluster::origin().
- Doxygen on both node_id_for and node_ids documents the topology-snapshot semantics: a node_id returned by one API may transiently not appear in the other across a rebalance; consistency holds only against a stable topology, so sweep loops should require an absence across two consecutive snapshots before retiring tracker state.

Internal layering:

- core::topology::configuration::effective_node_ids(transport) is the pure helper: node iteration, KV-port filter, per-node call to effective_node_id(transport), shrink-to-fit on the result. Lives next to node::effective_node_id for symmetry and is unit-testable without an io_context or a cluster.
- core::impl::resolve_node_ids_from_config maps a topology snapshot to (error_code, vector<node_id>), alongside resolve_node_id_from_config, so the null-config / empty-set / success branches are unit-testable.
- core::impl::with_bucket_config_or_timeout<Payload> is a shared async scaffold owning the atomic-done guard, steady_timer, and asio::post hop back to the io_context. Both collection_impl::node_id_for and collection_impl::node_ids use it, so exactly one of {timer, config callback} fires the user handler and the timing logic lives in one place. The timer is cancelled only on its own executor (the io_context), since asio::steady_timer is not thread-safe.
- collection::node_ids forwards to the impl; the future form wraps the handler form in the standard promise pattern used elsewhere in collection.

### DIFF
--- a/core/impl/resolve_node_id.hxx
+++ b/core/impl/resolve_node_id.hxx
@@ -25,6 +25,8 @@
 #include <memory>
 #include <string>
 #include <system_error>
+#include <utility>
+#include <vector>
 
 namespace couchbase::core::impl
 {
@@ -78,5 +80,46 @@ resolve_node_id_from_config(const std::shared_ptr<topology::configuration>& conf
   // constructor is selected and `resolved` is genuinely moved (a leading `{}`
   // would bind the const-reference constructor instead, defeating the move).
   return { std::error_code{}, std::move(resolved) };
+}
+
+/**
+ * Pure helper that enumerates the KV-serving nodes of a bucket configuration
+ * snapshot. Extracted out of @c collection_impl::node_ids so the
+ * error-mapping branches (no config, empty KV-serving set, success) can be
+ * unit-tested without an io_context or a live cluster — and so the branch
+ * matches the contract of the sibling resolver above:
+ *
+ * - A missing/null config snapshot is reported as
+ *   @c errc::network::configuration_not_available (it must not be a silent
+ *   empty success: the caller cannot distinguish "the cluster has zero KV
+ *   nodes" from "we don't know the topology yet").
+ * - An empty effective-node-ids set is reported the same way; this is the
+ *   mid-rebalance transient that the original commit silently surfaced as
+ *   a successful empty vector.
+ *
+ * @param config  bucket configuration snapshot (may be null)
+ * @param t       which port pair (plain/TLS) to use when deriving identity
+ * @return pair of (error_code, vector<node_id>). On success the @c
+ *         error_code is default and the vector is non-empty; on failure the
+ *         vector is empty.
+ */
+[[nodiscard]] inline auto
+resolve_node_ids_from_config(const std::shared_ptr<topology::configuration>& config,
+                             topology::transport t)
+  -> std::pair<std::error_code, std::vector<couchbase::node_id>>
+{
+  if (!config) {
+    return { errc::network::configuration_not_available, {} };
+  }
+  auto nids = config->effective_node_ids(t);
+  if (nids.empty()) {
+    // Tighter than the original "return {}" — a momentarily-empty
+    // topology (e.g. mid-rebalance) is reported as unavailable rather
+    // than as a successful empty list, so registry-sweep loops keyed on
+    // node_id don't conclude the cluster has zero KV-serving nodes and
+    // discard every entry.
+    return { errc::network::configuration_not_available, {} };
+  }
+  return { {}, std::move(nids) };
 }
 } // namespace couchbase::core::impl
--- a/core/impl/with_bucket_config_or_timeout.hxx
+++ b/core/impl/with_bucket_config_or_timeout.hxx
@@ -1,0 +1,180 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026-Present Couchbase, Inc.
+ *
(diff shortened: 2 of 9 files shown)


core/impl/resolve_node_id.hxx
```diff
@@ -79,4 +81,45 @@ resolve_node_id_from_config(const std::shared_ptr<topology::configuration>& conf
...
+    return { errc::network::configuration_not_available, {} };
+  }
+  auto nids = config->effective_node_ids(t);
+  if (nids.empty()) {
+    // Tighter than the original "return {}" — a momentarily-empty
+    // topology (e.g. mid-rebalance) is reported as unavailable rather
+    // than as a successful empty list, so registry-sweep loops keyed on
+    // node_id don't conclude the cluster has zero KV-serving nodes and
+    // discard every entry.
+    return { errc::network::configuration_not_available, {} };
```
thejcfactor, line 121: Would having a different error here be better?  `no_kv_nodes`?  Seems that it might be useful to differentiate between not having a config and having a config that did not provide any effective nodes.

core/topology/configuration.cxx
```diff
@@ -179,6 +179,21 @@ configuration::node::effective_node_id(transport t) const -> couchbase::node_id
...
 }
 
+auto
+configuration::effective_node_ids(transport t) const -> std::vector<couchbase::node_id>
+{
+  const bool is_tls = (t == transport::tls);
+  std::vector<couchbase::node_id> result;
+  result.reserve(nodes.size());
+  for (const auto& n : nodes) {
+    if (n.port_or(service_type::key_value, is_tls, 0) == 0) {
```
thejcfactor, line 189: Would it be beneficial to cache the `port_or` result?  If we don't continue it will be called again (`n.effective_node_id(t)`) alter when we add the `node_id `to the result.

core/impl/with_bucket_config_or_timeout.hxx
```diff
@@ -0,0 +1,180 @@
...
+ *                      Invoked at most once, on the config-callback path
+ *                      when @c ec is falsy.
+ */
+template<typename Payload, typename Handler, typename Resolver>
+void
+with_bucket_config_or_timeout(const couchbase::core::cluster& core,
+                              const std::string& bucket_name,
+                              std::chrono::milliseconds timeout,
+                              Handler handler,
+                              Resolver resolver)
```
thejcfactor, line 77: Does it make sense to use `Handler&&` and `Resolver&&`?  Then we `std::forward<>` when the handler/resolver are used (e.g. moved) later in the method?

avsej: With Handler&& an lvalue argument would deduce Handler to Handler&, and that flows into struct state's member `Handler handler;`, making it a reference member. state outlives the `with_bucket_config_or_timeout` call (it's held by the timer and the deferred post), so that reference would dangle.

Yes, changing it to `&&` be valid today only because so far we use rvalues everywhere, but by-value makes Handler always decay to an owning type regardless of the argument.

core/impl/resolve_node_id.hxx
avsej: I think it would be only confusing for users to have second code, which essentially should be handled in the same way "back off and retry". That condition transient anyway.

core/topology/configuration.cxx
avsej: port_or inside is a single branch anyway. The caching will not give a lot, and the port is a number, not some heap object. Lets keep it simple here.

core/impl/resolve_node_id.hxx
avsej: to have logging statement this function have to be de-inlined, and moved into .cxx where the logger headers will be added. And in addition to that, the caller would need to pass logger object into the function.

thejcfactor (approved):

PR #970, opened 2026-06-24
Make integration tests pass under Valgrind

avsej:
Valgrind instrumentation slows client-side work enough that a few tests race server-side TTLs that comfortably hold at native speed:

- Range-scan tests populated documents with a 30s expiry that could lapse before the slowed scan finished, leaving fewer keys than expected. Raise it to 300s, matching the sibling sampling-scan tests.
- The pessimistic-locking test locked a key for 10s, which auto-expired partway through the slowed sequence and turned the expected cas_mismatch into document_not_locked. Lock for the server maximum of 30s.

Two fixes are independent of timing:

- Analytics tests only gated on the server version supporting analytics, so they failed with service_not_available on a cluster that has no analytics node. Skip when the service is absent as well, matching the query suite; clusters that deploy analytics still exercise them.
- Suppress the benign "possibly lost" glibc TLS block allocated for the detached thread spawned by cluster::connect(). It is freed at thread exit, but memcheck reports it via _dl_allocate_tls/pthread_create.

### DIFF
--- a/test/test_integration_analytics.cxx
+++ b/test/test_integration_analytics.cxx
@@ -43,8 +43,8 @@ TEST_CASE("integration: analytics query", "[integration]")
     SKIP("elixir deployment does not support analytics");
   }
 
-  if (!integration.cluster_version().supports_analytics()) {
-    SKIP("cluster does not support analytics");
+  if (!integration.has_analytics_service()) {
+    SKIP("cluster does not have analytics service");
   }
 
   test::utils::open_bucket(integration.cluster, integration.ctx.bucket);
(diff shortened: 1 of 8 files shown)


test/test_integration_analytics.cxx
DemetrisChr: This check looks redundant now

avsej: thanks. should be fixed now.

DemetrisChr (approved):

DemetrisChr:
The `"example: search"` test is still failing because of an unambiguous timeout:
```
[97](https://github.com/couchbase/couchbase-cxx-client/actions/runs/28121724733/job/83284984651?pr=970#step:7:598)
unable to upsert the document "prancing-pony": ambiguous_timeout (13) | {"bucket":"travel-sample","collection":"hotel","id":"prancing-pony","last_dispatched_from":"172.17.0.1:42476","last_dispatched_to":"172.17.0.4:11210","opaque":8,"operation_id":"01/27109f-5803-304c-fe5b-cfc4e3cb724e14","retry_attempts":0,"retry_reasons":[],"scope":"inventory"}
```

That test is using a `github_actions_configuration_profile`, rather than the wan development config profile that other tests are using when running them with valgrind. That profile uses the default KV timeout.

I think the `github_actions_configuration_profile` should be updated to have an increased KV timeout, or replaced with the wan_development profile.

DemetrisChr (dismissed):
Update config profile for search example (comment above)

PR #974, opened 2026-06-29
CXXCBC-842: mcbp_session: use-after-free in message_handler::handle() when reauthentication fails

thejcfactor:
Changes
-------
* Keep a local strong reference (`shared_from_this()`) across `message_handler::handle()` and `bootstrap_handler::handle()` so a re-entrant `stop()` cannot free the handler mid-call
* In `do_read()`, take a local copy of the handler before dispatching so the call is skipped when the handler is null and the handler stays alive for the duration of `handle()`

### DIFF
--- a/core/io/mcbp_session.cxx
+++ b/core/io/mcbp_session.cxx
@@ -324,10 +324,6 @@ class mcbp_session_impl
 
     void handle(mcbp_message&& msg)
     {
-      // A nested stop() (e.g. on reauth failure) can release the session's last
-      // strong ref to this handler, which would otherwise free `this` mid-call.
-      // Holding a local ref keeps this handler alive for the duration (CXXCBC-842).
-      auto self = shared_from_this();
       if (stopped_ || !session_) {
         return;
       }
@@ -704,10 +700,6 @@ class mcbp_session_impl
 
     void handle(mcbp_message&& msg)
     {
-      // A nested stop() (e.g. on reauth failure) can release the session's last
-      // strong ref to this handler, which would otherwise free `this` mid-call.
-      // Holding a local ref keeps this handler alive for the duration (CXXCBC-842).
-      auto self = shared_from_this();
       if (stopped_ || !session_) {
         return;
       }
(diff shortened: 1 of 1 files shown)


DemetrisChr:
I think the reasoning behind these strong references could be forgotten over time since on the surface they don't appear to have a purpose (though of course, they do).

Maybe a better solution could be if the handler wasn't the one closing the session. Theoretically, the session is the one that should own its handler, not the other way round (although in practice both have a strong reference to each other). Would it be possible to instead have `handle` return a signal (e.g. the retry reason) to the caller, i.e. the session itself, that it should do the `stop`?

Happy to +2 this patch though as well, and file a ticket for an improvement, since this also does the job. Though it would be good to add the `[[maybe_unused]]` tags (compilation might fail depending on the configuration I think) & a brief comment explaining why those shared pointers exist.

thejcfactor:
Thanks for the feedback! I added a comment (sorry, should have done that on the first go-round).  But, I think adding `[[maybe_unused]]` would be misleading.  The addition is definitely used (we want its destructor's side-effect).  Also, I was under the impression that we would not be worried about `-Wunused-variable` for this b/c the `shared_ptr` is not "trivially destructable"?  Please lemme know if I'm thinking about this incorrectly though.

I do think a CXXCBC for the tweaking of handler/session ownership would be useful though.  I thin the path taken to signal the stop will not be trivial in some cases, so we will need to put some thought (and tests) into the implementation.   For instance the case I was looking into that bubbled this up was `handle() -> session_->handle_request() -> reauth commmand-handler lambda -> self->stop()`.

DemetrisChr:
You are right about `[[maybe_unused]]`, the compiler doesn't warn in this case because the destructor has an effect (We have that warning enabled in our CMakeLists so we'd have seen it if that was the case).

A separate CXXCBC to revisit the handler/session ownership sounds good.

DemetrisChr (approved):

avsej:
I see the problem, but I agree with original @DemetrisChr that this lifetime extension without using it reads as a hack. I will try to check if there maybe better solution, but in the worst case, we could define in-file simple guard struct, that would hold the extra pointer for the duration of the handler scope RAII-style.

something like
```
life_time_guard guard(shared_from_this());
```

avsej:
we could actually copy it right before invoking the handler, which potentially can stop the session. let me try to sketch the patch. I will push it here for review

avsej:
@thejcfactor something like this in https://github.com/couchbase/couchbase-cxx-client/pull/974/commits/48ff38c56c903da5f8d09df847ebc095e4deb811

I kept your commit in the history paired with revert. So the final PR should be cleaned up. But I think this explicit copying is an equivalent of the previous change, but there might be other place where it still have to be done (although I do not see them right now)

avsej:
I think the real problem here was using handler like this `self->handler_->` so it does not really copy smart pointer for the given scope

DemetrisChr (approved):

PR #975, opened 2026-07-08
CXXCBC-843: fix GCC 16 -Warray-bounds in client_request::generate_payload

avsej:
generate_payload() sized the MCBP request buffer from body_.size() but filled it by copying four independent regions (framing extras, extras, key, value). Because the allocation length and the written length came from separate expressions, GCC 16 could not relate them: for zero-length bodies (e.g. noop) it folds body_.size() to 0, proves the buffer is exactly header_size bytes, sees the write iterator one-past-the-end, and reports an out-of-bounds memmove. The write was safe only by the unchecked invariant that every body's size() equals the sum of its copied regions.

Derive the buffer length from those same four regions instead. The byte layout is unchanged, but the buffer is now provably large enough for the copies, so the diagnostic is a proof rather than a suppressed false positive and an over-small allocation becomes structurally impossible.

Drop the project-wide -Wno-error=array-bounds demotion now that our code no longer trips it, so a genuine out-of-bounds elsewhere fails the build.

### DIFF
--- a/core/protocol/client_request.hxx
+++ b/core/protocol/client_request.hxx
@@ -117,21 +117,36 @@ public:
 private:
   [[nodiscard]] auto generate_payload(bool try_to_compress) -> std::vector<std::byte>
   {
+    // Bind the body's four wire regions once. Note that value()/extras() on some bodies (e.g.
+    // hello, mutate_in) lazily populate their buffers on first access, so this also materialises
+    // them before we measure and copy.
+    const auto& framing_extras = body_.framing_extras();
+    const auto& extras = body_.extras();
+    const auto& key = body_.key();
+    const auto& value = body_.value();
+
+    // Size the buffer from the same regions that get copied into it, rather than from the
+    // independent body_.size(). body_.size() is defined as exactly this sum for every body type,
+    // so the byte layout is unchanged; deriving the length here makes the buffer provably large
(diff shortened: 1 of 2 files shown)


core/protocol/client_request.hxx
```diff
@@ -117,21 +117,36 @@ public:
...
+    const std::size_t body_size_bytes =
+      framing_extras.size() + extras.size() + key.size() + value.size();
+
     // SA: for some reason GCC 8.5.0 on CentOS 8 sees here null-pointer dereference
     // JC: BoringSSL changes, noticed the same when building w/ GCC 11.3.0; TODO:  is 12 okay?
 #if defined(__GNUC__) && __GNUC__ >= 8 && __GNUC__ < 12
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wnull-dereference"
 #endif
-    std::vector<std::byte> payload(header_size + body_.size(), std::byte{});
```
DemetrisChr, line 126: Should we remove the `size()` method from all the body types entirely? I think it might be unused now, and judging from this we probably shouldn't use it anywhere else in the future, so removing it would make the intention clearer.

Other than this, the changes look good to me.

DemetrisChr (approved):

core/protocol/client_request.hxx
avsej: let it be there, and we can do it later

PR #977, opened 2026-07-09
CXXCBC-846: fix cluster/bucket shared_ptr cycle leaked by parked deferred ops

avsej:
An in-flight operation (e.g. lookup_in_any_replica) that opens a bucket parks its continuation in bucket_impl::deferred_commands_ when the bucket is not yet configured. That continuation strongly captures the cluster_impl (core). Because cluster_impl owns the bucket (buckets_) which owns the deferred queue which owns the continuation which owns the cluster_impl, the graph forms a reference cycle.

with_configuration()/the deferred-enqueue path did not check closed_ under the deferred_commands_mutex_, so a command enqueued after bucket_impl::close() had already drained the queue is never drained again: the cycle survives process exit and LeakSanitizer reports the whole cluster/bucket/configuration graph as leaked (all indirect, zero direct - the signature of a cycle).

Reject enqueues into an already-closed bucket by invoking the handler with a cancellation instead of parking it, closing the race with close().


core/bucket.cxx
```diff
@@ -1049,9 +1063,18 @@ class bucket_impl
...
+      const std::scoped_lock lock_for_deferred_commands(deferred_commands_mutex_);
+      if (!closed_) {
+        deferred_commands_.emplace(std::move(command));
+        return {};
+      }
+    }
+    // The bucket was closed (and its deferred queue already drained) before this command could be
+    // parked. Enqueuing now would strand the command forever, leaking whatever its continuation
+    // captures. Complete it with a cancellation instead so its state is released (CXXCBC-846).
+    command(errc::common::request_canceled);
```
DemetrisChr, line 1066: Would `errc::network::bucket_closed` be more appropriate? That's the error `direct_dispatch` returns at the initial check about whether the bucket has been closed.

Though if we were to change the error here, the calls to `defer_command` would need to be updated to handle any error, currently they only treat `errc::common::request_canceled` as a failure.

core/bucket.cxx
```diff
@@ -602,23 +602,37 @@ class bucket_impl
...
       return handler(errc::network::configuration_not_available, nullptr);
     }
-    const std::scoped_lock lock(deferred_commands_mutex_);
-    deferred_commands_.emplace(
-      [self = shared_from_this(), handler = std::move(handler)](std::error_code ec) mutable {
-        if (ec == errc::common::request_canceled || self->closed_ || !self->configured_) {
-          return handler(errc::network::configuration_not_available, nullptr);
-        }
+    {
+      const std::scoped_lock lock(deferred_commands_mutex_);
```
DemetrisChr: I think we might be able to DRY this part slightly. It repeats the same logic that's done by `defer_command`. Could it be become something like the following?:
```cpp
   defer_command([handler = std::move(handler), self = shared_from_this()](std::error_code ec) mutable {
      if (ec == errc::common::request_canceled || self->closed_ || !self->configured_) {
        return handler(errc::network::configuration_not_available, nullptr);
      }

      std::shared_ptr<topology::configuration> config{};
      {
        const std::scoped_lock config_lock(self->config_mutex_);
        config = self->config_;
      }
      if (config) {
        return handler({}, config);
      }
      return handler(errc::network::configuration_not_available, nullptr);
    });
```

core/bucket.cxx
avsej: yes. this is why I used request_canceled, as it is the only special code that internally goes through some kind of fast path to exit, while other codes might be retried or just prolong the lifetime of the connection.

core/bucket.cxx
avsej: yes. this really makes sense. thank you

DemetrisChr (approved):

PR #978, opened 2026-07-10
ci: speed up the clang-tidy job and de-flake integration tests

avsej:
Three independent CI and test-stability improvements.

### Run clang-tidy on changed files in pull requests

The linter job re-analyzed the entire library on every PR, which was slow and made the failure output hard to attribute. Pull requests now analyze only the C++ sources changed against the PR base; push to a protected branch still runs the full library analysis. Both modes share the same setup — configure a compile database and build the `couchbase_cxx_client` target once (the static and shared libraries share a source list, so this covers every translation unit) so generated headers exist before analysis. The diff-based mode (`CB_TIDY_DIFF_BASE`) is scoped to `core/` and `couchbase/` so it lints exactly the translation units the full run covers.

CI reports findings as a `clang-tidy-fixes.yaml` artifact instead of rewriting the tree; `-fix` (in-place rewrite) stays opt-in for local use. The full run drives clang-tidy through `run-clang-tidy`, which analyzes each translation unit in a separate process and merges the per-unit fixes into a single file — CMake's build-integrated clang-tidy cannot export fixes safely, since its parallel per-unit invocations would all clobber one shared file. The changed-files run is a single clang-tidy process, so it writes its fixes without contention. One finding surfaced by this is fixed (`const std::string atr_id` in the lost-attempts cleanup loop).

### Harden flaky multi-threaded and replica-read integration tests

The concurrent open/upsert/close thread-safety test now tolerates ambiguous/unambiguous timeouts rather than requiring unconditional success, so it exercises concurrency without flaking under load. The all-replica read tests poll until every replica reports the document before asserting full entry coverage, so replication lag no longer produces a spurious short result.

### Disambiguate matrix job step names

Matrix legs shared identical step names ("Build tests", "Run tests"), making parallel runs indistinguishable in the Actions UI. Step names now carry their matrix dimension (os / sanitizer / tls / suite).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
         uses: hendrikmuhs/ccache-action@v1.2
         with:
           key: ${{ github.job }}-${{ matrix.os }}
-      - name: Build tests
+      - name: Build tests (${{ matrix.os }})
         timeout-minutes: 60
         env:
           CB_NUMBER_OF_JOBS: 8
@@ -53,7 +53,7 @@ jobs:
         with:
           key: ${{ github.job }}-${{ matrix.os }}
           variant: sccache
-      - name: Build tests
+      - name: Build tests (${{ matrix.os }})
         timeout-minutes: 120
         env:
           CB_CACHE_OPTION: sccache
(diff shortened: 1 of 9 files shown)


avsej:
extracted from #930

DemetrisChr (approved):

PR #979, opened 2026-07-10
Add lenient JSON transcoder for binary transactions

avsej:
Since ExtBinarySupport was added, the transactions `content_as()` accessors defaulted to `codec::default_json_transcoder`, which rejects documents whose common flags are not JSON. Per the ExtBinarySupport spec the SDK ignores common flags when reading in transactions, so a JSON document stored with binary common flags could no longer be decoded — a regression from the prior behaviour.

This restores the lenient read by giving the transaction read results (`transaction_get_result`, `transaction_get_multi_result`, and `transaction_get_multi_replicas_from_preferred_server_group_result`) a no-transcoder `content_as<Document>()` overload that decodes without validating common flags, as before ExtBinarySupport. The transcoder overloads no longer default their transcoder, so passing one explicitly still enforces its flag handling. Ordinary KV reads keep using the strict `json_transcoder`.

The lenient decode is implemented by `codec::lenient_json_transcoder`, marked `@internal` — it backs the no-transcoder overload and is not intended to be named by applications, so no committed public API is added.


avsej:
extracted from #930

tools/fit_performer/src/commands/common.hxx
```diff
@@ -171,30 +172,36 @@ result_to_content(const Result& result,
...
     overloaded{
       [&](std::monostate) {
+        // No transcoder specified: the reference SDK uses a bare JsonSerializer here, which does
+        // not validate the document's common flags when reading (see ExtBinarySupport spec). Use
+        // the lenient JSON transcoder so e.g. JSON content stored with binary flags still reads.
+        using lenient = couchbase::codec::default_lenient_json_transcoder;
         switch (proto_content_as.as_case()) {
           case protocol::shared::ContentAs::kAsString: {
-            proto_res_content->set_content_as_string(result.template content_as<std::string>());
+            proto_res_content->set_content_as_string(
```
DemetrisChr: `std::monostate` is intended to represent the case where the user does not provide a transcoder. Providing the lenient transcoder here violates that, which means that we no longer test the default case where a transcoder is not provided.

The real issue we have is that since ExtBinarySupport was added, the default transcoder was changed to the `codec::default_json_transcoder` instead of the prior behavior which was equivalent to the `lenient_transcoder`. We have to make this change in the transactions library itself.

So the change would be needed at all the points where we use the `codec:default_json_transcoder` as a default for decoding the documents:
* https://github.com/couchbase/couchbase-cxx-client/blob/d1d575d2276628676f0e14da8e2f12906d91ae13/couchbase/transactions/transaction_get_result.hxx#L69-L76
* https://github.com/couchbase/couchbase-cxx-client/blob/d1d575d2276628676f0e14da8e2f12906d91ae13/couchbase/transactions/transaction_get_multi_result.hxx#L47-L61

DemetrisChr (changes requested):

tools/fit_performer/src/commands/common.hxx
avsej: but in this case, I will bring back lenient transcoder into public api. I will make it uncommitted, and use only here in transactions api

DemetrisChr: It might be simpler to add a non-transcoder overload instead, and remove the default value for the transcoder parameter in the other overload, like what we used to have before ExtBinarySupport:
```cpp
template<typename Document, std::enable_if_t<!codec::is_transcoder_v<Document>, bool> = true>
[[nodiscard]] auto content_as() const -> typename Transcoder::document_type
{
  return codec::tao_json_serializer::deserialize<Document>(content().data);
}
```

This avoids the need to introduce a new transcoder in the Public API.

Although if you prefer, having the transcoder is fine too, but I would mark it as `@internal` rather than `@uncommitted`.

avsej: it is `@internal` now, and I add that overload.

couchbase/transactions/transaction_get_multi_replicas_from_preferred_server_group_result.hxx
```diff
@@ -46,42 +46,43 @@ public:
     -> transaction_get_multi_replicas_from_preferred_server_group_result& = default;
 
   /**
-   * Content of the document.
+   * Content of the document, decoded without inspecting the common flags.
+   *
+   * When no transcoder is given, transactions read the document leniently (like a bare
+   * JSON serializer), matching the ExtBinarySupport spec's read behaviour. Pass a transcoder
```
DemetrisChr: I would just remove the "matching the ExtBinarySupport spec's read behaviour" part from the API reference, it's an implementation detail

DemetrisChr (approved):
Looks good to me, I just left a minor doc-related comment

PR #980, opened 2026-07-14
CXXCBC-845: Add wait_until_ready to the public cluster and bucket API

avsej:
Expose cluster::wait_until_ready and bucket::wait_until_ready (blocking-future and callback overloads) so applications can block until a freshly connected cluster, or a freshly created bucket, is ready to serve rather than racing bootstrap. Behaviour is configured through wait_until_ready_options (desired_state, service_types) and the cluster_state enum.

Readiness is assessed by pinging the requested services, or every service the topology reports when none are given: the online target requires all endpoints of each requested service to be responsive, degraded requires at least one. cluster_state::offline is not a valid target and is rejected with errc::common::invalid_argument.

The bucket variant opens the bucket and, for the online target, additionally waits until the vbucket map is fully placed -- every vbucket has its active and all replica copies assigned -- the readiness that majority-durable writes require and that a new bucket does not satisfy immediately; the degraded target is ping-only.

The operation is fully asynchronous, polling on the cluster io_context and bounded by the timeout. The handler runs exactly once: empty on success, errc::common::unambiguous_timeout on expiry, and errc::network::cluster_closed if the io_context is torn down while the wait is in flight. The feature is also wired into the transactions FIT performer.


tools/fit_performer/src/service.cxx
```diff
@@ -294,6 +294,67 @@ to_bucket(ConnectionPtr conn, std::string name)
...
+        service_types.insert(couchbase::service_type::view);
+        break;
+      case wur::MANAGER:
+        service_types.insert(couchbase::service_type::management);
+        break;
+      case wur::EVENTING:
+        service_types.insert(couchbase::service_type::eventing);
+        break;
+      default:
+        // e.g. BACKUP and any future service types have no couchbase::service_type.
```
DemetrisChr: Might be better to throw a `performer_exception::unimplemented` here (that's caught further up and converted to a gRPC status).

couchbase/wait_until_ready_options.hxx
```diff
@@ -0,0 +1,112 @@
...
+
+namespace couchbase
+{
+/**
+ * Options for @ref cluster::wait_until_ready() and @ref bucket::wait_until_ready().
+ *
+ * @since 1.4.0
+ * @committed
+ */
+struct wait_until_ready_options : public common_options<wait_until_ready_options> {
```
DemetrisChr, line 37: `common_options` also exposes `timeout`, which should not be in the options block for `wait_until_ready` (as it's already a required parameter in the method).

avsej: I think we just ignore that fact for now. Maybe later we can assert it. I think it is clear that user must provide timeout as a mandatory parameter, and not in the options.

DemetrisChr: We don't have the option to ever remove this though, it would be a breaking change on the API

PR #984, opened 2026-07-15
CXXCBC-853: Fix handler-drop lifetime in async public-API operations

avsej:
Several self-owning async operations register a completion continuation that the core parks in a bucket-bootstrap queue (open_bucket / with_bucket_configuration) while also holding a strong reference to the core, or rely on a self-owned timer with no fail-closed path. Because close() does not drain the bootstrap queue, tearing the cluster down while such an operation is in flight neither invokes nor releases the continuation: the caller's handler never runs -- a std::future caller observes broken_promise, or blocks forever when the dropped continuation is also leaked, and a callback caller hangs.

Affected: observe-based legacy durability (persist_to / replicate_to on every mutation), collection scan (which also captured a raw this), watch_query_indexes during its polling backoff, and bucket-level ping.

Adopt the with_bucket_config_or_timeout pattern: capture a weak_ptr into the core-parked callback, anchor the operation's lifetime on an io_context timer, and deliver errc::network::cluster_closed from a fail-closed destructor when the operation is dropped.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -1149,11 +1149,21 @@ public:
             if (auto bucket = cluster->find_bucket_by_name(bucket_name.value()); bucket) {
               return bucket->ping(collector, timeout);
             }
+            // Capture a weak_ptr, not a strong cluster: open_bucket parks this callback in the
+            // bucket-bootstrap queue, which close() does not drain. A strong self-capture would
+            // form a cluster<->parked-callback cycle that pins the ping collector past cluster
+            // close, so its destructor -- the backstop that reports the ping result -- would never
+            // run and the caller would hang. With a weak capture the callback is released when the
+            // cluster is torn down, the collector is destroyed, and its destructor delivers the
+            // report.
             cluster->open_bucket(
-              bucket_name.value(), [collector, cluster, bucket_name, timeout](std::error_code ec) {
+              bucket_name.value(),
+              [collector, weak = std::weak_ptr(cluster), bucket_name, timeout](std::error_code ec) {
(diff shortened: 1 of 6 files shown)


avsej:
### Update

Addressed the latest Copilot review comments:
- Bounded the regression test's future wait (`wait_for` + `FAIL` on timeout) so a regression fails fast instead of hanging the run.
- Added trailing periods to the two scan error messages for consistency.

**On the failing AddressSanitizer job:** I root-caused it, and it is **not** a defect in this change — it is the pre-existing bucket-bootstrap session leak fixed separately in **CXXCBC-852 (#983)**.

The regression test here scans a bucket that never opens and tears the cluster down mid-bootstrap. The handler-drop fix in this PR works correctly (the scan future is completed with `cluster_closed` rather than being stranded), but opening a bucket starts a KV session bootstrap, and tearing the cluster down while that session is still bootstrapping strands it — the exact all-indirect (cyclic) leak that #983 addresses. Every KV handler-drop path (scan/observe/ping) must open a bucket to be parked in the bootstrap window, so this scenario cannot be exercised without also triggering the #983 leak.

Verified locally under ASan/LSan:

| Configuration | Functional test | LeakSanitizer |
|---|---|---|
| this branch alone | passes (`cluster_closed` delivered) | ~268 KB leaked in `mcbp_session` / `open_bucket` / `bucket_impl::bootstrap` |
| + `core/bucket.cxx` from #983 | passes | clean |

These two changes are intentionally independent PRs. This one will go green on ASan once **#983** lands; I'll rebase this branch on `main` at that point.

PR #997, opened 2026-07-20
test: verify CXXCBC-846 and align subdoc assertions with modern server

avsej:
- Added a robust unit regression test to test_unit_bucket_deferred_queue.cxx validating that with_configuration() on closed buckets completes with configuration_not_available and properly breaks reference cycles.
- Removed invalid CAS empty assertions on subdoc lookup error paths (which return subdoc_multi_path_failure) in test_integration_subdoc.cxx to support Couchbase Server 7.6.0+ protocol behavior.

Original fixing commit: 1fdbd5ac90385be6004d750f72f9449b0331e227
Follow-up fixing commit: 5897881759937ef982e9e7d24c35b2e5d2ed9ae7


thejcfactor (approved):
LGTM 👍

PR #1001, opened 2026-07-20
CXXCBC-865: Fix get_multi read-skew detection and resolution

avsej:
Extracted from #930 (CXXCBC-810) as an independent, self-contained fix.

transaction_get_multi could not detect read skew and could crash or loop
while resolving it:

- Documents were fetched with the MAV get path (get_optional), which performs
  its own ATR lookup and resolves staged content, stripping the transactional
  metadata the orchestrator needs. Fetch raw documents (body + metadata, no
  ATR lookup) via get_doc instead, as the spec requires, and run the
  forward-compatibility checks the MAV path used to provide (GM_G and G on the
  document, GETS_READING_ATR on T1's ATR entry).
- Read-skew victims were mis-identified: the candidate refetch set was built
  from documents that already carried T1's metadata, the opposite of the spec,
  which selects documents fetched WITHOUT T1's metadata that T1's ATR entry
  records as mutated. A document read before T1 committed was never refetched
  and kept its pre-commit value.
- reset_and_retry cleared the results vector but left it sized zero while
  refetch handlers assign by spec index, an out-of-bounds write. Resize back to
  the spec count after clearing.
- The transaction_id used to group documents by the transaction that staged
  them carried the per-mutation operation id, so two documents staged by the
  same attempt counted as two transactions and read skew looped on
  reset-and-retry until timeout. A transaction attempt is identified by
  (transaction, attempt); drop the operation id from its identity.
- When the read-skew resolution bound (the 100ms latency bound, or the
  transaction timeout in detection mode) is exceeded, return the best-effort
  snapshot already fetched instead of failing the whole operation, per the
  spec's BoundExceeded signal. Per-document fetch errors are still surfaced by
  the completion callback.
- Set the read-skew bound once, on the transition out of the first-doc-fetch
  phase, rather than recomputing it every fetch round (which in latency mode
  pushed the 100ms bound indefinitely into the future).
- Read replicas without the accessDeleted flag. Pre-8.0 servers reject a
  replica read combined with accessDeleted (MB-63241), and the get_multi spec
  directs clients to keep omitting it even on 8.0 -- which fixes MB-63241 --
  until cross-version handling has been specified. The replica path now always
  clears accessDeleted and drops the previous capability probe, which gated on
  the generic subdoc_access_deleted capability (advertised since 6.6), not the
  8.0 replica-read fix, and so would have sent the illegal combination on
  6.6-7.x. Tombstones are thus invisible on the replica path, which the spec
  accepts as its documented best-effort behaviour.

Also add doc comments mapping the orchestrator's methods to the named sections
of the ExtGetMulti spec they implement.

**Testing.** Library builds clean. The replica `accessDeleted` change was cross-checked against the transactions spec (get_multi, *Fetching an individual document*): per MB-63241 the flag must be omitted on replica reads, and the spec directs keeping it omitted even on 8.0 pending cross-version work; the prior probe gated on the generic `subdoc_access_deleted` capability and is removed. FIT get_multi is the runtime oracle (CI).


thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. Findings were compile-verified where noted; please treat the reasoning as a starting point for discussion, not ground truth.

---

Really nice fix — the two central corrections (grouping by `(transaction, attempt)` instead of including the operation id, and flipping the COMMITTED victim-selection to build `were_in_t1` from docs *without* T1 metadata) both check out cleanly against the ExtGetMulti spec, and the doc-comments mapping each method to its spec section make this much easier to follow. The `results_` resize, the added forward-compat checks (`GM_G`→`G`, `GETS_READING_ATR`), the deadline-set-once, and the `accessDeleted` omission all look correct too.

While reviewing I traced a few things in the surrounding error/timeout paths that I wanted to raise — none of them are in the fixes themselves, so mostly these are "is this intended for this PR or a follow-up?" questions rather than blockers.

**1. Missing ATR *document* can hang the operation.** When exactly one T1 is implicated we enter `resolve_read_skew()`, and if T1's ATR *document* is not-found, `active_transaction_record::get_atr` returns `cb({}, {})` (falsy `ec`, `nullopt` atr). Back in the callback, the `if (ec) return reset_and_retry();` guard is skipped, and the whole remaining body is inside `if (atr) { … }` with no `else` — so no `completed()`/`retry()`/`reset_and_retry()`/`invoke_callback()` ever fires and the operation callback never completes. This predates the PR, but it's right in the section you're reworking, and the spec's "ATR entry missing" discussion (committed-then-removed vs. cleaned-up-pre-commit) seems intended to cover the whole-document-absent case too. Worth handling here?

**2. Per-document error/timeout handling in the new `get_doc` adapter.** Switching the fetch to `get_doc` + the new classified-error adapter is the right call, but the adapter doesn't yet implement a couple of the spec's per-doc semantics from "Fetching an individual document":
   - `FAIL_TRANSIENT` — spec says retry the individual doc with exponential backoff (from 1ms). Currently it falls into the catch-all `else` → `rollback=true, retry=false`, so a transient KV blip fails and rolls back the whole `get_multi`.
   - `TimeoutException` — spec maps it to `BoundExceeded` (or `FAIL_TRANSIENT` under `PRIORITISE_READ_SKEW_DETECTION`); currently it becomes the same hard rollback failure, so a first-fetch timeout that the spec says should retry the transaction fails hard instead.
   - The per-doc read isn't bounded by `min(remaining deadline, 2.5s)` (`req.timeout` is never set), so a `PRIORITISE_LATENCY` resolution round can overshoot the 100ms bound by up to ~2.5s. (The comment at the `BoundExceeded` return says the bound is "enforced per individual document in the fetch path" — that part doesn't seem to hold currently.)

   Are these intended to land with this PR, or tracked separately?

**3. Test coverage.** The `transaction_id` unit test is a nicely targeted regression guard for the reset-and-retry hang. Since extracting that logic into a free-function header is exactly what made it testable, would it be worth doing the same for the victim-selection predicate and the `reset_and_retry` resize? Both operate on plain data (`atr_entry` is already unit-tested cluster-free in `test_unit_atr_entry.cxx`), so they'd be unit-testable with the same technique and would guard the two fixes most likely to silently regress.

Minor: the `accessDeleted=false` change also affects the normal replica MAV path (`get_replica_from_preferred_server_group` → `do_get(allow_replica=true)` → same branch), so replica MAV reads no longer see tombstones — just flagging in case that broader effect wasn't intended.

avsej:
All four hold up. Handling:

**1. Missing ATR document hangs the operation.** Confirmed and fixed in this PR. `active_transaction_record::get_atr` reports `document_not_found` as an empty record (`cb({}, {})`, no error), and the resolution callback wrapped its whole body in `if (atr)` with no `else`, so an absent ATR document fell through without completing, retrying, or resetting. It now treats an absent ATR document identically to an absent ATR *entry* — there's no ATR record for T1 either way — routing through the existing "ATR entry missing" path, matching the spec's discussion of that case.

**2. Per-document error/timeout semantics.** All three are real gaps in the new adapter, but they're spec-completeness work rather than part of the read-skew correctness fix, and (c) needs a timeout parameter on the shared `get_doc`. Tracking in CXXCBC-874 rather than expanding this PR:
- `FAIL_TRANSIENT` → per-document retry with exponential backoff (currently hard-fails the whole op).
- `TimeoutException` → `BoundExceeded` (or `FAIL_TRANSIENT` under `PRIORITISE_READ_SKEW_DETECTION`).
- Per-document read bounded by `min(remaining, 2.5s)`.

I also corrected the comment that claimed the bound was "enforced per individual document in the fetch path" — it isn't yet; the bound is checked between rounds and an individual fetch runs to the normal KV timeout, so that's now stated accurately.

**3. Test coverage.** Agreed the victim-selection predicate and the `reset_and_retry` resize deserve the same free-function + cluster-free unit-test treatment as `transaction_id`. Folded into CXXCBC-874.

**Minor: accessDeleted on the replica MAV path.** Intended. Per MB-63241 the replica+accessDeleted combination is illegal pre-8.0, and the get_multi spec directs keeping it omitted even on 8.0 until cross-version handling is specified — so the replica path always clears it (the old probe gated on the wrong capability). Tombstones being invisible there is the spec's documented best-effort behaviour.

thejcfactor (approved):

PR #1004, opened 2026-07-20
CXXCBC-868: transactions: harden atr_entry::has_expired against missing timestamps and overflow

avsej:
Extracted from #930 (CXXCBC-810) as an independent, self-contained fix.

Require both timestamp_start_ms and expires_after_ms before declaring an ATR entry expired, and compute the effective TTL in 64-bit so large TTLs or safety margins do not overflow.

**Testing.** `test_unit_atr_entry` (new) — 9 cases pass, including the large-TTL overflow regression and missing-timestamp/expiry cases.


thejcfactor (approved):
LGTM 👍

PR #1007, opened 2026-07-21
CXXCBC-816: Fix path traversal in bin/api.rb

avsej:
Aikido flagged a potential path traversal vulnerability in bin/api.rb at line 164. The method ensure_search_index_created accepts an index_definition_path and reads it without validation.

To mitigate this, sanitize and canonicalize the input path. Ensure the resolved path resides strictly within the safe bin/ directory and contains no '..' segments.

Embed a lightweight unit test suite inside an if __FILE__ == $0 block to verify the fix.

### DIFF
--- a/bin/api.rb
+++ b/bin/api.rb
@@ -156,12 +156,23 @@ def service_address_for_bucket(api, service, bucket, options)
 end
 
 def ensure_search_index_created(index_definition_path, options)
+  # Sanitize to prevent path traversal. File.expand_path collapses any "." / ".."
+  # segments, so an exact-directory comparison is enough: legitimate inputs are
+  # always files directly inside bin/. (A prefix match would also accept sibling
+  # dirs such as bin-evil/.)
+  resolved_path = File.expand_path(index_definition_path)
+  safe_dir = File.expand_path(__dir__)
+
+  unless File.dirname(resolved_path) == safe_dir
+    raise ArgumentError, "Invalid index definition path: outside #{safe_dir}"
+  end
+
   loop do
     catch :retry do
       search_api = API.new(options)
 
-      puts "using index definition: #{File.basename(index_definition_path)}"
-      index_definition = JSON.load(File.read(index_definition_path))
+      puts "using index definition: #{File.basename(resolved_path)}"
+      index_definition = JSON.load(File.read(resolved_path))
       search_api.put_json("/api/index/travel-sample-index", index_definition)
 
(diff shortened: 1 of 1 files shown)


thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. The behavior below was verified by running the actual comparison in Ruby; please treat the reasoning as a starting point for discussion, not ground truth.

The fix is correct for the real callers (`init-cluster` / `create-search-index` both pass `File.join(__dir__, "travel-sample-index*.json")`), so this is safe to merge. One suggestion while we're hardening this path, though:

**`start_with?(safe_dir)` is a prefix match, not a directory-containment check.** `safe_dir` has no trailing separator, so a sibling directory that merely shares the prefix passes validation:

```ruby
safe_dir = "/repo/bin"
input    = "/repo/bin-evil/secret.json"   # note: no ".." in the raw input
resolved = File.expand_path(input)
resolved.start_with?(safe_dir) && !input.include?("..")
# => true   (escapes bin/ but is accepted)
```

Since the only legitimate inputs are files directly inside `bin/`, an exact-directory check is both stricter and simpler:

```ruby
resolved_path = File.expand_path(index_definition_path)
safe_dir      = File.expand_path(__dir__)

unless File.dirname(resolved_path) == safe_dir
  raise ArgumentError, "Invalid index definition path: outside #{safe_dir}"
end
```

That also makes the separate `!index_definition_path.include?("..")` clause unnecessary — `File.expand_path` already collapses `..`, and the raw substring check can false-reject a legitimate filename that happens to contain two consecutive dots (e.g. `my..cfg.json`).

Minor, non-blocking: `File.expand_path` doesn't resolve symlinks (`File.realpath` would), but that only matters once an attacker can write into `bin/`, so it's just defense-in-depth to consider.

avsej:
Thanks @thejcfactor — good catch on both points. Switched to an exact-directory check:

```ruby
unless File.dirname(resolved_path) == safe_dir
  raise ArgumentError, "Invalid index definition path: outside #{safe_dir}"
end
```

This drops the prefix-match (`bin-evil/` no longer slips through) and removes the `!include?("..")` clause, since `File.expand_path` already collapses `..` and the substring check could false-reject a legit filename like `my..cfg.json`. Added a sibling-prefix case to the embedded test block. The `realpath`/symlink note is noted as defense-in-depth.

thejcfactor (approved):
LGTM 👍

PR #1008, opened 2026-07-21
CXXCBC-817: Fix template injection in GitHub Action

avsej:
Aikido flagged a potential template injection vulnerability in the remove-cluster GitHub Action at line 12. The shell step interpolated inputs.id directly into the command, enabling shell injection.

To mitigate this, map inputs.id to the CLUSTER_ID environment variable and reference it as "$CLUSTER_ID" inside the shell script. This prevents arbitrary shell commands from executing.


thejcfactor (approved):
LGTM 👍

PR #1012, opened 2026-07-22
CXXCBC-797: Fix lookup_in_all_replicas subdoc path capture and total failure handling

avsej:
### Description & Problem
In the `lookup_in_all_replicas` multi-node subdocument read operation, the client concurrently dispatches requests to the active primary and all effective replica nodes, registering asynchronous handlers with Asio. When a response is received, the handler checks if `resp.fields` is empty to handle cases where the document wasn't retrieved at all (e.g. `document_not_found` or access violation).

There were three critical bugs in this workflow:
1. **Dangling Reference Race / Use-After-Free**: To generate a synthetic field-level error entry for replica nodes returning document-level failures, we need the subdocument path requested. However, capturing requested specifications by reference results in a serious memory safety race, as the request structure can be destroyed on the calling thread before the Asio background handler runs.
2. **Synthetic Field Generation**: When `resp.ctx.ec()` is set and `resp.fields` is empty (indicating a document-level failure), the response was previously silently ignored if there were other outstanding replica requests. The fix populates synthetic `lookup_in_entry` entries carrying the requested subdoc paths and the specific error codes to satisfy test and API expectations.
3. **Total Failure Regression**: Unconditionally generating synthetic error fields on all node failures breaks normal `document_not_found` handling. If a document is completely missing from all nodes (both active and replicas), the client would generate synthetic fields for every node and append them to the results. Since the results list is no longer empty, the client would return overall success with synthetic errors instead of the expected `document_not_found` error.

### Solution
1. **Coordination Context Optimization**: Stored requested subdoc commands inside `replica_context` and capture only `ctx` in completion callbacks to guarantee lifetime safety and completely eliminate hot-path vector allocations.
2. **Extracted Helper Method**: Created `append_synthetic_error_fields(...)` as a member function of `replica_context` to encapsulate synthetic field generation, removing duplicate code and preventing path/field drift between replica and active-node handlers.
3. **Track Successful Node Responses**: Added `successful_responses_` inside `replica_context` to track how many nodes successfully retrieved the document.
4. **Enforce Total Failure Behavior**: If the final response completes and `successful_responses_ == 0`, we clear the results vector `ctx->result_`. This guarantees that if all nodes return document-level errors, the library returns the true document error and empty entries (passing standard missing document tests), while still surfacing partial failures correctly.

### Verification
- Added a dedicated test section `"CXXCBC-797: missing document / document-level failure with multiple specs"` in `test_integration_subdoc.cxx` to verify that total document-level failures with multiple requested specs correctly bubble up as `document_not_found` and empty result sets.
- Compiled `test_integration_subdoc` target successfully with zero errors/warnings.
- Formatted modified files using `clang-format`.

### DIFF
--- a/core/operations/document_lookup_in_all_replicas.hxx
+++ b/core/operations/document_lookup_in_all_replicas.hxx
@@ -139,20 +139,44 @@ struct lookup_in_all_replicas_request {
             using handler_type = utils::movable_function<void(response_type)>;
 
             struct replica_context {
-              replica_context(handler_type handler, std::size_t expected_responses)
+              replica_context(handler_type handler,
+                              std::size_t expected_responses,
+                              std::vector<couchbase::core::impl::subdoc::command> specs)
                 : handler_(std::move(handler))
                 , expected_responses_(expected_responses)
+                , specs_(std::move(specs))
               {
+              }
+
+              void append_synthetic_error_fields(
+                lookup_in_all_replicas_response::entry& entry,
+                std::error_code ec,
+                std::optional<key_value_status_code> status_code) const
+              {
+                entry.fields.reserve(specs_.size());
+                for (std::size_t i = 0; i < specs_.size(); ++i) {
+                  const auto& spec = specs_[i];
+                  lookup_in_all_replicas_response::entry::lookup_in_entry error_field{};
+                  error_field.path = spec.path_;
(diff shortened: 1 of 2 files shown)


thejcfactor (approved):
LGTM 👍

PR #1016, opened 2026-07-22
CXXCBC-876: detect and tear down dead idle HTTP connections promptly

avsej:
Pooled HTTP service sessions (query/search/analytics/management) parked between requests armed only an idle stop-timer and kept **no read pending**, and `check_out()` reused a session without a liveness check. When a peer closed an idle keep-alive connection, the SDK did not notice until the idle timer fired (`idle_http_connection_timeout`, default 4.5s) or the dead socket was handed to the next request, which then stalled to the service timeout. Reference SDKs avoid this (Netty `channelInactive`, Go `net/http`, Rust `hyper` pooled-connection liveness).

## Changes

- Keep a read armed while a session is idle so a peer FIN/RST tears it down and removes it from the pool immediately: `set_idle()` marks the session idle and posts `do_read()`; `reset_idle()` clears the flag on checkout so the armed read becomes the response read; `do_read()` treats any completion while idle (EOF/error or unexpected data) as a dead connection and stops the session; `write_and_subscribe()` surfaces a retryable `request_canceled` if the session was stopped between checkout and dispatch.
- Correct two teardown paths for the now-armed read: `update_config()` collects idle sessions to departed nodes under the lock and `stop()`s them afterward (posted per-session executor) instead of only dropping them from the pool; and `close()` calls `stop()` on idle sessions rather than only `reset_idle()` — the latter previously left the armed read pending so the `io_context` never drained, hanging the "destroy cluster without waiting for close" shutdown until CI timeout.
- Adds unit tests: idle peer-close detection, eviction close, manager-close drain.

## Reviewer notes

- The armed idle `do_read()` is posted to `get_executor()`; the `reading_` guard is a plain atomic check-and-set, not a CAS, so correctness relies on all `do_read` invocations for a session being serialized on the same strand — worth an explicit sign-off since an idle read and a response read must never overlap.
- `idle_` is read on the io thread and cleared on a manager thread under `sessions_mutex_`; it is atomic and both observed values funnel to `stop()` + the `request_canceled` guard, but this pool backs **all** HTTP services, so the concurrency deserves a careful look.

Extracted from the CXXCBC-810 stack (PR #930); single commit, `core/io` + new unit tests only, no dependency on the other extracted changes.


test/test_unit_http_session_manager_eviction.cxx
```diff
@@ -0,0 +1,165 @@
...
+      asio::async_write(
+        server, asio::buffer(*resp), [&, rbuf, resp](std::error_code w_ec, std::size_t) {
+          if (w_ec) {
+            return;
+          }
+          server.async_read_some(asio::buffer(*rbuf), [&](std::error_code r2_ec, std::size_t) {
+            if (r2_ec) { // EOF/reset => the SDK closed its end
+              server_saw_close = true;
+            }
+          });
```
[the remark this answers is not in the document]
avsej: Fixed — the final `async_read_some` handler now captures `rbuf` by value, keeping the buffer alive for the duration of the pending read. Verified: the eviction unit test passes.

test/test_unit_http_session_manager_eviction.cxx
thejcfactor, line 1: Cause of the linting failure.

core/io/http_session_manager.hxx
```diff
@@ -470,8 +485,11 @@ public:
     for (auto& [type, sessions] : idle_sessions) {
       for (auto& s : sessions) {
         if (s) {
-          s->reset_idle();
-          s.reset();
+          // stop() (not just reset_idle()) so the read armed by set_idle() is
+          // torn down: reset_idle() only cancels the idle timer, leaving the
+          // pending async read alive, which keeps a shared_ptr to the session
+          // and prevents the io_context from draining on close.
+          s->stop();
```
thejcfactor: Should this be w/in an `asio::post()` so we queu up `stop()` after any strand handlers?
```cpp
asio::post(s->get_executor(), [s]() {
  s->stop();
 });
```

test/test_unit_http_session_manager_eviction.cxx
avsej: Fixed — the eviction test (the final async_read_some handler I re-wrapped) is now clang-formatted; `clang-format --dry-run --Werror` is clean across all changed files. Rebased onto latest main.

core/io/http_session_manager.hxx
avsej: Good call — done. close() now posts stop() to the session's executor (`asio::post(s->get_executor(), [s]{ s->stop(); })`) so it runs after any in-flight strand handler (e.g. the armed read's completion) instead of racing it from the caller thread, matching update_config()'s eviction path. The io_context still drains: the posted stop() runs as an io task, cancels the armed read, and there is no more work left.

core/io/http_session.cxx
```diff
@@ -466,7 +466,7 @@ void
 http_session::set_idle(std::chrono::milliseconds timeout)
 {
   idle_timer_.expires_after(timeout);
-  return idle_timer_.async_wait([self = shared_from_this()](std::error_code ec) {
+  idle_timer_.async_wait([self = shared_from_this()](std::error_code ec) {
     if (ec == asio::error::operation_aborted) {
       return;
     }
```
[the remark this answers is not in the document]
avsej: Fixed — the idle-timeout handler is now bound to the session strand via asio::bind_executor(get_executor(), ...). idle_timer_ runs on the raw io_context in the plaintext constructor, so without this its stop() could run concurrently with the strand-armed liveness read's completion on a multi-threaded io_context; binding serializes idle-timeout teardown with the read/write handlers.

core/io/http_session.cxx
```diff
@@ -466,21 +466,45 @@ void
...
 {
+  // The session is leaving the pool to service a request, so the read armed by
+  // set_idle() is no longer an idle liveness probe -- it becomes the read that
+  // receives the response.  Clear the flag first so do_read() treats any
+  // subsequent completion as normal request/response traffic, not a stale
+  // connection.
+  idle_ = false;
   // Return true if cancel() is successful. Since the idle_timer_ has a single pending
   // wait per session, we know the timer has already expired if cancel() returns 0.
   return idle_timer_.cancel() != 0;
```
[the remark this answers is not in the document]
avsej: Fixed — reset_idle() now restores idle_ when idle_timer_.cancel() returns 0. A zero cancel means the timer already fired and its (strand-bound) handler is about to stop() the session, so it is being torn down rather than checked out; keeping idle_ true leaves the still-armed liveness read on the idle path (which tears the connection down) instead of feeding it to the response parser as request traffic.

thejcfactor (approved):
LGTM 👍

PR #1018, opened 2026-07-22
CXXCBC-878: fit_performer: fix transaction-stream completion hang

avsej:
The `transactionStream` RPC never closed the server stream after writing the final result, so the Java FIT driver parked in its blocking `Read()` (`blockOnResult()`) until the test timed out.

## Change

- Capture the `grpc::ServerContext` into the worker thread and, after writing the final result, call `context->TryCancel()` to close the stream so the driver's blocking read unblocks.

`TryCancel()` surfaces as `CANCELLED` on the driver; this mirrors the reference .NET performer and is intended (see the explaining comment in the code).

Extracted from the CXXCBC-810 stack (PR #930); single commit, `tools/fit_performer/src/service.cxx` only, no dependency on the other extracted changes.

### DIFF
--- a/tools/fit_performer/src/service.cxx
+++ b/tools/fit_performer/src/service.cxx
@@ -1328,7 +1328,7 @@ TxnService::clientRecordProcess(grpc::ServerContext* /*context*/,
 
 grpc::Status
 TxnService::transactionStream(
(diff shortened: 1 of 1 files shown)


thejcfactor (approved):
LGTM 👍

PR #1019, opened 2026-07-22
CXXCBC-879: fit_performer: propagate parallelize batch sub-command errors

avsej:
`execute_command` returns a `couchbase::error`, and the **serial** command path already propagates it (`if (op_err) { return op_err; }`). The **parallel** (`parallelize`) path discarded the result and ended with `return {}`, so a non-throwing operation error inside a parallel batch was silently swallowed — the transaction was neither failed nor rolled back, unlike the serial path.

## Change

- In the parallelize block, capture each sub-command's `couchbase::error`, keep the first non-empty one (under the existing batch lock), and return it so the batch fails/rolls back consistently with the serial path.

Extracted from the CXXCBC-810 stack (PR #930); single commit, `tools/fit_performer/src/service.cxx` only, no dependency on the other extracted changes.

### DIFF
--- a/tools/fit_performer/src/service.cxx
+++ b/tools/fit_performer/src/service.cxx
@@ -1215,6 +1215,12 @@ TxnService::execute_command(ConnectionPtr conn,
     uint32_t in_flight = 0;
     std::mutex mtx;
     std::condition_variable cv;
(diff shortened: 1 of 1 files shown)


thejcfactor (approved):
LGTM 👍

PR #1021, opened 2026-07-22
CXXCBC-881: fit_performer: serialize concurrent cleanup-hook RPCs (fix double-free)

avsej:
The test-only cleanup RPCs (`transactionCleanup`, `transactionCleanupATR`, `clientRecordProcess`) inject per-request hooks by overwriting the shared per-connection core-transactions config's hook `shared_ptr`s (`txn->cleanup().config().cleanup_hooks = …`, `txn->config().attempt_context_hooks = …`) and then immediately run a cleanup operation that reads them. The core transactions object is shared across all concurrent gRPC worker threads on a connection, so concurrent cleanup RPCs move-assign the same `shared_ptr` from multiple threads, releasing its control block twice — double-free / heap corruption.

## Change

- Add a performer-side `cleanup_config_mutex_` and hold it across each set-hooks-then-cleanup critical section (the three RPC handlers). In `clientRecordProcess` the lock is held across `run_with_timeout` so `get_active_clients` observes the hooks set for that request.

Notes: performer-only — uses `main`'s direct `config().cleanup_hooks = …` assignment, so it does not depend on any core change. The unrelated `get_multi_replicas` clang-format reflow from the original branch draft is dropped.

Extracted from the CXXCBC-810 stack (PR #930); single commit, `tools/fit_performer/src/service.{cxx,hxx}` only, no dependency on the other extracted changes.

### DIFF
--- a/tools/fit_performer/src/service.cxx
+++ b/tools/fit_performer/src/service.cxx
@@ -434,6 +434,9 @@ TxnService::transactionCleanup(grpc::ServerContext* /*context*/,
   try {
     auto txn =
       couchbase::core::transactions::get_core_transactions(conn->cluster()->transactions());
+    // Serialize the shared-config mutation and the cleanup that reads it (see
+    // cleanup_config_mutex_).
+    const std::scoped_lock<std::mutex> cleanup_lock(cleanup_config_mutex_);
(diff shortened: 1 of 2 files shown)


thejcfactor (approved):
LGTM 👍

PR #1024, opened 2026-07-28
CXXCBC-886: vendor Protostellar schema and build gRPC/protobuf stubs

avsej:
The couchbase2:// transport needs generated protobuf/gRPC stubs from the
Protostellar schema. Add cmake/Protostellar.cmake, which for VCS builds fetches
the schema via CPM (couchbase/protostellar, DOWNLOAD_ONLY) — the source-tarball
step captures the CPM cache so offline builds need no network — and generates
C++ stubs for the couchbase.* services plus the sole googleapis import,
google/rpc/status.proto (from googleapis/api-common-protos); google.protobuf
well-known types come from protoc. The output is the couchbase_cxx_protostellar
static library, linking gRPC::grpc++ and protobuf::libprotobuf and exposing the
generated headers by proto path (e.g. <couchbase/kv/v1/kv.pb.h>).

Gated by COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2 (default OFF, opt-in until a
transport consumer lands). When enabled, cmake/GrpcProtobuf.cmake is included
before OpenTelemetry, matching the FIT performer path, so protobuf targets are
shared. A codegen smoke test in the CNG harness instantiates a message and a
service type to prove the stubs compile and link.

---
Stacked PR **02/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–02; the change specific to this PR is its top commit. Depends on #1023 — merge the series bottom-up.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,12 @@ jobs:
         timeout-minutes: 60
         env:
           CB_NUMBER_OF_JOBS: 8
+          # Compile the generated Protostellar stubs on Apple clang too. These legs already build
+          # gRPC and protobuf from source for fit_performer -- BUILD_FIT_PERFORMER defaults to
+          # BUILD_TOOLS, so ThirdPartyDependencies.cmake pulls in GrpcProtobuf.cmake regardless --
+          # so the only added cost is generating and compiling the schema. Apple clang is otherwise
+          # never exercised against the generated code.
+          CB_COUCHBASE2: 1
         run: ./bin/build-tests
   windows:
     strategy:
@@ -69,4 +75,11 @@ jobs:
           # the runner's processor count via Etc.nprocessors (4 on windows-2022/2025).
           # Combined with /MP added in CMakeLists.txt this enables both axes of
           # parallelism instead of the previous /m:2 + serial-cl.exe cap.
+          #
+          # Compile the generated Protostellar stubs on MSVC. This leg already builds gRPC and
+          # protobuf from source for fit_performer, so the only added cost is generating and
+          # compiling the 14 services -- and MSVC is the only toolchain where the generated
+          # protobuf/gRPC code needs special handling (see the MSVC branch of
+          # cmake/Protostellar.cmake), which nothing else in CI would exercise.
+          CB_COUCHBASE2: 1
         run: ruby ./bin/build-tests.rb
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,6 +41,11 @@ jobs:
         run: ./bin/build-tests
         env:
           CB_NUMBER_OF_JOBS: 8
+          # Generate and compile the Protostellar gRPC/protobuf stubs, and the CNG tests that use
+          # them. The gRPC/protobuf dev packages this needs are installed above. Without it the
+          # generated stubs are never built by CI, so codegen breakage would only ever surface
+          # locally; later commits add the couchbase2:// transport on top of the same option.
+          CB_COUCHBASE2: 1
       - name: Upload build artifacts
         if: ${{ success() }}
         uses: actions/upload-artifact@v4
@@ -56,7 +61,10 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y libssl-dev cmake gcc g++ curl gdb
+          # libgrpc++-dev/libprotobuf-dev pull in the gRPC runtime needed to *run* the gRPC-linked
+          # binaries in the prebuilt artifact -- currently cng_codegen_smoke_test. (A later commit
+          # links the transport into the client library itself, which extends this to every binary.)
+          sudo apt-get install -y libssl-dev cmake gcc g++ curl gdb libprotobuf-dev libgrpc-dev libgrpc++-dev
       - uses: actions/checkout@v4
         with:
           submodules: recursive
@@ -72,6 +80,45 @@ jobs:
         run: |
           chmod -R +x ./cmake-build-tests
           ./bin/run-unit-tests
+      - name: Run CNG tests
+        # The CNG tests carry LABELS "cng" (test/cng/CMakeLists.txt), which matches none of the
+        # --label-regex filters in bin/run-{unit,integration,transaction,benchmark}-tests. Without
+        # this step they are compiled and linked by the build job but never executed by any job.
+        #
+        # --no-tests=error is load-bearing: a --label-regex that matches nothing makes ctest exit 0
+        # (measured: exit 0 without the flag, 8 with it), so a renamed label or a BUILD_CNG_TESTS
+        # that silently went OFF would leave this step passing vacuously -- the same "green means
+        # nothing ran" failure this step exists to close. Needs ctest >= 3.20; CI has 3.28.3.
+        #
+        # --output-junit resolves relative to --test-dir, so the report lands beside the build tree.
+        # The name deliberately differs from the results.xml that bin/run-unit-tests has already
+        # written into the same directory, and it is uploaded as its own artifact rather than folded
+        # into unit-test-results: a CNG test is not a unit test.
+        #
+        # always(): a unit-test failure must not mask CNG results.
+        if: ${{ always() }}
+        timeout-minutes: 10
+        env:
+          TEST_LOG_LEVEL: trace
+        run: |
+          chmod -R +x ./cmake-build-tests
+          ctest --test-dir ./cmake-build-tests --output-on-failure --label-regex 'cng' \
+                --no-tests=error --output-junit cng-results.xml
+      - name: Upload test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: unit-test-results
+          # bin/run-unit-tests cds into the build dir before ctest, so --output-junit lands there.
+          path: ./cmake-build-tests/results.xml
+          if-no-files-found: warn
+      - name: Upload CNG test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: cng-test-results
+          path: ./cmake-build-tests/cng-results.xml
+          if-no-files-found: warn
   integration:
     needs: build
     strategy:
@@ -92,7 +139,10 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y libssl-dev cmake gcc g++ curl gdb
+          # libgrpc++-dev/libprotobuf-dev pull in the gRPC runtime needed to *run* the gRPC-linked
+          # binaries in the prebuilt artifact -- currently cng_codegen_smoke_test. (A later commit
+          # links the transport into the client library itself, which extends this to every binary.)
+          sudo apt-get install -y libssl-dev cmake gcc g++ curl gdb libprotobuf-dev libgrpc-dev libgrpc++-dev
       - uses: actions/checkout@v4
         with:
           submodules: recursive
--- a/cmake/Protostellar.cmake
+++ b/cmake/Protostellar.cmake
@@ -1,0 +1,197 @@
+# Vendors the Couchbase Protostellar (couchbase2://) protobuf/gRPC schema and builds the generated
+# C++ stubs into the couchbase_cxx_protostellar static library.
+#
+# For VCS builds the schema is fetched via CPM (DOWNLOAD_ONLY, no CMakeLists), mirroring how
+# tools/fit_performer vendors couchbaselabs/fit-protocol.
+#
+# NOTE: offline/tarball builds are NOT yet fully wired for couchbase2. Before this feature ships
+# enabled, the source-tarball step must (a) enable COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2 in the
+# inner packaging configure so these CPM packages are fetched into the cache, and (b) add
+# protostellar/api_common_protos glob entries to cmake/tarball_glob.txt so they are embedded in
+# the tarball (mirroring fit_protocol). Until then a FETCHCONTENT_FULLY_DISCONNECTED build with
+# this option ON will fail. To be addressed as part of the ship-enable work.
+#
+# gRPC::grpc++, gRPC::grpc_cpp_plugin, protobuf::libprotobuf and protobuf::protoc are provided by
+# cmake/GrpcProtobuf.cmake, which CMakeLists.txt includes before this module.
+
+if(NOT TARGET gRPC::grpc++)
+  message(FATAL_ERROR "gRPC::grpc++ not found; cmake/GrpcProtobuf.cmake must be included first")
+endif()
+if(NOT TARGET protobuf::libprotobuf)
+  message(FATAL_ERROR "protobuf::libprotobuf not found; cmake/GrpcProtobuf.cmake must be included first")
+endif()
+
+# The Protostellar schema. The .proto files live at the repository root under couchbase/**.
+# WARNING: do NOT add GIT_SHALLOW — a shallow clone cannot reach an arbitrary pinned commit.
+cpmaddpackage(
+  NAME
+  protostellar
+  GITHUB_REPOSITORY
+  "couchbase/protostellar"
+  GIT_TAG
+  36d4d9aba03a388f7edb05290f1093d9497b1005
+  DOWNLOAD_ONLY
+  YES)
+
+# The schema imports google/rpc/status.proto (a googleapis common proto that protobuf itself does
+# not ship). api-common-protos is the small canonical source for it. The google/protobuf/* imports
+# (any, duration, timestamp) are well-known types resolved by protoc from the protobuf install.
+cpmaddpackage(
+  NAME
+  api_common_protos
+  GITHUB_REPOSITORY
+  "googleapis/api-common-protos"
+  GIT_TAG
+  3332dec527759859840a3a2ff108c67a54708130
+  DOWNLOAD_ONLY
+  YES)
+
+set(_ps_out "${CMAKE_CURRENT_BINARY_DIR}/protostellar_generated")
+
+if(TARGET gRPC::grpc_cpp_plugin)
+  set(_ps_grpc_plugin "$<TARGET_FILE:gRPC::grpc_cpp_plugin>")
+  # Depend on the target so it is built before codegen runs.
+  set(_ps_grpc_plugin_dep gRPC::grpc_cpp_plugin)
+else()
+  find_program(_ps_grpc_plugin grpc_cpp_plugin REQUIRED)
+  # No target to depend on; the discovered executable path is the dependency.
+  set(_ps_grpc_plugin_dep "${_ps_grpc_plugin}")
+endif()
+
+set(_ps_import_flags -I "${protostellar_SOURCE_DIR}" -I "${api_common_protos_SOURCE_DIR}")
+# When protobuf is built from source (not a system package) protoc has no built-in path to the
+# well-known .proto files; GrpcProtobuf.cmake sets PROTOBUF_IMPORT_DIRS in that case.
+if(DEFINED PROTOBUF_IMPORT_DIRS)
+  foreach(_dir ${PROTOBUF_IMPORT_DIRS})
+    list(APPEND _ps_import_flags -I "${_dir}")
+  endforeach()
+endif()
+
+# Service protos (generate both message and gRPC stubs), relative to the Protostellar root.
+set(_ps_service_protos
+    couchbase/admin/analytics/v1/analytics.proto
+    couchbase/admin/bucket/v1/bucket.proto
+    couchbase/admin/collection/v1/collection.proto
+    couchbase/admin/query/v1/query.proto
+    couchbase/admin/search/v1/search.proto
+    couchbase/analytics/v1/analytics.proto
+    couchbase/internal/hooks/v1/hooks.proto
+    couchbase/internal/xdcr/v1/xdcr.proto
+    couchbase/kv/v1/kv.proto
+    couchbase/query/v1/query.proto
+    couchbase/routing/v2/routing.proto
+    couchbase/search/v1/search.proto
+    couchbase/transactions/v1/transactions.proto
+    couchbase/view/v1/view.proto)
+
+set(_ps_srcs "")
+set(_ps_hdrs "")
+
+# Emit the protobuf (message) sources for one proto given its import-root base and relative path.
+# Appends the outputs to _ps_srcs / _ps_hdrs in the caller scope.
+function(_ps_generate_cpp base rel out_srcs out_hdrs)
+  get_filename_component(_rel_dir "${rel}" DIRECTORY)
+  get_filename_component(_rel_we "${rel}" NAME_WE)
+  file(MAKE_DIRECTORY "${_ps_out}/${_rel_dir}")
+  set(_cc "${_ps_out}/${_rel_dir}/${_rel_we}.pb.cc")
+  set(_h "${_ps_out}/${_rel_dir}/${_rel_we}.pb.h")
+  add_custom_command(
+    OUTPUT "${_cc}" "${_h}"
+    COMMAND protobuf::protoc
+            --cpp_out "${_ps_out}" ${_ps_import_flags} "${base}/${rel}"
+            --experimental_allow_proto3_optional
+    DEPENDS "${base}/${rel}" protobuf::protoc
+    COMMENT "protoc cpp: ${rel}"
+    VERBATIM)
+  set(${out_srcs} ${${out_srcs}} "${_cc}" PARENT_SCOPE)
+  set(${out_hdrs} ${${out_hdrs}} "${_h}" PARENT_SCOPE)
+endfunction()
+
+# Emit the gRPC service sources for one proto.
+function(_ps_generate_grpc base rel out_srcs out_hdrs)
+  get_filename_component(_rel_dir "${rel}" DIRECTORY)
+  get_filename_component(_rel_we "${rel}" NAME_WE)
+  file(MAKE_DIRECTORY "${_ps_out}/${_rel_dir}")
+  set(_cc "${_ps_out}/${_rel_dir}/${_rel_we}.grpc.pb.cc")
+  set(_h "${_ps_out}/${_rel_dir}/${_rel_we}.grpc.pb.h")
+  add_custom_command(
+    OUTPUT "${_cc}" "${_h}"
+    COMMAND protobuf::protoc
+            --grpc_out "${_ps_out}" ${_ps_import_flags}
+            "--plugin=protoc-gen-grpc=${_ps_grpc_plugin}" "${base}/${rel}"
+            --experimental_allow_proto3_optional
+    DEPENDS "${base}/${rel}" protobuf::protoc ${_ps_grpc_plugin_dep}
+    COMMENT "protoc grpc: ${rel}"
+    VERBATIM)
+  set(${out_srcs} ${${out_srcs}} "${_cc}" PARENT_SCOPE)
+  set(${out_hdrs} ${${out_hdrs}} "${_h}" PARENT_SCOPE)
+endfunction()
+
+# google/rpc/status.proto — message stubs only (no service).
+_ps_generate_cpp("${api_common_protos_SOURCE_DIR}" "google/rpc/status.proto" _ps_srcs _ps_hdrs)
+
+# google/rpc/error_details.proto — the typed detail blocks (ErrorInfo, PreconditionFailure,
+# ResourceInfo, …) that CNG packs into Status.details; RFC 77 error handling needs them decoded.
+_ps_generate_cpp(
+  "${api_common_protos_SOURCE_DIR}" "google/rpc/error_details.proto" _ps_srcs _ps_hdrs)
+
+foreach(_proto ${_ps_service_protos})
+  _ps_generate_cpp("${protostellar_SOURCE_DIR}" "${_proto}" _ps_srcs _ps_hdrs)
+  _ps_generate_grpc("${protostellar_SOURCE_DIR}" "${_proto}" _ps_srcs _ps_hdrs)
+endforeach()
+
+set_source_files_properties(${_ps_srcs} ${_ps_hdrs} PROPERTIES GENERATED TRUE)
+
+add_library(couchbase_cxx_protostellar STATIC ${_ps_srcs} ${_ps_hdrs})
+add_library(couchbase::cxx_protostellar ALIAS couchbase_cxx_protostellar)
+# Consumers include the generated headers by their proto path, e.g.
+# <couchbase/kv/v1/kv.pb.h>. Exposed SYSTEM so the (warning-heavy) generated code does not trip
+# the project's warnings-as-errors in downstream targets.
+target_include_directories(couchbase_cxx_protostellar SYSTEM PUBLIC "${_ps_out}")
+target_link_libraries(couchbase_cxx_protostellar PUBLIC gRPC::grpc++ protobuf::libprotobuf)
+set_target_properties(
+  couchbase_cxx_protostellar
+  PROPERTIES POSITION_INDEPENDENT_CODE ON
+             COMPILE_WARNING_AS_ERROR OFF
+             CXX_CLANG_TIDY ""
+             CXX_INCLUDE_WHAT_YOU_USE "")
+# NOTE (ABI visibility): nothing links this library into couchbase_cxx_client yet -- at this commit
+# it stands alone and only the codegen smoke test links it. Once a later commit embeds the transport
+# in the client, the generated protobuf/gRPC symbols will land in the exported ABI of
+# libcouchbase_cxx_client.so with default visibility, which can clash with an application linking
+# its own protobuf/gRPC. The fix is hidden visibility here + --exclude-libs on the shared client
+# target, BUT that will require reworking how the white-box CNG tests link (they will link this
+# static lib AND the client lib that embeds it; default visibility merges the two protobuf
+# descriptor copies into one, hidden visibility does not -> duplicate descriptor registration
+# abort). Deferred to the ship-enable ticket; the option is OFF by default.
+if(MSVC)
+  # /bigobj: the generated translation units are large -- search.pb.cc is 18k lines and yields a
+  # 7.4 MB object -- and protobuf-generated code this size can exceed the COFF section limit
+  # (fatal error C1128). CI builds RelWithDebInfo, but a Debug MSVC build (the default for
+  # bin/build-tests) produces markedly larger objects. tools/fit_performer gets this flag from
+  # set_project_options() for the same class of file; this target does not call that function.
+  #
+  # Do NOT add the /FI<proto_utils.h> force-include that tools/fit_performer uses. It was tried
+  # here and breaks the build: proto_utils.h reaches grpc/support/port_platform.h, which includes
+  # <windows.h> unconditionally ("Get windows.h included everywhere (we need it)"), so forcing it
+  # ahead of the translation unit's own includes puts winnt.h's STATUS_TIMEOUT macro in scope before
+  # couchbase/query/v1/query.pb.h is parsed. That header declares an enum constant of the same name,
+  # so
+  #   static constexpr Status STATUS_TIMEOUT = ...
+  # expands to ((DWORD)0x00000102L): C2143, plus a member named DWORD (C2789). query.pb.h is the
+  # only generated header carrying such a collision. In normal include order it is parsed before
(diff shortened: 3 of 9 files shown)


test/cng/framework/test_runner.cxx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Run one case on a worker thread, enforcing `timeout`. On timeout the worker is detached: the
+// packaged_task moved into it co-owns the future's shared state, so the detached thread writes
+// only to that heap state (never to this stack frame) and the result is simply never read.
+auto
+run_case(void (*func)(), std::chrono::milliseconds timeout) -> case_result
+{
+  std::packaged_task<void()> task(func);
+  auto future = task.get_future();
```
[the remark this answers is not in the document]
avsej: The detach is deliberate: a C++ thread running arbitrary test code cannot be safely cancelled, so on timeout we detach rather than block teardown. The worker writes only to the packaged_task's heap-owned shared state (never the test's stack frame) and the timed-out result is simply never read (see the comment on run_case). Keeping as-is.

cmake/Protostellar.cmake
```diff
@@ -0,0 +1,165 @@
...
+  set(_h "${_ps_out}/${_rel_dir}/${_rel_we}.grpc.pb.h")
+  add_custom_command(
+    OUTPUT "${_cc}" "${_h}"
+    COMMAND protobuf::protoc
+    ARGS --grpc_out "${_ps_out}" ${_ps_import_flags}
+         "--plugin=protoc-gen-grpc=${_ps_grpc_plugin}" "${base}/${rel}"
+         --experimental_allow_proto3_optional
+    DEPENDS "${base}/${rel}" protobuf::protoc gRPC::grpc_cpp_plugin
+    COMMENT "protoc grpc: ${rel}"
+    VERBATIM)
```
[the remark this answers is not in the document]
avsej: Intentional: this codegen drives protoc directly with the couchbase2-specific import roots (api-common-protos for google/rpc/*) and --experimental_allow_proto3_optional, emitting into a private tree. Folding it into protobuf_Generate/gRPC_Generate would couple those helpers to this schema's needs. Keeping as-is.

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. Mixed verification, flagged inline: the codegen inventory and import closure were checked against both pinned commits via the GitHub API; the vendoring/codegen/link path is confirmed by this PR's own green `build` leg plus a local `CB_COUCHBASE2=1` build; and finding 1 is confirmed by local `ctest` runs cross-checked against this PR's `unit` job log. Two exceptions, both flagged where they land: finding 2 I could **not** execute at all — I have no MSVC, and no CI leg builds couchbase2 on Windows — and finding 4 is a question rather than a measurement I made. Please treat the reasoning as a starting point for discussion, not ground truth.

Reviewed at `f81419cd`, incrementally against `main`@`488c7502` (#1023 having merged), so this is just the one commit.

The codegen module holds up well, and I'd rather be specific than say "looks good":

- The 14 entries in `_ps_service_protos` are **exactly** the 14 `.proto` files present in `couchbase/protostellar`@`36d4d9ab` — nothing missing, nothing stale.
- The import closure is complete. `google/rpc/status.proto` and `error_details.proto` are generated, `google/protobuf/{any,duration,timestamp}.proto` are well-known types from libprotobuf, and `couchbase/kv/v1/kv.proto` — imported by `admin/bucket/v1` and `query/v1` — is in the generated set. `google/rpc/code.proto` is correctly *not* generated. No dangling import.
- Generating `error_details.proto` even though nothing imports it is the right call and worth the comment you gave it — `Status.details` is `repeated Any`, so the concrete detail types are needed to unpack it.
- Hand-rolling `_ps_generate_cpp`/`_ps_generate_grpc` rather than reusing `protobuf_generate_cpp_optional`/`grpc_generate_cpp` is correct, and not obviously so: those flatten output to `${DEST}/<stem>.pb.h`, which would collide on this schema's three duplicate stems (`analytics`, `query`, `search` each appear under both `couchbase/admin/*/v1/` and `couchbase/*/v1/`) and would break the `<couchbase/kv/v1/kv.pb.h>` proto-path convention, which needs a single `-I` at the schema root.
- The `GIT_SHALLOW` warning and the `PROTOBUF_IMPORT_DIRS` branch both match how `GrpcProtobuf.cmake` actually behaves.
- The offline/tarball `NOTE` is accurate, and I appreciate it being in the module rather than only in the PR description — `cmake/tarball_glob.txt:30` does carry only `fit_protocol/*` globs, and the option being `OFF` means tarball builds are unaffected today.

Five things below. Findings 1 and 2 are the ones I'd like resolved before this merges; neither is a reason to redesign anything.

---

## 1. `major` — CI now compiles the CNG tests but no CI job ever runs them

`add_cng_test` tags each test `LABELS "cng"` (`test/cng/CMakeLists.txt:45`, from #1023). Every CI runner filters by label:

| Script | Filter |
|---|---|
| `bin/run-unit-tests:40` | `--label-regex 'unit'` |
| `bin/run-integration-tests:121` | `--label-regex 'integration'` |
| `bin/run-transaction-tests:34` | `--label-regex 'transaction'` |
| `bin/run-benchmark-tests:30` | `--label-regex 'benchmark'` |

`cng` matches none of them. `bin/build-tests` builds the default `all` target, so `cng_codegen_smoke_test` is compiled and **linked** in CI — which genuinely does prove the stubs generate and resolve — but its three assertions never execute in any CI job.

Confirmed, from a local `CB_COUCHBASE2=1` build at `f81419cd`:

```console
$ ctest -N --label-regex 'cng'
  Test #708: cng_framework_selftest_test
  Test #709: cng_codegen_smoke_test
Total Tests: 2

$ ctest -N --label-regex 'unit' | grep -c cng_
0

$ ctest -R 'cng_' --output-on-failure
1/2 Test #708: cng_framework_selftest_test ......   Passed    0.38 sec
2/2 Test #709: cng_codegen_smoke_test ...........   Passed    0.58 sec
100% tests passed, 0 tests failed out of 2
```

And on this PR's own `unit` job, ctest ran **364 tests with `cng` appearing nowhere in the job log**.

Two things follow that I think make this straightforward rather than contentious. Both tests **pass** when invoked, so this is purely a wiring gap and not a test that's quietly broken. And they cost **1.01 s combined**, so there's no runtime budget argument for keeping them out of a leg that already spends 17.6 s.

What makes me read it as an oversight rather than a decision is the other half of your own diff: you added `libprotobuf-dev libgrpc-dev libgrpc++-dev` to the **`unit` and `integration`** jobs. Those jobs only download a prebuilt artifact and run ctest. The only reason they'd need a gRPC *runtime* is to execute a gRPC-linked test binary — and under the label filter they never do. That change reads as though written expecting the smoke test to run in the `unit` leg.

Two reasons to fix it here rather than later:

- **It scales.** This is PR 02 of 29. Every CNG test added downstream (#1030–#1042 especially) inherits the same label and is silently excluded. By PR 12 there'd be a substantial CNG suite that has never run, and nothing in the output would say so.
- **It's the #1023 "silently green" problem one level up.** That PR was hardened so a binary executing nothing exits non-zero. The guard can't fire if the binary is never invoked — ctest reports no matching tests and the job goes green.

Smallest fix, if you want them in the existing `unit` leg (where the apt packages you added would then be doing real work):

```cmake
# test/cng/CMakeLists.txt:45 — only the label string changes
set_tests_properties(${target} PROPERTIES SKIP_RETURN_CODE 77 LABELS "cng;unit")
```

Or keep `cng` distinct and add a workflow step — `ctest --label-regex 'cng' --output-on-failure` — which keeps CNG failures attributable and lets the leg be gated on `CB_COUCHBASE2`.

One incidental ask while you're in there: the `unit` job passes `--output-junit results.xml` but never uploads it, which is why I had to reconstruct the selection locally rather than read it off the run. Uploading it would make this class of question answerable from CI directly.

---

## 2. `major` — the MSVC branch is missing the `proto_utils.h` force-include that this repo already needed for generated gRPC sources

`cmake/Protostellar.cmake:165-169` gives MSVC only `/W0 /FS`. But `tools/fit_performer/CMakeLists.txt:119-126` documents a second, non-obvious MSVC requirement for exactly this class of file:

> On MSVC, `proto_utils.h` must be force-included before gRPC streaming headers to work around non-conformant two-phase lookup that causes `SerializationTraits` to be undefined when `call_op_set.h` is processed before `proto_utils.h` **in generated gRPC source files.**

fit_performer compiles its generated sources as part of the `fit_performer` target (`:111`), so the `/FI${_msvc_grpc_fix_header}` at `:126` covers them. `couchbase_cxx_protostellar` compiles the same class of file — 14 services' worth of `*.grpc.pb.cc` — in its own target, and carries no equivalent.

The trigger is streaming RPCs, and this schema has them: `couchbase/query/v1/query.proto:25` is `rpc Query(QueryRequest) returns (stream QueryResponse)`, with analytics, search and view alike. So the generated sources will pull in the streaming headers that provoke it.

I can't execute this one — no MSVC here — so I'm not claiming it definitely fails. What I am claiming is that the repo has a documented workaround for this precise file class and toolchain, and the new module doesn't carry it, and **nothing in CI would tell you either way**, because `CB_COUCHBASE2` is set in exactly one place: the `build` job of `tests.yml`. That's ubuntu-24.04, Debug, gcc. The `windows (windows-2022)` and macOS legs, and the sanitizer builds, never define it, so their green ticks say nothing about couchbase2. The codegen path is currently exercised on one platform in one configuration.

Suggested: lift the same `/FI` shape into `Protostellar.cmake`'s MSVC branch, and consider setting `CB_COUCHBASE2` on the Windows build leg while the surface is still 170 lines of cmake. Three lines now versus an archaeology exercise at PR 25, when the option gets flipped on for ship. (I checked whether `/bigobj` was also needed — fit_performer doesn't use it, so I'm not asking for it.)

---

## 3. `minor` — three new comments describe a state this commit doesn't have

Each reads as a statement of fact about the tree as it stands, and none is true yet:

- `.github/workflows/tests.yml:44-46` — *"Without it CI never compiles `core/protostellar/**` at all"*. There is no `core/protostellar/` in this commit; `CB_COUCHBASE2` here compiles the generated stubs library and the smoke test.
- `.github/workflows/tests.yml` (both apt comments) — *"with `CB_COUCHBASE2` the shared client library has a NEEDED entry for `libgrpc++`"*. Nothing links `couchbase_cxx_protostellar` or gRPC into `couchbase_cxx_client`; `include(cmake/Protostellar.cmake)` only defines the target. The only gRPC-linked artifact is `cng_codegen_smoke_test`.
- `cmake/Protostellar.cmake:161-162` — *"today they link this static lib AND the client lib that embeds it"*. No CNG test links the client library here.

These are all true of your full branch, so this is a small thing. The reason I'd still change them: the wrong cause is the one that *survives* fixing finding 1. If the apt packages are there for the smoke test's own `NEEDED` entries, that justification stays valid once the test runs; the client-library justification won't be true until a later PR, and meanwhile a reviewer who checks it finds it false and loses track of which claims still need verifying. Present tense → "a later commit will…" fixes all three.

---

## 4. `minor` — the couchbase2 build products ride the artifact to 9 jobs that can't use them

`bin/build-tests` builds `Debug` by default, and the `build` job uploads all of `./cmake-build-tests`. With `CB_COUCHBASE2=1` that now carries Debug objects for 14 services' worth of protobuf + gRPC stubs plus `libcouchbase_cxx_protostellar.a`. It's downloaded by the `unit` job and by the `integration` matrix — 4 server versions × 2 suites = 8 more jobs — none of which link any of it (and, per finding 1, none of which run the one test that does).

I haven't measured this, which is really the ask. I'd *not* trim `_ps_service_protos` — the client needs all 14 eventually and churning that list across 27 PRs would cost more than the bytes. But could you check what it did to artifact size and to the `build` leg's wall clock before it's baked into the rest of the stack? If it's material, `path:` on the upload step could exclude `protostellar_generated/`, or the integration matrix could take a slimmer artifact.

---

## 5. `nit` — small things in the smoke test

- **`assert_true(sizeof(...) > 0)`** (`codegen_smoke.cxx:69`) asserts a compile-time property at runtime. `static_assert(sizeof(::couchbase::kv::v1::KvService::Stub) > 0)` says what you mean, can't be filtered out, and — while finding 1 stands — is the only form that actually executes. The real check is that an incomplete type fails to compile, which holds either way.
- **All three cases take the 5 s `network` default timeout**, because they supply only name and function and `default_timeout = timeout::network` (`test/cng/framework/test_framework.hxx:113,121`). Three in-memory assertions want `timeout::instant` (`:111`, 500 ms); expressing that is rather the point of the presets #1023 added.
- **`resp.set_content_uncompressed(...)`** (`:47`) is set and never read, while `cas` on the next line is asserted. Either assert it or drop it.
- **`target_link_libraries(cng_codegen_smoke_test PRIVATE couchbase_cxx_protostellar)`** (`test/cng/CMakeLists.txt:54`) reaches past `add_cng_test` to a target name the function derives internally, and uses the raw name rather than the `couchbase::cxx_protostellar` alias you defined for it. Worth raising because `CMakeLists.txt:655` still advertises a `LINK_CLIENT` argument to `add_cng_test` that doesn't exist — this PR shows what the linking pattern is actually turning out to be, so it may be time to either add the argument or fix that comment.

avsej:
> **Corrected below.** Three statements in this comment do not hold: the `/FI` force-include it
> describes was reverted and has since been shown unnecessary, the streaming-RPC list is incomplete
> (10, not 6), and the alias call-site counts are wrong (31 raw-name uses, not 16). See the
> [errata](https://github.com/couchbase/couchbase-cxx-client/pull/1024#issuecomment-5133680686)
> later in this thread. The original text is left unchanged below.

All five findings and all four nits are addressed below. Each is quoted, followed by the resolution
and the evidence for it. Two items were resolved differently from the suggestion, and one was
deferred; both are stated explicitly, as is the one change that remains unverified.

## 1. CI now compiles the CNG tests but no CI job ever runs them

> `cng` matches none of them. `bin/build-tests` builds the default `all` target, so
> `cng_codegen_smoke_test` is compiled and **linked** in CI — which genuinely does prove the stubs
> generate and resolve — but its three assertions never execute in any CI job.

Confirmed. `LABELS "cng"` is set at `test/cng/CMakeLists.txt:45`, and the four runners filter on
`unit` / `integration` / `transaction` / `benchmark` at `run-unit-tests:40`,
`run-integration-tests:121`, `run-transaction-tests:34` and `run-benchmark-tests:30`. None of those
regexes matches `cng`.

> Smallest fix, if you want them in the existing `unit` leg (where the apt packages you added would
> then be doing real work):
> ```cmake
> set_tests_properties(${target} PROPERTIES SKIP_RETURN_CODE 77 LABELS "cng;unit")
> ```
> Or keep `cng` distinct and add a workflow step

The second option was chosen: a `Run CNG tests` step in the `unit` job invoking
`ctest --label-regex 'cng'`. Relabelling was rejected for two reasons — a CNG test is not a unit
test, and `bin/run-unit-tests` writes its `--output-junit` as *unit* results, so the label would
file CNG outcomes under the wrong suite. The live tests arriving later in the stack would also
appear as permanently skipped entries in the unit leg. The step is `if: always()`, so a unit-test
failure cannot mask CNG results.

> It's the #1023 "silently green" problem one level up. That PR was hardened so a binary executing
> nothing exits non-zero. The guard can't fire if the binary is never invoked

Accurate, and the two are the same failure mode at different levels: neither is visible in a green
tick.

> One incidental ask while you're in there: the `unit` job passes `--output-junit results.xml` but
> never uploads it

Added. One correction to the mechanics: `bin/run-unit-tests` performs `cd "${BUILD_DIR}"` before
ctest (`:38`), so the file is written to `cmake-build-tests/results.xml` rather than the repository
root. The upload path reflects that.

## 2. The MSVC branch is missing the `proto_utils.h` force-include

> What I am claiming is that the repo has a documented workaround for this precise file class and
> toolchain, and the new module doesn't carry it

Confirmed on all three premises. `_ps_generate_grpc` emits `.grpc.pb.cc` into the target
(`Protostellar.cmake:115`); the MSVC branch carried only `/W0 /FS`; and
`tools/fit_performer/CMakeLists.txt:119-126` writes a header containing
`#include <grpcpp/impl/proto_utils.h>` and force-includes it via `/FI`. The same construct is now
present in `Protostellar.cmake`.

The complete set of streaming RPCs in the pinned schema is `Query`, `AnalyticsQuery`, `SearchQuery`,
`ViewQuery`, `WatchBarrier` and `WatchRouting`.

> nothing in CI would tell you either way, because `CB_COUCHBASE2` is set in exactly one place: the
> `build` job of `tests.yml`. That's ubuntu-24.04, Debug, gcc. […] consider setting `CB_COUCHBASE2`
> on the Windows build leg while the surface is still 170 lines of cmake.

Done, and the cost is smaller than it appears. The Windows leg already builds gRPC and protobuf from
source for fit_performer — 71k `grpc-build` lines and 406 `fit_performer` mentions in the
windows-2022 log — because `ThirdPartyDependencies.cmake:42-46` includes `GrpcProtobuf.cmake` when
*either* fit_performer *or* couchbase2 is enabled. The expensive dependency is therefore already
paid for, and the only addition is generating and compiling the 14 services. Enabling it also
required adding the `CB_COUCHBASE2` switch to `bin/build-tests.rb`, since the Windows legs use the
Ruby path while `tests.yml` uses the shell one.

On verification: this change has **not** been executed on MSVC — no MSVC toolchain was available
here either. What has been done is to reproduce a workaround the repository already demonstrates it
requires for the identical file class, and to arrange for CI to exercise it. If the `/FI` is wrong or
insufficient, the Windows legs will now report it.

## 3. Three new comments describe a state this commit doesn't have

> Each reads as a statement of fact about the tree as it stands, and none is true yet

All three are false at this commit and have been rewritten. Verification:

- `core/protostellar/` contains **0 files** at this commit.
- `couchbase_cxx_protostellar` appears in exactly two places: its definition in
  `Protostellar.cmake`, and `target_link_libraries(cng_codegen_smoke_test …)`. Nothing links it into
  `couchbase_cxx_client`, so the client library carries no gRPC `NEEDED` entry here.

The apt-package comment had been justified by a `readelf` check showing
`libcouchbase_cxx_client.so.1` requiring `libgrpc++`. That check was performed at the **tip of the
stack**, where a later commit links the transport into the client, and it does not hold at this
commit. The comment now states the reason that is true here — the smoke-test binary's own `NEEDED`
entries — which is also the justification that survives fixing finding 1.

The `Protostellar.cmake` ABI note is reworded to future tense. The top-level `LINK_CLIENT` sentence
is removed: `cmake_parse_arguments(ARG "LINK_CLIENT" …)` first appears in **commit 03**, so the claim
was one commit early. The replacement wording holds both before and after that point and needs no
later revision.

## 4. The couchbase2 build products ride the artifact to 9 jobs that can't use them

> I haven't measured this, which is really the ask. […] could you check what it did to artifact size
> and to the `build` leg's wall clock before it's baked into the rest of the stack?

Measured.

**Artifact size: 2303.6 → 2339.7 MiB, +36.1 MiB (+1.6%).** Locally the couchbase2 contribution is
~187 MB of a 9.9 GB Debug tree (generated sources 9.7 MB, objects 86 MB,
`libcouchbase_cxx_protostellar.a` 91 MB); the uploaded delta is far smaller because the artifact is
compressed.

**Build wall clock: +2 to +3 min (~10%).** A direct A/B of the two runs suggests 25.5 → 52.4 min
(+105%), but that figure does not survive a wider sample:

| branch | couchbase2 | build |
|---|---|---|
| CXXCBC-836 (non-CNG) | no | 25.1, 26.1 min |
| 03, 04, 06, 15, 16, 17, 22, 23 | yes | 26.9–28.8 min |
| 10 | yes | 52.5 min |
| 1024 | yes | 52.4 min |

Eight branches carrying the identical couchbase2 configuration, submitted in the same batch,
completed in 26.9–28.8 min. The two ~52-minute runs are therefore unexplained variance rather than a
cost of the option.

Neither figure is material, so the upload path is unchanged and `protostellar_generated/` is not
excluded. `_ps_service_protos` is likewise left intact.

## 5. Nits

> **`assert_true(sizeof(...) > 0)`** asserts a compile-time property at runtime.

Changed to `static_assert`, which holds whether or not the case is executed.

> **All three cases take the 5 s `network` default timeout**

All three now specify `timeout::instant`; they report `budget: 500.0ms` and complete in ≤0.2 ms.

> **`resp.set_content_uncompressed(...)`** is set and never read

Asserted rather than dropped: it is the only field on that response carrying a payload, so the
round-trip is worth covering.

> **`target_link_libraries(cng_codegen_smoke_test PRIVATE couchbase_cxx_protostellar)`** […] uses the
> raw name rather than the `couchbase::cxx_protostellar` alias

Deferred, for two reasons. Consistency currently argues the other way: by commit 05 there are 15
further `LIBS couchbase_cxx_protostellar` call sites, all using the raw name, so changing this single
line reduces uniformity rather than increasing it. More decisively, the one-token change was
attempted twice and conflicts across **25 downstream commits** — the line sits immediately above the
block inserted by commit 05, so every descendant's rebase encounters it. The appropriate form is a
single mechanical sweep of all 16 call sites once the stack has landed; that can be tracked as a
follow-up if preferred.

> `CMakeLists.txt:655` still advertises a `LINK_CLIENT` argument to `add_cng_test` that doesn't exist

Resolved as described under finding 3: the claim is removed rather than left one commit early.

---

Verification for this round: `cmake --build build-cng` completes clean and
`ctest --test-dir build-cng --label-regex 'cng'` reports 34/34 — the same invocation the new CI step
runs — plus a direct run of the smoke test to confirm the revised timeouts and the added assertion.
The MSVC `/FI` is the sole change not exercised locally; the Windows legs cover it for the first time
on this push.

avsej:
Correction to finding 2. Enabling `CB_COUCHBASE2` on the Windows legs did its job immediately: the
`/FI` force-include described in the previous comment **breaks the build** and has been removed.

`windows (windows-11-arm)` failed on the first run with the couchbase2 stubs enabled:

```
protostellar_generated\couchbase\query\v1\query.pb.h(1085,27):
  error C2143: syntax error: missing ')' before 'constant'
  error C2789: 'couchbase::query::v1::QueryResponse_MetaData::DWORD':
               an object of const-qualified type must be initialized
  error C2059: syntax error: ')'
[couchbase_cxx_protostellar.vcxproj] -- FAILED
```

The mechanism, traced through the headers rather than inferred:

1. `grpcpp/impl/proto_utils.h` includes `grpc/byte_buffer_reader.h`, which reaches
   `grpc/support/port_platform.h`. That header includes `<windows.h>` unconditionally on MSVC — its
   own comment reads *"Get windows.h included everywhere (we need it)"*.
2. Force-including it via `/FI` therefore puts `winnt.h`'s macros in scope **before** the
   translation unit's own includes are processed, where
   `#define STATUS_TIMEOUT ((DWORD   )0x00000102L)`.
3. `QueryResponse.MetaData.Status` declares enum values `STATUS_SUCCESS` and `STATUS_TIMEOUT`, so the
   generated `static constexpr Status STATUS_TIMEOUT = …` expands to
   `static constexpr Status ((DWORD   )0x00000102L) = …` — which is both a syntax error and a
   declaration of a member named `DWORD`, exactly as reported.

In normal include order the generated `.pb.h` is parsed before anything drags in `windows.h`, so
there is no clash. The three macOS legs passed on the same run, consistent with `winnt.h` being the
cause.

The force-include has been removed and the MSVC branch is back to `/W0 /FS`, with a comment recording
why it must not be reinstated on its own. `CB_COUCHBASE2` stays enabled on the Windows legs — it is
what surfaced this, and it is what will show whether the `SerializationTraits` problem the
force-include was meant to solve actually affects this target. No such error appeared in the failing
log; the build stopped at the macro collision first, so that question is still open rather than
answered.

If `SerializationTraits` does turn out to be undefined in a `*.grpc.pb.cc` here, the correct form is
the force-include **plus** `#undef` of the colliding `STATUS_*` macros, not the force-include alone.

The premise of the original finding is unaffected: this library compiles 14 services' worth of
generated gRPC sources and the schema does use server streaming, so if the toolchain needs the
workaround, it needs it here too. Only the remedy was wrong.

avsej:
The question the previous comment left open is now answered, and the answer is negative.

> No such error appeared in the failing log; the build stopped at the macro collision first, so that
> question is still open rather than answered.

Measured, on run
[30504965067](https://github.com/couchbase/couchbase-cxx-client/actions/runs/30504965067) with the
force-include removed. All three Windows legs and all three macOS legs completed `success`. Two of
the Windows legs performed full compiles and carry the evidence:

| | windows-2022 | windows-11-arm |
|---|---|---|
| compile lines | 34,969 | 29,242 |
| `couchbase_cxx_protostellar.vcxproj` | 118 | 116 |
| library linked | yes | yes |
| `SerializationTraits` diagnostics | 0 | 0 |
| `C1128` | 0 | 0 |
| `STATUS_TIMEOUT` / `C2143` / `C2789` | 0 | 0 |
| `error Cnnnn` of any kind | none | none |

Both legs compile all 30 of the library's translation units — the count is confirmed by
`error_details.pb.cc` and `routing.grpc.pb.cc`, which are unique to this schema and absent from
fit_performer's. `windows-2025` also reported `success` but ran only 382 compile lines against the
other legs' ~30,000, so it was largely served from cache; it corroborates rather than provides
independent evidence.

The `windows-11-arm` result is the significant one: that leg is where the macro collision was
observed, and its `C2143`/`C2789` count is now zero. The revert therefore removed the regression
rather than relocating it.

So the workaround this target was reported to be missing is not required. The end state is no
force-include, and `/permissive-` recorded in the MSVC branch as the remedy should the diagnostic
ever appear.

Three further points came out of tracing the mechanism, two of them defects in the commit as
reviewed.

**The remedy named in the previous comment was wrong.** It proposed reinstating the force-include
together with `#undef` of the colliding macros. That addresses the symptom. `tools/fit_performer`
calls `set_project_options()` — which on MSVC contributes only `/bigobj /FS` — and deliberately does
not call `set_project_warnings()`, where `/permissive-` is defined
(`cmake/CompilerWarnings.cmake:33`). Its generated gRPC sources are therefore compiled in MSVC's
default, non-conformant mode, which is the condition its own comment identifies. `/permissive-`
enables conformant two-phase lookup and needs no include-order inversion, so it addresses the cause
without dragging `<windows.h>` ahead of the generated headers. This is inference from the flag audit
and fit_performer's comment, not a measured result — the diagnostic never occurred, so no remedy was
exercised.

**`/bigobj` was absent and has been added.** `set_project_options()` supplies it to fit_performer for
the same class of file; this target does not call that function and so received nothing.
`search.pb.cc` is 18,482 lines and yields a 7.4 MB object. To be explicit: no `C1128` was observed,
and none is expected at the `RelWithDebInfo` the Windows legs build. The flag covers the Debug MSVC
path — the default for `bin/build-tests` — which CI does not exercise and where objects are
substantially larger.

**The collision surface was overstated.** The comment claimed the query, analytics and search
schemas declare enum constants matching `winnt.h` macros. Only `couchbase/query/v1/query.pb.h` does;
a scan of the generated tree returns that single file. The comment now names it specifically. A
related stale reference in `.github/workflows/build.yml`, which justified `CB_COUCHBASE2` by "the
/FI proto_utils.h workaround in cmake/Protostellar.cmake", has been reworded — the workaround it
cited no longer exists.

Verification for this round: `cmake --build build-cng` clean and `ctest --label-regex 'cng'` 34/34,
plus the CI run above.

avsej:
Three corrections to the first response, from re-checking each claim at commit 02 rather than at the
working copy.

**1. The `/FI` force-include is no longer present.** That comment stated "the same construct is now
present in `Protostellar.cmake`". It was reverted after it broke `windows-11-arm`, and measurement
then showed it is not required at all: across two runs, all three Windows legs compile this library's
30 translation units and link it with no `SerializationTraits` diagnostic. The MSVC branch is
`/W0 /FS /bigobj`. The two comments following that one carry the detail.

**2. The streaming-RPC list was incomplete — there are 10, not 6.** The four omitted are
`GetAllReplicas` (`kv.proto:95`), `GetVbucketInfo` (`xdcr.proto:32`), `WatchCollections`
(`xdcr.proto:36`) and `WatchRequests` (`hooks.proto:44`). All ten are server-streaming; the schema
declares no client-streaming RPC. Checked against the pinned tag `36d4d9ab`.

**3. The call-site counts in the alias nit were wrong.** "15 further `LIBS
couchbase_cxx_protostellar` call sites" by commit 05, and "all 16 call sites", do not hold. At commit
05 there is exactly **one** such site; at the tip of the stack there are **30**, plus the smoke test's
own `target_link_libraries` — **31** raw-name uses. The conclusion is unaffected and in fact
strengthened, since the sweep is larger than stated. The stated reason for deferring is unchanged: the
line sits immediately above the block commit 05 inserts, so the one-token change conflicts across 25
descendants.

Nothing else in that comment changed on re-check. The `LABELS "cng"` citation at
`test/cng/CMakeLists.txt:45` is correct **at commit 02** — it is line 63 at the tip, because later
commits add registrations — as are the four runner `--label-regex` references, `cd "${BUILD_DIR}"` at
`bin/run-unit-tests:38`, and `Protostellar.cmake:115`. The artifact-size and build-time measurements
stand as reported.

.github/workflows/build.yml
thejcfactor, line 1: Do we want to build on the mac machines?

.github/workflows/tests.yml
```diff
@@ -72,6 +80,27 @@ jobs:
...
+        # this step they are compiled and linked by the build job but never executed by any job --
+        # and ctest reports "no tests matched" as success, so nothing would say so.
+        # always(): a unit-test failure must not mask CNG results.
+        if: ${{ always() }}
+        timeout-minutes: 10
+        env:
+          TEST_LOG_LEVEL: trace
+        run: |
+          chmod -R +x ./cmake-build-tests
+          ctest --test-dir ./cmake-build-tests --output-on-failure --label-regex 'cng'
```
thejcfactor: This is missing the `--output-junit` options.

Also, do we need `--no-tests=error`?  i think w/o it we still exits with 0.

avsej:
Both notes are acted on. One of them also applied to a sibling step that was not commented on; that
is covered below.

## 1. Building on the mac machines

> Do we want to build on the mac machines?

Yes — enabled, and the marginal cost is small because the expensive part is already being paid.

Those legs already build gRPC and protobuf from source: `COUCHBASE_CXX_CLIENT_BUILD_FIT_PERFORMER`
defaults to `BUILD_TOOLS`, which defaults to `MASTER_PROJECT`, so
`cmake/ThirdPartyDependencies.cmake:42-46` includes `GrpcProtobuf.cmake` whether or not couchbase2 is
requested. Measured in the `macos-15` leg of the previous run: 4,505 `grpc` mentions, 1,094
`protobuf`, 367 `fit_performer` — and **0** `protostellar`. So the dependency was being built and the
schema was not.

The gap this closes is toolchain coverage. The generated protobuf/gRPC sources were being compiled by
gcc (the `tests.yml` build job) and MSVC (`build.yml` windows), but never by Apple clang, which is the
third toolchain the project ships against. Baseline wall clock for those legs was 21–23 min
(macos-14 22m54s, macos-15 21m09s, macos-26 23m27s); the delta will be visible on this run.

## 2. `--output-junit` and `--no-tests=error`

> This is missing the `--output-junit` options.

Added. One detail worth recording: `--output-junit` resolves relative to `--test-dir`, not the
working directory, so the report lands in `./cmake-build-tests/` rather than the workspace root —
verified locally rather than assumed, since the equivalent path was got wrong once already in this
PR.

The filename deliberately differs from `results.xml`. `bin/run-unit-tests` cds into the build tree and
writes `results.xml` there, and the CNG step runs in the same job against the same directory, so
reusing that name would have overwritten the unit report. It is uploaded as its own artifact
(`cng-test-results`) rather than folded into `unit-test-results`, consistent with the reasoning for
keeping the `cng` label out of the `unit` label: a CNG test is not a unit test, and filing its results
under that name would misreport the suite.

> Also, do we need `--no-tests=error`?  i think w/o it we still exits with 0.

Correct, and measured:

```
ctest --test-dir build-cng --label-regex 'no-such-label-xyz'                    -> exit 0
ctest --test-dir build-cng --label-regex 'no-such-label-xyz' --no-tests=error   -> exit 8
```

So without it the step would pass vacuously if the label were renamed or `BUILD_CNG_TESTS` went OFF —
the same "green because nothing ran" failure this step was added to close, one level down. Added.
`--no-tests=` requires ctest >= 3.20 and `--output-junit` >= 3.21; CI has 3.28.3 (apt
`cmake_3.28.3-1build7` on ubuntu-24.04), so both are available.

**The same defect existed in a step that was not commented on.** The live-cluster `cng` job added in
commit 10 (PR #1032) carried the identical invocation with neither flag. A vacuous pass there is worse
than in the unit job, because that job's purpose is to prove the tests ran against a real gateway — a
green tick would have asserted exactly what had not happened. It now carries both flags and uploads
`cng-live-results.xml` as `cng-live-test-results`; the artifact name differs because upload-artifact
names must be unique within a run.

---

Verification: both workflow files parse (`yaml.safe_load`), all four upload-artifact names are unique,
`cmake --build build-cng` is clean on the new base and `ctest --label-regex 'cng' --no-tests=error
--output-junit` reports 34/34 with exit 0, writing a report of 34 tests / 0 failures / 11 skipped.

Separately, the stack has been rebased onto `main@origin` `841c5cae` with 0 conflicts across all 28
commits.

.github/workflows/build.yml
avsej: Enabled. These legs already build gRPC and protobuf from source for fit_performer — `BUILD_FIT_PERFORMER` defaults to `BUILD_TOOLS`, so `GrpcProtobuf.cmake` is included regardless of couchbase2 — measured in the previous run's `macos-15` log as 4,505 `grpc` / 1,094 `protobuf` / 367 `fit_performer` mentions and **0** `protostellar`. So the dependency was already being paid for and the schema was not being compiled. The gap it closes is toolchain coverage: gcc and MSVC compiled the generated sources, Apple clang never did. Detail in [this comment](https://github.com/couchbase/couchbase-cxx-client/pull/1024#issuecomment-5135412633).

.github/workflows/tests.yml
avsej: Both added, and the `--no-tests=error` suspicion was correct — measured on a label matching nothing: exit **0** without the flag, exit **8** with it. So the step would have passed vacuously if the label were renamed or `BUILD_CNG_TESTS` went OFF, which is the same "green because nothing ran" failure the step exists to close. `--no-tests=` needs ctest >= 3.20 and `--output-junit` >= 3.21; CI has 3.28.3.

One detail on `--output-junit`: it resolves relative to `--test-dir`, so the report lands in `./cmake-build-tests/` rather than the workspace root. The filename differs from `results.xml` deliberately, since `bin/run-unit-tests` already writes that name into the same directory in the same job, and it is uploaded as its own artifact rather than folded into `unit-test-results`.

The same two flags were also absent from the live-cluster `cng` job added in commit 10 (#1032), which was not commented on. A vacuous pass there is worse — that job's purpose is to prove the tests ran against a real gateway — so it now carries both flags and uploads its own report. Detail in [this comment](https://github.com/couchbase/couchbase-cxx-client/pull/1024#issuecomment-5135412633).

thejcfactor (approved):

.github/workflows/build.yml
avsej: The numbers, from run
[30575152300](https://github.com/couchbase/couchbase-cxx-client/actions/runs/30575152300) against the
previous run's macOS legs as baseline.

Apple clang compiles the generated sources cleanly. All three legs are identical: 116 `protostellar`
mentions (0 before), 70 `couchbase_cxx_protostellar`, `libcouchbase_cxx_protostellar.a` linked, and
**0** `error:` lines.

| leg | baseline | with couchbase2 | delta |
|---|---|---|---|
| macos-14 | 22m54s | 21m29s | −1m25s |
| macos-15 | 21m09s | 22m30s | +1m21s |
| macos-26 | 23m27s | 22m13s | −1m14s |

**No measurable cost.** The deltas change sign and are smaller than the spread between the legs, which
is variance rather than signal — compiling 30 extra translation units cannot make a build faster, so
the net −1m18s is not a saving and is not claimed as one. `ccache` keys on
`${{ github.job }}-${{ matrix.os }}` and is shared across runs, so most of both builds was warm while
the new translation units were necessarily cold; that makes the marginal cost small but not cleanly
isolatable from this comparison.

The `--output-junit` and `--no-tests=error` changes are confirmed working in the same run: the `tests`
workflow produced `cng-test-results` (827 bytes) alongside `unit-test-results` (17,007 bytes), so the
two reports no longer collide on one filename.

One thing worth flagging before it reads as a regression: that report records **2** tests, not the 34
quoted earlier. Both are correct. Commit 02 is where the second CNG test appears, so the label matches
exactly two there (`cng_framework_selftest_test`, `cng_codegen_smoke_test`); the 34 figure is a local
tree built at the tip of the stack. `--no-tests=error` passing is itself the evidence that the label
matched real tests rather than nothing.

PR #1025, opened 2026-07-28
CXXCBC-887: recognize the couchbase2:// connection-string scheme

avsej:
The Protostellar transport is selected by the couchbase2:// scheme, which the
connection-string parser did not recognize. Add it: default port 18098, TLS
implied (propagated to cluster options), and a typed protocol marker
(connection_string::protocol_type plus uses_protostellar()) so the connect path
can branch to the gRPC client without a string comparison. The classic schemes
(couchbase/couchbases/http/https) are unchanged.

Covered by a connection-string unit test on the CNG harness.

---
Stacked PR **03/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–03; the change specific to this PR is its top commit. Depends on #1024 — merge the series bottom-up.

### DIFF
--- a/core/utils/connection_string.cxx
+++ b/core/utils/connection_string.cxx
@@ -91,6 +91,12 @@ struct action<scheme> {
       cs.default_port = 18091;
       cs.default_mode = connection_string::bootstrap_mode::http;
       cs.tls = true;
+    } else if (cs.scheme == "couchbase2") {
+      // Protostellar: the gateway does routing, so there is no MCBP-style config bootstrap.
+      cs.default_port = 18098;
+      cs.default_mode = connection_string::bootstrap_mode::unspecified;
+      cs.tls = true;
+      cs.protocol = connection_string::protocol_type::protostellar;
     } else {
       cs.default_mode = connection_string::bootstrap_mode::unspecified;
       cs.default_port = 0;
@@ -364,6 +370,19 @@ extract_options(connection_string& connstr)
       connstr.bootstrap_nodes[0].type != connection_string::address_type::dns) {
     connstr.options.enable_dns_srv = false;
   }
+  // couchbase2:// never resolves DNS *SRV*: there are no _couchbaseXXX._tcp records for a gateway,
+  // which is addressed directly and does its own routing. This does not turn off name resolution --
+  // ordinary A/AAAA lookup still happens inside gRPC, which is how a CNG cluster behind a network
+  // load balancer is reached (one hostname, one address per instance).
+  //
+  // cluster::open() already returns into open_protostellar() before reaching the DNS SRV branch, so
+  // this changes no behaviour today; it keeps the parsed options honest -- to_json() reports
+  // enable_dns_srv in diagnostics -- and stops the option becoming live if that ordering changes.
+  // Placed with the rule above, before the parameter loop, so an explicit enable_dns_srv= still
+  // wins, same as the single-node rule it sits next to.
+  if (connstr.protocol == connection_string::protocol_type::protostellar) {
+    connstr.options.enable_dns_srv = false;
+  }
   for (const auto& [name, value] : connstr.params) {
 #ifdef COUCHBASE_CXX_CLIENT_COLUMNAR
     if (name == "security.trust_only_pem_file") {
@@ -598,6 +617,34 @@ extract_options(connection_string& connstr)
     }
   }
 }
+// couchbase2:// addresses one CNG endpoint. The gateway performs routing, so there is no node list
+// to rotate through and no MCBP bootstrap mode to select per node. Reject strings that ask for
+// either instead of silently honouring only the first host. couchbase-jvm-clients does the same,
+// though it validates at transport construction (CoreProtostellar) rather than while parsing; here
+// connection_string already carries an error field, so failing at the parse boundary keeps the
+// diagnosis next to the input that caused it.
+void
+validate_protostellar(connection_string& connstr)
+{
+  if (connstr.error.has_value() ||
+      connstr.protocol != connection_string::protocol_type::protostellar) {
+    return;
+  }
+  if (connstr.bootstrap_nodes.size() != 1) {
+    connstr.error =
+      fmt::format(R"(connection string with scheme "couchbase2" must have exactly one host, )"
+                  R"(but got {}: "{}")",
+                  connstr.bootstrap_nodes.size(),
+                  connstr.input);
+    return;
+  }
+  if (connstr.bootstrap_nodes[0].mode != connection_string::bootstrap_mode::unspecified) {
+    connstr.error =
+      fmt::format(R"(connection string with scheme "couchbase2" does not accept a bootstrap mode )"
+                  R"(suffix, because the gateway does not use MCBP or HTTP config bootstrap: "{}")",
+                  connstr.input);
+  }
+}
 } // namespace
 
 auto
@@ -630,6 +677,7 @@ parse_connection_string(const std::string& input, cluster_options options) -> co
     }
   }
   extract_options(res);
+  validate_protostellar(res);
   return res;
 }
 } // namespace couchbase::core::utils
--- a/test/cng/connection_string/couchbase2_scheme.cxx
+++ b/test/cng/connection_string/couchbase2_scheme.cxx
@@ -1,0 +1,152 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Unit tests for the couchbase2:// connection-string scheme (CXXCBC-887). Pure parsing, no
+// server, so these run in every mode (env-agnostic).
+
+#include "framework/test_runner.hxx"
+
+#include "core/utils/connection_string.hxx"
+
+#include <cstddef>
+#include <cstdint>
+#include <string>
+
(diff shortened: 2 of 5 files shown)


test/cng/framework/test_runner.cxx
```diff
@@ -0,0 +1,145 @@
...
+auto
+human(std::chrono::nanoseconds ns) -> std::string
+{
+  return fmt::format("{}ms", std::chrono::duration_cast<std::chrono::milliseconds>(ns).count());
+}
+
+// Run one case on a worker thread, enforcing `timeout`. On timeout the worker is detached: the
+// packaged_task moved into it co-owns the future's shared state, so the detached thread writes
+// only to that heap state (never to this stack frame) and the result is simply never read.
+auto
```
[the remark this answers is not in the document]
avsej: Deliberate and documented. The CNG harness runs each case on a worker thread with a timeout as a backstop; on timeout it reports failure and detaches rather than hard-terminating, because the framework self-test (framework_selftest) exercises the timeout path IN-PROCESS -- a quick_exit/_Exit or subprocess-isolation model would abort the whole runner and make that self-test (and reporting the remaining cases) impossible. A genuinely hung case is a test bug surfaced as a failure; this in-tree harness is not a sandbox for adversarial test bodies, so subprocess isolation was considered and rejected as disproportionate. Keeping as-is.

core/utils/connection_string.cxx
```diff
@@ -91,6 +91,12 @@ struct action<scheme> {
       cs.default_port = 18091;
       cs.default_mode = connection_string::bootstrap_mode::http;
       cs.tls = true;
+    } else if (cs.scheme == "couchbase2") {
```
thejcfactor, line 94: Should we be disabling DNS?  The RFC does not mention it, but for some reason I'm thinking we don't want to do DNS w/ CNG...

test/cng/connection_string/couchbase2_scheme.cxx
```diff
@@ -0,0 +1,119 @@
...
+{
+  const auto cs = parse_connection_string("couchbase2://example.com:12345");
+  assert_false(cs.error.has_value(), "parses without error");
+  assert_eq(cs.bootstrap_nodes.size(), std::size_t{ 1 }, "one bootstrap node");
+  assert_eq(cs.bootstrap_nodes.at(0).address, std::string{ "example.com" }, "address parsed");
+  assert_eq(cs.bootstrap_nodes.at(0).port, std::uint16_t{ 12345 }, "explicit port overrides");
+}
+
+void
+couchbase2_multiple_hosts()
```
thejcfactor: I thought CNG does not support multiple hosts?  Looks like Java does not allow it: https://github.com/couchbase/couchbase-jvm-clients/blob/4f9d4d7f286968fdc0df1b36b048cf507fb8f93b/core-io/src/main/java/com/couchbase/client/core/CoreProtostellar.java#L104-L109

avsej:
Both points hold. Following them up surfaced a third case of the same shape, described below. All
three are now rejected at parse time rather than silently accepted.

## 1. DNS

> Should we be disabling DNS?  The RFC does not mention it, but for some reason I'm thinking we don't
> want to do DNS w/ CNG...

Disabled, with one distinction worth stating because it changes what "disabling DNS" means here: what
is switched off is DNS **SRV**, not name resolution.

`enable_dns_srv` defaults to `true` (`cluster_options.hxx:69`), and `extract_options()` only clears it
when the string is *not* a single DNS-typed host. `couchbase2://localhost` is exactly the shape that
keeps it on, so the option was surviving as `true` for every couchbase2 string. It is now forced off
for `protocol_type::protostellar`.

No behaviour changes today: `cluster::open()` returns into `open_protostellar()` at `cluster.cxx:525`,
before the `enable_dns_srv` branch at `:528`, so the flag was already unreachable on this path. What
it fixes is that the parsed options were dishonest — `origin::to_json()` reports `enable_dns_srv` in
diagnostics — and that the option would become live if that ordering were ever rearranged.

Ordinary A/AAAA resolution is untouched and must stay that way: gocbcoreps documents CNG being
deployed as a cluster behind a network load balancer, where each instance is one address of a single
hostname, resolved inside gRPC. Turning off SRV does not affect that.

On the RFC: it does not cover this, and its Usage section is the whole of what it asks for here —
"The only thing the user needs to do is use couchbase2:// in their connection string." There is no
host-count rule, and DNS SRV is not mentioned anywhere in the document; the only connection-string
example given, `couchbase2://invalid_hostname` in the error-handling section, uses a single hostname.
Beyond the scheme, the spec's connection-string content is the timeout parameters. So that reading
was right, and the decision below rests on the other SDKs rather than on the spec.

## 2. Multiple hosts

> I thought CNG does not support multiple hosts?  Looks like Java does not allow it: [CoreProtostellar.java#L104-L109]

Confirmed. This is now an error rather than a silently dropped host list.

`CoreProtostellar.java:104-109` throws `InvalidArgumentException` when `hosts().size() != 1`. The Go
side agrees structurally rather than by validation: `gocbcoreps.Dial()` takes a single `target
string`, and gocb builds it from a singular `resolved.Couchbase2Host` (`client_ps.go:62`). Whether Go
*errors* or quietly takes the first host was not established: that decision lives in `gocbconnstr`,
which was not inspected. Stated as unverified rather than assumed.

What settles it is the topology, from gocbcoreps' README: a CNG cluster is reached through **one**
hostname behind a load balancer, with each instance appearing as an address of that name. Extra hosts
in the connection string are not how CNG scales out, so accepting them could only mask a
misconfiguration.

Previous behaviour was tolerate-and-log: `open_protostellar()` used the first node and logged the
rest at INFO. That is now unreachable from a parsed string; the guard stays for origins assembled
programmatically (`set_nodes()`, the hostname/port constructors), and its log now reports how many
nodes were ignored. `COMPATIBILITY.md` said "additional hosts are ignored" and has been corrected.

## 3. A third case, from the same code you linked

The lines immediately after the host check in `CoreProtostellar.java` (`:112-116`) also reject a
**port type**. Our equivalent is the per-node bootstrap-mode suffix — `couchbase2://host=mcd`,
`=http`, `=cccp`, `=gcccp` — which the grammar accepts and which cannot mean anything for a gateway
that uses neither MCBP nor HTTP config bootstrap. Also rejected now. Flagged explicitly, since it was
not part of what was raised.

## Where the validation lives

In the parser, not at transport construction as in Java. `connection_string` already carries an
`error` field, so failing at the parse boundary keeps the diagnosis attached to the input that caused
it, and every caller checks that field already. The classic schemes are untouched — both new rules
are gated on `protocol_type::protostellar`.

## Tests

`couchbase2_multiple_hosts` asserted the old behaviour and is replaced. Three cases, each carrying a
classic-scheme control assertion so a regression in shared parsing shows up here:

| case | asserts | control |
|---|---|---|
| `couchbase2_rejects_multiple_hosts` | error names "exactly one host" | `couchbase://a,b` still parses, 2 nodes |
| `couchbase2_rejects_mode_suffix` | `=mcd` / `=http` rejected | `couchbase://host=mcd` still accepted |
| `couchbase2_disables_dns_srv` | `enable_dns_srv` off | `couchbase://localhost` keeps it **on** |

Verification: `connection_string_couchbase2` goes 5 → 7 cases, all passing;
`ctest --label-regex 'cng' --no-tests=error` 34/34. The classic suite is not built in this tree
(`BUILD_TESTS=OFF`), so the controls above are the local guard and CI's `unit` job is the backstop.
`clang-format-22` clean.

Note these two findings also apply to #1026, whose diff is cumulative and includes this commit; both
PRs carry the same fix rather than diverging.

core/utils/connection_string.cxx
avsej: Disabled, though the distinction matters: what is switched off is DNS **SRV**, not name resolution.

`enable_dns_srv` defaults to `true` (`cluster_options.hxx:69`), and `extract_options()` only clears it when the string is *not* a single DNS-typed host — which is precisely the shape `couchbase2://localhost` has, so the option was surviving as `true`. It is now forced off for `protocol_type::protostellar`.

Behaviour is unchanged today: `cluster::open()` returns into `open_protostellar()` (`cluster.cxx:525`) before the `enable_dns_srv` branch (`:528`), so the flag was already unreachable. What is fixed is that the parsed options were dishonest — `origin::to_json()` reports `enable_dns_srv` in diagnostics — and that the option would become live if that ordering were rearranged.

Ordinary A/AAAA resolution is deliberately untouched. gocbcoreps documents CNG deployed as a cluster behind a network load balancer, where each instance is one address of a single hostname resolved inside gRPC; disabling SRV does not affect that, and must not.

On the RFC: it does not cover this. Its Usage section is the whole of what it asks for here — "The only thing the user needs to do is use couchbase2:// in their connection string." No host-count rule, and DNS SRV is not mentioned anywhere in the document; the one connection-string example given, `couchbase2://invalid_hostname` in the error-handling section, uses a single hostname. The decision therefore rests on the other SDKs rather than on the spec. Full detail in the round-2 comment on this PR.

test/cng/connection_string/couchbase2_scheme.cxx
avsej: Confirmed — and it is now rejected at parse time rather than silently reduced to the first host. `couchbase2_multiple_hosts` asserted the old behaviour and has been replaced by `couchbase2_rejects_multiple_hosts`.

The linked check (`CoreProtostellar.java:104-109`) throws for `hosts().size() != 1`. Go agrees structurally rather than by validation: `gocbcoreps.Dial()` takes a single `target string`, built from a singular `resolved.Couchbase2Host` (`client_ps.go:62`). Whether Go errors or quietly takes the first host was not established: that decision lives in `gocbconnstr`, which was not inspected. Stated as unverified rather than assumed.

What settles it is the topology, from gocbcoreps' README: a CNG cluster is reached through **one** hostname behind a load balancer, each instance appearing as an address of that name. Extra hosts are not how CNG scales out, so accepting them could only mask a misconfiguration.

The lines immediately following the host check (`:112-116`) also reject a **port type**. The equivalent here is the per-node bootstrap-mode suffix (`couchbase2://host=mcd`, `=http`, `=cccp`, `=gcccp`), meaningless for a gateway that uses neither MCBP nor HTTP config bootstrap; that is rejected too. Flagged since it was not part of what was raised.

Previously `open_protostellar()` used the first node and logged the rest at INFO. That path is now unreachable from a parsed string; the guard remains for origins assembled programmatically (`set_nodes()`, the hostname/port constructors) and its log reports how many nodes were ignored. `COMPATIBILITY.md` stated "additional hosts are ignored" and has been corrected.

Since #1026's diff is cumulative it carries this change too, so the two PRs do not diverge.

thejcfactor (approved):

PR #1026, opened 2026-07-28
CXXCBC-888: carry the protostellar protocol marker on origin

avsej:
The connect path needs to know a cluster is couchbase2:// without re-parsing the
connection string. Thread the marker from the parsed connection string onto
origin as uses_protostellar(), set in the connection-string constructor and
preserved across copy and move. No behaviour change on the MCBP path.

Covered by an origin unit test on the CNG harness.

---
Stacked PR **04/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–04; the change specific to this PR is its top commit. Depends on #1025 — merge the series bottom-up.

### DIFF
--- a/core/origin.cxx
+++ b/core/origin.cxx
@@ -319,9 +319,15 @@ origin::to_json() const -> std::string
 couchbase::core::origin::origin(origin&& other) noexcept
   : options_(std::move(other.options_))
   , nodes_(std::move(other.nodes_))
-  , next_node_(other.next_node_)
+  // next_node_ is an iterator into nodes_. Container move construction transfers the buffer, so
+  // iterators into the source stay valid and now denote elements of nodes_ -- other.next_node_ can
+  // be adopted as-is, preserving the round-robin position. The empty case is guarded separately
+  // because that guarantee does not extend to a past-the-end iterator. nodes_ is declared before
+  // next_node_, so it is already initialized here.
+  , next_node_(nodes_.empty() ? nodes_.begin() : other.next_node_)
   , exhausted_(other.exhausted_)
   , connection_string_(std::move(other.connection_string_))
+  , uses_protostellar_(other.uses_protostellar_)
   , credentials_(std::move(other.credentials_))
 {
 }
@@ -330,20 +336,38 @@ auto
 couchbase::core::origin::operator=(origin&& other) noexcept -> origin&
 {
   if (this != &other) {
+    const auto offset = other.nodes_.empty() ? 0 : other.next_node_ - other.nodes_.begin();
     options_ = std::move(other.options_);
     nodes_ = std::move(other.nodes_);
-    next_node_ = other.next_node_;
+    next_node_ = nodes_.begin() + offset;
     exhausted_ = other.exhausted_;
     connection_string_ = std::move(other.connection_string_);
+    uses_protostellar_ = other.uses_protostellar_;
     credentials_ = std::move(other.credentials_);
   }
   return *this;
 }
 
+// A copy is a fresh origin that has not attempted anything, so the rotation cursor is deliberately
+// reset rather than carried. This matters: next_address() marks an origin exhausted as it hands out
+// the *last* node, which on a single-node cluster is the very first call, so a cluster origin is
+// normally exhausted immediately after bootstrap. open_bucket() copies that origin into every
+// bucket and on into every session, and mcbp_session::initiate_bootstrap() tests exhausted()
+// before it ever calls next_address() -- so carrying the flag would send a session that has tried
+// nothing down the "reached the end of the list of bootstrap nodes" path: an unconditional 500 ms
+// backoff, plus a spurious no_endpoints_left notification under COUCHBASE_CXX_CLIENT_COLUMNAR.
+//
+// Moving, by contrast, does carry it: a move transfers identity rather than producing an
+// independent origin.
+//
+// connection_string_ and uses_protostellar_ *are* copied. They were being dropped entirely, which
+// left connection_string() returning "" on any copied origin.
 couchbase::core::origin::origin(const couchbase::core::origin& other)
   : options_(other.options_)
   , nodes_(other.nodes_)
   , next_node_(nodes_.begin())
+  , connection_string_(other.connection_string_)
+  , uses_protostellar_(other.uses_protostellar_)
 {
   update_credentials(other.credentials());
 }
@@ -378,6 +402,7 @@ couchbase::core::origin::origin(couchbase::core::cluster_credentials auth,
                                 const couchbase::core::utils::connection_string& connstr)
   : options_(connstr.options)
   , connection_string_(connstr.input)
+  , uses_protostellar_(connstr.uses_protostellar())
   , credentials_(std::move(auth))
 {
   nodes_.reserve(connstr.bootstrap_nodes.size());
@@ -396,10 +421,13 @@ couchbase::core::origin::operator=(const couchbase::core::origin& other) -> couc
 {
   if (this != &other) {
     options_ = other.options_;
+    connection_string_ = other.connection_string_;
     update_credentials(other.credentials());
     nodes_ = other.nodes_;
+    // Rotation state is reset, not carried -- see the copy constructor for why.
     next_node_ = nodes_.begin();
     exhausted_ = false;
+    uses_protostellar_ = other.uses_protostellar_;
   }
   return *this;
 }
@@ -407,6 +435,11 @@ auto
 couchbase::core::origin::connection_string() const -> const std::string&
 {
   return connection_string_;
+}
+auto
+couchbase::core::origin::uses_protostellar() const -> bool
+{
+  return uses_protostellar_;
 }
 auto
 couchbase::core::origin::username() const -> std::string
@@ -502,6 +535,11 @@ couchbase::core::origin::set_nodes_from_config(const topology::configuration& co
     shuffle_nodes();
   }
   next_node_ = nodes_.begin();
+  // Matches set_nodes(): the node list has just been replaced, so any exhaustion recorded against
+  // the previous list is meaningless. Without this an origin that was exhausted stays exhausted
+  // across every config update -- cluster.cxx and io/config_tracker.cxx both call this on
+  // long-lived origins that next_address() will have marked exhausted.
+  exhausted_ = false;
 }
 void
 couchbase::core::origin::update_credentials(cluster_credentials auth)
--- a/core/origin.hxx
+++ b/core/origin.hxx
@@ -60,6 +60,9 @@ struct origin {
   auto operator=(const origin& other) -> origin&;
 
   [[nodiscard]] auto connection_string() const -> const std::string&;
+  // True when built from a couchbase2:// connection string (Protostellar transport). Lets the
+  // connect path branch to the gRPC client without re-parsing the connection string.
+  [[nodiscard]] auto uses_protostellar() const -> bool;
   [[nodiscard]] auto username() const -> std::string;
   [[nodiscard]] auto password() const -> std::string;
   [[nodiscard]] auto certificate_path() const -> std::string;
@@ -92,6 +95,7 @@ private:
   node_list::iterator next_node_{};
   bool exhausted_{ false };
   std::string connection_string_{};
+  bool uses_protostellar_{ false };
   cluster_credentials credentials_{};
   mutable std::shared_mutex credentials_mutex_{};
 };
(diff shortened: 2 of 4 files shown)


test/cng/connection_string/couchbase2_scheme.cxx
```diff
@@ -0,0 +1,119 @@
...
+{
+  const auto cs = parse_connection_string("couchbase2://example.com:12345");
+  assert_false(cs.error.has_value(), "parses without error");
+  assert_eq(cs.bootstrap_nodes.size(), std::size_t{ 1 }, "one bootstrap node");
+  assert_eq(cs.bootstrap_nodes.at(0).address, std::string{ "example.com" }, "address parsed");
+  assert_eq(cs.bootstrap_nodes.at(0).port, std::uint16_t{ 12345 }, "explicit port overrides");
+}
+
+void
+couchbase2_multiple_hosts()
```
thejcfactor: I was thinking for CNG we don't support multiple hosts?  Looks like Java does not: https://github.com/couchbase/couchbase-jvm-clients/blob/4f9d4d7f286968fdc0df1b36b048cf507fb8f93b/core-io/src/main/java/com/couchbase/client/core/CoreProtostellar.java#L104-L109

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. This one is a static trace — I read the code and followed the call paths by hand, and did **not** build or run anything. That matters most for the finding below, which asserts a runtime regression I have not measured; I've included the exact way to confirm or refute it and I'd rather you check it than take my word. Please treat the reasoning as a starting point for discussion, not ground truth.

Reviewed at `a95867bf`, incrementally against #1025's head `94552a6e`, so this is the one commit.

The marker plumbing is right, and one part of the diff is a real fix worth naming: the copy constructor and copy-assignment operator were both dropping `connection_string_` outright, so `origin::connection_string()` returned `""` on any copied origin. That's the same latent bug class as the marker itself — a field added to the class but not to the special members — and `uses_protostellar()` would have inherited it silently. `protocol.cxx:74-76` asserting the connection string is the right instinct.

One thing I'd want resolved before this merges, plus a test gap that follows from it.

---

## 1. `major` — carrying `exhausted_` across copies adds an unconditional 500 ms stall to every new `mcbp_session` on single-node clusters

Of the three things this commit does — add the marker, fix `connection_string_` being dropped on copy, and preserve `next_node_`/`exhausted_` rotation state across copy and move — only the first two are needed for the marker. The third changes MCBP bootstrap behavior, and I think not intentionally.

The chain, four steps:

**1.** `next_address()` sets `exhausted_ = true` when it hands out the *last* node (`origin.cxx:541-542`). With a single-node connection string that is the **first** call, so a cluster-level `origin_` is left `exhausted_ == true` immediately after a successful bootstrap (`cluster.cxx:1041`).

**2.** Before this PR the copy constructor did not copy `exhausted_` — it reset `next_node_` to `nodes_.begin()` and let `exhausted_` take its `false` default. Now it copies both (`origin.cxx:354-356`), as do copy-assignment (`415-417`) and the move members (`327-328`, `342-343`).

**3.** `open_bucket` copies the cluster origin, on both branches:

```cpp
auto origin = origin_;                       // cluster.cxx:579
if (session_ && session_->has_config()) {
  if (const auto config = session_->config(); config.has_value()) {
    origin = { origin_, config.value() };    // cluster.cxx:583
  }
}
```

The two-argument constructor calls `set_nodes_from_config()`, which resets `next_node_` (`origin.cxx:525`) but **not** `exhausted_` — unlike `set_nodes()`, which resets both (`497-498`). So `exhausted_ == true` survives into the bucket's origin and from there into every session: `io::mcbp_session(client_id_, {}, ctx_, origin_, ...)` (`bucket.cxx:627`), taken by value and stored via the move constructor (`mcbp_session.cxx:867`, `912`).

**4.** `initiate_bootstrap()` tests `origin_.exhausted()` (`mcbp_session.cxx:1291`) **before** it ever calls `next_address()` (`1319`). So the session's very first bootstrap attempt takes the exhausted branch:

```cpp
auto backoff = std::chrono::milliseconds(500);
CB_LOG_DEBUG("{} reached the end of list of bootstrap nodes, waiting for {}ms before restart", ...);
retry_backoff_.expires_after(backoff);
retry_backoff_.async_wait([self = ...] { self->origin_.restart(); self->initiate_bootstrap(); });
return;
```

500 ms of dead time before a session that has not tried a single node connects. Nothing resets `exhausted_` in between — the only two `restart()` calls (`mcbp_session.cxx:1239`, `1314`) are both inside retry handlers reached *after* this branch.

Two things sharpen it:

- **Single-node is the common dev/CI shape**, and it's the case that always triggers. With N nodes it only triggers when bootstrap consumed the whole list.
- **Under `COUCHBASE_CXX_CLIENT_COLUMNAR` it's worse than latency**: the same branch fires `notify_bootstrap_error({errc::network::no_endpoints_left, "Reached end of list of bootstrap nodes.", ...})` (`mcbp_session.cxx:1297-1302`) for a session that has attempted nothing. That's a spurious error notification, not just a delay.

The narrow fix, and arguably a latent bug on its own merits, is to make `set_nodes_from_config()` clear `exhausted_` the way `set_nodes()` already does:

```cpp
  next_node_ = nodes_.begin();
  exhausted_ = false;    // ← matches set_nodes(); the node list just changed
```

That covers the `cluster.cxx:583` branch but not `579`, so the alternative — just not carrying `next_node_`/`exhausted_` on copy, since the marker doesn't need it — may be the simpler answer.

**To confirm or refute:** open a bucket against a **single-node** cluster with debug logging on. Confirmed if the bucket session's first `initiate_bootstrap` logs `reached the end of list of bootstrap nodes, waiting for 500ms before restart`, and `open_bucket` gains roughly 500 ms. Refuted if that line does not appear before the first connect attempt.

## 2. `minor` — the tests cover the marker but not the fields whose semantics changed

`protocol.cxx` verifies `uses_protostellar()` survives copy, move, and copy-assignment, and that copy-assignment carries `connection_string_`. But `next_node_`/`exhausted_` — the fields this commit actually rewrote the rotation semantics of, and the subject of finding 1 — have no coverage, and move-assignment isn't exercised on any field. A case here would have caught finding 1 before I did.

Both accessors needed are already public:

```cpp
// three nodes, consume two, then copy: the copy must resume where the source left off
origin o{ cluster_credentials{}, parse_connection_string("couchbase://a,b,c") };
(void)o.next_address();
(void)o.next_address();
const origin copied{ o };
assert_false(copied.exhausted(), "not exhausted after two of three");
// and the exhausted case, which is what reaches finding 1
(void)o.next_address();
assert_true(o.exhausted(), "exhausted after the last node");
const origin copied_exhausted{ o };
assert_true(copied_exhausted.exhausted(), "exhausted state is carried");
```

`preserve_bootstrap_nodes_order` needs setting for a positional assertion to be deterministic — the constructor shuffles otherwise (`origin.cxx:402-404`).

---

If the `set_nodes_from_config` reset turns out to be the whole fix, finding 1 is a two-line change.

avsej:
Finding 1 is real and is fixed. The trace holds at every step; each was re-checked against the code
rather than taken from the description.

## 1. `major` — `exhausted_` carried across copies

> `next_address()` sets `exhausted_ = true` when it hands out the *last* node (`origin.cxx:541-542`).
> With a single-node connection string that is the **first** call

Confirmed. `next_address()` is `auto address = *next_node_; if (++next_node_ == nodes_.end()) {
exhausted_ = true; }` — one node means the first call exhausts it.

> `initiate_bootstrap()` tests `origin_.exhausted()` (`mcbp_session.cxx:1291`) **before** it ever
> calls `next_address()` (`1319`)

Confirmed, 28 lines apart, with the 500 ms backoff and the `COUCHBASE_CXX_CLIENT_COLUMNAR`
`no_endpoints_left` notification both inside that branch.

> The two-argument constructor calls `set_nodes_from_config()`, which resets `next_node_`
> (`origin.cxx:525`) but **not** `exhausted_` — unlike `set_nodes()`, which resets both (`497-498`).

Confirmed, and this asymmetry is a bug in its own right — see below.

**One correction, in favour of the simpler remedy.** The suggestion was that not carrying the state on
copy "covers the `cluster.cxx:583` branch but not `579`". It covers both. The two-argument constructor
takes its first parameter **by value**:

```cpp
origin::origin(origin other, const topology::configuration& config)
  : origin(std::move(other))          // move from a parameter that the copy ctor just produced
{
  set_nodes_from_config(config);
}
```

so `origin = { origin_, config.value() }` runs the copy constructor to materialise `other` before
anything is moved. With the copy resetting rotation state, `other` arrives already un-exhausted, and
the move that follows carries that. Both branches are therefore fixed by the copy-side change alone.

**What changed.** Copy construction and copy-assignment reset `next_node_` to `nodes_.begin()` and
`exhausted_` to `false`, which is what they did before this commit. `connection_string_` and
`uses_protostellar_` are still copied — dropping those was the actual bug this commit set out to fix,
and it is unaffected.

Move construction and move-assignment continue to carry the state. The asymmetry is deliberate and now
documented at the copy constructor: a move transfers identity, a copy produces an independent origin
that has attempted nothing. Only the first two of the three behaviours this commit changed were needed
for the marker, and the third is reverted.

**Also applied: the `set_nodes_from_config()` reset.** Not needed for the regression once the above is
in place, but it is a live pre-existing bug. `cluster.cxx:1196` and `io/config_tracker.cxx:123` both
call it on long-lived origins that `next_address()` will have marked exhausted, so an origin that
became exhausted stayed exhausted across every subsequent config update while its node list was
replaced underneath. It now clears the flag, matching `set_nodes()`.

**On measurement.** The origin-level behaviour is now pinned by a test (below) rather than by
inspection. The 500 ms consequence itself was **not** measured — no single-node cluster was driven with
debug logging, so the confirm/refute recipe has not been executed. That step is inference from the
call path, which is unambiguous but is not the same as an observation. The change is a revert to the
pre-existing semantics, so the risk of acting without that measurement is low; stated plainly in case
it matters.

## 2. `minor` — no coverage for the fields whose semantics changed

> A case here would have caught finding 1 before I did.

Added: `rotation_state_resets_on_copy_and_travels_on_move`. It walks a two-node origin to
exhaustion and then asserts all four special members — copy, copy-assignment, move, move-assignment.
Move-assignment is now exercised for the marker and connection string too, which the existing case
omitted.

**The suggested assertions would now fail, by design.** The sketch expects a copy to carry the flag:

```cpp
const origin copied_exhausted{ o };
assert_true(copied_exhausted.exhausted(), "exhausted state is carried");
```

Under the fix a copy is deliberately *not* exhausted, so the new case asserts the opposite for copies
and keeps the carrying assertion for moves. `preserve_bootstrap_nodes_order` is set as noted, though
the assertions are counts rather than positions, so it guards a future extension rather than the
present case.

## 3. Inline — multiple hosts

> I was thinking for CNG we don't support multiple hosts?

Answered in full on #1025, which owns the parsing commit:
`couchbase2://` now requires exactly one host and rejects a bootstrap-mode suffix, both at parse time.
Since #1026's diff is cumulative it carries that change too, so the two PRs do not diverge on this.

---

Verification: `cmake --build build-cng` clean; `ctest --label-regex 'cng' --no-tests=error` 34/34 with
`origin_protocol` going 3 → 4 cases and `connection_string_couchbase2` 5 → 7, all passing;
`clang-format-22` clean on every changed file.

test/cng/connection_string/couchbase2_scheme.cxx
avsej: Confirmed — and it is now rejected at parse time rather than silently reduced to the first host. `couchbase2_multiple_hosts` asserted the old behaviour and has been replaced by `couchbase2_rejects_multiple_hosts`.

The linked check (`CoreProtostellar.java:104-109`) throws for `hosts().size() != 1`. Go agrees structurally rather than by validation: `gocbcoreps.Dial()` takes a single `target string`, built from a singular `resolved.Couchbase2Host` (`client_ps.go:62`). Whether Go errors or quietly takes the first host was not established: that decision lives in `gocbconnstr`, which was not inspected. Stated as unverified rather than assumed.

What settles it is the topology, from gocbcoreps' README: a CNG cluster is reached through **one** hostname behind a load balancer, each instance appearing as an address of that name. Extra hosts are not how CNG scales out, so accepting them could only mask a misconfiguration.

The lines immediately following the host check (`:112-116`) also reject a **port type**. The equivalent here is the per-node bootstrap-mode suffix (`couchbase2://host=mcd`, `=http`, `=cccp`, `=gcccp`), meaningless for a gateway that uses neither MCBP nor HTTP config bootstrap; that is rejected too. Flagged since it was not part of what was raised.

Previously `open_protostellar()` used the first node and logged the rest at INFO. That path is now unreachable from a parsed string; the guard remains for origins assembled programmatically (`set_nodes()`, the hostname/port constructors) and its log reports how many nodes were ignored. `COMPATIBILITY.md` stated "additional hosts are ignored" and has been corrected.

Since #1026's diff is cumulative it carries this change too, so the two PRs do not diverge.

thejcfactor (approved):

PR #1030, opened 2026-07-28
CXXCBC-892: KV converter between core operations and kv.v1 messages

avsej:
Add core/protostellar/kv_converter: translate the core get/insert/upsert/replace/
remove request and response structs to and from the couchbase.kv.v1 protobuf
messages -- document location, value, content flags, expiry (seconds), CAS,
durability level (none left unset), and the mutation token. Responses are built
directly from the proto plus the component-supplied error context, not via the
MCBP-bound encode_to/make_response. Compressed content decoding is deferred until
compression negotiation lands.

Covered by converter unit tests on the CNG harness.

---
Stacked PR **08/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–08; the change specific to this PR is its top commit. Depends on #1029 — merge the series bottom-up.

### DIFF
--- a/core/protostellar/kv_converter.hxx
+++ b/core/protostellar/kv_converter.hxx
@@ -1,0 +1,274 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates the core KV operation request/response structs to and from the couchbase.kv.v1
+// protobuf messages. The response builders take an already-populated key_value_error_context (the
+// component derives it from the gRPC status) and fill the value/cas/token, deliberately not
+// reusing the MCBP-bound encode_to/make_response.
+
+#include "core/document_id.hxx"
+#include "core/error_context/key_value.hxx"
+#include "core/operations/document_get.hxx"
+#include "core/operations/document_insert.hxx"
+#include "core/operations/document_remove.hxx"
+#include "core/operations/document_replace.hxx"
+#include "core/operations/document_upsert.hxx"
+#include "core/utils/binary.hxx"
+
+#include <couchbase/cas.hxx>
+#include <couchbase/durability_level.hxx>
+#include <couchbase/error_codes.hxx>
+#include <couchbase/mutation_token.hxx>
+
+#include <couchbase/kv/v1/kv.pb.h>
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace couchbase::core::protostellar::kv
+{
+namespace v1 = ::couchbase::kv::v1;
+
+// Value bytes go through core::utils::to_binary / to_string so the couchbase2 path shares one
+// conversion with the rest of core rather than carrying its own memcpy helpers.
+
+template<typename Request>
+void
+set_location(Request& proto, const document_id& id)
+{
+  proto.set_bucket_name(id.bucket());
+  proto.set_scope_name(id.scope());
+  proto.set_collection_name(id.collection());
+  proto.set_key(id.key());
+}
+
+// The expiry encoding below is dictated by the gateway, which is what translates these messages
+// into MCBP and is therefore the authority on what each shape means. Line references are to
+// couchbase/stellar-gateway at commit 910af15aefcc094378636972dd22875c5e485ee9, and are pinned to
+// the exact file contents by blob hash so they stay checkable after the files move:
+//
+//   kvserver.go  git cat-file -p dde6353d06439fa4d8e0445f2362e8b460ff1265
+//   helpers.go   git cat-file -p 5bd3ad7dcaf3bf2ada354ad9846cf4a87aa73a31
+//
+// The core request carries expiry with the memcached convention: a value of at most the 30-day
+// cutoff is a relative duration in seconds; above it, it is an absolute Unix-epoch timestamp (this
+// is how core/impl/expiry.cxx encodes durations >= 30 days and absolute time points).
+//
+// The proto oneof is not that shape. expiry_secs is *always* a duration: the gateway passes it
+// through when it is <= 2592000 and otherwise re-bases it onto "now + N seconds"
+// (helpers.go:48-67, secsExpiryToGocbcorex), so an already-absolute core value sent through
+// expiry_secs would be read as a ~50-year duration. Absolute values therefore have to use
+// expiry_time, which the gateway takes as an epoch second (helpers.go:44-46). The cutoff itself is
+// relative on both sides: the gateway's own boundary is `expiry <= 2592000`.
+inline constexpr std::uint32_t relative_expiry_cutoff_seconds = 60U * 60U * 24U * 30U; // 2,592,000
+
+// An absent expiry oneof is the gateway's encoding for "preserve the document's existing expiry",
+// not for "no expiry": both Upsert and Replace map a nil oneof to opts.PreserveExpiry = true
+// (kvserver.go:478-484 and :580-581). "No expiry" must therefore be an explicit zero, which the
+// gateway maps to never-expires (helpers.go:49-51); the classic transport serialises the same zero
+// into the extras (core/protocol/cmd_upsert.cxx:82-91). Touch and GetAndTouch reject an absent
+// oneof outright -- "Expiry time specification is unknown." (kvserver.go:297-303, :131-137) -- so
+// the field is not optional there either. The oneof is omitted only when the caller asked to
+// preserve, which for ReplaceRequest is the only available encoding: unlike UpsertRequest it has no
+// preserve_expiry_on_existing field at all.
+template<typename Proto>
+void
+set_expiry(Proto& proto, std::uint32_t expiry, bool preserve_expiry = false)
+{
+  if (preserve_expiry) {
+    return;
+  }
+  if (expiry <= relative_expiry_cutoff_seconds) {
+    proto.set_expiry_secs(expiry);
+  } else {
+    proto.mutable_expiry_time()->set_seconds(static_cast<std::int64_t>(expiry));
+  }
+}
+
+// couchbase::durability_level::none means "no requirement" — the proto field is left unset.
+inline auto
+to_proto_durability(couchbase::durability_level level) -> std::optional<v1::DurabilityLevel>
+{
+  switch (level) {
+    case couchbase::durability_level::none:
+      return std::nullopt;
+    case couchbase::durability_level::majority:
+      return v1::DURABILITY_LEVEL_MAJORITY;
+    case couchbase::durability_level::majority_and_persist_to_active:
+      return v1::DURABILITY_LEVEL_MAJORITY_AND_PERSIST_TO_ACTIVE;
+    case couchbase::durability_level::persist_to_majority:
+      return v1::DURABILITY_LEVEL_PERSIST_TO_MAJORITY;
+  }
+  return std::nullopt;
+}
+
+inline auto
+to_core_token(const v1::MutationToken& token) -> couchbase::mutation_token
+{
+  return couchbase::mutation_token{ token.vbucket_uuid(),
+                                    token.seq_no(),
+                                    static_cast<std::uint16_t>(token.vbucket_id()),
+                                    token.bucket_name() };
+}
+
+// Every response that carries a document body puts it in a content oneof, and the generated
+// accessor for the inactive arm returns the empty-string singleton. Reading content_uncompressed()
+// while the compressed arm is active would therefore produce an empty value with a valid CAS and
+// no error -- indistinguishable from an empty document. Compression negotiation lands in
+// CXXCBC-905; until then every body-bearing decoder refuses the compressed arm instead of losing
+// the value, which is the same fail-closed rule the request side applies.
+template<typename Proto>
+[[nodiscard]] auto
+content_is_compressed(const Proto& proto) -> bool
+{
+  return proto.content_case() == Proto::kContentCompressed;
+}
+
+// ── Get ───────────────────────────────────────────────────────────────────────
+
+inline auto
+encode(const operations::get_request& request) -> v1::GetRequest
+{
+  v1::GetRequest proto;
+  set_location(proto, request.id);
+  return proto;
+}
+
+inline auto
+decode(const v1::GetResponse& proto, key_value_error_context ctx) -> operations::get_response
+{
+  operations::get_response response;
+  response.ctx = std::move(ctx);
+  if (content_is_compressed(proto)) {
+    response.ctx.override_ec(errc::common::feature_not_available);
+    return response;
+  }
+  response.value = utils::to_binary(proto.content_uncompressed());
+  response.cas = couchbase::cas{ proto.cas() };
+  response.flags = proto.content_flags();
+  return response;
+}
+
+// ── Mutations (upsert / insert / replace / remove) ─────────────────────────────
+
+template<typename Response, typename Proto>
+inline auto
+decode_mutation(const Proto& proto, key_value_error_context ctx) -> Response
+{
+  Response response;
+  response.ctx = std::move(ctx);
+  response.cas = couchbase::cas{ proto.cas() };
+  if (proto.has_mutation_token()) {
+    response.token = to_core_token(proto.mutation_token());
+  }
+  return response;
+}
+
+inline auto
+encode(const operations::upsert_request& request) -> v1::UpsertRequest
+{
+  v1::UpsertRequest proto;
+  set_location(proto, request.id);
+  proto.set_content_uncompressed(utils::to_string(request.value));
+  proto.set_content_flags(request.flags);
+  // preserve_expiry is expressed by omitting the oneof, NOT through the proto's
+  // preserve_expiry_on_existing flag. The gateway rejects that flag outright whenever the oneof is
+  // absent -- "Cannot specify preserve expiry with no expiry, leave expiry undefined to preserve
+  // expiry." (kvserver.go:478-482) -- and again when the oneof carries a zero (:498-501), while a
+  // nil oneof already means preserve (:484). Setting both is an InvalidArgument, not a redundancy.
+  set_expiry(proto, request.expiry, request.preserve_expiry);
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+encode(const operations::insert_request& request) -> v1::InsertRequest
+{
+  v1::InsertRequest proto;
+  set_location(proto, request.id);
+  proto.set_content_uncompressed(utils::to_string(request.value));
+  proto.set_content_flags(request.flags);
+  // insert has no preserve_expiry: a document that does not exist yet has no expiry to preserve.
+  set_expiry(proto, request.expiry);
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+encode(const operations::replace_request& request) -> v1::ReplaceRequest
+{
+  v1::ReplaceRequest proto;
+  set_location(proto, request.id);
+  proto.set_content_uncompressed(utils::to_string(request.value));
+  proto.set_content_flags(request.flags);
+  if (request.cas.value() != 0) {
+    proto.set_cas(request.cas.value());
+  }
+  set_expiry(proto, request.expiry, request.preserve_expiry);
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+encode(const operations::remove_request& request) -> v1::RemoveRequest
+{
+  v1::RemoveRequest proto;
+  set_location(proto, request.id);
+  if (request.cas.value() != 0) {
+    proto.set_cas(request.cas.value());
+  }
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+decode(const v1::UpsertResponse& proto, key_value_error_context ctx) -> operations::upsert_response
+{
+  return decode_mutation<operations::upsert_response>(proto, std::move(ctx));
+}
+
+inline auto
+decode(const v1::InsertResponse& proto, key_value_error_context ctx) -> operations::insert_response
+{
+  return decode_mutation<operations::insert_response>(proto, std::move(ctx));
+}
+
+inline auto
+decode(const v1::ReplaceResponse& proto, key_value_error_context ctx)
+  -> operations::replace_response
+{
+  return decode_mutation<operations::replace_response>(proto, std::move(ctx));
+}
+
+inline auto
+decode(const v1::RemoveResponse& proto, key_value_error_context ctx) -> operations::remove_response
+{
+  return decode_mutation<operations::remove_response>(proto, std::move(ctx));
+}
+
+} // namespace couchbase::core::protostellar::kv
--- a/test/cng/protostellar/kv_converter.cxx
+++ b/test/cng/protostellar/kv_converter.cxx
@@ -1,0 +1,330 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Unit tests for the KV converter (CXXCBC-892). Pure translation, no server (env-agnostic).
+
+#include "framework/test_runner.hxx"
+
+#include "core/protostellar/kv_converter.hxx"
+#include "core/utils/binary.hxx"
+
+#include <couchbase/error_codes.hxx>
+
+#include <array>
+#include <cstdint>
+#include <string>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+namespace pk = ::couchbase::core::protostellar::kv;
+namespace ops = ::couchbase::core::operations;
+namespace v1 = ::couchbase::kv::v1;
+namespace cu = ::couchbase::core::utils;
+using ::couchbase::core::document_id;
+using ::couchbase::core::key_value_error_context;
+
+auto
+test_id() -> document_id
+{
+  return document_id{ "b", "s", "c", "k" };
+}
+
+void
+get_request_encodes_location()
+{
+  ops::get_request request;
+  request.id = document_id{ "travel-sample", "inventory", "airline", "airline_10" };
+  const auto proto = pk::encode(request);
+  assert_eq(proto.bucket_name(), std::string{ "travel-sample" }, "bucket");
+  assert_eq(proto.scope_name(), std::string{ "inventory" }, "scope");
+  assert_eq(proto.collection_name(), std::string{ "airline" }, "collection");
+  assert_eq(proto.key(), std::string{ "airline_10" }, "key");
+}
+
+void
+get_response_decodes_value_cas_flags()
+{
+  v1::GetResponse proto;
+  proto.set_content_uncompressed("{\"type\":\"airline\"}");
+  proto.set_cas(0x1234'5678'9abc'def0ULL);
+  proto.set_content_flags(0x0200'0006U);
+  const auto response = pk::decode(proto, key_value_error_context{});
+  assert_false(static_cast<bool>(response.ctx.ec()), "uncompressed content decodes without error");
+  assert_eq(cu::to_string(response.value), std::string{ "{\"type\":\"airline\"}" }, "value");
+  assert_eq(response.cas.value(), 0x1234'5678'9abc'def0ULL, "cas");
+  assert_eq(response.flags, static_cast<std::uint32_t>(0x0200'0006), "flags");
+}
+
+// GetResponse.content is a oneof, and the accessor for the inactive arm returns the empty-string
+// singleton. Reading content_uncompressed() while the compressed arm is active would yield an empty
+// value with a valid CAS and no error, which no caller can tell apart from an empty document. Until
+// snappy negotiation lands (CXXCBC-905) the decoder has to fail closed.
+void
+get_response_refuses_compressed_content()
+{
+  v1::GetResponse proto;
+  proto.set_content_compressed("not-really-snappy");
+  proto.set_cas(0x1234ULL);
+  proto.set_content_flags(0x0200'0006U);
+
+  const auto response = pk::decode(proto, key_value_error_context{});
+  assert_true(response.ctx.ec() == errc::common::feature_not_available,
+              "the compressed arm is refused rather than decoded as an empty value");
+  assert_true(response.value.empty(), "no partial value is handed back");
+}
+
+// The three branches of set_expiry, plus the exact cutoff. Every expectation below is behaviour
+// read out of the gateway; core/protostellar/kv_converter.hxx pins the commit and the blob hashes
+// the kvserver.go / helpers.go line references are against.
+//
+//   * a nil expiry oneof means "preserve the existing expiry" for both Upsert and Replace
+//     (kvserver.go:478-484, :580-581), so "no expiry" has to be an explicit zero or a default
+//     upsert/replace inherits a TTL that couchbase:// clears;
+//   * expiry_secs == 0 is never-expires (helpers.go:49-51);
+//   * expiry_secs is always a *duration* -- passed through when <= 2592000 and otherwise re-based
+//     onto now (helpers.go:60-66) -- so a core value above the cutoff, which is already an absolute
+//     epoch second, must travel in expiry_time instead (helpers.go:44-46);
+//   * the gateway's own relative/absolute boundary is `expiry <= 2592000`, which is why 2592000
+//     itself belongs on the relative arm.
+struct expiry_case {
+  std::uint32_t expiry;
+  bool relative;
+  std::uint32_t seconds;
+  std::int64_t timestamp_seconds;
+};
+
+constexpr std::array<expiry_case, 4> expiry_cases{ {
+  { 0U, true, 0U, 0 },
+  { 1U, true, 1U, 0 },
+  { 2'592'000U, true, 2'592'000U, 0 }, // exactly 30 days is still relative
+  { 2'592'001U, false, 0U, 2'592'001 },
+} };
+
+template<typename Proto>
+void
+assert_expiry(const Proto& proto, const expiry_case& expected, const std::string& what)
+{
+  assert_true(proto.expiry_case() != Proto::EXPIRY_NOT_SET,
+              what + ": the expiry oneof must be set unless preserve_expiry was requested");
+  if (expected.relative) {
+    assert_true(proto.expiry_case() == Proto::kExpirySecs, what + ": relative arm selected");
+    assert_eq(proto.expiry_secs(), expected.seconds, what + ": expiry seconds");
+  } else {
+    assert_true(proto.expiry_case() == Proto::kExpiryTime, what + ": absolute arm selected");
+    assert_eq(proto.expiry_time().seconds(), expected.timestamp_seconds, what + ": expiry time");
+  }
+}
+
+// upsert, insert and replace all share set_expiry, so every branch is pinned on each of them.
+void
+expiry_encodes_every_branch_explicitly()
+{
+  for (const auto& expected : expiry_cases) {
+    const auto label = "expiry=" + std::to_string(expected.expiry);
+
+    ops::upsert_request upsert;
+    upsert.id = test_id();
+    upsert.value = cu::to_binary("v");
+    upsert.expiry = expected.expiry;
+    assert_expiry(pk::encode(upsert), expected, "upsert " + label);
+
+    ops::insert_request insert;
+    insert.id = test_id();
+    insert.value = cu::to_binary("v");
+    insert.expiry = expected.expiry;
+    assert_expiry(pk::encode(insert), expected, "insert " + label);
+
+    ops::replace_request replace;
+    replace.id = test_id();
+    replace.value = cu::to_binary("v");
+    replace.expiry = expected.expiry;
+    assert_expiry(pk::encode(replace), expected, "replace " + label);
+  }
+}
+
+// preserve_expiry wins over expiry, which is what upsert_options and replace_options document, and
+// the gateway's encoding for it is an omitted oneof (kvserver.go:484 for upsert, :581 for replace).
+void
+preserve_expiry_omits_the_expiry_oneof()
+{
+  ops::upsert_request upsert;
+  upsert.id = test_id();
+  upsert.value = cu::to_binary("v");
+  upsert.expiry = 300U;
+  upsert.preserve_expiry = true;
+  const auto upsert_proto = pk::encode(upsert);
+  assert_true(upsert_proto.expiry_case() == v1::UpsertRequest::EXPIRY_NOT_SET,
+              "upsert expresses preserve by omitting the expiry oneof");
+
+  // ReplaceRequest has no preserve_expiry_on_existing field at all, so an absent oneof is the only
+  // available encoding there.
+  ops::replace_request replace;
+  replace.id = test_id();
+  replace.value = cu::to_binary("v");
+  replace.expiry = 300U;
+  replace.preserve_expiry = true;
+  assert_true(pk::encode(replace).expiry_case() == v1::ReplaceRequest::EXPIRY_NOT_SET,
+              "replace expresses preserve by omitting the expiry oneof");
+
+  // ...and with the flag off, a default replace must clear the TTL rather than inherit it.
+  ops::replace_request plain;
+  plain.id = test_id();
+  plain.value = cu::to_binary("v");
+  const auto plain_proto = pk::encode(plain);
+  assert_true(plain_proto.expiry_case() == v1::ReplaceRequest::kExpirySecs,
+              "a default replace sends an explicit expiry");
+  assert_eq(plain_proto.expiry_secs(), static_cast<std::uint32_t>(0), "...and it is zero");
+}
+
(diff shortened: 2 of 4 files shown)


core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,250 @@
...
+  proto.set_collection_name(id.collection());
+  proto.set_key(id.key());
+}
+
+// The core request carries expiry with the memcached convention: a value below the 30-day cutoff
+// is a relative duration in seconds; at or above it, it is an absolute Unix-epoch timestamp
+// (this is how core/impl/expiry.cxx encodes durations >= 30 days and absolute time points). The
+// proto expiry oneof distinguishes the two explicitly and the gateway does NOT apply the 30-day
+// heuristic, so the client must disambiguate here. 0 means "no expiry" and leaves the oneof unset.
+inline constexpr std::uint32_t relative_expiry_cutoff_seconds = 60U * 60U * 24U * 30U; // 2,592,000
```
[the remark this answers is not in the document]
avsej: kv_converter decodes the already-encoded memcached-convention expiry to pick the proto oneof branch; core/impl/expiry.cxx encodes a duration/time_point into that value. Only the well-known 30-day constant overlaps, and the comment documents the coupling. Keeping as-is.

core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,199 @@
...
+      pending.assign(outstanding_.begin(), outstanding_.end());
+    }
+    for (auto* context : pending) {
+      context->TryCancel();
+    }
+    std::unique_lock lock{ mutex_ };
+    idle_.wait(lock, [this] {
+      return outstanding_.empty();
+    });
+  }
```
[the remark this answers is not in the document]
avsej: gRPC guarantees a completion (CANCELLED) for every call after TryCancel, so the wait is bounded in practice. Adding a timeout would risk abandoning a call whose late completion then posts onto a freed io_context -- the exact use-after-free this drain exists to prevent.

core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,207 @@
...
+        *state->context,
+        state->response,
+        [state, handler_holder, io, tracker, context](grpc::Status status) {
+          // Deregister on the gRPC thread first: once removed, a concurrent cancel_and_drain() may
+          // return. The post that follows only touches shared state (io outlives the call).
+          tracker->remove(context);
+          asio::post(*io, [state, handler_holder, status = std::move(status)]() mutable {
+            (*handler_holder)(std::move(status), std::move(state->response));
+          });
+        });
```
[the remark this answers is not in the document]
avsej: Deliberate and required as written. The completion posts the handler and, in the streaming reactor, the posted task self-deletes the reactor -- which owns context_. Removing from the tracker *after* posting would let the io thread run that self-delete and free context_ before the gRPC thread reads context_.get() in remove(), reintroducing a use-after-free. remove()-before-post() is kept consistent across unary() and the reactor; the safety rests on the documented SDK invariant that the io_context outlives the dispatcher, so a post issued just before remove() can never target a destroyed io_context.

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. This was a static review — I traced the code paths by hand and cross-checked against the Java Protostellar implementation, nothing was compiled or run; please treat the reasoning as a starting point for discussion, not ground truth.

Reviewed the incremental commit only (`895122da`, +410/-0 vs #1029): `core/protostellar/kv_converter.hxx` plus its unit test.

**`replace`'s expiry oneof doesn't distinguish "preserve" from "clear."**

`ReplaceRequest` has no `preserve_expiry_on_existing` field (unlike `UpsertRequest`), and that's by design — Java's [`CoreProtostellarKeyValueRequests.replaceRequest`](https://github.com/couchbase/couchbase-jvm-clients/blob/ceba23c64840f44fd21a31548b5dfec5e056f111/core-io/src/main/java/com/couchbase/client/core/protostellar/kv/CoreProtostellarKeyValueRequests.java#L256-L316) uses the `expiry` oneof itself as the signal: leaving it **unset** means preserve; explicitly setting it to the zero `Timestamp` ([`NO_EXPIRY = Timestamp.getDefaultInstance()`](https://github.com/couchbase/couchbase-jvm-clients/blob/ceba23c64840f44fd21a31548b5dfec5e056f111/core-io/src/main/java/com/couchbase/client/core/protostellar/kv/CoreProtostellarKeyValueRequests.java#L82)) means clear; setting it to a value means a new expiry:

```java
if (!preserveExpiry) {
  expiry.when(
    absolute -> request.setExpiryTime(toExpiryTime(absolute)),
    relative -> request.setExpirySecs(toExpirySeconds(relative)),
    () -> request.setExpiryTime(NO_EXPIRY)
  );
}
// preserveExpiry == true -> oneof left completely unset
```

`encode(const operations::replace_request&)` here doesn't reproduce that three-way split. It always calls the shared `set_expiry()`, which no-ops (leaves the oneof unset) whenever `expiry == 0`, and it never reads `request.preserve_expiry` at all:

```cpp
inline auto
encode(const operations::replace_request& request) -> v1::ReplaceRequest
{
  ...
  set_expiry(proto, request.expiry);
  ...
```

Two concrete effects, in order of how often they'd actually be hit:

1. A **plain `replace()` with no options** — `preserve_expiry=false`, `expiry=0` by default — leaves the oneof unset, which per the scheme above reads as "preserve existing expiry." The expected/MCBP behavior for a bare `replace()` is to *clear* expiry to never-expire, so today the common case gets the wrong wire behavior, not just the `preserve_expiry(true)` path.
2. `replace_options().preserve_expiry(true).expiry(300s)` — a combination `replace_options.hxx`'s own doc comment says resolves to "preserve wins" — will send `expiry_secs=300` instead, since `preserve_expiry` is never consulted.

Matching Java's structure would need something like:

```cpp
if (!request.preserve_expiry) {
  if (request.expiry == 0) {
    proto.mutable_expiry_time();               // explicit zero Timestamp, not "unset"
  } else if (request.expiry < relative_expiry_cutoff_seconds) {
    proto.set_expiry_secs(request.expiry);
  } else {
    proto.mutable_expiry_time()->set_seconds(request.expiry);
  }
}
// else: leave the oneof untouched
```
(and leaving `insert`/`upsert` on the current shared `set_expiry()` — they have no preserve concept for `insert`, and `upsert` already carries its own dedicated `preserve_expiry_on_existing` bool, so the unset-vs-zero distinction doesn't matter for either of them.)

No test currently exercises `replace`'s expiry/preserve_expiry interaction either way, which is how this went unnoticed.

**Suggestion: add a test for `replace`'s expiry/preserve_expiry interaction.**

Given the above, it'd be worth a unit test alongside `upsert_request_encodes_all_fields`/`durability_none_is_left_unset` that pins down `replace`'s expiry oneof behavior — e.g. asserting a bare `replace_request` (defaults) leaves the oneof unset today (documenting the current gap), plus cases for an explicit relative/absolute expiry and for `preserve_expiry(true)`. That would both lock in whatever fix lands here and make the current gap visible in CI rather than only in review.

**Minor: `to_bytes`/`to_string` reinvent existing utilities.**

`core::utils::to_binary(std::string_view) -> binary` and `core::utils::to_string(const std::vector<std::byte>&) -> std::string` (`core/utils/binary.hxx`) already do exactly what this file's `to_bytes`/`to_string` hand-roll with `memcpy`. Reusing those would drop ~15 lines and keep one canonical byte↔string conversion instead of two.

Everything else looks solid: the durability-level mapping is exhaustive and matches `none → unset`, the relative/absolute expiry cutoff matches `core/impl/expiry.cxx`'s own 30-day boundary exactly, and `to_core_token`'s field mapping (uuid/seqno/partition-id/bucket) is correct and has a dedicated test — the place a copy-paste field swap would most likely hide.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,251 @@
...
+
+template<typename Proto>
+void
+set_expiry(Proto& proto, std::uint32_t expiry)
+{
+  if (expiry == 0) {
+    return;
+  }
```
avsej: **Blocker: an unset `expiry` oneof is the wire encoding for "preserve expiry", not for "no expiry".**

Returning early leaves the oneof unset, and the gateway reads an absent oneof as "preserve the document's existing expiry". gocb states this in a source comment (`kvprovider_ps.go:419`) and all three reference clients encode the zero explicitly instead — gocb `ExpirySecs: 0`, couchbase-jvm-clients its `NO_EXPIRY` constant, couchbase-net-client `ExpirySecs = 0`. The classic MCBP path in this repository also sends a real zero (`core/protocol/cmd_upsert.cxx:82-91`).

Consequence: a default `upsert` or `replace` over `couchbase2://` preserves a TTL that `couchbase://` would clear, so a re-written document silently expires.

Taking the preserve-expiry flag as a parameter fixes this and the dropped `replace_request.preserve_expiry` flag together:

```suggestion
set_expiry(Proto& proto, std::uint32_t expiry, bool preserve_expiry = false)
{
  if (preserve_expiry) {
    return;
  }
  if (expiry == 0) {
    proto.set_expiry_secs(0);
    return;
  }
```

The upsert encoder then passes `request.preserve_expiry` here instead of relying on `set_preserve_expiry_on_existing` alone, and the replace encoder passes it too.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,251 @@
...
+  proto.set_content_flags(request.flags);
+  set_expiry(proto, request.expiry);
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+encode(const operations::replace_request& request) -> v1::ReplaceRequest
```
avsej, line 219: `operations::replace_request::preserve_expiry` is never read here, while the upsert encoder honours its equivalent via `set_preserve_expiry_on_existing`.

`ReplaceRequest` has no such proto field, so omitting the expiry oneof is the only way to express preserve for replace — which is how gocb implements it (`kvprovider_ps.go:419-426`). Threading the flag through `set_expiry` covers both this and the zero-expiry case:

```cpp
set_expiry(proto, request.expiry, request.preserve_expiry);
```

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,251 @@
...
+}
+
+inline auto
+decode(const v1::GetResponse& proto, key_value_error_context ctx) -> operations::get_response
+{
+  operations::get_response response;
+  response.ctx = std::move(ctx);
+  // Decode uncompressed payload. Compression negotiation for content_compressed is handled in
+  // CXXCBC-905.
+  response.value = to_bytes(proto.content_uncompressed());
```
avsej: `GetResponse.content` is a oneof. When `content_compressed` is the active arm this accessor returns the empty-string singleton, so the response decodes to an empty value with a valid CAS, correct flags and no error — indistinguishable from an empty document.

Deferring compression support is fine; deferring it as a silent empty read is not. Fail closed until the compressed arm is handled:

```cpp
  if (proto.has_content_compressed()) {
    // Compression negotiation lands in CXXCBC-905; refuse rather than returning an empty
    // value that cannot be distinguished from an empty document.
    return errc::common::feature_not_available;
  }
  response.value = to_bytes(proto.content_uncompressed());
```

avsej (commented):
## Review — PR #1030 (`CXXCBC-892`)

This commit adds `core/protostellar/kv_converter.hxx`, the translation layer between the core KV request/response structs and `couchbase.kv.v1`. The mechanical mapping is careful and mostly matches the shipping Protostellar clients field for field — durability, CAS presence, content flags and the mutation token are all correct. One encoding decision is wrong in a way that causes silent data loss, and because this commit establishes the conventions that eight later converters copy, it is worth fixing here rather than downstream.

### Strengths

- The `encode` / `decode` / `can_encode` triad established here is the right shape, and the `can_encode` gate in particular is a good decision: request shapes with no couchbase2 mapping fail with `feature_not_available` instead of being silently truncated.
- `to_proto_durability()` (`core/protostellar/kv_converter.hxx:110`) returns `std::optional` and maps `durability_level::none` to an unset field rather than to an enum zero. This matches gocb, couchbase-jvm-clients and couchbase-net-client, and it is the kind of distinction that is easy to get wrong.
- CAS is written only when non-zero, so an unset CAS does not reach the wire as a literal `0`. Again consistent with all three reference implementations.
- `to_core_token()` builds `couchbase::mutation_token` in the correct argument order and checks `has_mutation_token()` before reading. 
- The 30-day relative/absolute disambiguation of the core `uint32 expiry` is a genuinely correct adaptation. The reference SDKs do not need it because they carry a tagged expiry type; deriving it from the memcached convention that `core/impl/expiry.cxx` encodes is the right call, and the reasoning is written down at `core/protostellar/kv_converter.hxx:85-93`.

### Findings

|---|---|---|---|---|
| 1 | blocker | protocol-parity | A zero expiry leaves the proto `expiry` oneof unset, which the gateway reads as "preserve expiry" | `core/protostellar/kv_converter.hxx:90` |
| 2 | high | protocol-parity | `encode(replace_request)` never reads `request.preserve_expiry` | `core/protostellar/kv_converter.hxx:196` |
| 3 | medium | correctness | The `GetResponse` decoder reads one arm of the `content` oneof unguarded | `core/protostellar/kv_converter.hxx:143` |
| 4 | medium | test-coverage | Neither the absolute-expiry branch nor `expiry == 0` has any test | `test/cng/protostellar/kv_converter.cxx:69` |
| 5 | low | protocol-parity | The 30-day cutoff uses `<` where the convention is "greater than 30 days" | `core/protostellar/kv_converter.hxx:93` |
| 6 | low | simplification | `to_bytes` / `to_string` duplicate existing core utilities | `core/protostellar/kv_converter.hxx:50` |

#### 1. A zero expiry leaves the `expiry` oneof unset, and the gateway reads that as "preserve expiry" (blocker)

```cpp
template<typename Proto>
void
set_expiry(Proto& proto, std::uint32_t expiry)
{
  if (expiry == 0) {
    return;                                  // leaves the oneof unset
  }
  if (expiry < relative_expiry_cutoff_seconds) {
    proto.set_expiry_secs(expiry);
  } else {
    proto.mutable_expiry_time()->set_seconds(static_cast<std::int64_t>(expiry));
  }
}
```

An absent `expiry` oneof is the gateway's encoding for "preserve the document's existing expiry", not for "no expiry". This is stated explicitly in the Go client — gocb `kvprovider_ps.go:419` carries the comment `// CNG interprets a nil expiry value as "preserve expiry"` — and all three reference implementations therefore encode the zero rather than omitting the field:

- gocb sets `ExpirySecs: 0` on the `else` branch for upsert (`kvprovider_ps.go:354-359`), replace (`:421-426`) and mutate_in (`:226-232`), guarded by `if !opts.PreserveExpiry`.
- couchbase-jvm-clients defines `static final Timestamp NO_EXPIRY = Timestamp.getDefaultInstance()` (`CoreProtostellarKeyValueRequests.java:82`) and sends it for `CoreExpiry.NONE` at `:294` (replace, inside `if (!preserveExpiry)`), `:351` (upsert) and `:483` (touch).
- couchbase-net-client sets `request.ExpirySecs = 0` (`StellarCollection.cs:765`, `:786`) with the same guard.

The classic transport in this repository behaves the same way: `core/protocol/cmd_upsert.cxx:82-91` and `cmd_replace.cxx:82-91` unconditionally serialise the expiry into the extras, so a zero reaches the server as a real "clear the TTL", and preserve-expiry is a separate frame info added only on request.

Failure scenario: a document is written with a 60-second TTL, then upserted with default options — no expiry, no `preserve_expiry`. Over `couchbase://`, and over `couchbase2://` with any of the three reference clients, the document becomes permanent. Over `couchbase2://` with this converter the gateway sees no expiry field, preserves the 60-second TTL, and the freshly written document disappears a minute later. The same applies to `replace`. This is the default path for every upsert and replace in the SDK.

`insert` is unaffected in practice, since a new document has no prior expiry to preserve.

The fix is to make "no expiry" explicit and to let the caller's preserve-expiry intent suppress the field instead:

```cpp
// An absent expiry oneof is the gateway's encoding for "preserve the existing expiry", so a
// request for "no expiry" must send the zero explicitly. Omit the oneof only when the caller
// asked to preserve.
template<typename Proto>
void
set_expiry(Proto& proto, std::uint32_t expiry, bool preserve_expiry = false)
{
  if (preserve_expiry) {
    return;
  }
  if (expiry == 0) {
    proto.set_expiry_secs(0);
  } else if (expiry <= relative_expiry_cutoff_seconds) {
    proto.set_expiry_secs(expiry);
  } else {
    proto.mutable_expiry_time()->set_seconds(static_cast<std::int64_t>(expiry));
  }
}
```

(inline comment on `core/protostellar/kv_converter.hxx:90`)

#### 2. `encode(replace_request)` never reads `request.preserve_expiry` (high)

`operations::replace_request` carries `bool preserve_expiry`, and `encode(const operations::upsert_request&)` honours the equivalent flag through `set_preserve_expiry_on_existing`. The replace encoder does not read the field at all.

`ReplaceRequest` has no `preserve_expiry_on_existing` field in the proto, so for replace the *absence* of the expiry oneof is the only way to express preserve — which is exactly how gocb implements it (`kvprovider_ps.go:419-426`). Two consequences follow from finding 1:

- `preserve_expiry = true` together with a non-zero expiry silently overwrites the TTL instead of preserving it.
- `preserve_expiry = false` with the default zero expiry accidentally preserves it.

Threading the flag into `set_expiry` as shown above fixes both cases at once:

```cpp
set_expiry(proto, request.expiry, request.preserve_expiry);
```

(inline comment on `core/protostellar/kv_converter.hxx:196`)

#### 3. The `GetResponse` decoder reads one arm of the `content` oneof unguarded (medium)

```cpp
response.value = to_bytes(proto.content_uncompressed());
```

`GetResponse.content` is a oneof over `content_uncompressed` and `content_compressed`. When the compressed arm is active, the generated accessor for the inactive arm returns the empty-string singleton, so the decode produces an empty value with a valid CAS, correct flags, and **no error**.

The commit message states that compressed-content decoding is deferred until compression negotiation lands. Deferring the capability is reasonable; deferring it as a silent empty read is not, because the failure is indistinguishable from an empty document. Until the compressed arm is handled, the deferral should be explicit:

```cpp
if (proto.has_content_compressed()) {
  // Compression negotiation is not implemented yet (CXXCBC-905); refuse rather than
  // returning an empty value that is indistinguishable from an empty document.
  return errc::common::feature_not_available;
}
response.value = to_bytes(proto.content_uncompressed());
```

This is the same fail-closed convention that `can_encode` applies on the request side.

(inline comment on `core/protostellar/kv_converter.hxx:143`)

#### 4. Neither the absolute-expiry branch nor `expiry == 0` is tested (medium)

`test/cng/protostellar/kv_converter.cxx` exercises expiry only with values in the middle of the range (`300`, `120`, `100`). The three-way split in `set_expiry` therefore has one covered branch out of three: the zero branch and the absolute-timestamp branch have no coverage anywhere in the series, and the boundary value `2592000` is untested. The ticket lists the absolute-versus-relative edge cases as an acceptance criterion.

A table-driven case over the branch boundaries would pin all of them, and would have caught finding 1:

```cpp
void
upsert_encodes_every_expiry_branch()
{
  struct { std::uint32_t expiry; bool expect_secs; std::uint32_t secs; std::int64_t abs; } cases[] = {
    { 0U,       true, 0U,       0 },          // "no expiry" must be explicit, not an unset oneof
    { 1U,       true, 1U,       0 },
    { 2592000U, true, 2592000U, 0 },          // exactly 30 days is still relative
    { 2592001U, false, 0U,      2592001 },    // above the cutoff is an absolute timestamp
  };
  for (const auto& c : cases) {
    operations::upsert_request request{};
    request.id = document_id{ "default", "_default", "_default", "key" };
    request.expiry = c.expiry;
    const auto proto = kv::encode(request);
    assert_true(proto.expiry_case() != v1::UpsertRequest::EXPIRY_NOT_SET,
                "expiry oneof must always be set unless preserve_expiry was requested");
    if (c.expect_secs) {
      assert_eq(proto.expiry_secs(), c.secs, "relative expiry in seconds");
    } else {
      assert_eq(proto.expiry_time().seconds(), c.abs, "absolute expiry timestamp");
    }
  }
}
```

(inline comment on `test/cng/protostellar/kv_converter.cxx:69`)

#### 5. The 30-day cutoff uses `<` where the convention is "greater than 30 days" (low)

```cpp
if (expiry < relative_expiry_cutoff_seconds) {
```

Under the memcached convention a value *greater than* 30 days is absolute, so `2592000` itself is a 30-day relative TTL. With `<`, that exact value is sent as `expiry_time.seconds = 2592000`, which is 1970-01-31.

The public API is insulated: `core/impl/expiry.cxx` applies the same comparison before this point, and `expiry_absolute` rejects any time point earlier than 31 days, so a legitimate absolute value can never be `2592000`. The exposure is limited to callers that populate `operations::*_request::expiry` directly with memcached semantics. `<=` is strictly safer with no reachable downside.

(inline comment on `core/protostellar/kv_converter.hxx:93`)

#### 6. `to_bytes` / `to_string` duplicate existing core utilities (low)

`core::utils::to_binary` and `core::utils::to_string` already provide these conversions, and one of the new helpers has an identical signature in a second namespace. Reusing the existing utilities keeps a single conversion path for value bytes.

(inline comment on `core/protostellar/kv_converter.hxx:50`)

### Notes on things that were checked and are not defects

- The header-only `.hxx` form was checked against `AGENTS.md`. The file-extension guidance covers which extensions to use, not a requirement that every definition live in a `.cxx`, and `core/` already contains header-only helpers. This is consistent with the codebase.
- `content_flags` passthrough, counter `initial` presence handling, and `lock_time_secs` units all match the reference implementations.

### Verdict

Needs changes. Finding 1 is a silent-data-loss defect on the default upsert and replace path and should be fixed before this lands, since eight later converters build on this file's conventions; findings 2 and 3 are the same root cause and the same small change.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,251 @@
...
+inline constexpr std::uint32_t relative_expiry_cutoff_seconds = 60U * 60U * 24U * 30U; // 2,592,000
+
+template<typename Proto>
+void
+set_expiry(Proto& proto, std::uint32_t expiry)
+{
+  if (expiry == 0) {
+    return;
+  }
+  if (expiry < relative_expiry_cutoff_seconds) {
```
avsej: Under the memcached convention a value *greater than* 30 days is absolute, so `2592000` is itself a 30-day relative TTL. With `<` it is encoded as `expiry_time.seconds = 2592000`, i.e. 1970-01-31.

The public API is insulated (`core/impl/expiry.cxx` rejects absolute values below 31 days), so only direct core callers are exposed. `<=` is strictly safer with no reachable downside.

```suggestion
  if (expiry <= relative_expiry_cutoff_seconds) {
```

test/cng/protostellar/kv_converter.cxx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+void
+upsert_request_encodes_all_fields()
+{
+  ops::upsert_request request;
+  request.id = document_id{ "b", "s", "c", "k" };
+  request.value = pk::to_bytes("document-body");
+  request.flags = 0x0000'0006U;
+  request.expiry = 300U;
```
avsej, line 239: Every expiry test in the series uses a value in the middle of the range, so one of `set_expiry`'s three branches is covered. The zero branch and the absolute-timestamp branch have no coverage anywhere, and the boundary `2592000` is untested — the ticket names these edge cases as an acceptance criterion.

A table-driven case asserting `proto.expiry_case() != EXPIRY_NOT_SET` across `{0, 1, 2592000, 2592001}` would pin all three branches and would have caught the zero-expiry defect above.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,251 @@
...
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace couchbase::core::protostellar::kv
+{
+namespace v1 = ::couchbase::kv::v1;
+
+inline auto
+to_bytes(const std::string& source) -> std::vector<std::byte>
```
avsej: `core::utils::to_binary` and `core::utils::to_string` already provide these conversions, and one of these helpers has an identical signature in a second namespace. Reusing them keeps a single conversion path for value bytes.

avsej:
## Response to the reviews on `CXXCBC-892`

All nine points across the two reviews are addressed.

Nothing below rests on reading client SDKs. The expiry contract is a property of the gateway, so it
was read out of the gateway source and then **measured against a running one**. Two references are
used throughout:

- **Source.** couchbase/stellar-gateway at commit `910af15aefcc094378636972dd22875c5e485ee9`. Line
  references are pinned to exact file contents by blob hash, so they stay checkable if the files
  move: `kvserver.go` is blob `dde6353d06439fa4d8e0445f2362e8b460ff1265`, `helpers.go` is blob
  `5bd3ad7dcaf3bf2ada354ad9846cf4a87aa73a31` (`git cat-file -p <blob>`).
- **Measurement.** A probe driving raw `couchbase.kv.v1` messages against a CNG gateway
  (`gateway-version 1.2.2-123`, deployed via the operator), so the behaviour of the *shipped* gateway
  is confirmed and not just that of `master`. Every table below is its recorded output.

The fix is applied at commit 08 (`CXXCBC-892`), because the encoding convention established here is
inherited by the later converters, and propagated up the stack where later commits extend or
supersede it. Commits touched: 08, 09, 10, 12, 25.

## Review `4841169479`

### 1. A zero expiry leaves the `expiry` oneof unset, and the gateway reads that as "preserve expiry" (blocker)

> An absent `expiry` oneof is the gateway's encoding for "preserve the document's existing expiry",
> not for "no expiry".

Confirmed, in source and by measurement. The gateway maps a nil oneof to `opts.PreserveExpiry = true`
for both Upsert (`kvserver.go:478-484`) and Replace (`:580-581`), and `expiry_secs == 0` to
never-expires (`helpers.go:49-51`). Measured against `1.2.2-123`, writing a document with a 60s TTL
and then re-writing it:

| shape sent | status | TTL afterwards |
| --- | --- | --- |
| `upsert{expiry:<unset>}` | OK | **`now+60` — preserved** |
| `upsert{expiry_secs:0}` | OK | **none — cleared** |
| `replace{expiry:<unset>}` | OK | **`now+60` — preserved** |
| `replace{expiry_secs:0}` | OK | **none — cleared** |

So the failure scenario in the finding is real and was reproduced: before this change a default
`upsert` or `replace` over `couchbase2://` inherited a TTL that `couchbase://` clears. The classic
path serialises the zero unconditionally for the same reason
(`core/protocol/cmd_upsert.cxx:82-91`, `cmd_replace.cxx:82-91`).

The fix is as suggested, with one simplification: once the cutoff comparison becomes `<=`
(finding 5), the separate `expiry == 0` branch is redundant, since `0 <= 2592000` already routes to
`set_expiry_secs(0)`.

```cpp
template<typename Proto>
void
set_expiry(Proto& proto, std::uint32_t expiry, bool preserve_expiry = false)
{
  if (preserve_expiry) {
    return;
  }
  if (expiry <= relative_expiry_cutoff_seconds) {
    proto.set_expiry_secs(expiry);
  } else {
    proto.mutable_expiry_time()->set_seconds(static_cast<std::int64_t>(expiry));
  }
}
```

**On which arm encodes "no expiry."** The two reviews differ — `expiry_secs = 0` versus
`expiry_time = NO_EXPIRY` (a zero `Timestamp`). Measured, the two are equivalent: the gateway
converts a `Timestamp` with `uint32(expiry.Unix())` (`helpers.go:44-46`), so a zero `Timestamp`
also becomes 0, and `upsert{expiry_time:0}` was observed to clear the TTL exactly as
`upsert{expiry_secs:0}` did. `expiry_secs = 0` is used, since it states the intent directly.

**`set_expiry` is shared by six encoders**, not the two named. The corrected behaviour therefore also
changes `touch`, `get_and_touch`, `increment` and `decrement`, which previously sent no expiry field
at all for a zero expiry. For `touch` that was not merely wrong but rejected — its arm dispatch falls
through to an error (`kvserver.go:297-303`), measured:

| shape sent | status |
| --- | --- |
| `touch{expiry:<unset>}` | **`InvalidArgument` — "Expiry time specification is unknown."** |
| `touch{expiry_secs:0}` | OK, TTL cleared |

So `touch(0)` over `couchbase2://` could not have worked at all before this change. `get_and_touch`
has the same dispatch (`:131-137`).

`insert` now also sends an explicit zero. As the finding notes, a new document has no expiry to
preserve, so this is immaterial; the gateway only reads the field when present
(`kvserver.go:355-362`). It was kept explicit for uniformity across the six encoders.

### A property neither review raised, found while verifying the above

`expiry_secs` is **always a duration**, never an epoch second. Above the 30-day cutoff the gateway
does not pass it through — it re-bases it onto `now` (`helpers.go:60-66`):

```go
// Send the duration as a unix timestamp of now plus duration.
return uint32(time.Now().Add(time.Duration(expiry) * time.Second).Unix())
```

Measured, `upsert{expiry_secs:2592001}` produced an observed expiry of `now+2592001`, not
`1970-01-31`. The core `expiry` field, by contrast, is *already* an absolute epoch second above the
cutoff (`core/impl/expiry.cxx`). Sending such a value through `expiry_secs` would therefore have the
gateway compute `now + ~56 years`. Routing values above the cutoff to `expiry_time` — which the
gateway takes as an epoch second, measured exact — is what prevents that, so the disambiguation this
converter performs is load-bearing rather than cosmetic. This is recorded in the header comment,
since it is the non-obvious half of why the cutoff exists at all.

### 2. `encode(replace_request)` never reads `request.preserve_expiry` (high)

> `ReplaceRequest` has no `preserve_expiry_on_existing` field in the proto, so for replace the
> *absence* of the expiry oneof is the only way to express preserve

Confirmed — the field exists at `core/operations/document_replace.hxx:59` and was not read. Fixed by
threading the flag, as suggested:

```cpp
set_expiry(proto, request.expiry, request.preserve_expiry);
```

The flag is threaded through for `upsert` too, and there the first attempt at this was **wrong** in a
way only the gateway could reveal. Setting `preserve_expiry_on_existing` alongside the omitted oneof
— the shape that looks like "say it twice, unambiguously" — is rejected outright:

| shape sent | status |
| --- | --- |
| `upsert{preserve:true, expiry:<unset>}` | **`InvalidArgument` — "Cannot specify preserve expiry with no expiry, leave expiry undefined to preserve expiry."** |
| `upsert{preserve:true, expiry_secs:0}` | **`InvalidArgument` — "Cannot specify preserve expiry with zero expiry, leave expiry undefined to preserve expiry."** |

Source: `kvserver.go:478-482` and `:498-501`. The gateway's own error text states the intended
encoding, so `preserve_expiry_on_existing` is now **never set** and preserve is expressed for upsert
exactly as for replace, by omitting the oneof (`:484`). The flag is only legal beside a non-zero
expiry, where it would mean the same thing, so it carries no information this encoder needs.

Worth flagging for the shipping clients, since it is the same trap: gocb sends
`expiry = nil, preserveExpiry = &true` when `PreserveExpiry` is set and `Expiry == 0`
(`kvprovider_ps.go:348-359`), and couchbase-jvm-clients sends `preserve_expiry_on_existing` together
with `expiry_time = NO_EXPIRY` for the equivalent case
(`CoreProtostellarKeyValueRequests.java:339`, `:348-351`). Both are the rejected shapes measured
above. That was not verified against those clients' own test suites and may be version-dependent —
it is raised only because it is what made copying their encoding the wrong approach here.

### 3. The `GetResponse` decoder reads one arm of the `content` oneof unguarded (medium)

> Deferring the capability is reasonable; deferring it as a silent empty read is not, because the
> failure is indistinguishable from an empty document.

Confirmed and fixed fail-closed. The error is surfaced through the context, since `decode` returns a
response rather than an error code:

```cpp
if (content_is_compressed(proto)) {
  response.ctx.override_ec(errc::common::feature_not_available);
  return response;
}
```

`base_error_context::override_ec` is the existing mechanism for this in `core/impl/collection.cxx`
and `core/impl/binary_collection.cxx`.

The guard was extracted into a predicate rather than inlined, because commit 12 (`CXXCBC-894`)
reproduces the same unguarded read three more times — `get_projected`, `get_and_lock` and
`get_and_touch`, whose responses carry the same oneof. All three are now guarded by it. This is the
case the review anticipated in noting that this file establishes conventions the later converters
copy.

At commit 25 (`CXXCBC-905`) the guard is superseded for the three that gain snappy support;
`get_projected` keeps failing closed, since it never negotiates compression. The predicate is reused
by `decode_content` and by the component's projection path, replacing a duplicated `content_case()`
comparison.

### 4. Neither the absolute-expiry branch nor `expiry == 0` is tested (medium)

> A table-driven case over the branch boundaries would pin all of them, and would have caught
> finding 1

Added as suggested, over `{0, 1, 2592000, 2592001}`, applied to all three encoders that share the
helper rather than to upsert alone. Each case asserts `expiry_case() != EXPIRY_NOT_SET` first, then
the arm and its value. Three further cases were added:

- `preserve_expiry_omits_the_expiry_oneof` — the flag for upsert and replace, plus a bare
  `replace_request`, which must now send an explicit zero;
- `upsert_never_emits_a_gateway_rejected_expiry_shape` — sweeps both `preserve_expiry` values across
  the whole expiry table and asserts the encoder can never produce either of the two shapes the
  gateway rejects with `InvalidArgument`;
- `touch_always_sends_an_expiry_arm` (commit 12, where `touch_request` exists) — the arm is never
  omitted, since Touch has no preserve concept and rejects an absent oneof.

Every expectation carries the gateway line reference it comes from, against the blob hashes pinned in
`core/protostellar/kv_converter.hxx`, so a future reader can check the claim rather than trust the
comment.

Each new assertion was confirmed to **fail** against the pre-fix code before being accepted.
Measured, with the reviewed behaviour reinstated:

| reinstated defect | failing assertion |
| --- | --- |
| `expiry == 0` early return | `upsert expiry=0: the expiry oneof must be set unless preserve_expiry was requested` |
| `preserve_expiry` ignored | `upsert expresses preserve by omitting the expiry oneof` |
| unguarded compressed arm | `the compressed arm is refused rather than decoded as an empty value` |
| `<` instead of `<=` | `upsert expiry=2592000: relative arm selected` |

The `<`/`<=` case had to be isolated: with the zero-expiry defect also present the suite fails on the
first table row and never reaches the boundary.

These remain encoder-level tests. They pin the *shape* sent, which is all a unit test can do — the
*meaning* of each shape belongs to the gateway, so no encoder test could have caught finding 1 on its
own. A cluster-gated integration test asserting the observed TTL end to end is drafted as
`CXXCBC-CNG-D9` and should follow once the stack has landed and integration tests can run against a
real cluster.

### 5. The 30-day cutoff uses `<` where the convention is "greater than 30 days" (low)

> `<=` is strictly safer with no reachable downside.

Confirmed and applied — and it is not merely safer, it matches the gateway, whose own boundary is
`if expiry <= 2592000` (`helpers.go:61`). Measured, `expiry_secs = 2592000` yielded an observed
expiry of `now+2592000`, so the value is relative on the gateway side too; with `<` this converter
would have sent it as `expiry_time = 2592000`, i.e. 1970-01-31. `core/impl/expiry.cxx:76` uses `<` on
the same constant, but it encodes a duration rather than classifying an already-encoded value, and
`expiry_absolute` rejects time points below 31 days (`:100-107`), so the two cannot disagree on a
value arriving through the public API. Covered by the `2'592'000` row of the table.

### 6. `to_bytes` / `to_string` duplicate existing core utilities (low)

> Reusing the existing utilities keeps a single conversion path for value bytes.

Both helpers are removed. `core/utils/binary.hxx` provides `to_binary(std::string_view) -> binary`
and `to_string(const std::vector<std::byte>&) -> std::string`, matching the removed signatures, and
the converter now calls those. This also removes the name that collided across namespaces.

The helpers had 27 call sites in the CNG tests across five commits (08, 09, 10, 12, 25); all were
migrated in the commit that introduces them, so no commit in the stack references a name that no
longer exists. Commit 25's snappy helper (`core/protostellar/kv_converter.cxx`) was migrated too.

## Comment `5152640259`

### `replace`'s expiry oneof doesn't distinguish "preserve" from "clear."

> `encode(const operations::replace_request&)` here doesn't reproduce that three-way split.

Confirmed. The cited Java structure was read directly — `replaceRequest` wraps the expiry in
`if (!preserveExpiry)` at `CoreProtostellarKeyValueRequests.java:290-295` and sends `NO_EXPIRY` for
the "none" case, as quoted. Both listed effects are real and both are fixed; the "plain `replace()`
with no options" case is the one reachable by default, and its corrected behaviour is measured in the
table under finding 1.

One part of the analysis does not hold:

> (and leaving `insert`/`upsert` on the current shared `set_expiry()` — they have no preserve concept
> for `insert`, and `upsert` already carries its own dedicated `preserve_expiry_on_existing` bool, so
> the unset-vs-zero distinction doesn't matter for either of them.)

It does matter for `upsert`. With the shared helper unchanged, a default `upsert()` — `expiry = 0`,
`preserve_expiry = false` — leaves the oneof unset, and the gateway maps that to
`opts.PreserveExpiry = true` (`kvserver.go:484`); measured, the TTL survived the re-write. The
dedicated bool does not rescue it, because it is only set when the caller asks to preserve, so on the
default path the gateway sees neither an expiry nor a flag. `upsert` needed the same fix as
`replace`.

The dedicated bool turns out to be unusable for this purpose anyway: setting it beside an absent or
zero oneof is `InvalidArgument`, as measured under finding 2.

`insert` is as described — no preserve concept, no reachable difference.

The suggested three-way `if/else` chain was not adopted verbatim, since it is equivalent to the
shared helper once the helper takes the flag: the `expiry == 0` and `expiry <= cutoff` arms collapse
into one. Keeping one helper for all six encoders is also what corrects `touch` — which, measured,
was failing with `InvalidArgument` for a zero expiry — rather than leaving it on the old behaviour.

### Suggestion: add a test for `replace`'s expiry/preserve_expiry interaction.

Added; see finding 4. One deviation:

> e.g. asserting a bare `replace_request` (defaults) leaves the oneof unset today (documenting the
> current gap)

The gap is fixed rather than documented, so the test asserts the corrected behaviour — a bare
`replace_request` sends `kExpirySecs` with value `0`. Pinning the old behaviour would have locked in
the defect.

### Minor: `to_bytes`/`to_string` reinvent existing utilities.

Confirmed and fixed; see finding 6. The signatures match as described, so the call sites needed no
adaptation beyond the rename.

## Also closed in this pass

From the earlier automated round, one comment recorded as low confidence and never answered:

> This header uses fixed-width integer types (std::uint32_t/std::int64_t) but does not include
> <cstdint> directly, relying on transitive includes.

Correct; `<cstdint>` is now included directly. `<cstring>` was dropped in the same pass, since the
`memcpy` calls that needed it are gone.

## Two further items from the earlier automated round, also closed

Both were initially set aside as out of scope. That was wrong in each case, for a different reason.

- **`decode_content` returned an empty value on malformed snappy.** This is the same shape as
  finding 3 — an empty value with a valid CAS, indistinguishable from an empty document — narrowed to
  a gateway emitting an undecodable payload. Now fixed in the commit that introduces snappy
  (`CXXCBC-905`), which is where it belongs: `decode_content` returns `std::optional` and its three
  callers surface `errc::common::internal_server_failure`, since an undecodable frame is a
  server-side fault rather than a user transcoding error. Covered by
  `malformed_compressed_content_fails_closed`. The earlier note claimed five callers and a different
  commit's review scope; both were wrong — there are three, and that commit is part of this same
  unmerged series.
- **Fixed temp filenames in `test/cng/protostellar/credentials.cxx`** (a parallel-ctest race on
  `cng_test_cert.pem` / `cng_test_key.pem`). The test now owns a private directory named from
  `std::random_device` and removes it with `remove_all`, so concurrent runs cannot collide and a
  leftover file from a killed run cannot be picked up by the next one. The file is already on `main`
  and untouched by the rest of the series, so the fix rides in the commit whose review raised it.

## Not changed here

- **`origin::operator=(const origin&)` not copying `connection_string_`**, from the earlier automated
  round. Re-checked against `main`: the assignment operator *does* copy it
  (`connection_string_ = other.connection_string_;`), alongside `uses_protostellar_`. The finding was
  fixed by a later commit that has since merged, so it describes code that no longer exists. No
  change needed.

## Verification

- **Every commit in the series verified individually**, not just the tip: all 22 commits (08–30)
  reconfigure, build with `-DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON`, pass `ctest -L cng`, and are
  clang-format clean over that commit's own changed files. This matters because the PRs are merged
  one at a time, so "the tip is green" is not evidence about any individual PR. Two commits needed
  follow-up fixes that only a per-commit build could surface — the write encoders gain a
  `compression_settings` parameter at commit 25, which the new tests written at commit 08 had to be
  threaded through.
- **Every new assertion was proved to fail against the pre-fix code** before being accepted; the four
  reinstated-defect runs are tabulated under finding 4.
- **34/34 CNG tests pass against a live gateway** with `TEST_CONNECTION_STRING` set, so the
  cluster-gated cases ran rather than skipped.
- **Expiry semantics measured against that gateway** (`gateway-version 1.2.2-123`) with a probe
  driving raw `couchbase.kv.v1` messages: 10 shapes across Upsert, Replace, Touch and Get, all
  reported above. The probe and its recorded output are kept alongside these notes so the numbers can
  be re-derived after a gateway bump, and so shapes the SDK deliberately no longer emits can still be
  checked.

Scope of the measurement: one gateway version, single node, `_default._default`, no durability. It
establishes what each encoding *means*; it is not a substitute for the end-to-end SDK-level coverage
drafted as `CXXCBC-CNG-D9`.

core/protostellar/kv_converter.hxx
avsej: Fixed. `set_expiry` now always selects an arm unless preserve was requested, and "no expiry" is
an explicit `expiry_secs = 0`.

Confirmed against a live gateway (`1.2.2-123`) rather than inferred: over a document holding a
60s TTL, `upsert{expiry:<unset>}` left the TTL at `now+60`, while `upsert{expiry_secs:0}` cleared
it. The gateway maps a nil oneof to `opts.PreserveExpiry = true` (`kvserver.go:478-484`).

core/protostellar/kv_converter.hxx
avsej: Fixed — the flag is threaded into `set_expiry`, for `upsert` as well as `replace`.

One correction to the approach: for `upsert` the proto's `preserve_expiry_on_existing` must
**not** be set beside an omitted oneof. The gateway rejects that pair outright — measured,
`InvalidArgument` "Cannot specify preserve expiry with no expiry, leave expiry undefined to
preserve expiry." (`kvserver.go:478-482`), and again when the oneof carries a zero (`:498-501`).
Preserve is therefore expressed by omitting the oneof for both operations, and the flag is never
set.

core/protostellar/kv_converter.hxx
avsej: Fixed, fail-closed as suggested, surfaced through the context since `decode` returns a response
rather than an error code.

The guard is a shared `content_is_compressed` predicate, because the same unguarded read is
reproduced three more times by the next commit (`get_projected`, `get_and_lock`,
`get_and_touch`). In `CXXCBC-905` it is superseded by real snappy support, and an undecodable
frame now reports `internal_server_failure` instead of an empty value there too.

core/protostellar/kv_converter.hxx
avsej: Fixed — `<=`.

It is not only safer: the gateway's own boundary is `if expiry <= 2592000`
(`helpers.go:61`), and measured, `expiry_secs = 2592000` yields an observed expiry of
`now+2592000`. With `<` this converter would have sent that value as `expiry_time = 2592000`,
i.e. 1970-01-31.

test/cng/protostellar/kv_converter.cxx
avsej: Added, over `{0, 1, 2592000, 2592001}` and applied to `upsert`, `insert` and `replace`, since all
three share the helper.

Three further cases came with it: the preserve matrix, a sweep asserting the encoder can never
emit either of the two shapes the gateway rejects, and (in the commit where `touch_request`
exists) that `touch` never omits the arm. Every new assertion was confirmed to fail against the
pre-fix code before being accepted.

core/protostellar/kv_converter.hxx
avsej: Fixed — both helpers are removed and the converter calls `core::utils::to_binary` /
`core::utils::to_string`, which also drops the name that collided across namespaces. The 27 call
sites were migrated in the commits that introduce them, so no commit in the series references a
name that no longer exists.

thejcfactor (approved):

PR #1031, opened 2026-07-28
CXXCBC-891: Protostellar KV component

avsej:
Add core/protostellar/component: the couchbase2 KV data-plane. It owns the
KvService stub over a channel and executes the core get/upsert/insert/replace/
remove operations by encoding the request (kv_converter), dispatching it through
the gRPC<->asio bridge with the credentials' authorization attached as call
metadata, and delivering a decoded response whose key_value_error_context
carries the mapped gRPC status. The request message is kept alive for the
duration of the in-flight call.

Covered by a CNG-harness test that runs real KV round-trips against an
in-process gRPC server: success delivered on the io_context thread, NOT_FOUND
mapped to document_not_found, and upsert returning CAS plus a mutation token.
Wiring the component into cluster_impl::open/execute for couchbase2:// connect
is a separate follow-up.

---
Stacked PR **09/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–09; the change specific to this PR is its top commit. Depends on #1030 — merge the series bottom-up.

### DIFF
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -1,0 +1,238 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#include "core/protostellar/component.hxx"
+
+#include "core/error_context/key_value.hxx"
+#include "core/protostellar/credentials.hxx"
+#include "core/protostellar/error_utils.hxx"
+#include "core/protostellar/kv_converter.hxx"
+
+#include <couchbase/error_codes.hxx>
+
+#include <couchbase/kv/v1/kv.grpc.pb.h>
+
+#include <asio/post.hpp>
+
+#include <chrono>
+#include <memory>
+#include <string>
+#include <utility>
+
+namespace couchbase::core::protostellar
+{
+namespace v1 = ::couchbase::kv::v1;
+
+// Defined here rather than in the header so the generated gRPC types stay out of every consumer of
+// component.hxx.
+struct component::stubs {
+  std::unique_ptr<v1::KvService::Stub> kv;
+};
+
+namespace
+{
+// A request whose time budget is already spent must not be dispatched. dispatcher::unary only sets
+// a deadline when the timeout is positive, so handing it a non-positive one produces a call with no
+// deadline at all -- an RPC that runs until the channel breaks, and that cluster::close() then
+// blocks on. Turning "no time left" into "no limit" is the wrong direction for the mistake to go.
+//
+// Nothing is sent, so the operation provably did not reach the server. That makes it an
+// unambiguous timeout for every operation, mutating ones included: the ambiguity that
+// ambiguous_timeout warns about is whether the server applied the write, and here it never saw it.
+template<typename Request, typename Handler>
+auto
+fail_expired(asio::io_context& io, const Request& request, Handler& handler) -> pending_call
+{
+  auto response = request.make_response(
+    make_key_value_error_context(errc::common::unambiguous_timeout, request.id),
+    typename Request::encoded_response_type{});
+  // Delivered through the io_context, not inline on the caller's thread: "completions arrive on the
+  // SDK's execution context" is the contract every other path here honours, and a synchronous
+  // callback out of execute() would also re-enter the caller before it has its pending_call back.
+  asio::post(io, [handler = std::move(handler), response = std::move(response)]() mutable {
+    handler(std::move(response));
+  });
+  return {};
+}
+} // namespace
+
+component::component(asio::io_context& io, component_config config)
+  : io_{ io }
+  , stubs_{ std::make_unique<stubs>(stubs{ v1::KvService::NewStub(config.channel) }) }
+  , authorization_{ authorization_header(config.credentials) }
+  , default_kv_timeout_{ config.default_kv_timeout }
+  // Initialised last (see the declaration order in the header), so the channel can be moved in.
+  , dispatcher_{ io, std::move(config.channel) }
+{
+}
+
+component::~component() = default;
+
+auto
+component::execute(operations::get_request request,
+                   utils::movable_function<void(operations::get_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  auto proto = std::make_shared<v1::GetRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::GetResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::GetResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Get(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    // `proto` is captured here too so the request outlives the in-flight call.
+    [handler = std::move(handler), id, proto](grpc::Status status, v1::GetResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      // A get reads and nothing more, so a deadline here is unambiguous: the document cannot have
+      // changed. kv::decode owns the compressed-content rule (it reports feature_not_available
+      // rather than handing back an empty value with a success status).
+      auto ctx = make_error_context(status, id, operation_kind::read_only);
+      handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+auto
+component::execute(operations::upsert_request request,
+                   utils::movable_function<void(operations::upsert_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  auto proto = std::make_shared<v1::UpsertRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::UpsertResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::UpsertResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Upsert(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    [handler = std::move(handler), id, proto](grpc::Status status,
+                                              v1::UpsertResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id, operation_kind::mutating);
+      handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+auto
+component::execute(operations::insert_request request,
+                   utils::movable_function<void(operations::insert_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  auto proto = std::make_shared<v1::InsertRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::InsertResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::InsertResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Insert(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    [handler = std::move(handler), id, proto](grpc::Status status,
+                                              v1::InsertResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id, operation_kind::mutating);
+      handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+auto
+component::execute(operations::replace_request request,
+                   utils::movable_function<void(operations::replace_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  auto proto = std::make_shared<v1::ReplaceRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::ReplaceResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::ReplaceResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Replace(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    [handler = std::move(handler), id, proto](grpc::Status status,
+                                              v1::ReplaceResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id, operation_kind::mutating);
+      handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+auto
+component::execute(operations::remove_request request,
+                   utils::movable_function<void(operations::remove_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  auto proto = std::make_shared<v1::RemoveRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::RemoveResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::RemoveResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Remove(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    [handler = std::move(handler), id, proto](grpc::Status status,
+                                              v1::RemoveResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id, operation_kind::mutating);
+      handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+} // namespace couchbase::core::protostellar
--- a/core/protostellar/component.hxx
+++ b/core/protostellar/component.hxx
@@ -1,0 +1,96 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// The Protostellar KV component: the couchbase2 counterpart to the MCBP data-plane. It owns the
+// gRPC stubs over a channel and executes core KV operations by encoding the request (kv_converter),
+// dispatching it through the gRPC<->asio bridge, and decoding the reply plus a mapped error
+// context. cluster_impl routes KV ops here when the cluster was opened with couchbase2://.
+
+#include "core/cluster_credentials.hxx"
+#include "core/operations/document_get.hxx"
+#include "core/operations/document_insert.hxx"
+#include "core/operations/document_remove.hxx"
+#include "core/operations/document_replace.hxx"
+#include "core/operations/document_upsert.hxx"
+#include "core/protostellar/dispatcher.hxx"
+#include "core/timeout_defaults.hxx"
+#include "core/utils/movable_function.hxx"
+
+#include <asio/io_context.hpp>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+namespace couchbase::core::protostellar
+{
+
+// Construction parameters for `component`, grouped so that adding one is a source-compatible
+// change. Every field a caller can reasonably omit carries a default, so a new trailing field
+// leaves existing initialisers compiling instead of silently shifting a positional argument onto
+// the wrong parameter -- which is the failure this shape exists to prevent, the constructor having
+// gained and lost trailing parameters repeatedly as services were added.
+struct component_config {
+  std::shared_ptr<grpc::Channel> channel{};
+  cluster_credentials credentials{};
+  std::chrono::milliseconds default_kv_timeout{ timeout_defaults::key_value_timeout };
+};
+
+class component
+{
+public:
+  component(asio::io_context& io, component_config config);
+  component(const component&) = delete;
+  component(component&&) = delete;
+  auto operator=(const component&) -> component& = delete;
+  auto operator=(component&&) -> component& = delete;
+  // Out of line: `stubs` is incomplete here, so its deleter cannot be instantiated in this header.
+  ~component();
+
+  auto execute(operations::get_request request,
+               utils::movable_function<void(operations::get_response)>&& handler) -> pending_call;
+  auto execute(operations::upsert_request request,
+               utils::movable_function<void(operations::upsert_response)>&& handler)
+    -> pending_call;
+  auto execute(operations::insert_request request,
+               utils::movable_function<void(operations::insert_response)>&& handler)
+    -> pending_call;
+  auto execute(operations::replace_request request,
+               utils::movable_function<void(operations::replace_response)>&& handler)
+    -> pending_call;
+  auto execute(operations::remove_request request,
+               utils::movable_function<void(operations::remove_response)>&& handler)
+    -> pending_call;
+
+private:
+  // The generated gRPC stubs are held behind an opaque pointer so the generated protobuf and gRPC
+  // surface stays out of every translation unit that includes this header -- none of it appears in
+  // the interface, and the stub count grows with every service the transport learns to speak.
+  struct stubs;
+
+  asio::io_context& io_;
+  std::unique_ptr<stubs> stubs_;
+  std::string authorization_;
+  std::chrono::milliseconds default_kv_timeout_;
+  // Declared last, so it is destroyed first: ~dispatcher() cancels and drains the calls issued
+  // through the stubs above, which must therefore still be alive while it runs.
+  dispatcher dispatcher_;
+};
+
+} // namespace couchbase::core::protostellar
--- a/test/cng/protostellar/component.cxx
+++ b/test/cng/protostellar/component.cxx
@@ -1,0 +1,693 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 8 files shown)


core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,207 @@
...
+      pending.assign(outstanding_.begin(), outstanding_.end());
+    }
+    for (auto* context : pending) {
+      context->TryCancel();
+    }
+    std::unique_lock lock{ mutex_ };
+    idle_.wait(lock, [this] {
+      return outstanding_.empty();
+    });
+  }
```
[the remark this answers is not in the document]
avsej: Addressed: call_tracker now stores shared_ptr<grpc::ClientContext> (keyed by the raw pointer) and cancel_and_drain() snapshots the owning shared_ptrs under the lock before calling TryCancel() with the lock released. The snapshot keeps every context alive across the unlocked cancel, so a concurrent completion that removes and releases its own reference cannot leave a snapshotted pointer dangling.

core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,212 @@
...
+      std::forward<Invoker>(invoker)(
+        *state->context,
+        state->response,
+        [state, handler_holder, io, tracker, context](grpc::Status status) {
+          // Deregister on the gRPC thread first: once removed, a concurrent cancel_and_drain() may
+          // return. The post that follows only touches shared state (io outlives the call).
+          tracker->remove(context);
+          asio::post(*io, [state, handler_holder, status = std::move(status)]() mutable {
+            (*handler_holder)(std::move(status), std::move(state->response));
+          });
```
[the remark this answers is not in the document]
avsej: Deliberate and required as written. The completion posts the handler and, in the streaming reactor, the posted task self-deletes the reactor -- which owns context_. Removing from the tracker *after* posting would let the io thread run that self-delete and free context_ before the gRPC thread reads context_.get() in remove(), reintroducing a use-after-free. remove()-before-post() is kept consistent across unary() and the reactor; safety rests on the documented SDK invariant that the io_context outlives the dispatcher, so a post issued just before remove() can never target a destroyed io_context.

core/protostellar/component.cxx
```diff
@@ -0,0 +1,185 @@
...
+      grpc::ClientContext& ctx, v1::GetResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Get(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    // `proto` is captured here too so the request outlives the in-flight call.
+    [handler = std::move(handler), id, proto](grpc::Status status, v1::GetResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id);
```
avsej: This factory has no parameter for read-only-ness, so it resolves `DEADLINE_EXCEEDED` to `errc::common::ambiguous_timeout` for every operation — including `get`. RFC 77's error table keys the distinction on read-only status specifically, and notes it must not be based on idempotency. couchbase-jvm-clients uses `request.readonly()` (`CoreProtostellarErrorHandlingUtil.java:285`), gocb a `readOnly` flag (`error_ps.go:141`); .NET keys it on idempotency and the implementation audit records that as a High-priority bug (CNG-2). The classic MCBP path here already reports `unambiguous_timeout` for reads.

The overload knows it is handling a `get_request`, so the information is available at this call site. Adding the bit to the factory now means the remaining ~43 overloads in the series inherit the correct behaviour:

```cpp
auto ctx = make_error_context(status, id, /* readonly */ true);
```

with `map_status_code` returning `readonly ? errc::common::unambiguous_timeout : errc::common::ambiguous_timeout`.

core/protostellar/component.cxx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+auto
+component::execute(operations::get_request request,
+                   utils::movable_function<void(operations::get_response)>&& handler)
+  -> pending_call
+{
+  auto proto = std::make_shared<v1::GetRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
```
avsej: A non-positive resolved timeout becomes a call with **no deadline at all**: `dispatcher::unary` guards with `if (timeout.count() > 0) { state->context->set_deadline(...); }`, so `request.timeout = 0ms` yields an unbounded RPC that runs until the channel breaks, and `cluster::close()` then waits on it. The classic path treats an exhausted budget as an immediate timeout.

Rejecting it here, where the value is resolved, keeps the check in one place rather than in the dispatcher:

```cpp
if (timeout <= std::chrono::milliseconds::zero()) {
  return handler(request.make_response(
    make_error_context(errc::common::unambiguous_timeout, id), {}));
}
```

Separately, this is a per-attempt duration rather than an absolute deadline, which is why the retry loop added in #1043 gives each re-issue a fresh full-length budget.

core/protostellar/component.hxx
```diff
@@ -0,0 +1,75 @@
...
+#include <memory>
+#include <string>
+
+namespace couchbase::core::protostellar
+{
+
+class component
+{
+public:
+  component(asio::io_context& io,
```
avsej: Four positional parameters with no defaults, the last a bare duration — and one more arrives with every service.

This has already cost the series: `test/cng/protostellar/live_kv.cxx:100` and `:183` were written against this four-argument form and have failed to compile since #1036 added a fifth parameter, through #1038 replacing the timeouts with an aggregate, to #1047 adding a fifth parameter again.

A config struct makes each addition source-compatible:

```cpp
struct component_config {
  std::shared_ptr<grpc::Channel> channel;
  cluster_credentials credentials;
  component_timeouts timeouts{};
};

component(asio::io_context& io, component_config config);
```

core/protostellar/component.hxx
```diff
@@ -0,0 +1,75 @@
...
+#include "core/cluster_credentials.hxx"
+#include "core/operations/document_get.hxx"
+#include "core/operations/document_insert.hxx"
+#include "core/operations/document_remove.hxx"
+#include "core/operations/document_replace.hxx"
+#include "core/operations/document_upsert.hxx"
+#include "core/protostellar/dispatcher.hxx"
+#include "core/utils/movable_function.hxx"
+
+#include <couchbase/kv/v1/kv.grpc.pb.h>
```
avsej: This is included so the class can declare a `std::unique_ptr<KvService::Stub>` member, which means every consumer of `component.hxx` — `core/cluster.cxx` included — preprocesses the whole generated gRPC and protobuf surface for a type that never appears in the interface.

Forward-declaring the stub types and moving the `unique_ptr` destructors into the `.cxx`, or holding the stubs behind an opaque implementation pointer, keeps the generated headers out of consumers. The cost grows from one stub here to nine by the end of the series.

core/protostellar/component.hxx
```diff
@@ -0,0 +1,75 @@
...
+    -> pending_call;
+  auto execute(operations::replace_request request,
+               utils::movable_function<void(operations::replace_response)>&& handler)
+    -> pending_call;
+  auto execute(operations::remove_request request,
+               utils::movable_function<void(operations::remove_response)>&& handler)
+    -> pending_call;
+
+private:
+  dispatcher dispatcher_;
```
avsej: Members are destroyed in reverse declaration order, so `kv_stub_` is destroyed first and `~dispatcher()` — whose `cancel_and_drain()` covers the calls made through that stub — runs last. A gRPC call holds its own channel reference, so this is unlikely to fault today, but the ordering states the opposite of the intended invariant.

Declaring `dispatcher_` last makes teardown order match the dependency, and is cheaper to do now than when there are nine stubs.

test/cng/protostellar/component.cxx
```diff
@@ -0,0 +1,505 @@
...
+  return document_id{ "b", "s", "c", std::move(key) };
+}
+
+void
+get_round_trips_on_the_io_thread()
+{
+  in_process_server server;
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  component comp{ io, server.channel(), cluster_credentials{}, 5000ms };
```
avsej: Every case constructs the component with `5000ms`, none sets `request.timeout`, and the in-process server answers immediately. Rewriting `component::execute` to ignore `request.timeout` entirely would leave all eleven cases passing — so this commit's timeout plumbing, cancellation, and the exactly-once completion guarantee all have no coverage.

The most valuable missing case is a deadline that actually fires against a delaying server, asserting both that the handler runs exactly once and that a timed-out `get` reports `unambiguous_timeout`. That single test covers the timeout resolution, the non-positive-timeout hole, and the read-only timeout classification.

avsej (commented):
## Review — PR #1031 (`CXXCBC-891`)

This commit introduces `protostellar::component`, the object that owns the gRPC stubs and turns a core KV request into a dispatched call. The lifetime discipline is correct — the request proto is kept alive by a `shared_ptr` captured into both the invoker and the completion, no lambda captures `this` or a reference to a local, and the completion is delivered through the dispatcher's post-then-deregister path rather than inline on a gRPC thread.

The substantive point about this commit is that it is a template. The five `execute` overloads here become roughly 43 by the tip of the series, so anything the shape cannot express has to be worked around 43 times rather than once. Two such limitations are worth fixing before the pattern is replicated.

### Strengths

- The request proto is held in a `shared_ptr` captured by both the invoker lambda and the completion lambda, with an explicit `(void)proto;` documenting why. This is the correct way to keep a protobuf request alive across an async gRPC call, and it is applied consistently.
- Completions are routed through `dispatcher::unary`, which posts to the `io_context` before deregistering from the call tracker. Keeping user callbacks off gRPC's threadpool is the right contract and it is honoured here.
- `component` holds no mutable state: the authorization header and the default timeout are both immutable after construction, which removes a whole class of races from a type that is shared across every operation.
- The in-process gRPC server used by `test/cng/protostellar/component.cxx` exercises the real dispatch path rather than mocking it, which is a materially better test design than asserting on the converter alone.

### Findings

|---|---|---|---|---|
| 1 | high | spec-compliance | `make_error_context(status, id)` has no seam for read-only-ness, so every timed-out read reports `ambiguous_timeout` | `core/protostellar/component.cxx:67` |
| 2 | high | correctness | A resolved timeout of zero or less produces a call with no deadline at all | `core/protostellar/component.cxx:52` |
| 3 | medium | api-design | `execute` takes a per-attempt duration rather than an absolute deadline, which the later retry loop cannot bound correctly | `core/protostellar/component.cxx:52` |
| 4 | medium | api-design | Four positional constructor parameters with no defaults | `core/protostellar/component.hxx:48` |
| 5 | medium | test-coverage | Nothing tests the timeout parameter, cancellation, or exactly-once completion | `test/cng/protostellar/component.cxx:202` |
| 6 | medium | efficiency | `component.hxx` exposes the full generated gRPC surface to every including translation unit | `core/protostellar/component.hxx:34` |
| 7 | low | correctness | Member order destroys the stub before the dispatcher drains calls made through it | `core/protostellar/component.hxx:69` |

#### 1. `make_error_context(status, id)` cannot express read-only-ness (high)

```cpp
auto ctx = make_error_context(status, id);
```

This idiom is introduced here and repeated identically for `get`, `upsert`, `insert`, `replace` and `remove`. It resolves through `map_status` to `map_status_code`, which returns `errc::common::ambiguous_timeout` for `DEADLINE_EXCEEDED` unconditionally. `error_utils.cxx` acknowledges the gap in a comment, noting that distinguishing the read-only case would need the request's readonly flag, "which is not available at this layer".

It is available at *this* layer, though: the overload knows it is handling a `get_request`. RFC 77's error table keys the distinction on read-only status specifically, and calls out that the choice must not be made on idempotency — an upsert is idempotent but mutates. couchbase-jvm-clients uses `request.readonly()` (`CoreProtostellarErrorHandlingUtil.java:285`) and gocb a `readOnly` flag (`error_ps.go:141`); .NET keys it on idempotency and the implementation audit records that as a High-priority bug (CNG-2). The classic MCBP path in this repository already reports `unambiguous_timeout` for reads.

Failure scenario: a `get` times out against a slow gateway. The user receives `ambiguous_timeout`, whose documented meaning is that the operation may have mutated state. A caller that branches on ambiguity — for example one that will not retry an ambiguous mutation — takes the wrong branch for an operation that provably changed nothing.

The cheapest fix that scales to 43 call sites is to give the factory the bit:

```cpp
// error_utils.hxx
[[nodiscard]] auto
make_error_context(const grpc::Status& status, const document_id& id, bool readonly)
  -> key_value_error_context;
```

with `map_status_code` returning `readonly ? errc::common::unambiguous_timeout : errc::common::ambiguous_timeout` for `DEADLINE_EXCEEDED`. Fixing it here means the remaining overloads inherit it; fixing it after the series means touching every one.

(inline comment on `core/protostellar/component.cxx:67`)

#### 2. A resolved timeout of zero or less produces a call with no deadline (high)

```cpp
const auto timeout = request.timeout.value_or(default_kv_timeout_);
```

The value is handed straight to `dispatcher::unary`, which sets the deadline only when it is positive:

```cpp
if (timeout.count() > 0) {
  state->context->set_deadline(std::chrono::system_clock::now() + timeout);
}
```

So a request carrying `timeout = 0ms` — or any non-positive value — becomes a gRPC call with no deadline whatsoever, which runs until the channel breaks. The classic path treats an already-expired budget as an immediate timeout. Turning "no time left" into "no limit" is the wrong direction for a mistake to go, and it also means `cluster::close()` waits on a call that has no reason to finish.

The component is the right place to reject it, since it is the layer that resolves the value:

```cpp
const auto timeout = request.timeout.value_or(default_kv_timeout_);
if (timeout <= std::chrono::milliseconds::zero()) {
  return handler(request.make_response(make_error_context(errc::common::unambiguous_timeout, id), {}));
}
```

(inline comment on `core/protostellar/component.cxx:52`)

#### 3. `execute` takes a duration, not an absolute deadline (medium)

`dispatcher::unary` converts the duration to `now + timeout` at the moment of dispatch, and the signature offers no way to say "this operation's budget ends at absolute time T". That is fine for a single attempt, but PR #1043 builds a retry loop on top of this interface, and each re-issue therefore receives a fresh full-length deadline rather than the remaining budget — so a retried operation can run for roughly twice its nominal timeout.

Accepting an optional absolute deadline alongside the duration, and preferring it when present, removes the problem at the source rather than requiring the retry layer to compensate:

```cpp
auto execute(operations::get_request request,
             std::optional<std::chrono::steady_clock::time_point> deadline,
             utils::movable_function<void(operations::get_response)>&& handler) -> pending_call;
```

#### 4. Four positional constructor parameters with no defaults (medium)

```cpp
component(asio::io_context& io,
          std::shared_ptr<grpc::Channel> channel,
          cluster_credentials credentials,
          std::chrono::milliseconds default_kv_timeout);
```

The trailing parameter is a bare duration, and every added service brings another. This is not hypothetical: by #1036 the list has grown to five parameters, by #1038 the timeouts have been consolidated into a `component_timeouts` aggregate, and by #1047 there is a fifth parameter again — and a call site in `test/cng/protostellar/live_kv.cxx` that was written against the four-parameter form has been failing to compile since #1036 as a result.

Grouping the configuration into one struct with named fields, or defaulting the trailing parameters, makes each addition source-compatible:

```cpp
struct component_config {
  std::shared_ptr<grpc::Channel> channel;
  cluster_credentials credentials;
  component_timeouts timeouts{};
  kv::compression_settings compression{};
};

component(asio::io_context& io, component_config config);
```

(inline comment on `core/protostellar/component.hxx:48`)

#### 5. Nothing tests the timeout parameter, cancellation, or exactly-once completion (medium)

Every case in `test/cng/protostellar/component.cxx` constructs the component with `5000ms` and never sets `request.timeout`, and the in-process server answers immediately. If `component::execute` were rewritten to ignore `request.timeout` entirely, all eleven cases would still pass — so the timeout plumbing this commit adds has no coverage, and neither does cancellation or the guarantee that the handler runs exactly once.

The highest-value missing case is a deadline that actually fires, since it covers findings 1, 2 and 3 at once:

```cpp
void
get_honours_the_request_timeout()
{
  in_process_server server{ /* delay every reply by 500ms */ };
  asio::io_context io;
  auto work = asio::make_work_guard(io);
  component comp{ io, server.channel(), cluster_credentials{}, 5000ms };

  std::optional<operations::get_response> result;
  operations::get_request request{ document_id{ "default", "_default", "_default", "key" } };
  request.timeout = 50ms;                       // must win over the 5000ms default
  comp.execute(std::move(request), [&result](auto response) {
    assert_false(result.has_value(), "handler must run exactly once");
    result = std::move(response);
  });
  io.run();

  assert_true(result.has_value(), "handler must run");
  assert_true(result->ctx.ec() == errc::common::unambiguous_timeout,
              "a timed-out read is unambiguous");
}
```

(inline comment on `test/cng/protostellar/component.cxx:202`)

#### 6. `component.hxx` exposes the full generated gRPC surface (medium)

The header includes the generated `*.grpc.pb.h` so the class can declare `std::unique_ptr<...::Stub>` members. Every translation unit that includes `component.hxx` — `core/cluster.cxx` among them — therefore preprocesses the entire generated gRPC and protobuf surface, none of which appears in the interface.

Holding the stubs behind an opaque implementation pointer, or forward-declaring the stub types and moving the `unique_ptr` destructors into the `.cxx`, keeps the generated headers out of consumers. This matters more as the stub count grows from one here to nine by the tip.

(inline comment on `core/protostellar/component.hxx:34`)

#### 7. Member order destroys the stub before the dispatcher drains (low)

```cpp
private:
  dispatcher dispatcher_;
  std::unique_ptr<::couchbase::kv::v1::KvService::Stub> kv_stub_;
```

Members are destroyed in reverse declaration order, so `kv_stub_` is destroyed first and `~dispatcher()` — which calls `cancel_and_drain()` on the calls that were made through that stub — runs last. In practice a gRPC call holds its own reference to the channel, so this is unlikely to fault today, but the ordering states the opposite of the intended invariant. Declaring `dispatcher_` last makes the teardown order match the dependency, and is worth doing before the stub count reaches nine.

(inline comment on `core/protostellar/component.hxx:69`)

### Note

`cmake/Protostellar.cmake` carries a note stating that nothing links the stubs library into `couchbase_cxx_client` yet, so the generated protobuf and gRPC symbols do not reach the shared library's export set. That stops being true once a later commit in this series embeds the transport in the client. The note should be revisited at that point rather than left describing a pre-series state.

### Verdict

Needs changes. Findings 1 and 2 are small and local now and expensive later, because this commit's shape is copied into roughly 43 call sites before the series ends.

avsej:
All seven findings are addressed in this commit, along with the note on `cmake/Protostellar.cmake`.
Two are fixed differently from the shape suggested, and one suggested classification turned out to
need a further distinction; both are set out below rather than silently varied.

Every new assertion was checked against the pre-fix code first and observed to fail, so none of them
passes vacuously. Each commit of the series was rebuilt and re-run individually, since they are
merged one at a time.

## 1. `make_error_context(status, id)` cannot express read-only-ness

> This factory has no parameter for read-only-ness, so it resolves `DEADLINE_EXCEEDED` to
> `errc::common::ambiguous_timeout` for every operation — including `get`. RFC 77's error table keys
> the distinction on read-only status specifically, and notes it must not be based on idempotency.

Fixed. The bit is threaded through `map_status_code`, `map_status` and `make_error_context` as a
scoped `operation_kind` rather than a bare `bool`, so the ~43 call sites that inherit it read
`operation_kind::read_only` instead of an unlabelled `true`.

The premise was checked against RFC 77 rather than taken on trust, and it holds verbatim — from the
"Other errors" table:

> Readonly ops: UnambiguousTimeoutException Otherwise: AmbiguousTimeoutException [...] Remember the
> "Ambiguous" refers to whether the operation may have mutated state. This needs to be based on a
> readonly status, not an idempotent status. An upsert op is idempotent but mutates state.

One refinement to the classification, which matters because it is not what "returns a value to the
caller" would suggest: **`get_and_lock` and `get_and_touch` are `mutating`**. Both read a document,
but one takes a server-side lock and the other rewrites the expiry, so a timeout on either genuinely
may have changed state. Of the KV surface only `get`, `get_projected` and `exists` are read-only.
The operation table is enforced rather than defaulted — an unclassified operation is a hard error, so
a future overload cannot silently inherit the wrong answer.

`query` and `analytics` are keyed on the `readonly` flag their requests already carry, which is the
same flag RFC 77 means; `search` and views cannot mutate and are fixed read-only. Management
operations are classified individually — the eight `*_get*` / `analyze_document` reads are read-only,
the seventeen create/drop/update/flush/control operations are not.

One correction to the supporting evidence, since it is a factual claim about this repository:

> The classic MCBP path here already reports `unambiguous_timeout` for reads.

That does not hold. The classic path keys on **idempotency**, not read-only-ness
(`core/io/mcbp_command.hxx`, `cancel()`: `request.retries.idempotent() || !opaque_.has_value()`),
via a compile-time `retry_context<bool>` baked into each request type
(`core/io/retry_context.hxx`). There is no read-only flag for KV anywhere in the classic path. So it
is not a precedent for the classification RFC 77 asks for — and RFC 77 explicitly rejects the
predicate it does use.

The couchbase2 path therefore deliberately does not mirror it, and diverges in one visible place:
`exists` is `retry_context<false>` on classic (reported ambiguous) but read-only here, because it
changes nothing. The classic marking is a retry-safety judgement being reused for classification;
here the two are separate concerns — `retry_reason_for` decides retryability, `operation_kind`
decides how a timeout is reported.

The classic path does corroborate finding 2: its `!opaque_.has_value()` disjunct yields
`unambiguous_timeout` regardless of idempotency when the command was never sent, which is exactly the
rule applied below.

Verified: reverting the refinement leaves the mutating assertions passing and fails exactly the two
read-only ones (`deadline_exceeded_depends_on_the_operation_kind`, and end to end in
`get_honours_the_request_timeout`). The full mapping was then dumped and read back operation by
operation rather than assumed — which is how a real defect in the first attempt was caught: the
management overloads wrap their signature onto a second line, the classifier only matched single-line
signatures, and all 28 of them silently inherited the previous overload's kind and came out
read-only. `bucket_create` and `bucket_drop` among them. That is the same defect this finding is
about, so the classifier now refuses an operation it does not recognise instead of defaulting.

## 2. A resolved timeout of zero or less produces a call with no deadline

> So a request carrying `timeout = 0ms` — or any non-positive value — becomes a gRPC call with no
> deadline whatsoever, which runs until the channel breaks. [...] Turning "no time left" into "no
> limit" is the wrong direction for a mistake to go

Fixed at the component, as suggested. Two deliberate differences from the sketch:

The guard is hoisted **above** the encode rather than placed after it, so a rejected upsert does not
first serialise its value into a protobuf that is immediately discarded.

The response is delivered through the `io_context` rather than returned inline. Completions arriving
on the SDK's execution context is the contract every other path here honours, and a synchronous
callback out of `execute()` would also re-enter the caller before it has its `pending_call` back.

On the error code: `unambiguous_timeout` is correct here for **every** operation, mutations included,
not only for reads. The ambiguity `ambiguous_timeout` warns about is whether the server applied the
write; on this path nothing was sent, so it provably did not.

Verified: removing the guard fails `an_exhausted_budget_is_rejected_before_dispatch` and
`an_exhausted_default_timeout_is_also_rejected`. Both assert `calls_received == 0` as well as the
error code, so they distinguish "rejected before dispatch" from "dispatched and then failed" — which
the error code alone cannot.

## 3. `execute` takes a duration, not an absolute deadline

> That is fine for a single attempt, but PR #1043 builds a retry loop on top of this interface, and
> each re-issue therefore receives a fresh full-length deadline rather than the remaining budget — so
> a retried operation can run for roughly twice its nominal timeout.

The defect is confirmed, and it is subtler than the retry loop's own deadline suggests: that loop
does stamp an overall deadline, but the `when < deadline` check gates only whether another attempt
may **start**. An attempt begun just inside the deadline is still handed the whole timeout again.

The fix is placed in the retry layer rather than in this signature. Before a re-issue,
`request.timeout` is narrowed to what remains of the operation's budget, so the duration the
component already accepts now means "until the operation's deadline". This bounds the operation
exactly as an absolute deadline would, without adding a parameter to ~43 overloads, and it composes
with finding 2: when the remainder is non-positive the component rejects the request outright rather
than dispatching an unbounded call.

The suggested signature change was not adopted because nothing other than the retry loop needs an
absolute deadline, and the retry loop is the layer that already owns one. If a second consumer
appears, the parameter can be added then with the same effect.

Covered by a new case that drives a real `core::cluster` against a localhost `KvService`, since the
budget is a property of routing, backoff, re-dispatch and the per-attempt deadline interacting and is
not observable from any one of them. With a 400ms budget and 150ms backoff against a server that
stops answering, the operation ends at ~400ms; before the fix it ended at ~700ms. Verified by
reverting the narrowing and observing the case fail.

## 4. Four positional constructor parameters with no defaults

> A config struct makes each addition source-compatible

Adopted. `component_config` carries the channel, credentials and timeouts, each field defaulted so a
new trailing field leaves existing initialisers compiling.

The sketch's nested shape is what the series converges on: from the commit that consolidates the
per-service timeouts, `component_config` holds a `component_timeouts` aggregate, and later a
`kv::compression_settings`.

One consequence worth flagging, because it is created by finding 2 rather than by this one:
`component_timeouts` previously value-initialised its fields to zero. Now that a zero timeout means
"budget already spent" rather than "no limit", a default-constructed aggregate would have rejected
every request that relies on a default. Each field therefore now defaults from `timeout_defaults`.

The two `live_kv.cxx` call sites that had been failing to compile are fixed as part of this.

## 5. Nothing tests the timeout parameter, cancellation, or exactly-once completion

> If `component::execute` were rewritten to ignore `request.timeout` entirely, all eleven cases would
> still pass

Correct, and no longer so. Added:

- `get_honours_the_request_timeout` — 50ms request timeout against a 5000ms component default and a
  server that delays 400ms, asserting the handler runs exactly once and a timed-out read reports
  `unambiguous_timeout`. This is the case suggested; it covers findings 1, 2 and 3 at once.
- `mutation_timeout_is_reported_as_ambiguous` — the other arm, kept separate so a single shared
  classification cannot satisfy the read-only assertion by accident.
- `an_exhausted_budget_is_rejected_before_dispatch` and
  `an_exhausted_default_timeout_is_also_rejected` — both arms of finding 2, asserting the server saw
  nothing.
- `cancellation_completes_the_handler_once` — a cancelled call still completes, exactly once, as
  `request_canceled`. This one passes against the pre-fix code as well; it is new coverage of
  existing behaviour rather than a regression test, and is listed as such.
- `deadline_exceeded_depends_on_the_operation_kind` in the error-utils suite, asserting both arms
  through `map_status_code`, `map_status` and `make_error_context` so no layer can quietly drop the
  bit.

The in-process server gained two knobs to make this possible: a reply delay, and a count of requests
that actually reached it.

## 6. `component.hxx` exposes the full generated gRPC surface

> Holding the stubs behind an opaque implementation pointer, or forward-declaring the stub types and
> moving the `unique_ptr` destructors into the `.cxx`, keeps the generated headers out of consumers.

Fixed with the opaque pointer. Forward declaration is not available here — `KvService::Stub` is a
nested class, which cannot be declared without its enclosing type — so the stubs live in a
`struct component::stubs` defined in the `.cxx`, with the destructor out of line.

`component.hxx` now includes no generated header at all. By the tip of the series that is nine
services kept out of every consumer, `core/cluster.cxx` included; `grpc::Channel`, the only gRPC name
left in the interface, arrives through `dispatcher.hxx`.

## 7. Member order destroys the stub before the dispatcher drains

> Declaring `dispatcher_` last makes the teardown order match the dependency

Done. `dispatcher_` is declared last, so it is destroyed first and `cancel_and_drain()` runs while
the stubs its in-flight calls were issued through are still alive. Later commits in the series add
members in front of it rather than after it, so the invariant is maintained as the list grows.

## Note on `cmake/Protostellar.cmake`

> `cmake/Protostellar.cmake` carries a note stating that nothing links the stubs library into
> `couchbase_cxx_client` yet [...] That stops being true once a later commit in this series embeds the
> transport in the client.

The note is already inaccurate, and not only prospectively: `couchbase_cxx_client` links
`couchbase_cxx_protostellar` today, and the transport sources are appended to the client's source
list — both guarded by `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2`, and both already present in `main`
rather than arriving later in this series.

The note has been corrected to say what actually holds: in a couchbase2-enabled build the generated
symbols already reach the client's exported ABI, and what keeps that off released artefacts is only
that the option defaults to OFF. The ABI remedy itself is unchanged and still deferred to the
ship-enable ticket.

## The same hole on the non-KV paths

Not raised in the review, but the query, analytics, search, view and management overloads resolve a
timeout exactly as the KV ones do and had the same hole. It is closed here rather than deferred:
`fail_expired_ctx` does for the services that carry their own error context what `fail_expired` does
for the key-value one. All 44 overloads across the series now reject an exhausted budget instead of
issuing a call with no deadline.

Covered where the scaffolding for it first exists — the in-process query server, in
`query_streaming.cxx` — asserting `unambiguous_timeout`, exactly one completion, delivery on the io
thread, and that the server received nothing. Verified by removing the guards and observing the case
fail.

core/protostellar/component.cxx
avsej: Fixed. The bit is threaded through `map_status_code`, `map_status` and `make_error_context` as a
scoped `operation_kind` rather than a bare `bool`, so the call sites that inherit it read
`operation_kind::read_only` rather than an unlabelled `true`.

Two corrections came out of checking this rather than assuming it. `get_and_lock` and
`get_and_touch` are **mutating** — both read a document, but one takes a lock and the other
rewrites the expiry — so of the KV surface only `get`, `get_projected` and `exists` are
read-only. And the classic MCBP path is not a precedent here: it keys on idempotency
(`mcbp_command.hxx` `cancel()`, via `retry_context<bool>`), which is the predicate RFC 77
explicitly rejects. There is no read-only flag for KV on that path at all.

core/protostellar/component.cxx
avsej: Fixed at the component, as suggested, with two differences: the guard sits **above** the encode
so a rejected upsert does not first serialise its value into a discarded protobuf, and the
response is posted to the `io_context` rather than delivered inline.

On the code: `unambiguous_timeout` is right for **every** operation here, mutations included —
nothing was sent, so the ambiguity `ambiguous_timeout` warns about cannot arise. The classic path
agrees, via the `!opaque_.has_value()` disjunct in `cancel()`.

The separate point about the per-attempt duration is real, and subtler than it looks: the retry
loop does stamp an overall deadline, but `when < deadline` gates only whether another attempt may
**start**. Fixed by narrowing `request.timeout` to the remaining budget before a re-issue, which
bounds the operation exactly as an absolute deadline would without adding a parameter to ~43
overloads.

core/protostellar/component.hxx
avsej: Adopted. `component_config` carries the channel, credentials and timeouts, every field defaulted
so a new trailing field leaves existing initialisers compiling. The nested shape suggested is
what the series converges on — from the commit that consolidates them, the config holds a
`component_timeouts` aggregate, and later a `kv::compression_settings`.

`component_timeouts` also had to gain real defaults rather than value-initialising to zero: now
that a zero timeout means "budget already spent", a default-constructed aggregate would have
rejected every request relying on a default.

The two `live_kv.cxx` call sites that had been failing to compile are fixed with it.

core/protostellar/component.hxx
avsej: Fixed with the opaque pointer. Forward declaration is not available here — `KvService::Stub` is a
nested class, which cannot be declared without its enclosing type — so the stubs live in a
`struct component::stubs` defined in the `.cxx`, with the destructor out of line.

`component.hxx` now includes no generated header at all; by the tip of the series that is nine
services kept out of every consumer, `core/cluster.cxx` included. `grpc::Channel`, the only gRPC
name left in the interface, arrives through `dispatcher.hxx`.

core/protostellar/component.hxx
avsej: Done. `dispatcher_` is declared last, so it is destroyed first and `cancel_and_drain()` runs
while the stubs its in-flight calls were issued through are still alive. Later commits add
members in front of it rather than after it, so the ordering holds as the list grows.

test/cng/protostellar/component.cxx
avsej: Correct, and no longer so. Added the deadline case suggested (50ms request timeout against a
5000ms default and a delaying server, asserting exactly one completion and `unambiguous_timeout`),
its mutating counterpart, both arms of the non-positive-timeout guard, and a cancellation case.
The in-process server gained a reply delay and a count of requests that actually reached it — the
latter is what distinguishes "rejected before dispatch" from "dispatched and then failed", which
the error code alone cannot.

Every new assertion was run against the pre-fix code first and observed to fail.
`cancellation_completes_the_handler_once` is the exception: it passes either way, so it is new
coverage of existing behaviour rather than a regression test.

thejcfactor (approved):

PR #1032, opened 2026-07-28
CXXCBC-891: tls_verify=none and live KV verification vs a CNG gateway

avsej:
To reach a real couchbase2 gateway whose certificate cannot be chained to a
trusted CA (the operator-served CNG cert on a k8s NodePort), honor
tls_verify=none: make_channel_credentials then builds an encrypted channel that
skips server-certificate and hostname verification (gRPC TlsCredentials with a
NoOpCertificateVerifier), instead of the CA-verifying SslCredentials.

Add a cluster_only live KV test that drives upsert -> get -> remove against the
endpoint in TEST_CONNECTION_STRING, exercising the whole transport (credentials,
dispatcher, converter, component, error mapping, Basic-auth metadata) against a
real gateway. It skips when TEST_CONNECTION_STRING is unset.

---
Stacked PR **10/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–10; the change specific to this PR is its top commit. Everything below it has merged except #1059, which is now the bottom of the open series and should merge first — this branch carries its commit as an ancestor and depends on it for a green `unit (tsan)` leg.

This commit previously also contained the `couchbase2://` open path and execute-time routing (now #1033) and a global MSVC link option (dropped from the series; see CXXCBC-916). #1033 no longer shares a head commit with this one.

### DIFF
--- a/core/protostellar/credentials.cxx
+++ b/core/protostellar/credentials.cxx
@@ -20,17 +20,22 @@
 #include "core/capella_ca.hxx"
 #include "core/cluster_credentials.hxx"
 #include "core/cluster_options.hxx"
+#include "core/logger/logger.hxx"
 #include "core/mozilla_ca_bundle.hxx"
 #include "core/platform/base64.h"
 #include "core/protostellar/dispatcher.hxx"
+#include "core/tls_verify_mode.hxx"
 
 #include <grpcpp/create_channel.h>
+#include <grpcpp/security/tls_certificate_provider.h>
+#include <grpcpp/security/tls_credentials_options.h>
 
 #include <fstream>
 #include <optional>
 #include <sstream>
 #include <stdexcept>
 #include <utility>
+#include <vector>
 
 namespace couchbase::core::protostellar
 {
@@ -166,16 +171,52 @@ authorization_header(const cluster_credentials& credentials) -> std::string
 }
 
 auto
+tls_peer_verification_disabled(const cluster_options& options) -> bool
+{
+  return options.enable_tls && options.tls_verify == tls_verify_mode::none;
+}
+
+auto
 make_channel_credentials(const cluster_options& options, const cluster_credentials& credentials)
   -> std::shared_ptr<grpc::ChannelCredentials>
 {
   if (!options.enable_tls) {
+    // Fail closed: never let username/password or a JWT ride a plaintext channel (base64 is not
+    // encryption). couchbase2:// implies TLS; a programmatic config that disables it while
+    // carrying credentials is a misconfiguration, not a downgrade we perform silently.
     if (credentials.uses_password() || credentials.uses_jwt() ||
         !credentials.certificate_path.empty() || !credentials.key_path.empty()) {
       throw std::invalid_argument(
         "couchbase2: refusing to send credentials over a plaintext channel; enable TLS");
     }
     return grpc::InsecureChannelCredentials();
+  }
+  if (tls_peer_verification_disabled(options)) {
+    // Encrypt the channel but skip server-certificate and hostname verification. Used against
+    // dev/test gateways whose certificate cannot be chained to a trusted CA (e.g. the CNG cert
+    // served by the Couchbase operator on a k8s NodePort).
+    //
+    // Warn here rather than only at the call site: this is the one point every couchbase2 channel
+    // passes through, including callers that build a component directly without going through
+    // cluster::open. A deployment that enabled this by accident has to be able to see it.
+    CB_LOG_WARNING("couchbase2: TLS certificate verification is DISABLED (tls_verify=none); the "
+                   "connection is encrypted but unauthenticated and can be intercepted. Do not use "
+                   "this outside development.");
+    grpc::experimental::TlsChannelCredentialsOptions tls;
+    tls.set_verify_server_certs(false);
+    tls.set_check_call_host(false);
+    tls.set_certificate_verifier(std::make_shared<grpc::experimental::NoOpCertificateVerifier>());
+    // Still present the configured client identity (mTLS) so certificate auth works even when the
+    // gateway's server certificate cannot be verified. Skipping server verification does not imply
+    // skipping our own authentication.
+    if (auto identity = read_client_identity(credentials)) {
+      tls.set_certificate_provider(
+        std::make_shared<grpc::experimental::StaticDataCertificateProvider>(
+          std::vector<grpc::experimental::IdentityKeyCertPair>{
+            { identity->first, identity->second } }));
+      tls.watch_identity_key_cert_pairs();
+    }
+    return grpc::experimental::TlsCredentials(tls);
   }
   return grpc::SslCredentials(build_ssl_options(options, credentials));
 }
--- a/test/cng/protostellar/live_kv.cxx
+++ b/test/cng/protostellar/live_kv.cxx
@@ -1,0 +1,330 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Live KV round-trip against a real couchbase2:// (Cloud Native Gateway) endpoint. cluster_only:
+// runs only when TEST_CONNECTION_STRING is set (the harness skips it otherwise). It drives the
+// full transport -- channel/credentials, dispatcher, converter, component, error mapping --
+// through an upsert -> get -> remove chain, so it fails loudly if any layer regresses against a
+// real gateway. Extra env: TEST_CB2_USERNAME/TEST_CB2_PASSWORD (default Administrator/password),
+// TEST_CB2_BUCKET (default "default"). TLS verification is disabled (dev gateway cert).
+
+#include "framework/test_runner.hxx"
+
+#include "core/cluster_credentials.hxx"
+#include "core/cluster_options.hxx"
+#include "core/document_id.hxx"
+#include "core/error_context/key_value.hxx"
+#include "core/protostellar/component.hxx"
+#include "core/protostellar/credentials.hxx"
+#include "core/protostellar/kv_converter.hxx"
+#include "core/tls_verify_mode.hxx"
+#include "core/utils/binary.hxx"
+#include "core/utils/connection_string.hxx"
+
+#include <couchbase/error_codes.hxx>
+
+#include <asio/executor_work_guard.hpp>
+#include <asio/io_context.hpp>
+
+#include <cstdlib>
+#include <string>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+namespace ops = ::couchbase::core::operations;
+namespace ps = ::couchbase::core::protostellar;
+namespace pk = ::couchbase::core::protostellar::kv;
+namespace cu = ::couchbase::core::utils;
+using ::couchbase::core::cluster_credentials;
+using ::couchbase::core::cluster_options;
+using ::couchbase::core::document_id;
+using ::couchbase::core::tls_verify_mode;
+using ::couchbase::core::protostellar::component;
+using ::couchbase::core::protostellar::component_config;
+using ::couchbase::core::utils::parse_connection_string;
+using namespace std::chrono_literals;
+
+auto
+env_or(const char* name, const char* fallback) -> std::string
+{
+  return safe_getenv(name).value_or(fallback);
+}
+
+void
+kv_crud_round_trip_against_live_gateway()
+{
+  const auto connstr_opt = safe_getenv("TEST_CONNECTION_STRING");
+  if (!connstr_opt.has_value()) {
+    skip("TEST_CONNECTION_STRING is not set");
+  }
+  const auto parsed = parse_connection_string(connstr_opt.value());
+  if (!parsed.uses_protostellar()) {
+    skip("TEST_CONNECTION_STRING is not a couchbase2:// endpoint");
+  }
+  if (parsed.bootstrap_nodes.empty()) {
+    skip("no nodes in TEST_CONNECTION_STRING");
+  }
+
+  const auto& node = parsed.bootstrap_nodes.front();
+  const auto port = node.port > 0 ? node.port : parsed.default_port;
+  const std::string endpoint = node.address + ":" + std::to_string(port);
+
+  cluster_options options;
+  options.enable_tls = true;
+  options.tls_verify = tls_verify_mode::none; // dev gateway certificate is not chainable
+
+  cluster_credentials credentials;
+  credentials.username = env_or("TEST_CB2_USERNAME", "Administrator");
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  // Named-field assignment rather than positional aggregate initialisation: component_config
+  // gains fields as the series progresses, and a positional list silently rebinds when it does.
+  component_config config;
+  config.channel = channel;
+  config.credentials = credentials;
+  config.default_kv_timeout = 20000ms;
+  component comp{ io, config };
+
+  std::string failure;    // non-empty => the round-trip failed at some stage
+  bool completed = false; // the whole chain ran
+  couchbase::cas upsert_cas;
+
+  ops::upsert_request upsert;
+  upsert.id = id;
+  upsert.value = cu::to_binary(document);
+  upsert.flags = 0x06U;
+
+  comp.execute(std::move(upsert), [&](ops::upsert_response upsert_result) {
+    if (upsert_result.ctx.ec()) {
+      failure = "upsert: " + upsert_result.ctx.ec().message();
+      work.reset();
+      return;
+    }
+    upsert_cas = upsert_result.cas;
+
+    ops::get_request get;
+    get.id = id;
+    comp.execute(std::move(get), [&](ops::get_response get_result) {
+      if (get_result.ctx.ec()) {
+        failure = "get: " + get_result.ctx.ec().message();
+        work.reset();
+        return;
+      }
+      if (cu::to_string(get_result.value) != document) {
+        failure = "get: value mismatch, got '" + cu::to_string(get_result.value) + "'";
+        work.reset();
+        return;
+      }
+      // The upsert set these deliberately so the document is written as JSON. Checking only the
+      // value bytes would pass against a converter that dropped content_flags entirely.
+      if (get_result.flags != 0x06U) {
+        failure = "get: flags mismatch, got " + std::to_string(get_result.flags);
+        work.reset();
+        return;
+      }
+
+      ops::remove_request remove;
+      remove.id = id;
+      comp.execute(std::move(remove), [&](ops::remove_response remove_result) {
+        if (remove_result.ctx.ec()) {
+          failure = "remove: " + remove_result.ctx.ec().message();
+        } else {
+          completed = true;
+        }
+        work.reset();
+      });
+    });
+  });
+
+  io.run();
+
+  assert_true(failure.empty(), failure);
+  assert_true(completed, "upsert -> get -> remove chain completed");
+  assert_true(upsert_cas.value() != 0, "upsert returned a non-zero CAS");
+}
+
+void
+insert_and_replace_round_trip_against_live_gateway()
+{
+  const auto connstr_opt = safe_getenv("TEST_CONNECTION_STRING");
+  if (!connstr_opt.has_value()) {
+    skip("TEST_CONNECTION_STRING is not set");
+  }
+  const auto parsed = parse_connection_string(connstr_opt.value());
+  if (!parsed.uses_protostellar() || parsed.bootstrap_nodes.empty()) {
+    skip("TEST_CONNECTION_STRING is not a couchbase2:// endpoint");
+  }
+  const auto& node = parsed.bootstrap_nodes.front();
+  const auto port = node.port > 0 ? node.port : parsed.default_port;
+  const std::string endpoint = node.address + ":" + std::to_string(port);
+
+  cluster_options options;
+  options.enable_tls = true;
+  options.tls_verify = tls_verify_mode::none;
+
+  cluster_credentials credentials;
+  credentials.username = env_or("TEST_CB2_USERNAME", "Administrator");
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-insert-replace-key" };
+  const auto missing_id = document_id{ bucket, "_default", "_default", "cng-live-missing-key" };
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  // Named-field assignment rather than positional aggregate initialisation: component_config
+  // gains fields as the series progresses, and a positional list silently rebinds when it does.
+  component_config config;
+  config.channel = channel;
+  config.credentials = credentials;
+  config.default_kv_timeout = 20000ms;
+  component comp{ io, config };
+
+  std::string failure;
+  bool completed = false;
+  bool duplicate_insert_failed_correctly = false;
+  bool replace_missing_failed_correctly = false;
+  bool remove_missing_failed_correctly = false;
+
+  // 1. Clean slate
+  ops::remove_request pre_remove;
+  pre_remove.id = id;
+  comp.execute(std::move(pre_remove), [&](ops::remove_response) {
+    // 2. Insert new document
+    ops::insert_request insert;
+    insert.id = id;
+    insert.value = cu::to_binary("doc1");
+    comp.execute(std::move(insert), [&](ops::insert_response insert_result) {
+      if (insert_result.ctx.ec()) {
+        failure = "insert: " + insert_result.ctx.ec().message();
+        work.reset();
+        return;
+      }
+
+      // 3. Insert existing document should fail with document_exists
+      ops::insert_request dup_insert;
+      dup_insert.id = id;
+      dup_insert.value = cu::to_binary("doc1_dup");
+      comp.execute(std::move(dup_insert), [&](ops::insert_response dup_result) {
+        if (dup_result.ctx.ec() == couchbase::errc::key_value::document_exists) {
+          duplicate_insert_failed_correctly = true;
+        } else {
+          failure = "duplicate insert did not fail with document_exists, got: " +
+                    dup_result.ctx.ec().message();
+          work.reset();
+          return;
+        }
+
+        // 4. Replace existing document
+        ops::replace_request replace;
+        replace.id = id;
+        replace.value = cu::to_binary("doc2");
+        comp.execute(std::move(replace), [&](ops::replace_response replace_result) {
+          if (replace_result.ctx.ec()) {
+            failure = "replace: " + replace_result.ctx.ec().message();
+            work.reset();
+            return;
+          }
+
+          // 5. Replace missing document should fail with document_not_found
+          ops::replace_request missing_replace;
+          missing_replace.id = missing_id;
+          missing_replace.value = cu::to_binary("doc_missing");
+          comp.execute(std::move(missing_replace), [&](ops::replace_response missing_result) {
+            if (missing_result.ctx.ec() == couchbase::errc::key_value::document_not_found) {
+              replace_missing_failed_correctly = true;
+            } else {
+              failure = "replace missing did not fail with document_not_found, got: " +
+                        missing_result.ctx.ec().message();
+              work.reset();
+              return;
+            }
+
+            // 6. Remove existing document
+            ops::remove_request remove;
+            remove.id = id;
+            comp.execute(std::move(remove), [&](ops::remove_response remove_result) {
+              if (remove_result.ctx.ec()) {
+                failure = "remove: " + remove_result.ctx.ec().message();
+                work.reset();
+                return;
+              }
+
+              // 7. Remove missing document should fail with document_not_found
+              ops::remove_request missing_remove;
+              missing_remove.id = missing_id;
+              comp.execute(
+                std::move(missing_remove), [&](ops::remove_response missing_remove_result) {
+                  if (missing_remove_result.ctx.ec() ==
+                      couchbase::errc::key_value::document_not_found) {
+                    remove_missing_failed_correctly = true;
+                    completed = true;
+                  } else {
+                    failure = "remove missing did not fail with document_not_found, got: " +
+                              missing_remove_result.ctx.ec().message();
+                  }
+                  work.reset();
+                });
+            });
+          });
+        });
+      });
+    });
+  });
+
+  io.run();
+
+  assert_true(failure.empty(), failure);
+  assert_true(duplicate_insert_failed_correctly, "duplicate insert failed with document_exists");
+  assert_true(replace_missing_failed_correctly, "replace missing failed with document_not_found");
+  assert_true(remove_missing_failed_correctly, "remove missing failed with document_not_found");
+  assert_true(
+    completed,
+    "insert -> dup_insert -> replace -> missing_replace -> remove -> missing_remove completed");
+}
+
+} // namespace
+
+auto
+tests() -> test_suite
+{
+  return {
+    "protostellar_live_kv",
+    {
+      { "kv_crud_round_trip_against_live_gateway",
+        kv_crud_round_trip_against_live_gateway,
+        timeout::integration,
+        test_env::cluster_only },
+      { "insert_and_replace_round_trip_against_live_gateway",
+        insert_and_replace_round_trip_against_live_gateway,
+        timeout::integration,
+        test_env::cluster_only },
+    },
+  };
+}
+
+} // namespace couchbase::cng::test
--- a/.github/actions/create-cng-cluster/action.yml
+++ b/.github/actions/create-cng-cluster/action.yml
@@ -1,0 +1,107 @@
+name: create-cng-cluster
+description: >
+  Creates a Couchbase cluster fronted by Cloud Native Gateway (the couchbase2:// gRPC endpoint)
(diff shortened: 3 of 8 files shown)


core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,207 @@
...
+        *state->context,
+        state->response,
+        [state, handler_holder, io, tracker, context](grpc::Status status) {
+          // Deregister on the gRPC thread first: once removed, a concurrent cancel_and_drain() may
+          // return. The post that follows only touches shared state (io outlives the call).
+          tracker->remove(context);
+          asio::post(*io, [state, handler_holder, status = std::move(status)]() mutable {
+            (*handler_holder)(std::move(status), std::move(state->response));
+          });
+        });
```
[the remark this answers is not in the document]
avsej: Deliberate and required as written. The completion posts the handler and, in the streaming reactor, the posted task self-deletes the reactor -- which owns context_. Removing from the tracker *after* posting would let the io thread run that self-delete and free context_ before the gRPC thread reads context_.get() in remove(), reintroducing a use-after-free. remove()-before-post() is kept consistent across unary() and the reactor; the safety rests on the documented SDK invariant that the io_context outlives the dispatcher, so a post issued just before remove() can never target a destroyed io_context.

core/protostellar/credentials.cxx
```diff
@@ -170,13 +174,36 @@ make_channel_credentials(const cluster_options& options, const cluster_credentia
...
+    // encryption). couchbase2:// implies TLS; a programmatic config that disables it while
+    // carrying credentials is a misconfiguration, not a downgrade we perform silently.
     if (credentials.uses_password() || credentials.uses_jwt() ||
         !credentials.certificate_path.empty() || !credentials.key_path.empty()) {
       throw std::invalid_argument(
         "couchbase2: refusing to send credentials over a plaintext channel; enable TLS");
     }
     return grpc::InsecureChannelCredentials();
   }
+  if (options.tls_verify == tls_verify_mode::none) {
```
avsej: This branch disables server certificate verification, and it lands with no test that can distinguish it from the verifying path. The only unit test touching `make_channel_credentials` asserts `!= nullptr` for `enable_tls` true and false — a condition every branch satisfies, including this one and including an edit that returned insecure credentials unconditionally.

Two things are needed:

1. A test that pins the *distinction* — that the default options produce a verifying channel and that `tls_verify=none` is the only combination that reaches this branch. That likely means factoring the decision into a small predicate the test can call, rather than only being able to inspect the opaque `ChannelCredentials`.
2. A warning-level log when this branch is taken. A connection running without certificate verification should be visible in the logs of a deployment that has it enabled by accident.

Related: when no trust certificate is configured the couchbase2 path seeds only the Capella root and the bundled Mozilla list, while the classic path also calls `set_default_verify_paths()`. RFC 77 (Bootstrapping → Security) requires the two default sets to be identical. A user whose gateway uses a platform-trusted or private CA therefore finds `couchbase2://` failing where `couchbase://` worked, with `tls_verify=none` as the switch that makes it work — which is how verification ends up disabled in production. CNG-14 records the trust-anchor question as unsettled across SDKs, so the default is a cross-SDK decision; the error message naming the missing trust configuration is not.

core/cluster.cxx
```diff
@@ -493,6 +498,11 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
                  origin_.connection_string(),
                  origin_.to_json());
     setup_observability();
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+    if (origin_.uses_protostellar()) {
```
avsej: `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` defaults to `OFF` (`CMakeLists.txt:114`), so this `#ifdef` is inactive in the shipping configuration — but the connection-string parser accepts the scheme unconditionally (`core/utils/connection_string.cxx:94-100` sets port 18098, `tls = true`, `protocol = protostellar` with no feature guard, and `validate_protostellar` checks only host count and bootstrap suffix).

So in a default build, `couchbase2://gateway` parses successfully, falls past this block, and enters MCBP bootstrap against the gateway's gRPC port. The user sees a bootstrap timeout or a TLS error that says nothing about the real cause.

Reject the scheme where it is parsed instead:

```cpp
#ifndef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
  if (connstr.protocol == connection_string::protocol_type::protostellar) {
    connstr.error = R"(couchbase2:// requires a build with couchbase2 support enabled)";
    return;
  }
#endif
```

core/cluster.cxx
```diff
@@ -704,6 +714,18 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
       return handler(request.make_response(
         make_key_value_error_context(errc::network::cluster_closed, request.id), response_type{}));
     }
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+    if (protostellar_) {
```
avsej: `cluster_impl::execute` runs on the caller's thread — nothing posts it — and reads this plain `std::shared_ptr` member directly, then dereferences it. `close()` resets the same member on the io thread. `stopped_` being `std::atomic_bool` narrows the window but does not order this access, and concurrent read/write of one `shared_ptr` object is a data race independent of refcount atomicity.

Scenario: a worker thread calls `collection.upsert(...)` while the io thread runs `protostellar_.reset()` from `close()`. The reader can observe a torn control-block pointer, or pass this test and then dereference a component whose last owner has just dropped.

A local copy fixes the dereference and documents the invariant:

```cpp
if (auto component = protostellar_; component) {
  component->execute(std::move(request), std::forward<Handler>(handler));
  return;
}
```

The load itself still needs to be safe — `std::atomic<std::shared_ptr<...>>`, a mutex, or a documented rule that `execute` is io-thread-only.

core/cluster.cxx
```diff
@@ -825,6 +856,76 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+                     "couchbase2; the connection is vulnerable to interception and must not be "
+                     "used in production.",
+                     id_);
+    }
+
+    std::shared_ptr<grpc::Channel> channel;
+    try {
+      channel = protostellar::make_channel(endpoint, options, origin_.credentials());
+    } catch (const std::exception& e) {
+      CB_LOG_ERROR(R"([{}]: failed to establish couchbase2 channel: {})", id_, e.what());
```
avsej: This catch block sets `stopped_ = true` before reporting the error to the handler. `close()` tests `stopped_` and returns early, so the caller's subsequent `cluster.close()` does nothing — and `setup_observability()` has already run at this point, so whatever it started keeps running.

Reporting the failure without latching `stopped_` lets the normal `close()` path perform the teardown it is written to do.

test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  component comp{ io, channel, credentials, 20000ms };
```
avsej: This four-argument positional construction (also at `:183`) binds a bare duration to the fourth parameter of a constructor with no defaults, in a file only a `CB_COUCHBASE2=1` build compiles.

It has since been broken three times without being touched: #1036 added a fifth parameter, #1038 replaced the timeouts with the `component_timeouts` aggregate so `20000ms` no longer converts, and #1047 added a fifth parameter again. The `build` check has failed on every pull request from #1036 upward as a result, on every platform leg.

Naming the field removes the positional coupling:

```cpp
  component comp{ io, channel, credentials, component_timeouts{ .key_value = 20000ms }, {} };
```

The durable fix is on the constructor rather than here — a configuration struct, or defaults on the trailing parameters, so that adding a parameter is not a silent breaking change to every call site.

avsej (commented):
## Review — PR #1032 (`CXXCBC-891`)

This commit does three separable things: it adds the `couchbase2://` open and execute routing in `core/cluster.cxx`, it adds a `tls_verify=none` path for the gRPC channel, and it adds live KV and connect verification tests. The routing is well guarded — operations that are not mapped return `feature_not_available` rather than falling through to the never-bootstrapped MCBP session manager — and the live-test scaffolding is reasonable.

The security-relevant part needs work. `tls_verify=none` is a switch that turns off certificate verification for a database connection, and it lands with no automated coverage of any kind. The default trust anchors also differ from the classic transport in a way that makes that switch look like the natural workaround rather than a debugging aid.

PR #1033 has a byte-identical head commit to this one. One of the two should be closed, or the two commits separated, so that review and CI effort is not duplicated.

### Strengths

- The execute-time routing fails closed. `component_supports_v` gates the KV surface and unmapped operations return `feature_not_available`, which is much safer than letting a request reach an MCBP session manager that was never bootstrapped.
- `open_protostellar` is a distinct code path rather than a set of conditionals threaded through the existing bootstrap, so the classic path is not at risk from it.
- The live tests skip cleanly when no gateway is configured, and `test/cng/framework/live_fixture.hxx` is the right instinct — a shared fixture rather than per-file duplication.

### Findings

|---|---|---|---|---|
| 1 | high | test-coverage | `tls_verify=none` ships with no test that can distinguish it from a verifying channel | `core/protostellar/credentials.cxx:187` |
| 2 | high | spec-compliance | The couchbase2 default trust anchors are not the classic set, which makes `tls_verify=none` the path of least resistance | `core/protostellar/credentials.cxx:187` |
| 3 | high | correctness | `couchbase2://` is accepted and routed to MCBP bootstrap when the feature is compiled out — the default build configuration | `core/cluster.cxx:502` |
| 4 | high | race | `protostellar_` is read from caller threads and reset from the io thread with no synchronisation | `core/cluster.cxx:718` |
| 5 | medium | correctness | The channel-construction failure path sets `stopped_`, which makes the subsequent `close()` a no-op | `core/cluster.cxx:905` |
| 6 | medium | api-design | A four-argument positional construction is written twice in a test that only a feature-gated CI leg compiles | `test/cng/protostellar/live_kv.cxx:100` |
| 7 | medium | test-coverage | The live KV round-trip sets `content_flags` deliberately and never reads them back | `test/cng/protostellar/live_kv.cxx:109` |
| 8 | low | scope | An unrelated global MSVC link option is folded into this commit | `CMakeLists.txt:60` |

#### 1. `tls_verify=none` ships with no test that can distinguish it from a verifying channel (high)

```cpp
if (options.tls_verify == tls_verify_mode::none) {
```

This branch selects channel credentials that do not verify the server certificate. The only unit test that touches `make_channel_credentials` asserts `!= nullptr` for `enable_tls = true` and for `enable_tls = false`, which every branch of the function satisfies — including this one, and including a hypothetical future edit that returned insecure credentials unconditionally.

So the most security-sensitive branch added by this commit has no coverage, and the test that appears to cover the area cannot fail. For a switch that disables certificate verification, the test needs to assert the *distinction*, not non-nullness:

```cpp
void
tls_verify_none_is_reachable_only_when_explicitly_requested()
{
  cluster_credentials creds{};
  cluster_options verifying{};
  verifying.enable_tls = true;
  // tls_verify defaults to peer: the returned credentials must verify.
  assert_true(ps::verifies_peer(ps::make_channel_credentials(verifying, creds)),
              "default TLS options must verify the peer certificate");

  cluster_options unverified{};
  unverified.enable_tls = true;
  unverified.tls_verify = tls_verify_mode::none;
  assert_false(ps::verifies_peer(ps::make_channel_credentials(unverified, creds)),
              "tls_verify=none must be the only way to reach an unverified channel");
}
```

That requires exposing enough of the decision to be observable — for example by factoring the branch into a small predicate that the test can call directly, rather than only being able to look at the opaque `ChannelCredentials`. The important property to pin is that no other combination of options reaches the unverified branch.

A log line at warning level when the branch is taken would also be appropriate; a connection running without certificate verification should be visible in the logs of any deployment that has it on by accident.

(inline comment on `core/protostellar/credentials.cxx:187`)

#### 2. The couchbase2 default trust anchors are not the classic set (high)

When no trust certificate is configured, the couchbase2 path seeds `pem_root_certs` from the Capella root plus the bundled Mozilla CA list. The classic path additionally calls `set_default_verify_paths()`, so it trusts the operating system store as well.

RFC 77, Bootstrapping and Connections → Security, is explicit: "The set of Root CAs that SDKs trust by default for couchbase2 must be identical to the one used for Classic." The implementation audit records this area as unsettled across SDKs (CNG-14) — gocb falls back to the system pool (`routingconn.go:67`), couchbase-jvm-clients requires an explicit trust configuration — so the *right* default is a cross-SDK decision rather than something to settle in this review.

What is not a matter of opinion is the interaction with finding 1: a user whose gateway presents a certificate signed by a private or platform-trusted CA finds that `couchbase2://` fails where `couchbase://` worked, and the same commit hands them `tls_verify=none` as the switch that makes it work. That combination pushes people toward disabling verification for a reason that is really a missing trust anchor. Matching the classic default set removes the pressure; if the divergence is deliberate, it needs a comment saying so and an error message that names the missing trust configuration rather than a generic handshake failure.

#### 3. `couchbase2://` is accepted in builds that cannot serve it (high)

```cpp
#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
    if (origin_.uses_protostellar()) {
      return open_protostellar(std::move(handler));
    }
#endif
```

`COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` defaults to `OFF` (`CMakeLists.txt:114`), so this is the shipping configuration. The connection-string parser, however, accepts the scheme unconditionally — `core/utils/connection_string.cxx:94-100` sets `default_port = 18098`, `tls = true`, `protocol = protostellar` with no feature guard, and `validate_protostellar` only checks the host count and the bootstrap-mode suffix.

Failure scenario: a user builds the library with default options, connects with `couchbase2://gateway.example.com`, and the open call falls straight past this `#ifdef` into the MCBP bootstrap path — which then attempts an MCBP handshake against the gateway's gRPC port. The result is a bootstrap timeout or a TLS error that says nothing about the actual problem, which is that this build has no couchbase2 support.

The scheme should be rejected at parse or validation time when the feature is absent:

```cpp
#ifndef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
  if (connstr.protocol == connection_string::protocol_type::protostellar) {
    connstr.error = R"(couchbase2:// requires a build with couchbase2 support enabled)";
    return;
  }
#endif
```

(inline comment on `core/cluster.cxx:502`)

#### 4. `protostellar_` is read from caller threads and reset from the io thread (high)

```cpp
if (protostellar_) {
```

`cluster_impl::execute` runs on the caller's thread — nothing posts it — and reads this plain `std::shared_ptr` member directly, both to test it and then to dereference it. `close()` resets the same member on the io thread. `stopped_` is `std::atomic_bool`, which narrows the window but does not order the shared_ptr access, and concurrent read/write of the same `shared_ptr` object is a data race regardless of the refcount's own atomicity.

Failure scenario: a worker thread calls `collection.upsert(...)` while another thread's `cluster.close()` handler runs `protostellar_.reset()` on the io thread. The reader can observe a torn control-block pointer, or pass the `if` and then dereference a component whose last owner has just been dropped. ThreadSanitizer would report this on the `cng` leg if any test exercised concurrent operations, which none currently does.

Taking a local copy is both the fix and a documentation of the invariant:

```cpp
if (auto component = protostellar_; component) {
  ...
  component->execute(std::move(request), std::forward<Handler>(handler));
  return;
}
```

That still needs the load itself to be safe — either an `std::atomic<std::shared_ptr<...>>`, or a mutex, or the documented rule that `execute` must only be called from the io thread.

(inline comment on `core/cluster.cxx:718`)

#### 5. The channel-construction failure path makes the subsequent `close()` a no-op (medium)

```cpp
CB_LOG_ERROR(R"([{}]: failed to establish couchbase2 channel: {})", id_, e.what());
```

The catch block sets `stopped_ = true` and releases the work guard before invoking the handler with an error. `close()` begins by testing `stopped_` and returning early, so the caller's later `cluster.close()` does nothing — which leaves whatever `setup_observability()` started still running, since that ran before this point.

Reporting the error without latching `stopped_` lets the normal `close()` path do the teardown it is written to do.

(inline comment on `core/cluster.cxx:905`)

#### 6. A four-argument positional construction, written twice, in a feature-gated test (medium)

```cpp
component comp{ io, channel, credentials, 20000ms };
```

This appears at `:100` and `:183`. It binds a bare duration to the fourth positional parameter of a constructor with no defaults, in a file that only a build with `CB_COUCHBASE2=1` compiles.

That combination has not aged well. The constructor gained a fifth parameter in #1036, had its timeout parameters replaced by the `component_timeouts` aggregate in #1038 — at which point `20000ms` no longer converts — and gained a fifth parameter again in #1047. This call site was never updated, and the `build` check has been failing on every pull request from #1036 upward as a result, on every platform leg.

The call site is not at fault by itself; the constructor's shape is. Naming the field removes the positional coupling, and a configuration struct on the component would remove it permanently:

```cpp
component comp{ io, channel, credentials, component_timeouts{ .key_value = 20000ms }, {} };
```

(inline comment on `test/cng/protostellar/live_kv.cxx:100`)

#### 7. The live KV round-trip never reads back the flags it sets (medium)

```cpp
upsert.flags = 0x06U;
```

The flags are set deliberately so the document is written as JSON, and the get callback then checks only the value bytes. A converter that dropped `content_flags` entirely would pass this test. Asserting `get_result.flags == 0x06U` after the round trip costs one line and makes the test cover the field it was written to exercise.

(inline comment on `test/cng/protostellar/live_kv.cxx:109`)

#### 8. An unrelated global MSVC link option is folded into this commit (low)

`add_link_options($<$<CONFIG:Debug,RelWithDebInfo>:/DEBUG:FASTLINK>)` is added at top level, so it applies to the client library, every example, every test binary, and every downstream consumer of the exported configuration. The change may well be necessary for the template-heavy gRPC stubs, but it is a global build-behaviour change with packaging implications, it is not mentioned in the pull request body, and it does not belong in a commit about TLS verification and live KV tests.

More generally, the pull request body describes `tls_verify=none` and the live KV test, while the diff also contains the entire `couchbase2://` open path, the execute-time routing, the CI workflow additions and this link option. Splitting the routing from the TLS change would make both easier to review and to revert.

### Verdict

Needs changes. Findings 1 and 2 together are the reason: a switch that disables certificate verification should not land without a test that can tell it apart from the verifying path, and it should not be the easiest way to work around a trust-anchor default that differs from the classic transport.

test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+  std::string failure;    // non-empty => the round-trip failed at some stage
+  bool completed = false; // the whole chain ran
+  couchbase::cas upsert_cas;
+
+  ops::upsert_request upsert;
+  upsert.id = id;
+  upsert.value = pk::to_bytes(document);
+  upsert.flags = 0x06U;
```
avsej, line 118: These flags are set deliberately so the document is written as JSON, but the get callback checks only the value bytes — a converter that dropped `content_flags` entirely would pass. Asserting `get_result.flags == 0x06U` after the round trip costs one line and makes the test cover the field it exists to exercise.

CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ endif()
 # abort with `nasm.exe ... /MP ... exited with code 1`.
 if(MSVC)
   add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/MP>)
+  # /DEBUG:FASTLINK avoids LNK1140 ("limit exceeded for program database") on MSVC Debug and
```
avsej: This adds a global MSVC link option at top level, so it applies to the client library, every example, every test binary and every downstream consumer of the exported configuration. It may well be necessary for the template-heavy generated stubs, but it is a global build-behaviour change with packaging implications, it is not mentioned in the pull request body, and it is unrelated to TLS verification or live KV tests.

More broadly, this commit contains the `couchbase2://` open path, the execute-time routing, the CI workflow additions and this link option in addition to what the body describes. Splitting the routing from the TLS change would make both easier to review and to revert independently.

avsej:
The commit has been split. This pull request now carries only the `tls_verify=none` change and the
live KV verification; the `couchbase2://` open path and the execute-time routing have moved to
#1033, and the MSVC link option has been dropped from the series altogether — see finding 8. Four
of the findings therefore have their fix in #1033 and are answered there rather than here.

> This commit does three separable things: it adds the `couchbase2://` open and execute routing in `core/cluster.cxx`, it adds a `tls_verify=none` path for the gRPC channel, and it adds live KV and connect verification tests.

> PR #1033 has a byte-identical head commit to this one. One of the two should be closed, or the two commits separated, so that review and CI effort is not duplicated.

Separated. The split was verified lossless: the three resulting commits, diffed against the merge
base, reproduce the original diff byte-for-byte apart from two deliberate edits — a comment on the
dispatcher test that the commit had deleted for no reason is restored, and a reference to a
"CXXCBC-891b" that does not exist is corrected.

### 1. `tls_verify=none` ships with no test that can distinguish it from a verifying channel

> This branch selects channel credentials that do not verify the server certificate. The only unit test that touches `make_channel_credentials` asserts `!= nullptr` for `enable_tls = true` and for `enable_tls = false`, which every branch of the function satisfies — including this one, and including a hypothetical future edit that returned insecure credentials unconditionally.

Correct, and the "hypothetical future edit" turned out to be measurable rather than hypothetical.
The decision is now a named predicate that `make_channel_credentials()` itself branches on:

```cpp
[[nodiscard]] auto
tls_peer_verification_disabled(const cluster_options& options) -> bool;
```

so a test pins the decision that is actually taken rather than a parallel reimplementation of it.

> That requires exposing enough of the decision to be observable — for example by factoring the branch into a small predicate that the test can call directly, rather than only being able to look at the opaque `ChannelCredentials`. The important property to pin is that no other combination of options reaches the unverified branch.

Two cases were added. `peer_verification_is_disabled_only_by_tls_verify_none` covers default
options, `tls_verify=peer` over TLS, `tls_verify=none` over TLS, and `tls_verify=none` *without*
TLS. `peer_verification_does_not_depend_on_the_credential_shape` sweeps the four credential shapes
(none, password, JWT, client certificate) against both verify modes, so no credential
configuration can steer the decision on its own.

The last of those four is a deliberate judgement rather than a mechanical enumeration:
`tls_verify=none` with `enable_tls=false` reports **false**. A plaintext channel presents no
certificate to verify, and `make_channel_credentials()` already refuses to put credentials on one;
reporting it as "unverified" would conflate two different failures and make the predicate useless
as a signal. This is stated in the header comment and asserted in the test.

Both new cases were negative-controlled. With the predicate replaced by `return false;` — a build
that compiles, which was checked, because a control that fails to compile silently reuses the
previous binary — both new cases fail and the pre-existing `channel_credentials_switch_on_tls`
still passes. That is the claim above, reproduced: the old test cannot fail. With the
`enable_tls &&` conjunct dropped, the plaintext case fails.

> A log line at warning level when the branch is taken would also be appropriate; a connection running without certificate verification should be visible in the logs of any deployment that has it on by accident.

Added, in `make_channel_credentials()` rather than at the call site. That is the single point every
couchbase2 channel passes through, so it also covers callers that construct a component directly
without going through `cluster_impl::open`. The near-duplicate warning that the routing commit
emitted has been removed with it; what remains there is only the Capella clamp message, which
reports something different — that a requested `tls_verify=none` was overridden.

### 2. The couchbase2 default trust anchors are not the classic set

> RFC 77, Bootstrapping and Connections → Security, is explicit: "The set of Root CAs that SDKs trust by default for couchbase2 must be identical to the one used for Classic."

The quotation is accurate. No change is being made here, because the point is already addressed on
`main` and has been since #1028 — in a file this pull request does not touch, which is why it is not
visible in the diff.

> if the divergence is deliberate, it needs a comment saying so

`core/protostellar/credentials.cxx:62-83` quotes that same RFC 77 sentence verbatim and then gives
the reasoning: gRPC accepts trust material only as a single PEM blob and treats a non-empty
`pem_root_certs` as the entire store, so there is nothing to layer the platform store onto and no
incremental API to layer with; and snapshotting the platform store would not reproduce the classic
set anyway, because under OpenSSL/BoringSSL "system roots" is not a defined collection but whatever
the host ships, relocatable at run time through `SSL_CERT_FILE`/`SSL_CERT_DIR`. The comment closes
with the cross-SDK survey — Java trusts its Capella root plus the JVM trust store, .NET adds its
Capella root to the chain's `ExtraStore`, Go trusts only `x509.SystemCertPool()` and ships no
Capella root. `core/protostellar/credentials.hxx:39-45` states the same contract on the declaration.

> an error message that names the missing trust configuration rather than a generic handshake failure

Not done, and not attempted rather than quietly dropped. A gRPC handshake failure arrives as
`UNAVAILABLE` with a provider-specific string, so attributing it to a missing trust anchor is a
heuristic on that string. Committing to a shape for that without measuring what the gateway
actually returns would be guesswork; it is recorded for separate work.

### 6. A four-argument positional construction, written twice, in a feature-gated test

> This four-argument positional construction (also at `:183`) binds a bare duration to the fourth parameter of a constructor with no defaults, in a file only a `CB_COUCHBASE2=1` build compiles.

The immediate breakage was already resolved in the #1031 round, which introduced `component_config`
and migrated both call sites. The durable problem described was not, and both sites have now been
changed to named-field assignment:

```cpp
component_config config;
config.channel = channel;
config.default_kv_timeout = 20000ms;
component comp{ io, config };
```

> ```cpp
>   component comp{ io, credentials, component_timeouts{ .key_value = 20000ms }, {} };
> ```

The designated-initialiser form is not available: this project is C++17
(`cmake/StandardProjectSettings.cmake:1`, `cmake/CompilerOptions.cmake:30`) and designated
initialisers are C++20. There are no occurrences of the syntax anywhere under `core/` or `test/`.
Named-field assignment gives the same property — immunity to a field being inserted or reordered —
within the language version in use.

This mattered during the rebase that followed. The later commits add `default_query_timeout`, then
replace the timeouts with an aggregate, then add compression settings; each of those produced a
conflict at these two call sites, and each was resolved by adding a named assignment rather than by
counting positions.

### 7. The live KV round-trip never reads back the flags it sets

> These flags are set deliberately so the document is written as JSON, but the get callback checks only the value bytes — a converter that dropped `content_flags` entirely would pass.

Fixed. The same gap was present in `live_connect.cxx`, which was not flagged; it is fixed too, in
#1033 where that file now lives.

### 8. An unrelated global MSVC link option is folded into this commit

> This adds a global MSVC link option at top level, so it applies to the client library, every example, every test binary and every downstream consumer of the exported configuration.

Removed from the series entirely. Checking it before moving it produced a result that changes what
the change is worth, so it is not being carried forward in any form here; the analysis and the
candidate patch are recorded in CXXCBC-916 instead.

On the
`windows (windows-11-arm)` leg of this pull request's own CI run (run 30867596374, job 91862700004),
`/DEBUG:FASTLINK` **is** applied — 130 link commands carry it — and `LNK1140` occurs anyway. The
flags on the exact command that fails are:

```
/DEBUG:FASTLINK /INCREMENTAL /PDB:".../test_integration_crud.pdb"
→ LINK : fatal error LNK1140: limit exceeded for program database
```

Seven seconds later the same job reports `there is not enough space on the disk` (FTK1005/FTK1006)
and `LNK1318 Unexpected PDB error; FILE_SYSTEM (3)` on a sibling target.

That leg is also a regression introduced by this commit: it passes on `main` at this pull request's
merge base (`f7921961`) and fails at this head, while `windows-2022` and `windows-2025` pass on the
same run.

So the justification in the comment — that the option avoids LNK1140 — does not hold as written, and
the change has been dropped rather than relocated.

What is *not* established is whether the option helps marginally, whether the disk exhaustion is
cause or consequence, whether `/DEBUG:FASTLINK` is even honoured by the ARM64 cross-toolchain, or
whether the trigger is the two additional gRPC-linked test binaries this commit adds. Removing it
outright is not obviously safe either, since the other 129 links would then produce full rather than
referenced PDBs. Each iteration on that leg costs roughly 75 minutes, so it is not settled by
assertion here: **CXXCBC-916** carries the measurements above, the candidate patch verbatim for
reference, and an explicit note not to apply it as-is.

The ARM64 leg is therefore still expected to fail on this series, and that is a real failure rather
than a known-flaky one. It is tracked separately because its cause is not this commit's subject and
is not yet understood.

### Findings answered in #1033

Findings 3 (`couchbase2://` accepted in builds that cannot serve it), 4 (the `protostellar_` data
race), and 5 (the channel-construction failure path) are all in `core/cluster.cxx`, which has moved
to #1033. They are answered there. Two of the three are fixed differently from the shape suggested,
with the reasoning given in that response.

### Verification

Every commit in the series was rebuilt and re-tested individually after the split and the rebase
onto `main`, not only the tip: configure, build, `ctest -L cng`, and a `clang-format` check over
each commit's own changed files. The new `tls_verify` cases run on every leg; the new build-support
case added in #1033 runs on the **default** legs, where couchbase2 is off.

core/protostellar/credentials.cxx
avsej: Fixed, and one half of it needed no change.

The decision is now a named predicate that `make_channel_credentials()` itself branches on,
`tls_peer_verification_disabled(const cluster_options&)`, so a test pins the branch that is
actually taken rather than a reimplementation of it. Two cases were added: the 2x2 of
`enable_tls` against `tls_verify`, and a sweep of all four credential shapes against both verify
modes so no credential configuration can steer the decision on its own. `tls_verify=none` with
`enable_tls=false` deliberately reports **false** — a plaintext channel presents no certificate
to verify, and reporting it as "unverified" would conflate two different failures.

Both cases were negative-controlled. Under `return false;` both fail while the pre-existing
`channel_credentials_switch_on_tls` still passes, which is exactly the point made here about it
being unable to fail.

The warning was added inside `make_channel_credentials()` rather than at the call site: that is
the one point every couchbase2 channel passes through, including callers that build a component
directly. The near-duplicate warning in the routing commit was removed with it.

On the trust anchors: the quotation from RFC 77 is accurate, but the remedy asked for is already
on `main` and has been since #1028, in a file this pull request does not touch —
`core/protostellar/credentials.cxx:62-83` quotes that same sentence verbatim, explains why the
platform store cannot be layered in (gRPC takes roots only as one PEM blob that replaces its
store outright), and surveys how Java, .NET and Go each land differently. The remaining
sub-point, an error message naming the missing trust configuration, is **not** done: a gRPC
handshake failure arrives as `UNAVAILABLE` with a provider-specific string, so attributing it
would be a heuristic on that string, and it is recorded rather than guessed at.

core/cluster.cxx
avsej: **This file moved to #1033 in the split**, so this thread is outdated here. Fixed there, but not in the parser.

Rejecting the scheme at parse time would break eight existing assertions:
`test/cng/connection_string/couchbase2_scheme.cxx` and `test/cng/origin/protocol.cxx` are
registered **outside** the couchbase2 guard (`test/cng/CMakeLists.txt:89,92`) and therefore build
and run in the default configuration — the very one being protected. It would also make parsing
behaviour depend on the build mode, and the intended contract is that couchbase2 support is
transparent: no API or ABI differs by build configuration. That currently holds — the macro
appears in exactly one library translation unit and in no public header.

The rejection is instead in `cluster_impl::open`, with the condition unconditional and only the
outcome varying, so parsing is identical in both builds and the user gets
`feature_not_available` plus a log line naming the missing build option.

A new test, registered outside the guard so it runs on the default legs, pins both halves. With
the fix reverted it does not merely return the wrong code — it **times out**, which is the
bootstrap timeout described in this comment.

core/cluster.cxx
avsej: **This file moved to #1033 in the split**, so this thread is outdated here. Fixed there. `std::atomic<std::shared_ptr<...>>` is C++20 and unavailable — this project is
C++17 — so of the remaining options the mutex was taken, because the same file already solves
the identical problem the same way one branch earlier in the very same function:
`find_bucket_by_name` (`core/cluster.cxx:878`) locks and returns the `shared_ptr` by value.

Every read now goes through a `protostellar_component()` accessor, `is_protostellar()` included.
`close()` moves the pointer out under the lock and destroys the component **outside** it, since
`~component` drains in-flight gRPC calls and holding the mutex across that drain would block
concurrent `execute()`. The retry helpers added later in the series needed the same treatment:
they re-enter themselves from a timer, so each attempt re-reads through the accessor.

Stated plainly: no assertion-based regression test is offered. A data race is not
deterministically reproducible from a unit test, and one that passed would be evidence of
nothing. ThreadSanitizer is the right instrument, as noted here.

core/cluster.cxx
avsej: **This file moved to #1033 in the split**, so this thread is outdated here. Fixed there, though the diagnosis turned out to be different.

The consequence described is real. The attribution is not: latching `stopped_` is this file's
established contract for a failed open — `open()` already has two sibling early-failure paths
(`core/cluster.cxx:483-487` and `:542-546`) that do exactly `stopped_ = true; work_.reset();`.
Removing the latch would diverge from both and leave a failed-open cluster answering `execute()`
through an MCBP path that was never bootstrapped.

The actual defect is ordering: both siblings latch **before** `setup_observability()`, and the
couchbase2 catch block latched after it. Fixed by moving the branch above
`setup_observability()` and calling it inside `open_protostellar()` once the channel is built, so
the failure path has the same shape as its siblings and nothing has been started when it latches.
Moving that call is load-bearing rather than cosmetic — without it a *successful* couchbase2 open
would start no observability at all.

test/cng/protostellar/live_kv.cxx
avsej: The immediate breakage was fixed in the #1031 round, which introduced `component_config` and
migrated both call sites. The durable problem described here was not, and both sites now use
named-field assignment on a default-constructed config.

The suggested `component_timeouts{ .key_value = 20000ms }` is not available: designated
initialisers are C++20 and this project is C++17 (`cmake/StandardProjectSettings.cmake:1`), with
no occurrences of the syntax anywhere under `core/` or `test/`. Named-field assignment gives the
same immunity to a field being inserted or reordered within the language version in use.

That mattered immediately: the later commits add `default_query_timeout`, then replace the
timeouts with an aggregate, then add compression settings, and each produced a conflict at these
two sites which was resolved by adding a named assignment rather than by counting positions.

test/cng/protostellar/live_kv.cxx
avsej: Fixed. The same gap was present in `live_connect.cxx`, which was not flagged; it is fixed too,
in #1033 where that file now lives.

CMakeLists.txt
avsej: Removed from the series entirely, rather than relocated — checking it first showed the change
does not do what its comment claimed.

On this pull request's own CI run (30867596374, job 91862700004), `/DEBUG:FASTLINK` **is**
applied — 130 link commands carry it — and `LNK1140` occurs anyway. The flags on the exact
command that fails are `/DEBUG:FASTLINK /INCREMENTAL /PDB:".../test_integration_crud.pdb"`, and
seven seconds later the same job reports `there is not enough space on the disk` plus `LNK1318`
on a sibling target.

That leg is also a regression introduced by this commit: `windows (windows-11-arm)` passes on
`main` at this pull request's merge base and fails at this head, while `windows-2022` and
`windows-2025` pass on the same run. So the ARM64 failure is real rather than known-flaky, and
it is expected to persist on this series.

What is not established is whether the option helps marginally, whether the disk exhaustion is
cause or consequence, whether the flag is honoured by the ARM64 cross-toolchain at all, or
whether the trigger is the two extra gRPC-linked test binaries this commit adds. Removing it
outright is not obviously safe either, since the other 129 links would then produce full rather
than referenced PDBs. Each iteration costs roughly 75 minutes, so it is tracked in
**CXXCBC-916** with the measurements and the candidate patch recorded verbatim and marked not to
be applied as-is.

thejcfactor (approved):

PR #1034, opened 2026-07-28
CXXCBC-894: complete the KV unary operation surface over couchbase2

avsej:
Add the remaining single-document KV operations to the converter and component:
touch, exists, get_and_lock, unlock, get_and_touch, increment, decrement,
append, and prepend, each with its kv.v1 request/response mapping (lock time,
expiry, counter delta/initial, append/prepend content, CAS, durability). Each is
routed by cluster_impl::execute via a component_supports_v specialization; the
five CRUD ops from CXXCBC-891/892 already covered the rest.

Covered by converter unit tests and two live cluster_only round-trips against a
CNG gateway (CRUD, and increment/exists/remove).

---
Stacked PR **12/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–12; the change specific to this PR is its top commit. Depends on #1044 — merge the series bottom-up.

**Not included: subdoc.** `lookup_in` and `mutate_in` are not routed over `couchbase2://` and return `errc::common::feature_not_available`. "The KV unary operation surface" here means the whole-document operations; the proto does define `LookupInRequest` and `MutateInRequest`, and the reference clients implement them, so this is a gap to be closed by a separate ticket rather than a limitation of the transport. Failing cleanly rather than silently degrading is deliberate.

### DIFF
--- a/core/protostellar/kv_converter.hxx
+++ b/core/protostellar/kv_converter.hxx
@@ -24,10 +24,20 @@
 
 #include "core/document_id.hxx"
 #include "core/error_context/key_value.hxx"
+#include "core/operations/document_append.hxx"
+#include "core/operations/document_decrement.hxx"
+#include "core/operations/document_exists.hxx"
 #include "core/operations/document_get.hxx"
+#include "core/operations/document_get_and_lock.hxx"
+#include "core/operations/document_get_and_touch.hxx"
+#include "core/operations/document_get_projected.hxx"
+#include "core/operations/document_increment.hxx"
 #include "core/operations/document_insert.hxx"
+#include "core/operations/document_prepend.hxx"
 #include "core/operations/document_remove.hxx"
 #include "core/operations/document_replace.hxx"
+#include "core/operations/document_touch.hxx"
+#include "core/operations/document_unlock.hxx"
 #include "core/operations/document_upsert.hxx"
 #include "core/utils/binary.hxx"
 
@@ -166,6 +176,37 @@ decode(const v1::GetResponse& proto, key_value_error_context ctx) -> operations:
   return response;
 }
 
+inline auto
+encode(const operations::get_projected_request& request) -> v1::GetRequest
+{
+  v1::GetRequest proto;
+  set_location(proto, request.id);
+  for (const auto& path : request.projections) {
+    proto.add_project(path);
+  }
+  return proto;
+}
+
+inline auto
+decode(const v1::GetResponse& proto,
+       key_value_error_context ctx,
+       const operations::get_projected_request& /* request */) -> operations::get_projected_response
+{
+  operations::get_projected_response response;
+  response.ctx = std::move(ctx);
+  if (content_is_compressed(proto)) {
+    response.ctx.override_ec(errc::common::feature_not_available);
+    return response;
+  }
+  response.value = utils::to_binary(proto.content_uncompressed());
+  response.cas = couchbase::cas{ proto.cas() };
+  response.flags = proto.content_flags();
+  if (proto.has_expiry()) {
+    response.expiry = static_cast<std::uint32_t>(proto.expiry().seconds());
+  }
+  return response;
+}
+
 // ── Mutations (upsert / insert / replace / remove) ─────────────────────────────
 
 template<typename Response, typename Proto>
@@ -269,6 +310,214 @@ inline auto
 decode(const v1::RemoveResponse& proto, key_value_error_context ctx) -> operations::remove_response
 {
   return decode_mutation<operations::remove_response>(proto, std::move(ctx));
+}
+
+// ── Touch / Exists / lock / counters / append / prepend ────────────────────────
+
+inline auto
+encode(const operations::touch_request& request) -> v1::TouchRequest
+{
+  v1::TouchRequest proto;
+  set_location(proto, request.id);
+  set_expiry(proto, request.expiry);
+  return proto;
+}
+
+inline auto
+decode(const v1::TouchResponse& proto, key_value_error_context ctx) -> operations::touch_response
+{
+  operations::touch_response response;
+  response.ctx = std::move(ctx);
+  response.cas = couchbase::cas{ proto.cas() };
+  return response;
+}
+
+inline auto
+encode(const operations::exists_request& request) -> v1::ExistsRequest
+{
+  v1::ExistsRequest proto;
+  set_location(proto, request.id);
+  return proto;
+}
+
+inline auto
+decode(const v1::ExistsResponse& proto, key_value_error_context ctx) -> operations::exists_response
+{
+  operations::exists_response response;
+  response.ctx = std::move(ctx);
+  response.document_exists = proto.result();
+  response.cas = couchbase::cas{ proto.cas() };
+  return response;
+}
+
+inline auto
+encode(const operations::get_and_lock_request& request) -> v1::GetAndLockRequest
+{
+  v1::GetAndLockRequest proto;
+  set_location(proto, request.id);
+  proto.set_lock_time_secs(request.lock_time);
+  return proto;
+}
+
+inline auto
+decode(const v1::GetAndLockResponse& proto, key_value_error_context ctx)
+  -> operations::get_and_lock_response
+{
+  operations::get_and_lock_response response;
+  response.ctx = std::move(ctx);
+  if (content_is_compressed(proto)) {
+    response.ctx.override_ec(errc::common::feature_not_available);
+    return response;
+  }
+  response.value = utils::to_binary(proto.content_uncompressed());
+  response.cas = couchbase::cas{ proto.cas() };
+  response.flags = proto.content_flags();
+  return response;
+}
+
+inline auto
+encode(const operations::unlock_request& request) -> v1::UnlockRequest
+{
+  v1::UnlockRequest proto;
+  set_location(proto, request.id);
+  proto.set_cas(request.cas.value());
+  return proto;
+}
+
+inline auto
+decode(const v1::UnlockResponse& /* proto */, key_value_error_context ctx)
+  -> operations::unlock_response
+{
+  operations::unlock_response response;
+  response.ctx = std::move(ctx);
+  return response;
+}
+
+inline auto
+encode(const operations::get_and_touch_request& request) -> v1::GetAndTouchRequest
+{
+  v1::GetAndTouchRequest proto;
+  set_location(proto, request.id);
+  set_expiry(proto, request.expiry);
+  return proto;
+}
+
+inline auto
+decode(const v1::GetAndTouchResponse& proto, key_value_error_context ctx)
+  -> operations::get_and_touch_response
+{
+  operations::get_and_touch_response response;
+  response.ctx = std::move(ctx);
+  if (content_is_compressed(proto)) {
+    response.ctx.override_ec(errc::common::feature_not_available);
+    return response;
+  }
+  response.value = utils::to_binary(proto.content_uncompressed());
+  response.cas = couchbase::cas{ proto.cas() };
+  response.flags = proto.content_flags();
+  return response;
+}
+
+template<typename Request, typename Proto>
+inline void
+set_counter_fields(Proto& proto, const Request& request)
+{
+  set_location(proto, request.id);
+  proto.set_delta(request.delta);
+  set_expiry(proto, request.expiry);
+  if (request.initial_value.has_value()) {
+    proto.set_initial(static_cast<std::int64_t>(request.initial_value.value()));
+  }
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+}
+
+template<typename Response, typename Proto>
+inline auto
+decode_counter(const Proto& proto, key_value_error_context ctx) -> Response
+{
+  Response response;
+  response.ctx = std::move(ctx);
+  response.content = static_cast<std::uint64_t>(proto.content());
+  response.cas = couchbase::cas{ proto.cas() };
+  if (proto.has_mutation_token()) {
+    response.token = to_core_token(proto.mutation_token());
+  }
+  return response;
+}
+
+inline auto
+encode(const operations::increment_request& request) -> v1::IncrementRequest
+{
+  v1::IncrementRequest proto;
+  set_counter_fields(proto, request);
+  return proto;
+}
+
+inline auto
+decode(const v1::IncrementResponse& proto, key_value_error_context ctx)
+  -> operations::increment_response
+{
+  return decode_counter<operations::increment_response>(proto, std::move(ctx));
+}
+
+inline auto
+encode(const operations::decrement_request& request) -> v1::DecrementRequest
+{
+  v1::DecrementRequest proto;
+  set_counter_fields(proto, request);
+  return proto;
+}
+
+inline auto
+decode(const v1::DecrementResponse& proto, key_value_error_context ctx)
+  -> operations::decrement_response
+{
+  return decode_counter<operations::decrement_response>(proto, std::move(ctx));
+}
+
+inline auto
+encode(const operations::append_request& request) -> v1::AppendRequest
+{
+  v1::AppendRequest proto;
+  set_location(proto, request.id);
+  proto.set_content(utils::to_string(request.value));
+  if (request.cas.value() != 0) {
+    proto.set_cas(request.cas.value());
+  }
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+decode(const v1::AppendResponse& proto, key_value_error_context ctx) -> operations::append_response
+{
+  return decode_mutation<operations::append_response>(proto, std::move(ctx));
+}
+
+inline auto
+encode(const operations::prepend_request& request) -> v1::PrependRequest
+{
+  v1::PrependRequest proto;
+  set_location(proto, request.id);
+  proto.set_content(utils::to_string(request.value));
+  if (request.cas.value() != 0) {
+    proto.set_cas(request.cas.value());
+  }
+  if (const auto level = to_proto_durability(request.durability_level); level) {
+    proto.set_durability_level(*level);
+  }
+  return proto;
+}
+
+inline auto
+decode(const v1::PrependResponse& proto, key_value_error_context ctx)
+  -> operations::prepend_response
+{
+  return decode_mutation<operations::prepend_response>(proto, std::move(ctx));
 }
 
 } // namespace couchbase::core::protostellar::kv
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -29,8 +29,11 @@
 #include <asio/post.hpp>
 
 #include <chrono>
+#include <cstdint>
+#include <limits>
 #include <memory>
 #include <string>
+#include <system_error>
 #include <utility>
 
 namespace couchbase::core::protostellar
@@ -112,6 +115,39 @@ component::execute(operations::get_request request,
       // rather than handing back an empty value with a success status).
       auto ctx = make_error_context(status, id, operation_kind::read_only);
       handler(kv::decode(resp, std::move(ctx)));
+    });
+}
+
+auto
+component::execute(operations::get_projected_request request,
+                   utils::movable_function<void(operations::get_projected_response)>&& handler)
+  -> pending_call
+{
+  auto proto = std::make_shared<v1::GetRequest>(kv::encode(request));
+  const auto id = request.id;
+  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired(io_, request, handler);
+  }
+  const auto auth = authorization_;
+  auto* stub = stubs_->kv.get();
+  return dispatcher_.unary<v1::GetResponse>(
+    timeout,
+    [stub, proto, auth](
+      grpc::ClientContext& ctx, v1::GetResponse& resp, std::function<void(grpc::Status)> cb) {
+      if (!auth.empty()) {
+        ctx.AddMetadata("authorization", auth);
+      }
+      stub->async()->Get(&ctx, proto.get(), &resp, std::move(cb));
+    },
+    [handler = std::move(handler), id, proto, request](grpc::Status status,
+                                                       v1::GetResponse resp) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      auto ctx = make_error_context(status, id, operation_kind::read_only);
+      if (status.ok() && resp.content_case() == v1::GetResponse::kContentCompressed) {
+        ctx = make_key_value_error_context(errc::common::feature_not_available, id);
+      }
+      handler(kv::decode(resp, std::move(ctx), request));
     });
 }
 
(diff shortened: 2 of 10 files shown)


core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,459 @@
...
+template<typename Request, typename Proto>
+inline void
+set_counter_fields(Proto& proto, const Request& request)
+{
+  set_location(proto, request.id);
+  proto.set_delta(request.delta);
+  set_expiry(proto, request.expiry);
+  if (request.initial_value.has_value()) {
+    proto.set_initial(static_cast<std::int64_t>(request.initial_value.value()));
+  }
```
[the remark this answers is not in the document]
avsej: initial is a counter's starting value, bounded well below INT64_MAX in practice, and the proto field is signed int64. Keeping as-is.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,488 @@
...
+inline auto
+decode(const v1::RemoveResponse& proto, key_value_error_context ctx) -> operations::remove_response
+{
+  return decode_mutation<operations::remove_response>(proto, std::move(ctx));
+}
+
+// ── Touch / Exists / lock / counters / append / prepend ────────────────────────
+
+inline auto
+encode(const operations::touch_request& request) -> v1::TouchRequest
```
avsej, line 318: `set_expiry()` returns without writing anything when `request.expiry == 0`, so `touch(id, 0)` — whose entire purpose is to clear the expiry — sends a `TouchRequest` with no expiry field and reports success without changing anything.

`TouchRequest` has no `preserve_expiry` field, so the expiry oneof is the only channel here; couchbase-jvm-clients sends its `NO_EXPIRY` constant for exactly this case (`CoreProtostellarKeyValueRequests.java:483`).

Scenario: a document with a 60-second TTL is touched with expiry 0 to make it permanent. Over `couchbase://` the TTL is cleared; over `couchbase2://` the document expires a minute later.

The fix is in `set_expiry` (see the review on #1030) — encoding the zero explicitly fixes `touch`, `get_and_touch` and the counter operations together. `get_and_touch` at the equivalent encoder has the same problem.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,488 @@
...
+
+template<typename Request, typename Proto>
+inline void
+set_counter_fields(Proto& proto, const Request& request)
+{
+  set_location(proto, request.id);
+  proto.set_delta(request.delta);
+  set_expiry(proto, request.expiry);
+  if (request.initial_value.has_value()) {
+    proto.set_initial(static_cast<std::int64_t>(request.initial_value.value()));
```
avsej, line 429: `initial_value` is `std::optional<std::uint64_t>` and the proto field is `int64`, so a value above `INT64_MAX` wraps negative and the gateway is asked to seed the counter with it. The classic path shares the representation mismatch, so this is not a couchbase2 regression — but the converter boundary is where it can be rejected rather than silently wrapped:

```cpp
  if (request.initial_value.has_value()) {
    const auto initial = request.initial_value.value();
    if (initial > static_cast<std::uint64_t>(std::numeric_limits<std::int64_t>::max())) {
      return {};  // surfaces as invalid_argument
    }
    proto.set_initial(static_cast<std::int64_t>(initial));
  }
```

avsej (commented):
## Review — PR #1034 (`CXXCBC-894`)

This commit completes the unary KV surface over couchbase2: `touch`, `exists`, `get_and_lock`, `unlock`,
`get_and_touch`, `increment`, `decrement`, `append` and `prepend`. The mapping is consistent with the
conventions established in #1030 and the operation list now matches what gocbcoreps exposes in
`impl_kv_v1.go`, with the exception noted below. `lock_time_secs` units are correct, and the append/prepend
CAS handling matches the reference clients.

The findings are all consequences of one shared helper rather than of this commit's own logic, but they first
become user-visible here, because this is the commit that adds the operations whose whole purpose is to
manipulate expiry.

### Strengths

- The operation list is essentially complete relative to gocbcoreps `impl_kv_v1.go` for the unary surface,
  and the additions reuse `set_location` / `to_proto_durability` / `set_expiry` rather than re-deriving them.
- `unlock` correctly sends the CAS as a required field rather than as an optional presence-guarded one, which
  is the right reading of the proto for that operation.
- Counter operations set `initial` only when the caller supplied one, matching couchbase-jvm-clients and
  couchbase-net-client. gocb always sends it, and gocb is the outlier here.

### Findings

|---|---|---|---|---|
| 1 | high | protocol-parity | `touch` and `get_and_touch` with expiry 0 send no expiry field, making the operation a no-op | `core/protostellar/kv_converter.hxx:291` |
| 2 | medium | api-surface | `lookup_in` and `mutate_in` are not routed, so all subdoc operations fail with `feature_not_available` | `core/protostellar/kv_converter.hxx:291` |
| 3 | low | correctness | The counter `initial` value is narrowed from `uint64` to `int64` without a range check | `core/protostellar/kv_converter.hxx:394` |

#### 1. `touch` and `get_and_touch` with expiry 0 send no expiry field (high)

Both encoders delegate to the shared `set_expiry()`, which returns without writing anything when
`request.expiry == 0`. For `upsert` and `replace` that produces the TTL-preservation defect reported on #1030.
For `touch` it is worse in kind: the entire purpose of `touch(id, 0)` is to clear the expiry and make the
document permanent, and with no expiry field on the wire the gateway has nothing to act on. The operation
becomes a no-op that reports success.

`TouchRequest` has no `preserve_expiry` field in the proto, so for touch the expiry oneof is the only channel;
couchbase-jvm-clients sends its `NO_EXPIRY` constant here (`CoreProtostellarKeyValueRequests.java:483`).

Failure scenario: a document is written with a 60-second TTL, then `collection.touch(id, 0s)` is called to make
it permanent. Over `couchbase://` the TTL is cleared. Over `couchbase2://` the gateway receives a
`TouchRequest` with no expiry, the call succeeds, and the document expires a minute later.

The fix belongs in `set_expiry` (see #1030) — encoding the zero explicitly fixes touch, get_and_touch and the
counters at once, and touch needs no `preserve_expiry` parameter because it has no such option.

(inline comment on `core/protostellar/kv_converter.hxx:291`)

#### 2. `lookup_in` and `mutate_in` are not routed (medium)

`component_supports_v` is specialised for the fifteen operations this commit completes, and subdoc is not among
them, so every `lookup_in` and `mutate_in` over `couchbase2://` returns `feature_not_available`. The proto
defines `LookupInRequest` and `MutateInRequest`, and all three reference clients implement them against those
RPCs — gocb in `kvprovider_ps.go:158-250`, which is also where its store-semantics and preserve-expiry
interaction is handled.

Failing cleanly is the right behaviour for an unimplemented operation, so this is not a defect in this commit.
It is worth stating in the pull request body and the ticket, though, because "complete the KV unary operation
surface" reads as though subdoc is included, and subdoc is a large part of what applications use KV for.

#### 3. The counter `initial` value is narrowed without a range check (low)

```cpp
proto.set_initial(static_cast<std::int64_t>(request.initial_value.value()));
```

`increment_request::initial_value` is `std::optional<std::uint64_t>` and the proto field is `int64`, so an
initial value above `INT64_MAX` wraps to a negative number and the gateway is asked to seed a counter with it.
The classic path has the same representation mismatch, so this is not a couchbase2 regression, but the
converter boundary is the natural place to reject it rather than silently wrap:

```cpp
if (request.initial_value.has_value()) {
  const auto initial = request.initial_value.value();
  if (initial > static_cast<std::uint64_t>(std::numeric_limits<std::int64_t>::max())) {
    return {};  // caller maps this to invalid_argument
  }
  proto.set_initial(static_cast<std::int64_t>(initial));
}
```

(inline comment on `core/protostellar/kv_converter.hxx:394`)

### Verdict

Needs changes. Finding 1 is the shared `set_expiry` defect, but it is worth calling out separately here because
`touch(id, 0)` silently doing nothing is a more direct contradiction of the operation's contract than the
upsert case, and this is the commit that makes it reachable.

avsej:
Two of the three findings needed a change here and both have one; the third is answered without a
code change, for the reason the review itself gives. The reviewed commit has since been rebased onto
`main`, which matters for finding 1: the shared helper it identifies was fixed in the meantime.

### 1. `touch` and `get_and_touch` with expiry 0 send no expiry field (high)

> Both encoders delegate to the shared `set_expiry()`, which returns without writing anything when
> `request.expiry == 0`.

Accurate as written. At the reviewed SHA `ca837941`, `set_expiry` opened with
`if (expiry == 0) { return; }`, so the described no-op was real.

> The fix belongs in `set_expiry` (see #1030) — encoding the zero explicitly fixes touch,
> get_and_touch and the counters at once, and touch needs no `preserve_expiry` parameter because it
> has no such option.

That is what happened, in the #1030 round, and it is now on `main`. The early return is gone; the
only remaining early return is for `preserve_expiry`, so expiry 0 reaches `set_expiry_secs(0)`.
Because `expiry_secs` sits inside `oneof expiry` in every one of these messages, setting it to zero
marks the oneof case and the field does travel — a bare proto3 scalar set to its default would not
have, so the oneof is doing load-bearing work here.

No code change is therefore made for this finding on this commit. What was genuinely missing was
coverage, and that is added:

- `touch` was already pinned, by `touch_always_sends_an_expiry_arm` in this commit, over all four
  expiry cases.
- `get_and_touch`, `increment` and `decrement` had **no expiry assertion at all**. A new case,
  `expiry_encodes_every_branch_for_get_and_touch_and_the_counters`, pins all three across the same
  four cases, asserting through the shared `assert_expiry` helper that the oneof is always set.

Negative-controlled: with `if (expiry == 0) { return; }` restored in `set_expiry` — a control that
was confirmed to compile before its result was read — the new case fails on
`get_and_touch expiry=0: the expiry oneof must be set unless preserve_expiry was requested`. So it is
a regression test for the reported defect, not merely new coverage.

The claim about the reference client was checked and holds: none of `GetAndTouchRequest`,
`IncrementRequest` or `DecrementRequest` has a `preserve_expiry` field in the proto, so for all of
them, as for `TouchRequest`, the oneof is the only channel.

### 2. `lookup_in` and `mutate_in` are not routed (medium)

> `component_supports_v` is specialised for the fifteen operations this commit completes, and subdoc
> is not among them, so every `lookup_in` and `mutate_in` over `couchbase2://` returns
> `feature_not_available`.

Confirmed: there is no occurrence of `lookup_in` or `mutate_in` anywhere under
`core/protostellar/`, so neither is routed and both take the unsupported-operation path.

> Failing cleanly is the right behaviour for an unimplemented operation, so this is not a defect in
> this commit. It is worth stating in the pull request body and the ticket, though, because "complete
> the KV unary operation surface" reads as though subdoc is included, and subdoc is a large part of
> what applications use KV for.

Agreed and no code change is made. The pull request body and `CXXCBC-894` are being updated to say
so explicitly, in the wording at the end of this comment, so that "complete the KV unary operation
surface" cannot be read as including subdoc.

### 3. The counter `initial` value is narrowed without a range check (low)

> `increment_request::initial_value` is `std::optional<std::uint64_t>` and the proto field is
> `int64`, so an initial value above `INT64_MAX` wraps to a negative number and the gateway is asked
> to seed a counter with it.

Confirmed, and fixed. `IncrementRequest.initial` is `optional int64` while `initial_value` is
`std::optional<std::uint64_t>`, so every value above `INT64_MAX` reached the gateway as a negative
seed.

> ```cpp
> if (request.initial_value.has_value()) {
>   const auto initial = request.initial_value.value();
>   if (initial > static_cast<std::uint64_t>(std::numeric_limits<std::int64_t>::max())) {
>     return {};  // caller maps this to invalid_argument
>   }
>   proto.set_initial(static_cast<std::int64_t>(initial));
> }
> ```

The fix is not placed there. The suggested shape does not hold at that location: the surrounding
`set_counter_fields` returns `void`, so `return {}` does not compile there, and in the enclosing
`encode` overload — which returns the request message by value — `return {}` would hand back a
*default-constructed request* rather than an error. That would send an empty `IncrementRequest`
instead of rejecting the call, which is worse than the wrap being reported.

The converter has no error channel to use. The rejection is therefore at the component boundary,
before anything is encoded, mirroring the existing pre-dispatch rejection for an exhausted time
budget:

```cpp
if (counter_initial_value_out_of_range(request)) {
  return fail_invalid_argument(io_, request, handler);
}
```

on both the `increment` and `decrement` overloads. The completion is delivered on the `io_context`
like every other completion rather than inline on the caller's thread, which is the contract the
sibling `fail_expired` path already honours.

Two cases were added:

- `a_counter_seed_above_int64_max_is_rejected_before_dispatch` asserts `invalid_argument` **and**
  `calls_received == 0`. The counter is what distinguishes the two stories: "refused locally" and
  "sent and rejected by the gateway" would both surface as `invalid_argument`.
- `counter_initial_value_is_optional_and_encodes_its_upper_bound` pins the boundary that must keep
  working — `INT64_MAX` encodes exactly, a seed of zero is still present on the wire, and an absent
  seed leaves the optional field absent. This one is **new coverage, not a regression test**: it
  passes against the pre-fix code.

Negative-controlled: with both guards removed — again confirming the control compiled first — the
rejection case fails on `a counter seed that cannot be represented is an invalid argument, not a
wrapped one`.

The observation that the classic path shares the representation mismatch was not treated as a reason
to leave it, and it is not being fixed on the classic path here.

### Verification

Every commit in the series was rebuilt and re-tested individually, not only the tip: configure,
build, `ctest -L cng`, and a `clang-format` check over each commit's own changed files. Both changes
land in this commit, which is the one that adds the operations concerned.

core/protostellar/kv_converter.hxx
avsej: Accurate as written: at the reviewed SHA `ca837941`, `set_expiry` opened with `if (expiry == 0) { return; }`, so the described no-op was real.

The fix landed where this comment says it belongs -- in the shared helper, during the #1030 round -- and is now on `main`. The only remaining early return is for `preserve_expiry`, so expiry 0 reaches `set_expiry_secs(0)`. That works only because `expiry_secs` sits inside `oneof expiry`: setting it to zero marks the oneof case and the field travels, where a bare proto3 scalar at its default would not.

No code change is therefore made here. What was missing was coverage, and not quite where this points: `touch` was **already** pinned by `touch_always_sends_an_expiry_arm` in this commit, while `get_and_touch`, `increment` and `decrement` had no expiry assertion at all. A new case covers those three over all four expiry branches.

Negative-controlled: with the early return restored -- a control confirmed to compile before its result was read -- the new case fails on `get_and_touch expiry=0: the expiry oneof must be set unless preserve_expiry was requested`.

The proto claim holds: none of `GetAndTouchRequest`, `IncrementRequest` or `DecrementRequest` has a `preserve_expiry` field, so for all of them the oneof is the only channel.

core/protostellar/kv_converter.hxx
avsej: Confirmed and fixed, though not at this location.

The suggested shape does not hold here: the surrounding `set_counter_fields` returns `void`, so `return {}` does not compile, and in the enclosing `encode` overload -- which returns the request message by value -- it would hand back a default-constructed request. That would send an empty `IncrementRequest` rather than reject the call, which is worse than the wrap being reported.

The converter has no error channel, so the rejection is at the component boundary before anything is encoded, mirroring the existing pre-dispatch rejection for an exhausted time budget, on both the `increment` and `decrement` overloads. The completion is delivered on the `io_context` like every other completion.

Two cases were added. The rejection case asserts `invalid_argument` **and** `calls_received == 0`, since "refused locally" and "sent and rejected by the gateway" are otherwise indistinguishable. A second case pins the boundary that must keep working: `INT64_MAX` encodes exactly, a seed of zero is still present on the wire, and an absent seed leaves the optional field absent. That second one is new coverage rather than a regression test -- it passes against the pre-fix code, and is labelled as such.

Negative-controlled: with both guards removed, the rejection case fails on `a counter seed that cannot be represented is an invalid argument, not a wrapped one`.

The observation that the classic path shares the representation mismatch was not treated as a reason to leave it; it is also not being fixed on the classic path here.

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. This was a static review — I traced the code paths by hand, nothing was compiled or run; please treat the reasoning as a starting point for discussion, not ground truth.

Reviewed the incremental commit only (`6abbf26f`, +908/-0 vs #1033): the nine new KV ops (`touch`/`exists`/`get_and_lock`/`unlock`/`get_and_touch`/`increment`/`decrement`/`append`/`prepend`) plus `get_projected`, wired through `kv_converter.hxx` and `component.{hxx,cxx}`.

Everything here follows the `execute()` pattern established in #1031/#1033 exactly — encode, timeout check, gRPC dispatch, decode+error-context — and nothing needed to change in `cluster.cxx`'s routing, which is a nice confirmation that #1033's `component_supports_v` funnel design is paying off. The `counter_initial_value_out_of_range` guard (rejecting a uint64 seed that can't fit in the proto's int64 `initial` field before it's ever encoded) is a good catch, and it's well tested.

**Question on test coverage:** of these nine ops, only `increment`'s pre-dispatch rejection and `get_projected` get an in-process `component_test` (the ones using `in_process_server`/the fake KV service). `touch`, `exists`, `get_and_lock`, `unlock`, `get_and_touch`, `decrement`, `append`, and `prepend` are only exercised at the pure encode/decode level in `kv_converter.cxx` — nothing drives them through `component::execute()`'s actual gRPC dispatch and error-context mapping in CI. In particular, `get_and_lock`/`get_and_touch` each got their own compressed-content fail-closed check copied from `get`'s (`resp.content_case() == v1::GetAndLockResponse::kContentCompressed`, etc.) — correct as written, but nothing would catch it if a future edit swapped in the wrong response type there.

The live `cluster_only` round-trips mentioned in the PR description cover `increment`/`exists`/`remove`, so most of the gap is `touch`/`get_and_lock`/`unlock`/`get_and_touch`/`decrement`/`append`/`prepend` specifically. Worth adding `component_test` cases (or extending the live round-trip) for those, similar to what `get`/`upsert`/`insert`/`replace`/`remove` already have?

thejcfactor (approved):

avsej:
> **Question on test coverage:** of these nine ops, only `increment`'s pre-dispatch rejection and `get_projected` get an in-process `component_test` (the ones using `in_process_server`/the fake KV service). `touch`, `exists`, `get_and_lock`, `unlock`, `get_and_touch`, `decrement`, `append`, and `prepend` are only exercised at the pure encode/decode level in `kv_converter.cxx` — nothing drives them through `component::execute()`'s actual gRPC dispatch and error-context mapping in CI.

Accepted, and it turned up a real defect. Two suites were added: `component_kv_operations.cxx`, 38 in-process cases driving all nine operations through `component::execute()`, and `live_kv_operations.cxx`, 9 `cluster_only` cases. Both were run; results below.

The expectations were not taken from this implementation, nor from another SDK's tests. They were transcribed from the gateway, because that is what the client actually has to agree with. `stellar-gateway/docs/ERRORS.md` states that "All errors produced by CNG are centrally located within two files", one of them `gateway/dataimpl/server_v1/errorhandler.go` — so that file defines every status, and `gateway/dataimpl/server_v1/kvserver.go` defines which of them each RPC can produce. Extracting the per-operation sets from `kvserver.go` gives an exact target: each operation is now driven once per status its own RPC can return, no more (which would test fiction) and no less.

**The defect.** `NewDocNotNumericStatus` is emitted only by `Increment` and `Decrement` — the two operations this commit adds — as `FAILED_PRECONDITION` with a `PreconditionFailure` violation of type `DOC_NOT_NUMERIC`. `map_precondition` in `error_utils.cxx` did not handle it, so incrementing a non-numeric document reported **`internal_server_failure`**: the SDK telling the user their cluster is broken when their document simply is not a counter. Confirmed against a real gateway before fixing:

```
"incrementing_a_non_numeric_document_is_reported_as_delta_invalid" FAILED:
  a counter operation on a non-numeric document is delta_invalid, got: internal_server_failure (5)
```

It now maps to `delta_invalid`. The same audit found `DURABILITY_IMPOSSIBLE` unmapped, also falling through to `internal_server_failure`; it now maps to `durability_impossible`. `POISONED_CAS` is left falling through deliberately, with a comment saying why — the SDK has no error code for it and inventing one would be a guess.

**The contested corners**, each pinned by a case that cites its source in a comment rather than asserting a bare value:

- *Unlock with a stale CAS is `cas_mismatch`, not `document_locked`.* The classic transport makes this opcode-dependent — `core/protocol/status.cxx` maps status `locked` to `cas_mismatch` when the opcode is `unlock`. The gateway arrives at the same answer differently: `kvserver.go` `Unlock` translates `memdx.ErrCasMismatch` into `NewDocCasMismatchStatus` and never emits `DocLocked` at all. Both transports agree, and the case asserts the positive and the negative.
- *A locked document is reported by the component, not absorbed by it.* This is the one that looks wrong at first glance. Classic does not surface a locked document at all — `core/bucket.cxx` and `core/io/mcbp_command.hxx` raise `retry_reason::key_value_locked`, which `core/impl/retry_reason.cxx` lists under both `always_retry` and `allows_non_idempotent_retry`, so it is retried until the lock or the deadline clears. couchbase2 gets the same behaviour, but one layer up: `cluster.cxx` turns `document_locked` into `retry_reason::key_value_locked` via `protostellar::retry_reason_for` (added in #1043). Retrying inside the component would put a second, invisible retry loop underneath the strategy. The case asserts the contract between the two layers and says so.
- *`exists` cannot fail with "not found".* `Exists` is the only one of the nine whose handler set in `kvserver.go` contains no `NewDocMissingStatus`, which is the server-side proof of the contract — a missing document is a successful `false`. Verified live: upsert, exists → true, remove, exists → false with no error.
- *A CAS mismatch travels as `ErrorInfo`, every other contested KV error as `PreconditionFailure`.* `NewDocCasMismatchStatus` is the one place the gateway uses that detail type for something the client must distinguish. The mapping is correct here by accident — `ABORTED` alone suffices — so there is now a case making the accident deliberate.
- *An unrecognised server error arrives as `request_canceled`.* `NewGenericStatus` is the gateway's fallback and its `context.Canceled` arm returns `codes.Canceled`. Surprising enough to be worth pinning.
- *`NewSyncWriteAmbiguousStatus` attaches no details at all*, so the bare `DEADLINE_EXCEEDED` is the only signal: ambiguous for a mutation, unambiguous for a read, since a read cannot have changed anything.

**On the hazard you named specifically** — a compressed-content guard left keyed on the wrong response type — that one cannot happen. Applying the mutation gives:

```
component.cxx:360: error: comparison between
  'enum couchbase::kv::v1::GetAndLockResponse::ContentCase' and
  'enum couchbase::kv::v1::GetAndTouchResponse::ContentCase' [-Werror=enum-compare]
```

The generated `ContentCase` enums are distinct types and `-Wenum-compare` under `-Werror` rejects the mismatch; dispatching to the wrong stub likewise does not compile. Worth noting the rule is nonetheless enforced **twice** — in `component.cxx` before decode and again in `kv_converter.hxx` inside it — and disabling either alone leaves the tests green. Defensible as defence in depth, but duplicated, and the `component.cxx` copy is the fragile one since it names the type explicitly while the converter's is a template that deduces it.

**Verification.** Both suites were run against a real Cloud Native Gateway (server `8.1.0-1743`, CNG `1.2.2-123`, operator `2.8.0`, via cbdinocluster's `cao` deployer on k3d): `ctest -L cng` reports **14/14 passing with `TEST_CONNECTION_STRING` set**, so the `cluster_only` cases genuinely executed rather than skipping. The in-process suite is 38/38.

The suites were also checked for the ability to fail, since a green suite that cannot go red proves nothing. Deliberate defects were injected one at a time: disabling the two new mappings turns exactly four cases red, each naming the right gateway status; five of six earlier mutations were caught and the sixth does not compile. One of those runs found a genuine weakness in a case of mine — it derived CAS presence from `cas() != 0`, which cannot distinguish "field absent" from "field set to zero" — now fixed to read `has_cas()`.

**On the two expiry cases**, which are the only ones with any timing in them. They do not sleep for a fixed period and hope. Each sets a one-second expiry and then polls until the document becomes unreadable, giving up after twenty seconds — so a slow or loaded host makes the case take longer but cannot make it fail, and the case still fails when expiry genuinely does not work. Nothing asserts on elapsed time, only on the state transition.

Two further defences, because "the document is gone" is a claim that a broken fixture can satisfy accidentally. Each case first asserts the document is readable *before* the expiry is applied, so the disappearance is attributable to the touch rather than to a seed that silently failed; and each seeds a second document that is never given an expiry and asserts it is *still* readable afterwards, which separates "the expiry worked" from "everything vanished" — a dropped collection or flushed bucket would otherwise make these pass for the wrong reason. The twenty-second budget is held below the harness's own thirty-second case timeout by a `static_assert`, so that a failure reports the assertion instead of a bare timeout; that assert was confirmed to fire by raising the deadline and watching the build stop.

The de-flaked versions are also faster than the fixed sleeps they replaced, at roughly 1.5 s and 2.1 s against a real gateway.

One genuine limit remains, stated rather than implied: the live cases were exercised against the build-number images named above rather than the release versions CI pins.

PR #1036, opened 2026-07-28
CXXCBC-897: route N1QL query over the couchbase2 streaming transport

avsej:
Query was rejected with feature_not_available on couchbase2: the HTTP execute
front-door gate covered every HTTP-shaped request, and nothing bridged the core
query_request onto the gRPC path. Wire it onto the server-streaming primitive.

Add a query converter (core operations <-> couchbase.query.v1): encode statement,
parameters, query_context (parsed into bucket/scope), scan consistency,
mutation-token scan vectors (consistent_with), profile, and tuning options;
decode the terminal metadata message (status, request_id, metrics, warnings).
Rows arrive across streamed messages and are buffered into query_response.rows.
The component gains a QueryService stub and a query execute() that drives
dispatcher::server_stream, mapping the terminal gRPC status to the error
context. cluster_impl routes query_request to the component; other HTTP services
keep returning feature_not_available.

Query features with no couchbase2 equivalent (raw passthrough, use_replica, the
streaming row_callback) are reported up front as feature_not_available rather
than silently dropped. Covered by a converter unit test and a live N1QL
round-trip against a CNG gateway.

---
Stacked PR **14/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–14; the change specific to this PR is its top commit. Depends on #1035 — merge the series bottom-up.

### DIFF
--- a/core/protostellar/query_converter.hxx
+++ b/core/protostellar/query_converter.hxx
@@ -1,0 +1,264 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates the core N1QL query request/response to and from the couchbase.query.v1 protobuf
+// messages. Rows arrive across streamed QueryResponse messages and are buffered into
+// query_response::rows; the terminal message carries the metadata. Deliberately does not reuse the
+// MCBP-bound encode_to/make_response (those build an HTTP body + parse a JSON reply).
+//
+// Query features with no couchbase2 equivalent are reported by can_encode() so the component can
+// surface feature_not_available rather than silently dropping them: the `raw` passthrough map,
+// use_replica, the streaming row_callback (only the buffered result path is wired), node targeting
+// via send_to_node, and tuning values too large for the proto's uint32 fields.
+
+#include "core/json_string.hxx"
+#include "core/operations/document_query.hxx"
+#include "core/protostellar/json_payload.hxx"
+#include "core/protostellar/query_proto.hxx"
+
+#include <couchbase/query_profile.hxx>
+#include <couchbase/query_scan_consistency.hxx>
+
+#include <chrono>
+#include <cstdint>
+#include <limits>
+#include <optional>
+#include <string>
+
+namespace couchbase::core::protostellar::query
+{
+namespace v1 = ::couchbase::query::v1;
+
+// The proto's tuning fields are uint32 while the core request holds them as uint64, so a value
+// that does not fit is not encodable: casting it would ask the gateway to use the wrapped number.
+[[nodiscard]] inline auto
+fits_uint32(const std::optional<std::uint64_t>& value) -> bool
+{
+  return !value.has_value() || *value <= std::numeric_limits<std::uint32_t>::max();
+}
+
+// Query request features the couchbase2 gateway schema cannot express yet. The component maps this
+// to feature_not_available instead of silently ignoring the caller's intent.
+[[nodiscard]] inline auto
+can_encode(const operations::query_request& request) -> bool
+{
+  return request.raw.empty() && !request.use_replica.value_or(false) &&
+         !request.row_callback.has_value() && !request.send_to_node.has_value() &&
+         fits_uint32(request.max_parallelism) && fits_uint32(request.pipeline_batch) &&
+         fits_uint32(request.pipeline_cap) && fits_uint32(request.scan_cap);
+}
+
+// Split the SDK query_context ("default:`bucket`.`scope`") into the proto's bucket_name/scope_name.
+// Any pair of backtick-delimited identifiers is accepted; a value that does not parse leaves both
+// unset (the statement is then expected to be fully qualified).
+inline void
+apply_query_context(const std::string& query_context, v1::QueryRequest& proto)
+{
+  std::string bucket;
+  std::string scope;
+  std::string* current = &bucket;
+  bool inside = false;
+  int identifiers = 0;
+  for (const char c : query_context) {
+    if (c == '`') {
+      if (inside) { // closing backtick ends an identifier
+        ++identifiers;
+        if (identifiers == 1) {
+          current = &scope;
+        }
+      }
+      inside = !inside;
+      continue;
+    }
+    if (inside) {
+      current->push_back(c);
+    }
+  }
+  // A well-formed context is exactly two balanced backtick-delimited identifiers
+  // (`bucket`.`scope`). Anything else -- unbalanced, too few, or too many -- leaves both unset
+  // rather than routing on a half- or mis-parsed context.
+  if (!inside && identifiers == 2 && !bucket.empty() && !scope.empty()) {
+    proto.set_bucket_name(bucket);
+    proto.set_scope_name(scope);
+  }
+}
+
+inline auto
+encode(const operations::query_request& request) -> v1::QueryRequest
+{
+  v1::QueryRequest proto;
+  proto.set_statement(request.statement);
+  if (request.client_context_id.has_value()) {
+    proto.set_client_context_id(*request.client_context_id);
+  }
+  if (request.query_context.has_value()) {
+    apply_query_context(*request.query_context, proto);
+  }
+  proto.set_read_only(request.readonly);
+  proto.set_flex_index(request.flex_index);
+  proto.set_preserve_expiry(request.preserve_expiry);
+  // adhoc=false means "use a prepared statement" on the MCBP path; the gateway prepares/caches on
+  // its side, so we only signal intent.
+  if (!request.adhoc) {
+    proto.set_prepared(true);
+  }
+
+  if (request.scan_consistency.has_value()) {
+    switch (*request.scan_consistency) {
+      case query_scan_consistency::not_bounded:
+        proto.set_scan_consistency(v1::QueryRequest_ScanConsistency_SCAN_CONSISTENCY_NOT_BOUNDED);
+        break;
+      case query_scan_consistency::request_plus:
+        proto.set_scan_consistency(v1::QueryRequest_ScanConsistency_SCAN_CONSISTENCY_REQUEST_PLUS);
+        break;
+    }
+  }
+  // A non-empty mutation_state carries at_plus scan vectors; the gateway takes them as
+  // consistent_with mutation tokens.
+  for (const auto& token : request.mutation_state) {
+    auto* proto_token = proto.add_consistent_with();
+    proto_token->set_bucket_name(token.bucket_name());
+    proto_token->set_vbucket_id(token.partition_id());
+    proto_token->set_vbucket_uuid(token.partition_uuid());
+    proto_token->set_seq_no(token.sequence_number());
+  }
+
+  if (request.profile.has_value()) {
+    switch (*request.profile) {
+      case query_profile::off:
+        proto.set_profile_mode(v1::QueryRequest_ProfileMode_PROFILE_MODE_OFF);
+        break;
+      case query_profile::phases:
+        proto.set_profile_mode(v1::QueryRequest_ProfileMode_PROFILE_MODE_PHASES);
+        break;
+      case query_profile::timings:
+        proto.set_profile_mode(v1::QueryRequest_ProfileMode_PROFILE_MODE_TIMINGS);
+        break;
+    }
+  }
+
+  for (const auto& parameter : request.positional_parameters) {
+    proto.add_positional_parameters(json_payload(parameter));
+  }
+  for (const auto& [name, value] : request.named_parameters) {
+    (*proto.mutable_named_parameters())[name] = json_payload(value);
+  }
+
+  // Tuning options: the SDK's `metrics` flag defaults to off, so disable_metrics is its inverse.
+  auto* tuning = proto.mutable_tuning_options();
+  tuning->set_disable_metrics(!request.metrics);
+  if (request.max_parallelism.has_value()) {
+    tuning->set_max_parallelism(static_cast<std::uint32_t>(*request.max_parallelism));
+  }
+  if (request.pipeline_batch.has_value()) {
+    tuning->set_pipeline_batch(static_cast<std::uint32_t>(*request.pipeline_batch));
+  }
+  if (request.pipeline_cap.has_value()) {
+    tuning->set_pipeline_cap(static_cast<std::uint32_t>(*request.pipeline_cap));
+  }
+  if (request.scan_cap.has_value()) {
+    tuning->set_scan_cap(static_cast<std::uint32_t>(*request.scan_cap));
+  }
+  if (request.scan_wait.has_value()) {
+    const auto secs = std::chrono::duration_cast<std::chrono::seconds>(*request.scan_wait);
+    const auto nanos =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(*request.scan_wait - secs);
+    auto* scan_wait = tuning->mutable_scan_wait();
+    scan_wait->set_seconds(secs.count());
+    scan_wait->set_nanos(static_cast<std::int32_t>(nanos.count()));
+  }
+
+  return proto;
+}
+
+// google.protobuf.Duration -> nanoseconds.
+[[nodiscard]] inline auto
+to_nanoseconds(const google::protobuf::Duration& duration) -> std::chrono::nanoseconds
+{
+  return std::chrono::seconds{ duration.seconds() } + std::chrono::nanoseconds{ duration.nanos() };
+}
+
+[[nodiscard]] inline auto
+status_to_string(v1::QueryResponse_MetaData_Status status) -> std::string
+{
+  switch (status) {
+    case v1::QueryResponse_MetaData_Status_STATUS_RUNNING:
+      return "running";
+    case v1::QueryResponse_MetaData_Status_STATUS_SUCCESS:
+      return "success";
+    case v1::QueryResponse_MetaData_Status_STATUS_ERRORS:
+      return "errors";
+    case v1::QueryResponse_MetaData_Status_STATUS_COMPLETED:
+      return "completed";
+    case v1::QueryResponse_MetaData_Status_STATUS_STOPPED:
+      return "stopped";
+    case v1::QueryResponse_MetaData_Status_STATUS_TIMEOUT:
+      return "timeout";
+    case v1::QueryResponse_MetaData_Status_STATUS_CLOSED:
+      return "closed";
+    case v1::QueryResponse_MetaData_Status_STATUS_FATAL:
+      return "fatal";
+    case v1::QueryResponse_MetaData_Status_STATUS_ABORTED:
+      return "aborted";
+    default:
+      return "unknown";
+  }
+}
+
+// Fill a query_meta_data from the terminal QueryResponse.MetaData message.
+inline void
+decode_meta_data(const v1::QueryResponse_MetaData& proto,
+                 operations::query_response::query_meta_data& meta)
+{
+  meta.request_id = proto.request_id();
+  meta.client_context_id = proto.client_context_id();
+  meta.status = status_to_string(proto.status());
+  if (!proto.signature().empty()) {
+    meta.signature = proto.signature();
+  }
+  if (!proto.profile().empty()) {
+    meta.profile = proto.profile();
+  }
+  if (proto.has_metrics()) {
+    const auto& m = proto.metrics();
+    operations::query_response::query_metrics metrics;
+    metrics.elapsed_time = to_nanoseconds(m.elapsed_time());
+    metrics.execution_time = to_nanoseconds(m.execution_time());
+    metrics.result_count = m.result_count();
+    metrics.result_size = m.result_size();
+    metrics.sort_count = m.sort_count();
+    metrics.mutation_count = m.mutation_count();
+    metrics.error_count = m.error_count();
+    metrics.warning_count = m.warning_count();
+    meta.metrics = metrics;
+  }
+  if (proto.warnings_size() > 0) {
+    std::vector<operations::query_response::query_problem> warnings;
+    warnings.reserve(static_cast<std::size_t>(proto.warnings_size()));
+    for (const auto& w : proto.warnings()) {
+      operations::query_response::query_problem problem;
+      problem.code = w.code();
+      problem.message = w.message();
+      warnings.push_back(std::move(problem));
+    }
+    meta.warnings = std::move(warnings);
+  }
+}
+
+} // namespace couchbase::core::protostellar::query
--- a/test/cng/protostellar/live_query.cxx
+++ b/test/cng/protostellar/live_query.cxx
@@ -1,0 +1,454 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Live end-to-end tests of N1QL query over the couchbase2:// streaming transport (CXXCBC-897).
+//
+// Each case drives the core::cluster public path -- open() builds the gRPC component,
+// execute(query_request) routes to the component's server-streaming query path -- against a real
+// CNG gateway. cluster_only: the whole suite skips unless TEST_CONNECTION_STRING points at a
+// couchbase2:// endpoint.
+//
+// Every expectation is taken from the gateway's own sources rather than from this client, so a case
+// asserts the answer the gateway defines rather than the behaviour this client happens to have. The
+// error-condition cases in particular are the gateway's acceptance suite
+// (stellar-gateway/gateway/test/query_test.go) re-expressed at this SDK's surface.
+
+#define COUCHBASE_CXX_CLIENT_IGNORE_CORE_DEPRECATIONS
+
+#include "framework/live_fixture.hxx"
+#include "framework/test_runner.hxx"
+
+#include "core/json_string.hxx"
+
+#include <couchbase/error_codes.hxx>
+
+#include <chrono>
+#include <cstddef>
+#include <string>
+#include <thread>
+#include <vector>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+namespace ops = ::couchbase::core::operations;
+using namespace std::chrono_literals;
+
+// Query parameters reach the core request as codec::binary, which selects json_string's binary
+// arm. Building the string arm here instead would exercise a path core/impl/query.cxx never
+// produces -- and would have hidden the empty-parameter defect this suite exists to catch.
+[[nodiscard]] auto
+binary_json(const std::string& json) -> couchbase::core::json_string
+{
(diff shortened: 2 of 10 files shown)


core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,288 @@
...
+  {
+    // Deregister on the gRPC thread before posting: once this returns gRPC is done with the call,
+    // so a concurrent cancel_and_drain() may proceed. The posted task only touches shared state.
+    tracker_->remove(context_.get());
+    // Copy the status into the closure -- it is a reference parameter, so the posted task must own
+    // its own copy rather than capture a reference that dangles once OnDone returns.
+    asio::post(io_, [this, status_copy = status]() {
+      on_done_(status_copy);
+      delete this;
+    });
```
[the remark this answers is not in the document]
avsej: Intentional: remove() must precede post() here precisely so a streaming reactor's self-delete (run by the posted task) can never race cancel_and_drain's TryCancel on an already-freed ClientContext. The io_context outlives the dispatcher and its run loop resumes after cancel_and_drain, so the CANCELLED completion is still delivered. No use-after-free.

core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,296 @@
...
+      context->set_deadline(std::chrono::system_clock::now() + timeout);
+    }
+    // Register before launch so shutdown can cancel it; the reactor deregisters in OnDone.
+    tracker_->add(context.get());
+    // Owned by gRPC for the duration of the call; deletes itself in OnDone.
+    auto* reactor = new streaming_reactor<Response>(
+      io_, tracker_, context, std::move(on_row), std::move(on_done));
+    std::forward<Invoker>(invoker)(*context, reactor);
+    reactor->begin();
+    return pending_call{ context };
```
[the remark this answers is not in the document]
avsej: Addressed: server_stream() now wraps invoker()+begin() in try/catch; on an exception it removes the tracker entry, deletes the heap-allocated reactor, and rethrows. A failed launch therefore neither leaks the reactor nor leaves an entry that would block cancel_and_drain() at shutdown -- matching the guard unary() already has.

core/protostellar/query_converter.hxx
```diff
@@ -0,0 +1,250 @@
...
+#include "core/operations/document_query.hxx"
+
+#include <couchbase/query_profile.hxx>
+#include <couchbase/query_scan_consistency.hxx>
+
+#include <couchbase/query/v1/query.pb.h>
+
+#include <chrono>
+#include <cstdint>
+#include <string>
```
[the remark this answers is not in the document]
avsej: query_converter.hxx does not actually use std::numeric_limits -- it has no range checks on the tuning fields, so no <limits> include is required. The proto tuning fields (max_parallelism, pipeline_batch/pipeline_cap, scan_cap) are uint32 by protocol design and are bounded to small values in practice, so encode() casts them directly. This differs from view_converter, where limit/skip are genuinely large result-set values -- there I did add a numeric_limits truncation gate plus the <limits> include.

core/protostellar/query_converter.hxx
```diff
@@ -0,0 +1,250 @@
...
+        proto.set_profile_mode(v1::QueryRequest_ProfileMode_PROFILE_MODE_PHASES);
+        break;
+      case query_profile::timings:
+        proto.set_profile_mode(v1::QueryRequest_ProfileMode_PROFILE_MODE_TIMINGS);
+        break;
+    }
+  }
+
+  for (const auto& parameter : request.positional_parameters) {
+    proto.add_positional_parameters(parameter.str());
```
avsej: **Blocker: every query parameter is sent as an empty string.**

`core::json_string` is a `std::variant<std::nullptr_t, std::string, std::vector<std::byte>>`, and `str()` returns a reference to a **static empty string** unless the variant holds the `std::string` arm (`core/json_string.hxx:47-53`).

Query parameters never hold that arm: `couchbase/query_options.hxx:70` declares `std::vector<codec::binary> positional_parameters`, and `core/impl/query.cxx:204` moves each `codec::binary` in, constructing the json_string from the **binary** arm. So this and the named-parameter line below both write `""`.

Scenario: `cluster.query("SELECT * FROM b WHERE id = $1", query_options{}.positional_parameters("abc"))` executes with `$1 = "abc"` over `couchbase://` and with an empty `$1` over `couchbase2://` — wrong rows, or a parse error blamed on the user's query. Every parameterised query is affected.

The classic path handles the variant via `utils::json::parse(value)` (`core/operations/document_query.cxx:47`). Reading the binary arm directly avoids a parse-and-regenerate round trip:

```cpp
// json_string holds either a std::string or the raw bytes the public API produced; str() yields
// an empty string for the binary arm, which is the arm query parameters always use.
[[nodiscard]] auto
json_payload(const core::json_string& value) -> std::string
{
  if (value.is_binary()) {
    const auto& bytes = value.bytes();
    return std::string{ reinterpret_cast<const char*>(bytes.data()), bytes.size() };
  }
  return value.str();
}
```

Worth auditing every other `json_string` read in the couchbase2 converters for the same pattern — the raw tuning passthrough here, and the analytics converter added in #1037.

Two further items on this commit, noted here rather than as separate anchors:

- This adds a fifth positional constructor parameter to `component` without updating `test/cng/protostellar/live_kv.cxx:100`/`:183`, so `cng_live_kv_test` stops compiling with `CB_COUCHBASE2=1`. The `build` check passes on #1035 and fails from here upward.
- The completion maps `DEADLINE_EXCEEDED` to `ambiguous_timeout` unconditionally. `request.readonly` is known at this call site, and RFC 77 keys the distinction on read-only status — a timed-out `SELECT` should be `unambiguous_timeout`.

avsej (commented):
## Review — PR #1036 (`CXXCBC-897`)

This commit routes N1QL query over the streaming transport added in #1035. The converter covers most of the
request surface, the metadata decode handles the terminal message, and the `can_encode` gate keeps unmappable
request shapes from being silently truncated. The streaming integration is also where this series does something
none of the reference SDKs do: the completion handler converts a mid-stream failure into `request_canceled`
rather than letting it be retried, which is RFC 77's rule and is recorded as unimplemented in gocb,
couchbase-jvm-clients and couchbase-net-client (CNG-16).

Two blockers. Every parameterised query sends empty parameters, and this is the commit where the series' build
goes red.

### Strengths

- The mid-stream error rule from RFC 77's Streaming section is implemented, which is ahead of all three
  reference clients.
- `can_encode` gating rather than silent field-dropping is the right convention, and it is applied consistently
  with the KV converters.
- The metadata decode handles the metrics `Duration` fields with an explicit seconds-plus-nanoseconds
  reconstruction rather than a lossy cast.

### Findings

|---|---|---|---|---|
| 1 | blocker | correctness | Every positional and named query parameter is sent to the gateway as an empty string | `core/protostellar/query_converter.hxx:143` |
| 2 | blocker | build | A fifth required constructor parameter is added without updating the existing call sites | `core/protostellar/query_converter.hxx:143` |
| 3 | high | spec-compliance | A timed-out read-only query reports `ambiguous_timeout` | `core/protostellar/query_converter.hxx:143` |
| 4 | medium | correctness | Tuning options are narrowed from 64-bit to `uint32` without a range check | `core/protostellar/query_converter.hxx:153` |
| 5 | medium | portability | `live_query.cxx` uses raw `std::getenv` instead of the framework's wrapper | `test/cng/protostellar/live_query.cxx:58` |

#### 1. Every query parameter is sent as an empty string (blocker)

```cpp
for (const auto& parameter : request.positional_parameters) {
  proto.add_positional_parameters(parameter.str());
}
for (const auto& [name, value] : request.named_parameters) {
  (*proto.mutable_named_parameters())[name] = value.str();
}
```

`core::json_string` is a `std::variant<std::nullptr_t, std::string, std::vector<std::byte>>`, and `str()`
returns a reference to a **static empty string** unless the variant currently holds the `std::string` arm
(`core/json_string.hxx:47-53`).

Query parameters never hold that arm. `couchbase/query_options.hxx:70` declares
`std::vector<codec::binary> positional_parameters`, and `core/impl/query.cxx:204` moves each `codec::binary`
into the request, which constructs the `json_string` from the **binary** arm. So `str()` returns `""` for every
parameter, positional and named alike.

Failure scenario: `cluster.query("SELECT * FROM b WHERE id = $1", query_options{}.positional_parameters("abc"))`.
Over `couchbase://` the statement executes with `$1 = "abc"`. Over `couchbase2://` the gateway receives
`positional_parameters: [""]` and executes the statement with an empty parameter — returning the wrong rows, or a
parse error attributed to the user's query. Every parameterised query is affected, which is most application
queries.

The classic path handles the variant correctly by going through `utils::json::parse(value)`
(`core/operations/document_query.cxx:47`), which has a `json_string` overload. The converter needs the same
awareness; reading the binary arm directly avoids a parse-and-regenerate round trip:

```cpp
namespace
{
// json_string holds either a std::string or the raw bytes the public API produced; str() yields an
// empty string for the binary arm, which is the arm query parameters always use.
[[nodiscard]] auto
json_payload(const core::json_string& value) -> std::string
{
  if (value.is_binary()) {
    const auto& bytes = value.bytes();
    return std::string{ reinterpret_cast<const char*>(bytes.data()), bytes.size() };
  }
  return value.str();
}
} // namespace
```

The same `str()` pattern should be checked anywhere else in the couchbase2 converters that reads a
`json_string` — the raw tuning passthrough and the analytics converter added in #1037 among them.

(inline comment on `core/protostellar/query_converter.hxx:143`)

#### 2. A fifth required constructor parameter without updating the call sites (blocker)

This commit adds `std::chrono::milliseconds default_query_timeout` to `component::component`, making it five
positional parameters, and does not update `test/cng/protostellar/live_kv.cxx:100` and `:183`, which pass four.
With `CB_COUCHBASE2=1` — the CI configuration — `cng_live_kv_test` fails to compile, and the `build` check passes
on #1034 and #1035 and fails here and on every pull request above.

This is the first of three commits to break that same untouched call site: #1038 later replaces the timeout
parameters with the `component_timeouts` aggregate, at which point `20000ms` stops converting even though the
arity returns to four, and #1047 adds a fifth parameter again. The series has been red for twelve consecutive
pull requests as a result.

The immediate fix is to update both call sites in this commit. The design fix is on the constructor: five
positional parameters with no defaults make every addition a silent breaking change, and a configuration struct
with named fields removes the class of failure.

#### 3. A timed-out read-only query reports `ambiguous_timeout` (high)

The completion handler builds the error from `map_status(status)`, which maps `DEADLINE_EXCEEDED` to
`errc::common::ambiguous_timeout` unconditionally. RFC 77's error table keys that on read-only status, and notes
it must not be based on idempotency; couchbase-jvm-clients uses `request.readonly()` and gocb a `readOnly` flag.
A `SELECT` that times out should be `unambiguous_timeout` — it provably changed nothing.

The information is available here: `request.readonly` is known at the point the completion is built, alongside
`statement` and `client_context_id`. Capturing it and post-correcting the mapping is a two-line change, and the
same correction applies to the analytics path in #1037.

#### 4. Tuning options are narrowed without a range check (medium)

```cpp
tuning->set_max_parallelism(static_cast<std::uint32_t>(*request.max_parallelism));
```

`max_parallelism`, `pipeline_batch`, `pipeline_cap` and `scan_cap` are 64-bit in the core request and `uint32` in
the proto, so a value above `UINT32_MAX` wraps silently and the gateway is asked to use the wrapped value. The
classic path sends the full value. Either reject out-of-range values in `can_encode` — which is where this
converter puts its other "cannot express this" decisions — or clamp explicitly with a comment.

(inline comment on `core/protostellar/query_converter.hxx:153`)

#### 5. `live_query.cxx` uses raw `std::getenv` (medium)

```cpp
const char* value = std::getenv(name); // NOLINT(concurrency-mt-unsafe)
```

The framework provides a wrapper for exactly this, and `live_connect.cxx` uses it. On MSVC, `std::getenv` is
deprecated and the Windows CI legs build with warnings as errors, so a hand-rolled helper here is a portability
hazard as well as a duplication — and the `NOLINT` suppresses the diagnostic that would have pointed at it.
Using the framework helper keeps the twelve live-test prologues consistent.

(inline comment on `test/cng/protostellar/live_query.cxx:58`)

### Verdict

Blocked. Finding 1 means no parameterised query returns correct results over couchbase2, and finding 2 means the
commit does not build in the configuration CI uses.

core/protostellar/query_converter.hxx
```diff
@@ -0,0 +1,250 @@
...
+  }
+  for (const auto& [name, value] : request.named_parameters) {
+    (*proto.mutable_named_parameters())[name] = value.str();
+  }
+
+  // Tuning options: the SDK's `metrics` flag defaults to off, so disable_metrics is its inverse.
+  auto* tuning = proto.mutable_tuning_options();
+  tuning->set_disable_metrics(!request.metrics);
+  if (request.max_parallelism.has_value()) {
+    tuning->set_max_parallelism(static_cast<std::uint32_t>(*request.max_parallelism));
```
avsej, line 167: `max_parallelism`, `pipeline_batch`, `pipeline_cap` and `scan_cap` are 64-bit in the core request and `uint32` in the proto, so a value above `UINT32_MAX` wraps silently and the gateway is asked to use the wrapped number. The classic path sends the full value.

`can_encode` is where this converter puts its other "cannot express this" decisions, so rejecting out-of-range values there is consistent:

```cpp
if (request.max_parallelism.value_or(0) > std::numeric_limits<std::uint32_t>::max() ||
    request.pipeline_batch.value_or(0) > std::numeric_limits<std::uint32_t>::max() ||
    request.pipeline_cap.value_or(0) > std::numeric_limits<std::uint32_t>::max() ||
    request.scan_cap.value_or(0) > std::numeric_limits<std::uint32_t>::max()) {
  return false;
}
```

test/cng/protostellar/live_query.cxx
```diff
@@ -0,0 +1,132 @@
...
+namespace ops = ::couchbase::core::operations;
+using ::couchbase::core::cluster_credentials;
+using ::couchbase::core::origin;
+using ::couchbase::core::tls_verify_mode;
+using ::couchbase::core::utils::parse_connection_string;
+
+auto
+env_or(const char* name, const char* fallback) -> std::string
+{
+  const char* value = std::getenv(name); // NOLINT(concurrency-mt-unsafe)
```
avsej: The framework already provides a wrapper for this and `live_connect.cxx` uses it. `std::getenv` is deprecated on MSVC and the Windows legs build with warnings as errors, so the hand-rolled helper is a portability hazard as well as duplication — and the `NOLINT` suppresses the diagnostic that would have pointed at it.

Using the framework helper also keeps the twelve live-test prologues consistent; several of them have drifted into slightly different forms of the same block.

core/protostellar/query_converter.hxx
avsej: > **Blocker: every query parameter is sent as an empty string.**

Confirmed and fixed. `json_payload()` reads the binary arm directly, as suggested; the proto field
is `repeated bytes`, so the payload is copied rather than parsed and regenerated. Both the
positional and the named line go through it.

> Worth auditing every other `json_string` read in the couchbase2 converters for the same pattern

Audited. `query_request::raw` is the only other `json_string` this converter can reach, and it is
never read — `can_encode` refuses a non-empty `raw` outright. The analytics converter in #1037 has
the same construct and is fixed in that commit.

Why this passed review and CI is worth recording. `encode_maps_parameters_and_context` built its
fixture with `positional_parameters.emplace_back(R"("pos")")`, which constructs the **string** arm;
the public API constructs the **binary** arm (`couchbase/query_options.hxx:70` →
`core/impl/query.cxx:204`). The case exercised an arm no caller produces, so it was right about the
arm it used and wrong about the code. The replacement builds the binary arm through a named helper
whose comment says why, and asserts the payload rather than its presence — including one containing
an embedded NUL and a multi-byte UTF-8 sequence, checked byte-for-byte and by length.

Live cases assert the value the gateway echoes back for `SELECT $1 AS echo, $2 AS count` and
`SELECT $word AS echo`, which is the level the defect bites at: an empty parameter produced a wrong
answer, not an error. Reverting either line to `str()` fails both the unit case and the live case
that name it.

On the two further items raised here:

- **The fifth constructor parameter.** Already resolved in the tree, by the design fix suggested
  rather than the immediate one: the constructor is now `component(asio::io_context&,
  component_config)` and the timeout is a named field, so adding one is no longer an arity change.
  Both `live_kv.cxx` call sites set it. A full build at this commit completes and `ctest -L cng`
  runs clean, `cng_live_kv_test` among it.
- **`DEADLINE_EXCEEDED` mapped unconditionally.** Also already resolved: `execute` captures
  `request.readonly ? operation_kind::read_only : operation_kind::mutating` and passes it to
  `map_status`. It had no query-level coverage, so two live cases now exercise both directions
  against a real gateway; hardcoding `operation_kind::mutating` fails the read-only one.

core/protostellar/query_converter.hxx
avsej: > `max_parallelism`, `pipeline_batch`, `pipeline_cap` and `scan_cap` are 64-bit in the core request
> and `uint32` in the proto, so a value above `UINT32_MAX` wraps silently

Fixed in `can_encode` as suggested, through a `fits_uint32` helper so the four checks read as one
rule.

The test covers both sides of the boundary: `UINT32_MAX` itself still encodes and survives the
cast, and each of the four fields is rejected individually at `UINT32_MAX + 1`. A single combined
case would have passed with three of the four checks missing.

test/cng/protostellar/live_query.cxx
avsej: > The framework already provides a wrapper for this and `live_connect.cxx` uses it.

> Using the framework helper also keeps the twelve live-test prologues consistent; several of them
> have drifted into slightly different forms of the same block.

Fixed by the second half of this rather than the first. Substituting `safe_getenv` would have left
the thirty-line prologue duplicated, so `live_fixture.hxx` gains a `live_cluster_fixture` alongside
the existing component-level one; it owns the parse, the skip, the credentials, the io thread,
`open()` and `close()`. `live_query.cxx` now has no environment handling in it at all, and the
analytics and search live suites will use the same fixture.

One instance of the pattern is **not** fixed here: `test/cng/protostellar/live_observability.cxx`
still calls `std::getenv` directly at `:224` and `:231`. That file belongs to CXXCBC-902, so the fix
belongs in that commit rather than in this pull request.

avsej:
## Response

Five findings, all addressed. Two of them — the build break and the read-only timeout — were
already fixed in the tree by earlier rounds and are confirmed below rather than changed again.

The round also produced a defect the review did not name, in the same commit, and a correction to
one of the review's own claims.

### 1. Every query parameter is sent as an empty string (blocker)

> **Blocker: every query parameter is sent as an empty string.**
>
> `core::json_string` is a `std::variant<std::nullptr_t, std::string, std::vector<std::byte>>`, and
> `str()` returns a reference to a **static empty string** unless the variant holds the
> `std::string` arm (`core/json_string.hxx:47-53`).

Confirmed and fixed. `json_payload()` reads the binary arm directly, as suggested; the proto field
is `repeated bytes`, so the payload is copied rather than parsed and regenerated. Named and
positional parameters both go through it.

> Worth auditing every other `json_string` read in the couchbase2 converters for the same pattern
> — the raw tuning passthrough here, and the analytics converter added in #1037.

Audited. `query_request::raw` is the only other `json_string` in this converter's reach and it is
never read — `can_encode` refuses a non-empty `raw` outright, so there is no second site here. The
analytics converter in #1037 has the same construct and will be fixed in that commit.

The reason this passed review and passed CI is worth recording, because it is a reusable trap.
`encode_maps_parameters_and_context` built its fixture with
`positional_parameters.emplace_back(R"("pos")")`, which constructs the **string** arm; the public
API constructs the **binary** arm (`couchbase/query_options.hxx:70` →
`core/impl/query.cxx:204`). The test exercised an arm no caller produces, so it asserted the
payload survived and was right about the arm it used and wrong about the code. The replacement
builds the binary arm through a named helper whose comment says why.

Coverage added: the payload is asserted, not merely its presence; a payload containing an embedded
NUL and a multi-byte UTF-8 sequence is checked byte-for-byte and by length; the string arm and a
default-constructed `json_string` are covered separately. Live, `SELECT $1 AS echo, $2 AS count`
and `SELECT $word AS echo` assert the values the gateway echoes back — which is the level the
defect bites at, since an empty parameter produces a wrong answer rather than an error.

Both mutations — reverting each of the two lines to `str()` — fail the unit case and the live case
that name them.

### 2. A fifth required constructor parameter without updating the call sites (blocker)

> This commit adds `std::chrono::milliseconds default_query_timeout` to `component::component`,
> making it five positional parameters, and does not update
> `test/cng/protostellar/live_kv.cxx:100` and `:183`, which pass four.

Already fixed in the tree, by the design fix the finding recommends rather than the immediate one.
The constructor is now `component(asio::io_context&, component_config)`, and the timeout is a named
field on `component_config`, so adding one is no longer an arity change. Both `live_kv.cxx` call
sites set `config.default_query_timeout`.

Verified rather than assumed: a full build at this commit completes, and the `cng` label runs
clean, `cng_live_kv_test` among it.

### 3. A timed-out read-only query reports `ambiguous_timeout` (high)

> The completion handler builds the error from `map_status(status)`, which maps
> `DEADLINE_EXCEEDED` to `errc::common::ambiguous_timeout` unconditionally. RFC 77's error table
> keys that on read-only status.

Already fixed in the tree: `execute` captures
`request.readonly ? operation_kind::read_only : operation_kind::mutating` and passes it to
`map_status`.

It was covered only by the KV-level unit test of `map_status_code`, so the query path had no
evidence of its own. Two live cases now cover both directions against a real gateway, and the
mutation that hardcodes `operation_kind::mutating` fails the read-only one.

### 4. Tuning options are narrowed without a range check (medium)

> `max_parallelism`, `pipeline_batch`, `pipeline_cap` and `scan_cap` are 64-bit in the core request
> and `uint32` in the proto, so a value above `UINT32_MAX` wraps silently and the gateway is asked
> to use the wrapped number.
>
> `can_encode` is where this converter puts its other "cannot express this" decisions, so rejecting
> out-of-range values there is consistent

Fixed as suggested, through a `fits_uint32` helper so the four checks read as one rule. The test
covers both sides of the boundary: `UINT32_MAX` itself still encodes and survives the cast, and
each of the four fields is rejected one at a time at `UINT32_MAX + 1` — a single combined case
would pass with three of the four checks missing.

### 5. `live_query.cxx` uses raw `std::getenv` (medium)

> The framework provides a wrapper for exactly this, and `live_connect.cxx` uses it. On MSVC,
> `std::getenv` is deprecated and the Windows CI legs build with warnings as errors
>
> Using the framework helper keeps the twelve live-test prologues consistent.

Fixed, and by the second half of the finding rather than the first. Swapping in `safe_getenv` would
have left the thirty-line prologue duplicated; `live_fixture.hxx` gains a `live_cluster_fixture`
alongside the existing component-level one, which owns the parse, the skip, the credentials, the
io thread, `open()` and `close()`. `live_query.cxx` has no environment handling left in it, and the
same fixture is what the analytics and search live suites will use.

One instance of the same pattern is **not** fixed here:
`test/cng/protostellar/live_observability.cxx:224` and `:231` still call `std::getenv` directly.
That file belongs to CXXCBC-902, three commits below this one, so the fix belongs in that commit
and not in this pull request.

## Not in the review

### The mid-stream retry rule is not implemented

The review credits this commit with it:

> the completion handler converts a mid-stream failure into `request_canceled` rather than letting
> it be retried, which is RFC 77's rule and is recorded as unimplemented in gocb,
> couchbase-jvm-clients and couchbase-net-client (CNG-16).

That does not hold. The completion handler maps the status with `map_status(status, kind)` and
nothing else; there is no test for "rows already delivered" and no conversion to
`errc::common::request_canceled`. RFC 77's Streaming section states the rule:

> no subsequent mid-stream error can be safely retried. If a mid-stream error is retryable
> according to the retry orchestrator … the operation should be failed with a `RequestCancelled`
> error instead.

At this commit the omission is inert: nothing retries a query yet, so a mid-stream failure is
simply reported. It stops being inert at CXXCBC-906, which adds retry for non-KV operations, and
that commit does not implement the rule either. It is therefore a finding against CXXCBC-906
rather than against this one, and is recorded there.

### The two paths that answer without sending anything disagreed with each other

`execute` can complete without dispatching in two cases: a spent time budget, and a request
`can_encode` refuses. The expired path posts through the io context, with a comment on
`fail_expired` giving the reason:

> Delivered through the io_context, not inline on the caller's thread: "completions arrive on the
> SDK's execution context" is the contract every other path here honours, and a synchronous
> callback out of execute() would also re-enter the caller before it has its pending_call back.

The `can_encode` path called the handler inline, in the same function and immediately after the expired path that posts. It also left the
error context blank, so a caller correlating failures by statement or `client_context_id` had
nothing to correlate on — while every other failure on this path carries both.

Both are fixed: the refusal is posted like every other completion, and both no-dispatch paths stamp
the statement and `client_context_id`. Neither is visible to an assertion on the response, so the
two live cases check the thread the handler ran on. The mutations that restore the inline call and
that blank the statement both fail.

## Coverage and how it was chosen

The expectations were taken from `queryserver.go`, `query.proto` and the gateway's own acceptance
suite rather than from this converter, on the principle that a test written by reading the code it
tests only reports that the code changed.

| | Before | After |
|---|---|---|
| `query_converter.cxx` | 5 cases | 16 |
| `live_query.cxx` | 1 case | 17 |

The live cases include the gateway's own rejections re-expressed at this SDK's surface — syntax
error, write under `read_only`, `DROP INDEX` on a missing index, scan consistency together with
mutation tokens, named together with positional parameters — plus the streaming shapes the gateway
produces: a result over the server's 1024-byte row-batch cap, checked complete and in order, and an
empty result that still carries metadata.

Every assertion added was negative-controlled: twelve mutations, each reverting one behaviour,
each observed to fail the case that claims to cover it. Four of them build against a live gateway.
The two that would have compiled to something plausible but wrong — the inline callback and the
`str()` revert — are the ones worth having.

avsej:
A correction to the `live_observability.cxx` paragraph above.

> One instance of the same pattern is **not** fixed here:
> `test/cng/protostellar/live_observability.cxx:224` and `:231` still call `std::getenv` directly.
> That file belongs to CXXCBC-902, three commits below this one, so the fix belongs in that commit
> and not in this pull request.

Where the fix belongs was stated correctly, and that is where it has now been made. The
implication that this pull request was therefore unaffected was wrong, and the paragraph should
not have been left as a scope note.

`live_observability.cxx` sits at the **base** of the series -- CXXCBC-902 is the bottom commit,
below CXXCBC-895 -- so the file is an ancestor of every branch in the series, this one included.
MSVC deprecates `getenv` (C4996), these builds compile the tests with warnings as errors, and the
resulting C2220 failed all three Windows legs. On the head current when that response was
written, `windows (windows-2022)`, `windows (windows-2025)` and `windows (windows-11-arm)` were
all failing for this reason. Calling it out-of-scope described the commit boundary accurately and
the consequence not at all.

The scope of it was also understated. Seven further live suites carried their own copy of the
same hand-rolled preamble, so thirteen pull requests in the series were failing every Windows
leg:

| Introducing commit | Pull request | File(s) |
|---|---|---|
| CXXCBC-899 | #1038 | `live_search.cxx`, `live_view.cxx` |
| CXXCBC-900 | #1039 | `live_bucket_admin.cxx` |
| CXXCBC-900 | #1040 | `live_collection_admin.cxx` |
| CXXCBC-901 | #1041 | `live_query_index_admin.cxx` |
| CXXCBC-901 | #1042 | `live_search_index_admin.cxx` |
| CXXCBC-908 | #1049 | `live_wait_until_ready.cxx` |

#1043, #1046, #1047, #1048, #1050, #1051 and #1057 introduced no further copy and were failing
by inheritance alone, so the six commits above cover them.

All of these now use the framework's `safe_getenv`, each fixed in the commit that introduced the
call rather than in one sweeping change, so every pull request in the series stays
self-consistent. `safe_getenv` was already `#ifdef`-guarded with a branch that does not call
`getenv` on MSVC, which is why the framework itself never triggered the warning and only the
hand-rolled copies did.

Nothing else in that response is affected.

test/cng/protostellar/live_query.cxx
```diff
@@ -0,0 +1,454 @@
...
+
+// Live end-to-end tests of N1QL query over the couchbase2:// streaming transport (CXXCBC-897).
+//
+// Each case drives the core::cluster public path -- open() builds the gRPC component,
+// execute(query_request) routes to the component's server-streaming query path -- against a real
+// CNG gateway. cluster_only: the whole suite skips unless TEST_CONNECTION_STRING points at a
+// couchbase2:// endpoint.
+//
+// What each case is checking, and why that is the expected answer rather than merely the current
+// one, is recorded in cng-dev-docs/CATALOGUE_QUERY.md against the gateway source it comes from.
```
thejcfactor: Probably want to remove this mention?

test/cng/protostellar/query_converter.cxx
```diff
@@ -0,0 +1,490 @@
...
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Unit tests for the N1QL query <-> couchbase.query.v1 converter (CXXCBC-897). Pure, no server.
+//
+// The expectations here are read off couchbase/query/v1/query.proto and the gateway's
+// queryserver.go, not off the converter -- an assertion derived from the code it tests only
+// notices that the code changed. cng-dev-docs/CATALOGUE_QUERY.md records which source each case
```
thejcfactor: Probably want to remove this mention?

thejcfactor (approved):

PR #1037, opened 2026-07-28
CXXCBC-898: route analytics over the couchbase2 streaming transport

avsej:
Analytics was rejected by the HTTP execute front-door gate on couchbase2. Wire
it onto the server-streaming primitive, mirroring the query path.

Add an analytics converter (core operations <-> couchbase.analytics.v1): encode
statement, parameters, scan consistency, read-only and priority; decode the
terminal metadata (status string -> core enum, request_id, metrics, warnings).
Rows are buffered into analytics_response.rows. The component gains an
AnalyticsService stub and an analytics execute() driving
dispatcher::server_stream; cluster_impl routes analytics_request to it alongside
query. The component constructor takes an analytics timeout; its direct-ctor
test callers are updated to match.

A non-success terminal status arriving with an OK gRPC trailer is turned into an
error through `map_analytics_error`, the same classifier the HTTP path uses, so a
caller cannot see a successful result whose own metadata says it failed. This is
conditional on a MetaData message having been received: `analytics_status`
numbers `running` as zero, so an absent status is otherwise indistinguishable
from a reported one, and the gateway does end a stream OK with no metadata when
its own `result.MetaData()` fails.

Features the couchbase2 schema cannot express are reported up front as
`feature_not_available` rather than silently dropped: the `raw` passthrough map
and any scope qualification.

**Scope-qualified analytics is not supported over couchbase2**, so
`scope.analytics_query(...)` returns `feature_not_available`. This is a schema
limitation rather than a missing mapping: at the pinned protostellar revision
(36d4d9ab) `AnalyticsQueryRequest` reserves field 8 (the former `bucket_name`)
and field 9 is `analytics_scope_name` -- a single analytics-scope identifier, not
the bucket/scope pair the core request carries. Collapsing a two-part
qualification into that one name would guess at which the gateway means, and
guessing wrong runs the statement against the wrong scope, so the request is
refused. Note that the protos vendored by couchbase-jvm-clients and
couchbase-net-client still carry the older `bucket_name = 8` / `scope_name = 9`
shape, so a mapping copied from those clients would target a schema this client
does not speak. Lifting this needs a schema decision, not a client change.

Note: the CNG gateway does not serve `couchbase.analytics.v1` at all.
`gateway/system/system.go` registers kv, query, search, the four admin services,
xdcr and routing, and no analytics service, so every method on it answers
UNIMPLEMENTED -- observed as `unknown service
couchbase.analytics.v1.AnalyticsService`, which the client maps to
`feature_not_available`.

The live suite therefore covers what this client does before anything is sent --
a scope-qualified request is refused, an unsupported option is refused, an
exhausted budget is refused, and each refusal is stamped with the statement and
delivered on the io context rather than inline. Those three cases run and assert
against today's gateway. Only the end-to-end round trip skips, and it skips on
the gateway's own UNIMPLEMENTED message rather than on the bare error code, so
deleting this commit's routing turns it into a failure instead of a skip.

---
Stacked PR **15/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–15; the change specific to this PR is its top commit. Depends on #1036 — merge the series bottom-up.

### DIFF
--- a/core/protostellar/analytics_converter.hxx
+++ b/core/protostellar/analytics_converter.hxx
@@ -1,0 +1,171 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates the core analytics request/response to and from the couchbase.analytics.v1 protobuf
+// messages, mirroring query_converter. Rows arrive across streamed AnalyticsQueryResponse messages
+// and are buffered into analytics_response::rows; the terminal message carries the metadata.
+//
+// Features with no couchbase2 equivalent are reported by can_encode() so the component can surface
+// feature_not_available rather than silently dropping them: the `raw` passthrough map, the
+// streaming row_callback (only the buffered result path is wired), and scoped analytics.
+//
+// Scoped analytics is refused because the schema cannot express it, not merely because it is
+// unimplemented. AnalyticsQueryRequest at the pinned protostellar revision retired the field the
+// older schema used for this: field 8 is `reserved // bucket_name`, and field 9 was renamed from
+// `scope_name` to `analytics_scope_name` -- one analytics-scope identifier rather than the
+// bucket/scope pair the core request carries in bucket_name/scope_name/scope_qualifier. Collapsing
+// a two-part qualification into that single name would be a guess about which the gateway means, so
+// the request is refused instead. (Vendored copies in couchbase-jvm-clients and
+// couchbase-net-client still carry the older bucket_name=8/scope_name=9 shape, so a mapping cribbed
+// from them would target a schema this client does not speak.)
+
+#include "core/operations/document_analytics.hxx"
+#include "core/protostellar/json_payload.hxx"
+
+#include <couchbase/analytics_scan_consistency.hxx>
+
+#include <couchbase/analytics/v1/analytics.pb.h>
+
+#include <chrono>
+#include <cstddef>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace couchbase::core::protostellar::analytics
+{
+namespace v1 = ::couchbase::analytics::v1;
+
+[[nodiscard]] inline auto
+can_encode(const operations::analytics_request& request) -> bool
+{
+  return request.raw.empty() && !request.row_callback.has_value() &&
+         !request.bucket_name.has_value() && !request.scope_name.has_value() &&
+         !request.scope_qualifier.has_value();
+}
+
+inline auto
+encode(const operations::analytics_request& request) -> v1::AnalyticsQueryRequest
+{
+  v1::AnalyticsQueryRequest proto;
+  proto.set_statement(request.statement);
+  proto.set_read_only(request.readonly);
+  proto.set_priority(request.priority);
+  if (request.client_context_id.has_value()) {
+    proto.set_client_context_id(*request.client_context_id);
+  }
+  if (request.scan_consistency.has_value()) {
+    switch (*request.scan_consistency) {
+      case analytics_scan_consistency::not_bounded:
+        proto.set_scan_consistency(
+          v1::AnalyticsQueryRequest_ScanConsistency_SCAN_CONSISTENCY_NOT_BOUNDED);
+        break;
+      case analytics_scan_consistency::request_plus:
+        proto.set_scan_consistency(
+          v1::AnalyticsQueryRequest_ScanConsistency_SCAN_CONSISTENCY_REQUEST_PLUS);
+        break;
+    }
+  }
+  for (const auto& parameter : request.positional_parameters) {
+    proto.add_positional_parameters(json_payload(parameter));
+  }
+  for (const auto& [name, value] : request.named_parameters) {
+    (*proto.mutable_named_parameters())[name] = json_payload(value);
+  }
+  return proto;
+}
+
+[[nodiscard]] inline auto
+to_nanoseconds(const google::protobuf::Duration& duration) -> std::chrono::nanoseconds
+{
+  return std::chrono::seconds{ duration.seconds() } + std::chrono::nanoseconds{ duration.nanos() };
+}
+
+// The gateway reports status as a lowercase string; map it onto the core enum.
+[[nodiscard]] inline auto
+status_from_string(const std::string& status) -> operations::analytics_response::analytics_status
+{
+  using s = operations::analytics_response::analytics_status;
+  if (status == "running") {
+    return s::running;
+  }
+  if (status == "success") {
+    return s::success;
+  }
+  if (status == "errors") {
+    return s::errors;
+  }
+  if (status == "completed") {
+    return s::completed;
+  }
+  if (status == "stopped") {
+    return s::stopped;
+  }
+  if (status == "timeout" || status == "timedout") {
+    return s::timedout;
+  }
+  if (status == "closed") {
+    return s::closed;
+  }
+  if (status == "fatal") {
+    return s::fatal;
+  }
+  if (status == "aborted") {
+    return s::aborted;
+  }
+  return s::unknown;
+}
+
+inline void
+decode_meta_data(const v1::AnalyticsQueryResponse_MetaData& proto,
+                 operations::analytics_response::analytics_meta_data& meta)
+{
+  meta.request_id = proto.request_id();
+  meta.client_context_id = proto.client_context_id();
+  meta.status = status_from_string(proto.status());
+  if (!proto.signature().empty()) {
+    meta.signature = proto.signature();
+  }
+  if (proto.has_metrics()) {
+    const auto& m = proto.metrics();
+    meta.metrics.elapsed_time = to_nanoseconds(m.elapsed_time());
+    meta.metrics.execution_time = to_nanoseconds(m.execution_time());
+    meta.metrics.result_count = m.result_count();
+    meta.metrics.result_size = m.result_size();
+    meta.metrics.error_count = m.error_count();
+    meta.metrics.processed_objects = m.processed_objects();
+    meta.metrics.warning_count = m.warning_count();
+  }
+  // Replaced rather than appended, matching query_converter: decode_meta_data runs once per
+  // MetaData message, and appending would accumulate duplicates without bound if a gateway ever
+  // sent more than one.
+  if (proto.warnings_size() > 0) {
+    std::vector<operations::analytics_response::analytics_problem> warnings;
+    warnings.reserve(static_cast<std::size_t>(proto.warnings_size()));
+    for (const auto& w : proto.warnings()) {
+      operations::analytics_response::analytics_problem problem;
+      problem.code = w.code();
+      problem.message = w.message();
+      warnings.push_back(std::move(problem));
+    }
+    meta.warnings = std::move(warnings);
+  }
+}
+
+} // namespace couchbase::core::protostellar::analytics
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -17,9 +17,12 @@
 
 #include "core/protostellar/component.hxx"
 
+#include "core/error_context/analytics.hxx"
 #include "core/error_context/key_value.hxx"
 #include "core/error_context/query.hxx"
+#include "core/operations/analytics_response_parsing.hxx"
 #include "core/operations/query_response_parsing.hxx"
+#include "core/protostellar/analytics_converter.hxx"
 #include "core/protostellar/credentials.hxx"
 #include "core/protostellar/error_utils.hxx"
 #include "core/protostellar/kv_converter.hxx"
@@ -28,6 +31,7 @@
 #include <couchbase/error_codes.hxx>
 
 #include "core/protostellar/query_proto.hxx"
+#include <couchbase/analytics/v1/analytics.grpc.pb.h>
 
 #include <couchbase/kv/v1/kv.grpc.pb.h>
 
@@ -46,12 +50,14 @@ namespace couchbase::core::protostellar
 {
 namespace v1 = ::couchbase::kv::v1;
 namespace query_v1 = ::couchbase::query::v1;
+namespace analytics_v1 = ::couchbase::analytics::v1;
 
 // Defined here rather than in the header so the generated gRPC types stay out of every consumer of
 // component.hxx.
 struct component::stubs {
   std::unique_ptr<v1::KvService::Stub> kv;
   std::unique_ptr<query_v1::QueryService::Stub> query;
+  std::unique_ptr<analytics_v1::AnalyticsService::Stub> analytics;
 };
 
 namespace
@@ -98,11 +104,14 @@ fail_expired_ctx(asio::io_context& io, Handler& handler, Response response) -> p
 
 component::component(asio::io_context& io, component_config config)
   : io_{ io }
-  , stubs_{ std::make_unique<stubs>(stubs{ v1::KvService::NewStub(config.channel),
-                                           query_v1::QueryService::NewStub(config.channel) }) }
+  , stubs_{ std::make_unique<stubs>(
+      stubs{ v1::KvService::NewStub(config.channel),
+             query_v1::QueryService::NewStub(config.channel),
+             analytics_v1::AnalyticsService::NewStub(config.channel) }) }
   , authorization_{ authorization_header(config.credentials) }
   , default_kv_timeout_{ config.default_kv_timeout }
   , default_query_timeout_{ config.default_query_timeout }
+  , default_analytics_timeout_{ config.default_analytics_timeout }
   // Initialised last (see the declaration order in the header), so the channel can be moved in.
   , dispatcher_{ io, std::move(config.channel) }
 {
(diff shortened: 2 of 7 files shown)


core/protostellar/component.cxx
```diff
@@ -0,0 +1,558 @@
...
+    // `proto` is captured here too so the request outlives the in-flight call.
+    [handler = std::move(handler), id, proto](grpc::Status status, v1::GetResponse resp) mutable {
+      auto ctx = make_error_context(status, id);
+      // Compressed content decode is not wired yet; surface an error rather than an empty value
+      // with a success status (which would be silent data loss).
+      if (status.ok() && resp.content_case() == v1::GetResponse::kContentCompressed) {
+        ctx = make_key_value_error_context(errc::common::feature_not_available, id);
+      }
+      handler(kv::decode(resp, std::move(ctx)));
+    });
```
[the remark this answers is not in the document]
avsej: Addressed: added (void)proto; to this lambda and every other execute() completion lambda that captures proto only for lifetime (incl. lines 480/541 and the query/analytics/search/view and management overloads). The capture is deliberate lifetime extension; the cast documents intent and silences -Wunused-lambda-capture under clang.

test/cng/protostellar/streaming.cxx
```diff
@@ -0,0 +1,169 @@
...
+        rows.push_back(row);
+      }
+      if (message.has_meta_data()) {
+        saw_metadata = true;
+      }
+    },
+    [&, request](grpc::Status status) {
+      final_status = std::move(status);
+      work.reset();
+    });
```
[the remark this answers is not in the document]
avsej: Addressed: added (void)request; the capture keeps the QueryRequest alive for the streaming call, and the cast documents intent while silencing the unused-capture warning.

core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,309 @@
...
+  void OnReadDone(bool ok) override
+  {
+    if (!ok) {
+      return; // stream finished; OnDone follows
+    }
+    auto message = std::make_shared<Response>(std::move(current_));
+    asio::post(io_, [this, message]() {
+      on_row_(std::move(*message));
+    });
+    this->StartRead(&current_);
```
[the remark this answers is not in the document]
avsej: Defensible as written. gRPC serializes a single call's reactions (OnReadDone completes before OnDone runs), so every row task is posted before the OnDone task. The SDK contract is a single-threaded io_context, which runs posted tasks in FIFO order, so the self-delete task always runs last -- after every row handler. `delete this` from the terminal post is the standard self-owning-reactor pattern; the row handlers never dereference a freed reactor.

core/protostellar/analytics_converter.hxx
```diff
@@ -0,0 +1,147 @@
...
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates the core analytics request/response to and from the couchbase.analytics.v1 protobuf
+// messages, mirroring query_converter. Rows arrive across streamed AnalyticsQueryResponse messages
+// and are buffered into analytics_response::rows; the terminal message carries the metadata.
+//
+// Features with no couchbase2 equivalent are reported by can_encode() so the component can surface
```
avsej, line 24: Four observations on this converter, all from comparing it against its query sibling:

1. **Parameters are sent empty (high).** Positional and named parameters are read with `json_string::str()`, which returns a static empty string unless the variant holds the `std::string` arm — and the public API populates them from `codec::binary`. Every analytics parameter therefore reaches the gateway as `""`. Full analysis on #1036; the fix is the same `json_payload` helper, applied in both converters.

2. **Warnings accumulate.** `decode_meta_data` appends to the warning list where the query converter replaces it, so a gateway sending more than one metadata message duplicates warnings on this path only. Replacing is the safer default and matches the sibling.

3. **A non-success terminal status is not turned into an error.** The analytics status enum reaches the response metadata, but a terminal `errors` / `fatal` / `aborted` / `timedout` arriving with an OK gRPC trailer leaves `ctx.ec` unset, so the caller sees a successful query whose metadata says it failed. The classic path derives an error from the status; the same check is worth adding on the query path too.

4. **Scoped analytics is rejected** although `AnalyticsQueryRequest` carries `bucket_name` and `scope_name` and couchbase-net-client maps them. Failing closed is correct for something unimplemented, so this is a scope note rather than a defect — but it belongs in the pull request body, since `scope.analytics_query(...)` is a normal way to use the API.

test/cng/protostellar/live_analytics.cxx
```diff
@@ -0,0 +1,139 @@
...
+
+  // Tear down before evaluating: assert_*/skip throw, and unwinding past a still-joinable io thread
+  // would std::terminate instead of reporting cleanly.
+  std::promise<void> closed;
+  cluster.close([&closed]() {
+    closed.set_value();
+  });
+  closed.get_future().get();
+
+  // Older CNG gateways do not serve analytics.v1 (UNIMPLEMENTED -> feature_not_available); the
```
avsej: The skip predicate is `feature_not_available`, which is the error the **client** produces for its own unrouted operations. If this commit's routing were deleted entirely, the test would report "Skipped" rather than failing — it cannot distinguish "the gateway does not implement analytics.v1" from "the client never sent anything".

Making the skip provably about the gateway is the fix: require an explicit opt-in environment variable for the gateway-capability skip so an unexpected `feature_not_available` fails by default, and assert that a dispatch was actually attempted.

Seven live suites in this series share this pattern, so it is worth fixing once in the shared fixture rather than per file.

avsej (commented):
## Review — PR #1037 (`CXXCBC-898`)

This commit routes analytics over the streaming transport, following the query commit that landed immediately
before it. The two implementations are deliberately parallel, which is the right choice — the wire shapes are
close enough that divergence would be a maintenance cost rather than a benefit — and the analytics-specific
fields (`priority`, the analytics scan-consistency enum, the analytics status enum) are handled.

Because it is a near-copy, the most useful thing to check is what the copy did not carry over, and the answer is
"a few things". The query commit's own defects are inherited too.

### Findings

|---|---|---|---|---|
| 1 | high | correctness | The `json_string` parameter defect from the query converter is inherited verbatim | `core/protostellar/analytics_converter.hxx:24` |
| 2 | medium | consistency | Warnings are appended on each metadata message rather than replaced, diverging from the query converter | `core/protostellar/analytics_converter.hxx:24` |
| 3 | medium | correctness | A non-success terminal analytics status is decoded but never turned into an error | `core/protostellar/analytics_converter.hxx:24` |
| 4 | medium | api-surface | Scoped analytics is rejected although the proto carries bucket and scope | `core/protostellar/analytics_converter.hxx:24` |
| 5 | medium | test-coverage | The live analytics test cannot fail, because it skips on the error the client itself produces | `test/cng/protostellar/live_analytics.cxx:109` |

#### 1. The `json_string` parameter defect is inherited (high)

The analytics converter reads positional and named parameters with `json_string::str()`, exactly as the query
converter does. `str()` returns a static empty string unless the variant holds the `std::string` arm, and the
public API populates these from `codec::binary`, so every analytics parameter is sent empty. The full analysis is
on #1036; the fix is the same helper, and it needs applying in both converters.

#### 2. Warnings are appended rather than replaced (medium)

The query converter's `decode_meta_data` replaces the warning list when a metadata message arrives; the analytics
copy appends to it. A gateway that sends more than one metadata message therefore produces duplicated warnings on
the analytics path and not on the query path. Neither behaviour is obviously wrong on its own, but they should
match, and replacing is the safer default because it cannot accumulate without bound.

#### 3. A non-success terminal status is decoded but not acted on (medium)

The analytics status enum is mapped into the response metadata, but a terminal status of `errors`, `fatal`,
`aborted` or `timedout` arriving with an OK gRPC trailer leaves `ctx.ec` unset — so the user sees a successful
query whose metadata says it failed. The classic path derives an error from the status. Deriving `ctx.ec` from a
non-success terminal status closes it, and the same check is worth adding on the query path.

#### 4. Scoped analytics is rejected although the proto supports it (medium)

`can_encode` refuses a request carrying a scope qualification, so `scope.analytics_query(...)` returns
`feature_not_available` over couchbase2. `AnalyticsQueryRequest` carries `bucket_name` and `scope_name`, and
couchbase-net-client maps them. Failing closed is the correct behaviour for something unimplemented, so this is
not a defect — but it is a gap worth stating in the pull request body and the ticket, because scope-level
analytics is a normal way to use the API.

#### 5. The live analytics test cannot fail (medium)

```cpp
// Older CNG gateways do not serve analytics.v1 (UNIMPLEMENTED -> feature_not_available); the
```

The test skips when the error is `feature_not_available`. That is the error the client produces for its *own*
unrouted operations, so if this commit's routing were deleted entirely the test would report "Skipped" rather
than failing. The skip predicate cannot distinguish "the gateway does not implement analytics" from "the client
never sent anything".

Making the skip provably about the gateway is what fixes it — for example by requiring an explicit opt-in
environment variable for the gateway-capability skip, so that the default outcome for an unexpected
`feature_not_available` is a failure, and by asserting that the client at least attempted a dispatch. Seven live
suites in this series share this pattern; it is worth fixing once in the shared fixture rather than per file.

(inline comment on `test/cng/protostellar/live_analytics.cxx:109`)

### Note on the macOS failure

This pull request had a completed `macos (macos-14)` failure at the time of review while the Linux legs were
still queued. The analytics files omit the `<cstdint>` and `<chrono>` includes that their query counterparts
carry, relying on transitive includes that libc++ and libstdc++ do not agree about — that is the most likely
candidate in this diff and is worth checking first.

### Verdict

Needs changes. Finding 1 is inherited from #1036 but has to be fixed in this file too, and findings 2 and 3 are
small divergences from the sibling implementation that are cheapest to correct while the two are still parallel.

avsej:
All five findings are addressed. One is answered with a correction: finding 4's
premise does not hold at the pinned schema, and the conclusion it reaches is
right for a different reason than the one given.

### 1. The `json_string` parameter defect is inherited (high)

> Positional and named parameters are read with `json_string::str()`, which returns a static empty string unless the variant holds the `std::string` arm — and the public API populates them from `codec::binary`. Every analytics parameter therefore reaches the gateway as `""`. Full analysis on #1036; the fix is the same `json_payload` helper, applied in both converters.

Fixed. The helper is now shared rather than copied: it moved out of
`query_converter.hxx` into `core/protostellar/json_payload.hxx`, and both
converters include it. A per-converter copy is how this defect reached the
analytics converter in the first place, and search will need the same helper, so
the duplication was the thing worth removing.

The pre-existing converter test did not catch this, and would not have caught it
after the fix either: it built its parameters from a string literal, which
selects the `std::string` arm that production never uses, so it passed against
both the broken and the correct reader. The replacement builds the binary arm and
asserts the payload length as well as its value, so an empty-against-empty
comparison cannot pass it.

Confirmed by mutation: reverting either parameter line to `.str()` fails
`parameters_from_the_binary_arm_carry_their_payload`.

### 2. Warnings are appended rather than replaced (medium)

> `decode_meta_data` appends to the warning list where the query converter replaces it, so a gateway sending more than one metadata message duplicates warnings on this path only. Replacing is the safer default and matches the sibling.

Fixed; the analytics decoder now builds a vector and assigns it, as the query
decoder does. A case decodes the same metadata message twice and asserts one
warning survives; reverting to `push_back` fails it.

### 3. A non-success terminal status is decoded but not acted on (medium)

> The analytics status enum reaches the response metadata, but a terminal `errors` / `fatal` / `aborted` / `timedout` arriving with an OK gRPC trailer leaves `ctx.ec` unset, so the caller sees a successful query whose metadata says it failed. The classic path derives an error from the status; the same check is worth adding on the query path too.

Fixed on both paths. Each terminal handler now derives the error from the status
through the classifier the HTTP path already uses -- `map_analytics_error` and
`map_query_error` -- so the two transports agree on the mapping rather than
growing a second one. Neither proto carries an errors list, and for that case both
classifiers return `internal_server_failure`, which is the fallback
`document_query.cxx` applies when a payload has no error code to classify. A
transport-level error already in `ctx.ec` is more specific and takes precedence.

One qualification, because the obvious form of this check is wrong. It is
conditional on a MetaData message having been received, not on the status value
alone. `Status status = 4` is a plain proto3 field and `analytics_status` numbers
`running` as zero, so "the server said running" and "the server said nothing" are
indistinguishable from the decoded value. That case is reachable rather than
theoretical: `queryserver.go` builds its `psMetaData` only when its own
`result.MetaData()` succeeds, and on failure sends the rows it has and returns
`nil` -- an OK trailer with no metadata. Classifying on the value alone would turn
that stream into `internal_server_failure`. A separate flag records whether
metadata arrived.

The success path is unaffected: the seventeen live query cases still pass against
a real gateway, which is what pins that `status` is `"success"` there rather than
absent.

This was not mentioned in the #1036 response, which is a gap in that reply rather
than in this one.

### 4. Scoped analytics is rejected although the proto supports it (medium)

> **Scoped analytics is rejected** although `AnalyticsQueryRequest` carries `bucket_name` and `scope_name` and couchbase-net-client maps them. Failing closed is correct for something unimplemented, so this is a scope note rather than a defect — but it belongs in the pull request body, since `scope.analytics_query(...)` is a normal way to use the API.

The premise does not hold at the revision this branch pins, and the request for a
body note is met.

At protostellar 36d4d9ab, `AnalyticsQueryRequest` carries neither of those
fields. Field 8 is `reserved // bucket_name` and field 9 was renamed from
`scope_name` to `analytics_scope_name` -- one analytics-scope identifier, not the
bucket/scope pair the core request holds in
`bucket_name`/`scope_name`/`scope_qualifier`.

The observation about couchbase-net-client is accurate but describes an older
schema. Its generated `Analytics.cs` lists `BucketName` and `ScopeName`, and the
proto vendored by couchbase-jvm-clients still declares `bucket_name = 8` /
`scope_name = 9`. Both predate the rename, so a mapping copied from either would
target a schema this client does not speak -- and field 9 keeps its tag while
changing meaning, so such a mapping would be accepted on the wire and interpreted
as something else.

The conclusion is therefore unchanged but rests on a stronger reason: this is not
failing closed on something merely unimplemented, it is refusing to guess which
of two names the single remaining field denotes, where guessing wrong runs the
statement against the wrong scope. Lifting it needs a schema decision, not a
client change. The converter comment now says this, the pull request body states
the gap, and a case asserts each of the three qualification fields is refused
individually -- the previous test covered only `scope_qualifier`, so dropping the
`scope_name` guard alone would have gone unnoticed.

### 5. The live analytics test cannot fail (medium)

> The skip predicate is `feature_not_available`, which is the error the **client** produces for its own unrouted operations. If this commit's routing were deleted entirely, the test would report "Skipped" rather than failing — it cannot distinguish "the gateway does not implement analytics.v1" from "the client never sent anything".

Fixed in the shared fixture header, as suggested.

`skip_unless_service_implemented()` requires the gateway's own answer before it
will skip. An UNIMPLEMENTED reply arrives as a gRPC status and the component
copies its message into `first_error_message`; the client-side refusals --
`cluster.cxx` declining to route a request type, and `can_encode()` rejecting an
option -- build their error context without one, because no status was ever
received. A `feature_not_available` with no message is now a failure.

The suggested environment-variable opt-in was not added, for this reason: an
opt-in records what the operator believes about the gateway, whereas the message
records what the gateway actually said. The second is the evidence the finding
asks for, it needs no configuration, and it cannot be satisfied by a mistaken
setting. Requiring both would make the default run red against a gateway that
correctly reports the service missing, which is a normal configuration today.

The claim that this is falsifiable is measured, and both halves were run. With
the client forced to refuse the request before dispatch -- producing the same
`feature_not_available` the gateway's UNIMPLEMENTED does -- the round-trip case
fails under the new predicate. The old predicate was then restored alongside the
same mutation, and the suite reported

```
Suite "protostellar_live_analytics": 3 passed, 1 skipped
EXIT=0
```

so it does report success against a client that never dispatches. That is the
difference the fix makes, rather than an argument that it should.

Against today's gateway the case skips with the gateway's own words:

```
"analytics_round_trip_against_live_gateway" skipped
  (gateway does not implement analytics.v1
   (unknown service couchbase.analytics.v1.AnalyticsService))
```

That message also settles what the earlier round only inferred. `system.go`
registers kv, query, search, the four admin services, xdcr and routing, and no
analytics service at all; `analyticsserver.go` is a bare
`UnimplementedAnalyticsServiceServer`. The skip is now evidence that a request
reached a gateway and was declined there.

The rest of the suite was rewritten onto `live_cluster_fixture`, which removes
its hand-rolled preamble and its raw `std::getenv`, and three cases were added
that do not need a gateway that serves analytics: a scope-qualified request is
refused, an unsupported option is refused, and an exhausted budget is refused,
each asserted to be stamped with the statement and delivered on the io context
rather than inline. Those three run and assert today. The suite previously had one
case, which always skipped.

The pattern is shared by eight live suites rather than seven -- analytics,
bucket_admin, collection_admin, kv_operations, query_index_admin, search,
search_index_admin and view. The single `feature_not_available` mention in
`live_query.cxx` is a deliberate assertion, not a skip. Only analytics is
converted here; the helper is in the shared header so the other seven can adopt
it in their own commits.

### Note on the macOS failure

> This pull request had a completed `macos (macos-14)` failure at the time of review while the Linux legs were still queued. The analytics files omit the `<cstdint>` and `<chrono>` includes that their query counterparts carry, relying on transitive includes that libc++ and libstdc++ do not agree about — that is the most likely candidate in this diff and is worth checking first.

All three macOS legs pass on the current head, so this no longer reproduces. The
missing includes were added to the analytics test anyway, since the reasoning
holds whether or not it was the cause.

The Windows legs do fail, on all three runners, and the cause is in this diff:

```
live_analytics.cxx(58,28): warning C4996: 'getenv': This function or variable may be unsafe
error C2220: the following warning is treated as an error [cng_live_analytics_test.vcxproj]
```

MSVC deprecates `getenv`, these builds treat warnings as errors, and the Linux
and macOS legs are unaffected -- which is why it surfaced an hour into CI. Line 58
was the hand-rolled `env_or` in the file, so moving the suite onto
`live_cluster_fixture` removes it.

The same warning fires for `live_observability.cxx`, which is not part of this
change. That file sits at the base of the series, below every one of these pull
requests, so the raw `std::getenv` in it fails the Windows legs on all of them.
It is fixed there, in its own commit, using the `safe_getenv` the framework
already provides. Flagging this as a correction to the #1036 reply, which stated
that call site was deliberately left alone: it is not confined to its own commit,
it blocks the whole series on Windows.

### Not raised in the review

Two defects on the analytics refusal paths, both already fixed on the query side
in #1036 and both missed when this commit was written:

- `fail_expired_ctx` was handed a default-constructed `analytics_response`, so a
  request refused for an exhausted budget reported a blank statement and
  client_context_id -- the two fields a caller has to correlate the failure by.
  The response is now stamped on every path that sends nothing.
- The `can_encode` refusal invoked the handler inline instead of posting it,
  re-entering the caller before it holds its `pending_call`. It is posted now,
  and each refusal case asserts the handler did not run on the calling thread.

### Verification

- 21/21 cng cases pass at this commit against a live gateway — cloud-native-gateway
  `1.2.2-123` in front of Couchbase Server `8.1.0-1743`.
- 5/5 mutations killed: four on the converter (both parameter lines, the warnings
  list, the scope guard) and one on the live suite (the falsifiability of the
  skip).
- The whole stack is re-verified per commit -- configure, build, `ctest -L cng`,
  clang-format on that commit's own changed files -- because these land one at a
  time and a green tip proves nothing about the commits under it. 13 of 17 are
  fully green. The other four (26, 27, 28, 30) are green on configure, build and
  format, and fail only `cng_streaming_test` with `Subprocess aborted`: the
  CXXCBC-910 hold-release race, measured at roughly a 5-in-6 abort rate and
  reported separately on #1048. It is not attributable to this change --
  `cng_streaming_test` exists from commit 13 and passes at every commit from 13
  through 25 with these changes applied, and begins aborting only where the
  backpressure hold is introduced.

core/protostellar/analytics_converter.hxx
avsej: Answered in full at https://github.com/couchbase/couchbase-cxx-client/pull/1037#issuecomment-5193094598 — a top-level comment rather than an inline reply, because the push that carries the fixes rewrites these lines and would mark a reply here outdated.

test/cng/protostellar/live_analytics.cxx
avsej: Answered in full at https://github.com/couchbase/couchbase-cxx-client/pull/1037#issuecomment-5193094598 — a top-level comment rather than an inline reply, because the push that carries the fixes rewrites these lines and would mark a reply here outdated.

thejcfactor (approved):

PR #1038, opened 2026-07-28
CXXCBC-899: route search and views over the couchbase2 streaming transport

avsej:
FTS search and map/reduce views were rejected by the HTTP execute front-door
gate on couchbase2. Wire both onto the server-streaming primitive.

Add a view converter (core <-> couchbase.view.v1): the request maps cleanly
(keys, ranges, reduce/group, consistency, order, on_error, namespace) and rows
buffer into document_view_response; raw/query_string/full_set have no couchbase2
home and are reported as feature_not_available. Add a search converter (core
<-> couchbase.search.v1): the core query is an opaque JSON string while the
proto models a structured Query oneof, so this translates the trivial forms and
reports every other shape as unmappable. Hits decode with their fragments, field
values, locations and array positions, so a request that asks for highlighting is
either answered with it or refused, never answered with it silently empty.

The component gains SearchService and ViewService stubs and streaming execute()
overloads; cluster_impl routes both alongside query and analytics. Refusals are
delivered on the io context, like every other completion, and every path that
sends nothing still stamps the request's identifiers, so a failure the caller
never saw dispatched is still attributable to the search or view that caused it.
The search and view
error contexts carry first_error_message so the gateway's status message reaches
the caller, which is also what keeps a client-side refusal distinguishable from a
service the gateway does not implement. The per-service default timeouts move
into a component_timeouts struct so the constructor stops growing an argument per
service.

Views are deprecated in the core API but remain routable, so the CNG headers that
name document_view_request suppress the core-deprecation attribute via the shared
portable macro.

### Supported search surface

The proto models the FTS query as a oneof of 21 types; the core request carries an
opaque JSON blob. This converter translates three of them, so the routing works
but search over couchbase2 is far from complete:

| | supported |
| --- | --- |
| query shapes | `query_string` (optionally boosted), `match_all`, `match_none` |
| not translated | the other 18 arms — `match`, `match_phrase`, `phrase`, `term`, `prefix`, `regexp`, `wildcard`, `boolean`, `boolean_field`, `conjunction`, `disjunction`, `date_range`, `numeric_range`, `term_range`, `doc_id`, and the three geo arms |
| request options | limit, skip, explain, disable_scoring, include_locations, highlight style and fields, fields, collections |
| gated | structured facets, structured sort specs, vector search, mutation-token consistency, `raw` |

Everything untranslated is reported as `feature_not_available` rather than dropped,
so no request runs as something other than what was asked for. A query that is not
JSON at all is the caller's own error and surfaces as `invalid_argument` instead.

Three of twenty-one understates what is reachable and overstates what is missing.
`query_string` is FTS's full query-string syntax, so fielded terms, boolean
operators, ranges, wildcards and phrases are all expressible through the supported
arm — `{"query": "name:couchbase AND age:>40"}` works today. What does not work is
the structured builder API: `couchbase::match_query("foo")` emits `{"match":"foo"}`,
which this converter does not recognise, so the most common way to compose a query
programmatically is refused. Translating the remaining arms is follow-up work, and
nothing external blocks it — the gateway already serves all of them.

### Gateway support

The three streaming services are in three different states:

| service | gateway state |
| --- | --- |
| `search.v1` | served, with all 21 query arms translated server-side |
| `view.v1` | not implemented — no server registered |
| `analytics.v1` | deliberately not served — a stub with no RPC methods, deregistered on purpose |

Answering a search at all takes a cluster running the search service: without one the query comes
back as an internal error rather than as a missing index, so CI's single node now runs `fts`.

So the live search test round-trips a real response against a 1.2.2 gateway and
pins the outcome to hits or a genuine index error; returning hits needs a
provisioned index and is deferred to index admin. The live view test skips on the
gateway's own UNIMPLEMENTED message — and a client that stopped routing views
fails there rather than skipping, which is what makes the skip meaningful.

---
Stacked PR **16/30** in the CNG (`couchbase2://`) transport series. Everything below it in the series is merged, so this branch sits directly on `main` and its single commit is the change under review. Next in the series to merge.

### DIFF
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -20,6 +20,8 @@
 #include "core/error_context/analytics.hxx"
 #include "core/error_context/key_value.hxx"
 #include "core/error_context/query.hxx"
+#include "core/error_context/search.hxx"
+#include "core/error_context/view.hxx"
 #include "core/operations/analytics_response_parsing.hxx"
 #include "core/operations/query_response_parsing.hxx"
 #include "core/protostellar/analytics_converter.hxx"
@@ -27,6 +29,8 @@
 #include "core/protostellar/error_utils.hxx"
 #include "core/protostellar/kv_converter.hxx"
 #include "core/protostellar/query_converter.hxx"
+#include "core/protostellar/search_converter.hxx"
+#include "core/protostellar/view_converter.hxx"
 
 #include <couchbase/error_codes.hxx>
 
@@ -34,6 +38,8 @@
 #include <couchbase/analytics/v1/analytics.grpc.pb.h>
 
 #include <couchbase/kv/v1/kv.grpc.pb.h>
+#include <couchbase/search/v1/search.grpc.pb.h>
+#include <couchbase/view/v1/view.grpc.pb.h>
 
 #include <asio/post.hpp>
 
@@ -51,6 +57,8 @@ namespace couchbase::core::protostellar
 namespace v1 = ::couchbase::kv::v1;
 namespace query_v1 = ::couchbase::query::v1;
 namespace analytics_v1 = ::couchbase::analytics::v1;
+namespace search_v1 = ::couchbase::search::v1;
+namespace view_v1 = ::couchbase::view::v1;
 
 // Defined here rather than in the header so the generated gRPC types stay out of every consumer of
 // component.hxx.
@@ -58,6 +66,8 @@ struct component::stubs {
   std::unique_ptr<v1::KvService::Stub> kv;
   std::unique_ptr<query_v1::QueryService::Stub> query;
   std::unique_ptr<analytics_v1::AnalyticsService::Stub> analytics;
+  std::unique_ptr<search_v1::SearchService::Stub> search;
+  std::unique_ptr<view_v1::ViewService::Stub> view;
 };
 
 namespace
@@ -104,14 +114,13 @@ fail_expired_ctx(asio::io_context& io, Handler& handler, Response response) -> p
 
 component::component(asio::io_context& io, component_config config)
   : io_{ io }
-  , stubs_{ std::make_unique<stubs>(
-      stubs{ v1::KvService::NewStub(config.channel),
-             query_v1::QueryService::NewStub(config.channel),
-             analytics_v1::AnalyticsService::NewStub(config.channel) }) }
+  , stubs_{ std::make_unique<stubs>(stubs{ v1::KvService::NewStub(config.channel),
+                                           query_v1::QueryService::NewStub(config.channel),
+                                           analytics_v1::AnalyticsService::NewStub(config.channel),
+                                           search_v1::SearchService::NewStub(config.channel),
+                                           view_v1::ViewService::NewStub(config.channel) }) }
   , authorization_{ authorization_header(config.credentials) }
-  , default_kv_timeout_{ config.default_kv_timeout }
-  , default_query_timeout_{ config.default_query_timeout }
-  , default_analytics_timeout_{ config.default_analytics_timeout }
+  , timeouts_{ config.timeouts }
   // Initialised last (see the declaration order in the header), so the channel can be moved in.
   , dispatcher_{ io, std::move(config.channel) }
 {
@@ -124,7 +133,7 @@ component::execute(operations::get_request request,
                    utils::movable_function<void(operations::get_response)>&& handler)
   -> pending_call
 {
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -159,7 +168,7 @@ component::execute(operations::get_projected_request request,
 {
   auto proto = std::make_shared<v1::GetRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -190,7 +199,7 @@ component::execute(operations::upsert_request request,
                    utils::movable_function<void(operations::upsert_response)>&& handler)
   -> pending_call
 {
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -220,7 +229,7 @@ component::execute(operations::insert_request request,
                    utils::movable_function<void(operations::insert_response)>&& handler)
   -> pending_call
 {
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -250,7 +259,7 @@ component::execute(operations::replace_request request,
                    utils::movable_function<void(operations::replace_response)>&& handler)
   -> pending_call
 {
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -280,7 +289,7 @@ component::execute(operations::remove_request request,
                    utils::movable_function<void(operations::remove_response)>&& handler)
   -> pending_call
 {
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -312,7 +321,7 @@ component::execute(operations::touch_request request,
 {
   auto proto = std::make_shared<v1::TouchRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -341,7 +350,7 @@ component::execute(operations::exists_request request,
 {
   auto proto = std::make_shared<v1::ExistsRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -371,7 +380,7 @@ component::execute(operations::get_and_lock_request request,
 {
   auto proto = std::make_shared<v1::GetAndLockRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -405,7 +414,7 @@ component::execute(operations::unlock_request request,
 {
   auto proto = std::make_shared<v1::UnlockRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -435,7 +444,7 @@ component::execute(operations::get_and_touch_request request,
 {
   auto proto = std::make_shared<v1::GetAndTouchRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -506,7 +515,7 @@ component::execute(operations::increment_request request,
   }
   auto proto = std::make_shared<v1::IncrementRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -539,7 +548,7 @@ component::execute(operations::decrement_request request,
   }
   auto proto = std::make_shared<v1::DecrementRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -569,7 +578,7 @@ component::execute(operations::append_request request,
 {
   auto proto = std::make_shared<v1::AppendRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -599,7 +608,7 @@ component::execute(operations::prepend_request request,
 {
   auto proto = std::make_shared<v1::PrependRequest>(kv::encode(request));
   const auto id = request.id;
-  const auto timeout = request.timeout.value_or(default_kv_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.key_value);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired(io_, request, handler);
   }
@@ -638,7 +647,7 @@ component::execute(operations::query_request request,
     return response;
   };
 
-  const auto timeout = request.timeout.value_or(default_query_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.query);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired_ctx(io_, handler, stamped());
   }
@@ -742,7 +751,7 @@ component::execute(operations::analytics_request request,
     return response;
   };
 
-  const auto timeout = request.timeout.value_or(default_analytics_timeout_);
+  const auto timeout = request.timeout.value_or(timeouts_.analytics);
   if (timeout <= std::chrono::milliseconds::zero()) {
     return fail_expired_ctx(io_, handler, stamped());
   }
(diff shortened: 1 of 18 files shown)


core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,309 @@
...
+  void OnDone(const grpc::Status& status) override
+  {
+    // Deregister on the gRPC thread before posting: once this returns gRPC is done with the call,
+    // so a concurrent cancel_and_drain() may proceed. The posted task only touches shared state.
+    tracker_->remove(context_.get());
+    // Copy the status into the closure -- it is a reference parameter, so the posted task must own
+    // its own copy rather than capture a reference that dangles once OnDone returns.
+    asio::post(io_, [this, status_copy = status]() {
+      on_done_(status_copy);
+      delete this;
```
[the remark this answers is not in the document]
avsej: Defensible as written. gRPC serializes a single call's reactions (OnReadDone completes before OnDone runs), so every row task is posted before the OnDone task. The SDK contract is a single-threaded io_context, which runs posted tasks in FIFO order, so the self-delete task always runs last -- after every row handler. `delete this` from the terminal post is the standard self-owning-reactor pattern; the row handlers never dereference a freed reactor.

test/cng/protostellar/live_search.cxx
```diff
@@ -0,0 +1,132 @@
...
+  closed.get_future().get();
+
+  if (result.ctx.ec == couchbase::errc::common::feature_not_available) {
+    skip("gateway does not implement search.v1 (feature_not_available)");
+  }
+  // Past the skip guard the gateway served search.v1: the transport encoded the query and decoded a
+  // real response. Without a provisioned FTS index the query returns an index error rather than
+  // hits; verifying actual hits needs index admin (CXXCBC-901).
+  assert_true(result.ctx.ec != couchbase::errc::common::feature_not_available,
+              "gateway serves search.v1 (transport round-trip verified)");
```
avsej: This assertion cannot fail. The guard immediately above it is:

```cpp
if (result.ctx.ec == couchbase::errc::common::feature_not_available) {
  skip("gateway does not implement search.v1 (feature_not_available)");
}
```

`skip()` leaves the function, so past that point `ec != feature_not_available` is necessarily true. Deleting the search routing entirely would leave this test passing. `live_view.cxx` has the same shape — both binaries pass by construction.

Assert the actual contract instead of the negation of the skip predicate. Without a provisioned index the honest expectation is a specific index error:

```cpp
  // Past the skip guard the gateway served search.v1. Without a provisioned FTS index the query
  // returns an index error; anything else means the transport mis-encoded the request.
  assert_true(result.ctx.ec == couchbase::errc::common::index_not_found || !result.ctx.ec,
              "search.v1 round trip yields either hits or a genuine index error");
```

Separately, the skip predicate itself is `feature_not_available` — the error the **client** produces for its own unrouted operations — so a client-side routing regression reports "Skipped" rather than failing. Seven further live suites share this, so it is worth fixing once in the shared fixture: require an explicit opt-in for the gateway-capability skip so an unexpected `feature_not_available` fails by default.

core/protostellar/search_converter.hxx
```diff
@@ -0,0 +1,182 @@
...
+    row.score = hit.score();
+    row.explanation = hit.explanation();
+    for (const auto& location : hit.locations()) {
+      operations::search_response::search_location out;
+      out.field = location.field();
+      out.term = location.term();
+      out.position = location.position();
+      out.start_offset = location.start();
+      out.end_offset = location.end();
+      row.locations.push_back(std::move(out));
```
avsej, line 219: The hit loop decodes `index`, `id`, `score`, `explanation` and `locations` and stops. The proto's `fragments` map (field 6), the per-hit `fields` map, and `Location.array_positions` are never read.

The asymmetry is the problem: `can_encode` does **not** reject a request that asks for highlighting or specific field values. So a caller setting `highlight_style` / `highlight_fields` / `fields` has the request accepted, sent, and answered with rows whose fragments and fields are silently empty — no error, no indication. That is the exact failure mode the request side of this file deliberately prevents.

Either decode them:

```cpp
    for (const auto& [field, fragment] : hit.fragments()) {
      row.fragments[field] = std::vector<std::string>{ fragment.content().begin(),
                                                       fragment.content().end() };
    }
```

or reject the request until they are decoded, so the whole path stays fail-closed:

```cpp
  return ... && !request.highlight_style.has_value() && request.highlight_fields.empty() &&
         request.fields.empty();
```

core/protostellar/search_converter.hxx
```diff
@@ -0,0 +1,182 @@
...
+  } catch (...) {
+    return false;
+  }
+  if (!parsed.is_object()) {
+    return false;
+  }
+  // Only the exact single-key trivial shapes map cleanly. A richer object (extra top-level keys
+  // such as boost/analyzer alongside the recognized one) would silently lose those fields, so
+  // report it as unmappable and let the component surface feature_not_available.
+  if (parsed.get_object().size() != 1) {
```
avsej: The single-key rule is a good decision and the comment explains it correctly — a richer object would silently lose the extra keys, so reporting it as unmappable is right.

Two notes rather than defects:

1. The supported set is three query shapes (`match_all`, `match_none`, bare `{"query": "..."}`) out of twenty-one, with no facets and no sort specs. That is fail-closed and fine, but "route search over the couchbase2 streaming transport" in the title and ticket reads as though search works. Stating the supported surface in the body, and tracking the remaining arms, sets the right expectation.

2. A `boost` alongside an otherwise-supported query makes it unmappable, because `boost` counts as a second top-level key — yet `boost` maps to a field on every proto query arm. Treating it as a modifier rather than an unrecognised sibling would widen the supported set at very little cost.

avsej (commented):
## Review — PR #1038 (`CXXCBC-899`)

This commit routes both FTS search and views over the streaming transport. At 1118 insertions across 11 files it
is the largest in the series and it bundles two independent services; splitting it would have made both easier to
review, and the views half in particular deserves a separate discussion (see the note below).

The request side is deliberately and correctly fail-closed: `can_encode` and `apply_query` reject anything the
converter cannot express, so unsupported shapes produce `feature_not_available` rather than a silently degraded
query. The reasoning for the single-key rule in `apply_query` is written down and is right.

The response side breaks that convention in the opposite direction, and one of the new tests cannot fail.

### Strengths

- `apply_query`'s single-key rule is a genuinely good decision, and the comment explains it: a richer query object
  with extra top-level keys alongside a recognised one would silently lose those keys, so it is reported as
  unmappable instead. That is the right instinct applied at the right place.
- `can_encode` rejecting facets, sort specs, mutation state and vector search wholesale is honest — those are not
  partially supported, they are not supported, and the caller finds out.
- The view converter handles the design-document namespace and the JSON key encodings without inventing its own
  serialisation.

### Findings

|---|---|---|---|---|
| 1 | high | test-coverage | The only assertion in `live_search` and `live_view` is unfalsifiable | `test/cng/protostellar/live_search.cxx:113` |
| 2 | high | correctness | Per-hit `fields` and `fragments` are never decoded, so requested values silently vanish while the request is accepted | `core/protostellar/search_converter.hxx:161` |
| 3 | medium | scope | Three of twenty-one FTS query types are mapped, which the pull request title does not convey | `core/protostellar/search_converter.hxx:77` |

#### 1. The only assertion in `live_search` and `live_view` is unfalsifiable (high)

```cpp
              "gateway serves search.v1 (transport round-trip verified)");
```

The lines immediately above are:

```cpp
if (result.ctx.ec == couchbase::errc::common::feature_not_available) {
  skip("gateway does not implement search.v1 (feature_not_available)");
}
assert_true(result.ctx.ec != couchbase::errc::common::feature_not_available, ...);
```

`skip()` leaves the function, so past that guard `ec != feature_not_available` is necessarily true. The assertion
cannot fail under any circumstances, including a change that deleted the search routing entirely. `live_view.cxx`
has the same shape. These two binaries pass by construction.

The fix is to assert the actual contract — the set of outcomes that prove the gateway parsed the request — rather
than the negation of the skip predicate. Without a provisioned index the honest expectation is a specific
index-level error, not a generic one:

```cpp
// Past the skip guard the gateway served search.v1. Without a provisioned FTS index the query
// returns an index error; anything else means the transport mis-encoded the request.
assert_true(result.ctx.ec == couchbase::errc::common::index_not_found ||
              !result.ctx.ec,
            "search.v1 round trip yields either hits or a genuine index error");
```

A related problem affects seven further live suites in this series, including the ones added here: they skip on
`feature_not_available`, which is the error the client produces for its *own* unrouted operations. A client-side
routing regression therefore reports "Skipped" rather than failing. Fixing the skip predicate once in the shared
fixture — for example requiring an explicit opt-in for the gateway-capability skip, so an unexpected
`feature_not_available` fails by default — would cover all of them.

(inline comment on `test/cng/protostellar/live_search.cxx:113`)

#### 2. Per-hit `fields` and `fragments` are never decoded (high)

```cpp
row.locations.push_back(std::move(out));
```

The hit loop decodes `index`, `id`, `score`, `explanation` and `locations`, and stops there. The proto's
`fragments` map (field 6) and the per-hit `fields` map are not read, and `Location.array_positions` is not read
either. The header comment acknowledges the omission and defers it.

The problem is that `can_encode` does not reject a request that asks for them. A caller setting `highlight_style`
and `highlight_fields`, or requesting specific `fields`, has that request accepted, sent, and answered with rows
whose `fields` and `fragments` are silently empty — no error, no indication. That is precisely the failure mode
the request side of this same file goes out of its way to prevent.

Either decode the maps:

```cpp
for (const auto& [field, fragment] : hit.fragments()) {
  row.fragments[field] = std::vector<std::string>{ fragment.content().begin(),
                                                   fragment.content().end() };
}
if (!hit.fields().empty()) {
  tao::json::value fields = tao::json::empty_object;
  for (const auto& [name, value] : hit.fields()) {
    fields[name] = value;
  }
  row.fields = utils::json::generate(fields);
}
```

or reject the request in `can_encode` until they are decoded, so the whole path is fail-closed:

```cpp
return ... && request.highlight_style == std::nullopt && request.highlight_fields.empty() &&
       request.fields.empty();
```

Either is acceptable; the current asymmetry is not.

(inline comment on `core/protostellar/search_converter.hxx:161`)

#### 3. Three of twenty-one FTS query types are mapped (medium)

```cpp
if (parsed.get_object().size() != 1) {
```

`apply_query` maps `match_all`, `match_none`, and a bare `{"query": "<string>"}`. Everything else — `match`,
`match_phrase`, `term`, `prefix`, `regexp`, `wildcard`, `numeric_range`, `date_range`, the three geo forms,
`boolean`, `conjunction`, `disjunction`, `doc_id`, and any query carrying a `boost` — returns
`feature_not_available`, as do all facets and sort specifications.

That is fail-closed and deliberate, and it is not a defect. It is a scope statement that the pull request title
and the ticket do not make: "route search and views over the couchbase2 streaming transport" reads as though
search works, when what is supported is three query shapes with no facets and no sorting. Stating the supported
surface explicitly in the body, and tracking the remaining arms, would set the right expectation for anyone
reading this later or deciding whether the feature is usable.

The single-key rule has one avoidable consequence worth noting: a `boost` alongside an otherwise-supported query
makes it unmappable, even though `boost` maps to a field on every proto query arm. Handling `boost` as a modifier
rather than as an unrecognised sibling key would widen the supported set at very little cost.

(inline comment on `core/protostellar/search_converter.hxx:77`)

### Note on views over couchbase2

Views are wired here, and gocb, couchbase-jvm-clients and couchbase-net-client all deliberately refuse them on
this protocol. Views are deprecated server-side, and the proto surface exists but is not something the other
clients expose. Before this lands it is worth confirming that exposing views over couchbase2 is intended rather
than incidental — if the gateway's view support is not something the product intends to carry forward, this is
API surface that will have to be withdrawn later, which is harder than not adding it.

### Verdict

Needs changes. Finding 1 is the priority: two of the test binaries this commit adds cannot fail, so the routing
they exist to verify is effectively untested. Finding 2 is a small decode addition or a small `can_encode`
addition, whichever the team prefers.

avsej:
All four open threads are addressed, along with the findings that appeared only under "Comments
suppressed due to low confidence". Closing them turned up two defects the review did not name, both
fixed here and described at the end.

### Gateway support, checked rather than assumed

The title groups search and views together, but the three streaming services are in three different
states, and that distinction drives several of the answers below:

| service | gateway state |
| --- | --- |
| `search.v1` | **served** — registered in the gateway, and its server translates all 21 query arms, facets and sort specs |
| `view.v1` | **not implemented** — no server, no registration; only a test client references the proto |
| `analytics.v1` | **deliberately not served** — the server type is a stub with no RPC methods, and its registration was removed on purpose |

Views resemble analytics in effect but not in cause, and search resembles neither.

### 1. `live_kv.cxx` — component_timeouts aggregate initialization

> `component` now expects a `component_timeouts` struct as the last constructor argument. Passing
> `20000ms` directly won’t compile; wrap it in braces so it aggregate-initializes the struct

Outdated. Both sites already read `config.timeouts = { 20000ms };` and the suite compiles and runs.
No change.

### 2. The assertion that cannot fail

> This assertion cannot fail. […] `skip()` leaves the function, so past that point
> `ec != feature_not_available` is necessarily true. Deleting the search routing entirely would leave
> this test passing. `live_view.cxx` has the same shape — both binaries pass by construction.

Accepted. Both suites were rewritten onto the shared live fixture and
`skip_unless_service_implemented()`.

Closing this exposed a blocker the helper could not work around: `error_context::search` and
`error_context::view` had no field carrying the service's message, and the completion handlers
discarded `status.error_message()` outright. Query and analytics both keep it in
`first_error_message`. The effect was that a failed search over couchbase2 reported a bare error code
with no diagnostic at all — `http_status` and `http_body` stay zero and empty on that transport. Both
contexts now carry `first_error_message`, populated from the gRPC status. That restores the
diagnostic and is also what lets the skip be evidence-based rather than a belief about the gateway.

The search case no longer skips against a current gateway. It asserts hits or `index_not_found`, and
that any error carries the gateway's own message.

> Separately, the skip predicate itself is `feature_not_available` — the error the **client** produces
> for its own unrouted operations — so a client-side routing regression reports "Skipped" rather than
> failing.

Verified by mutation rather than by inspection: removing `document_view_request` from the routing
list makes the view suite fail with *"view.v1 reported feature_not_available with no gateway message,
so the request was refused by the client and never reached the gateway"* instead of skipping.

> Seven further live suites share this, so it is worth fixing once in the shared fixture

The shared helper already exists, and is what both suites now use. The remaining suites belong to
other commits in the series and are left to them.

### 3. Hits decoded only in part

> The hit loop decodes `index`, `id`, `score`, `explanation` and `locations` and stops. The proto's
> `fragments` map (field 6), the per-hit `fields` map, and `Location.array_positions` are never read.
>
> The asymmetry is the problem: `can_encode` does **not** reject a request that asks for highlighting
> or specific field values.

Accepted, and resolved by decoding rather than by widening the gate — the gateway serves search, so
the decode side is verifiable.

`fragments` and `array_positions` map directly. `fields` needed care: the proto keys each field to
its value as raw JSON bytes, while the core row carries the whole set as a single JSON string, so the
object is reassembled. That matches what the gateway sends and what the JVM client does with the same
message. A value that is not JSON is kept verbatim as a string rather than allowed to throw, because
`decode()` runs on the io_context out of a stream callback where an escaping exception would take
down the loop.

Two unit cases cover it: one for the populated shape, one pinning that an absent fragment set leaves
`fields` empty and `array_positions` unset, so "not requested" stays distinguishable from "did not
match".

### 4. Supported query surface, and boost

> The supported set is three query shapes […] out of twenty-one […] "route search over the couchbase2
> streaming transport" in the title and ticket reads as though search works. Stating the supported
> surface in the body, and tracking the remaining arms, sets the right expectation.

Accepted. The description now states the supported surface. The header comment previously implied the
remaining arms were waiting on a gateway that serves search; that is not the case and it has been
corrected.

> A `boost` alongside an otherwise-supported query makes it unmappable, because `boost` counts as a
> second top-level key — yet `boost` maps to a field on every proto query arm. Treating it as a
> modifier rather than an unrecognised sibling would widen the supported set at very little cost.

Accepted in part.

The premise is right and worth fixing: `match_all_query().boost(2)` encodes as
`{"match_all":{},"boost":2}`, so an ordinary boosted query was unroutable. `boost` is now read as a
modifier of the recognised key rather than as an unrecognised sibling.

The second half does not hold against the pinned schema. `MatchAllQuery` and `MatchNoneQuery` are
empty messages with no `boost` field; only `QueryStringQuery` has one. Boost is therefore accepted for
`query_string` and still refused for `match_all`/`match_none`, since flattening it there would run the
query at a weight the caller did not ask for. A unit case pins each direction.

### Findings from the review bodies

Twenty-three entries never became threads. Each was checked against the current code.

Fixed here:

- Invalid query JSON surfaced as `feature_not_available`. A query the caller failed to serialize is
  now `invalid_argument`; a well-formed shape the converter cannot translate stays
  `feature_not_available`. Both directions are pinned by unit and live cases.
- Extra top-level keys are still refused, and `boost` is now handled explicitly rather than falling
  into the same bucket.

Already correct, no change:

- The stray `)` in the Capella warning is the raw-string terminator, not content; the message
  concatenates correctly.
- `origin::operator=` does copy `connection_string_`. The node-iteration reset is deliberate and the
  reason is documented above the operator.
- `component.hxx` already uses `push_macro`/`pop_macro`.
- `query_converter.hxx` already includes `<limits>`, and all four narrowed tuning fields are already
  guarded by `fits_uint32()`. Search's own `limit`/`skip` are `uint32` on both sides; the view ones
  are `uint64` in core and are guarded.
- `view_converter.hxx` already range-checks `limit`/`skip`.
- `streaming.cxx` already calls `server_->Wait()`.
- The `live_kv.cxx` findings describe a conflicted snapshot; the file has no conflict markers and
  compiles.

Real, but not this PR — all present before the series, so they are left for their own change rather
than folded into a search-and-views commit:

- IPv6 literals are not bracketed when the gRPC target is built as `hostname + ":" + port`, so a
  couchbase2:// bootstrap on an IPv6 address yields an invalid authority. `parse_connection_string`
  stores IPv6 addresses without brackets, so this is reachable.
- `get_nodes()` allocates a formatted vector purely to test a count.
- The component test's in-process server is shut down without `Wait()`, unlike its two siblings.

### Not in the review

1. **Search and view refusals were delivered inline out of `execute()`.** Query and analytics post
   theirs to the io context, with a comment giving the reason: a synchronous callback re-enters the
   caller before it has its `pending_call`. Both now post, and the live cases assert the handler did
   not run on the calling thread.
2. **The gateway's status message was discarded** on both paths, as described under thread 2.

One observation left alone: the routing front door also refuses an unrouted request type inline. It
is reachable only for request types this series has not wired yet, each of which a later commit
wires, so it is noted rather than changed here.

### Verification

Measured: `ctest -L cng` 25/25 at this commit; the search converter suite 9/9. Against a 1.2.2
gateway, the live search suite passes 3/3 without skipping, view is 1 passed and 1 skipped on the
gateway's own UNIMPLEMENTED message, query 17/17, analytics 3 passed and 1 skipped, KV 2/2. The
search round trip returns `index_not_found` with the message `Search index 'cng-index' not found.`,
which only a gateway that ran the query can produce.

Negative controls: four mutations of the converter each killed exactly one unit case, and deleting
the view routing turned the live view suite from "1 passed, 1 skipped" into two failures.

Not verified end to end: the new hit decoding. `fragments`, per-hit `fields` and `array_positions`
are covered by unit cases built from hand-constructed protos, and the raw-JSON shape of `fields` is
taken from the gateway's row construction and the JVM client rather than from a measured round trip.
A live check needs a queryable FTS index, which was not available.

Not verified: CI, including the Windows, macOS and sanitizer legs. Also not verified: the branch of
the live view case past the skip — no released gateway serves `view.v1`, so it cannot be exercised,
and the case says so rather than asserting a mapping that cannot run.

test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  component comp{ io, channel, credentials, 20000ms };
```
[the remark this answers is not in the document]
avsej: Outdated. Both sites already read `config.timeouts = { 20000ms };`, and the suite compiles and runs.
No change needed.

test/cng/protostellar/live_search.cxx
avsej: Fixed. Both suites were rewritten onto the shared live fixture and `skip_unless_service_implemented()`.

Closing this needed a change beyond the tests: `error_context::search` and `error_context::view` had
no field carrying the service's message, and the completion handlers discarded
`status.error_message()`, so there was nothing to distinguish a gateway UNIMPLEMENTED from a
client-side refusal — and nothing to diagnose a failed search with, since `http_status`/`http_body`
stay empty on this transport. Both contexts now carry `first_error_message`.

The search case no longer skips at all against a current gateway: it asserts hits or
`index_not_found`. The view case skips only on the gateway's own message. Checked by mutation rather
than inspection — removing `document_view_request` from the routing list makes the view suite fail
instead of skipping.

Detail in the summary comment on the PR.

core/protostellar/search_converter.hxx
avsej: Fixed by decoding rather than by widening the gate, since the gateway does serve search and the
decode side is verifiable.

`fragments` and `array_positions` map directly. `fields` is a proto `map<string, bytes>` whose values
are raw JSON, while the core row carries the whole set as one JSON string, so the object is
reassembled — matching what the gateway sends and what the JVM client does with the same message. A
value that is not JSON is kept verbatim as a string rather than allowed to throw, because `decode()`
runs on the io_context out of a stream callback.

Two unit cases: one for the populated shape, one pinning that an absent fragment set leaves `fields`
empty and `array_positions` unset, so "not requested" stays distinguishable from "did not match".

core/protostellar/search_converter.hxx
avsej: Both points addressed.

The supported surface is now stated in the PR description, including the 18 untranslated arms.

On boost: accepted in part. The premise is right and was worth fixing — `match_all_query().boost(2)`
encodes as `{"match_all":{},"boost":2}`, so an ordinary boosted query was unroutable; `boost` is now
read as a modifier of the recognised key. But it does not map to a field on every arm:
`MatchAllQuery` and `MatchNoneQuery` are empty messages in the pinned schema, so a boost on either
has nowhere to go. It is accepted for `query_string` and still refused for the other two, since
flattening it there would run the query at a weight the caller did not ask for. A unit case pins each
direction.

core/protostellar/component.cxx
```diff
@@ -827,4 +836,123 @@ component::execute(operations::analytics_request request,
...
 }
 
+auto
+component::execute(operations::search_request request,
+                   utils::movable_function<void(operations::search_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(timeouts_.search);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired_ctx(io_, handler, operations::search_response{});
```
thejcfactor: Shouldn't we also utilize the same stamped helper pattern instead of returning the bare `search_response{}`?

I think along these lines do we want the same `an_expired_budget_is_refused_on_the_io_context()` test in live_search like we have w/ query/analytics?

core/protostellar/component.cxx
```diff
@@ -827,4 +836,123 @@ component::execute(operations::analytics_request request,
...
+}
+
+auto
+component::execute(operations::document_view_request request,
+                   utils::movable_function<void(operations::document_view_response)>&& handler)
+  -> pending_call
+{
+  const auto timeout = request.timeout.value_or(timeouts_.view);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired_ctx(io_, handler, operations::document_view_response{});
```
thejcfactor: Shouldn't we also utilize the same stamped helper pattern instead of returning the bare `document_view_response{}`?

I think along these lines do we want the same `an_expired_budget_is_refused_on_the_io_context()` test in live_view like we have w/ query/analytics?

core/protostellar/component.cxx
```diff
@@ -827,4 +836,123 @@ component::execute(operations::analytics_request request,
...
+    },
+    [handler = std::move(handler), response, proto, index_name, query_str, client_context_id](
+      grpc::Status status) mutable {
+      (void)proto; // kept only to keep the request alive for the call
+      response->ctx.ec = map_status(status, operation_kind::read_only);
+      response->ctx.first_error_message = error_message(status);
+      response->ctx.index_name = std::move(index_name);
+      response->ctx.query = std::move(query_str);
+      response->ctx.client_context_id = std::move(client_context_id);
+      handler(std::move(*response));
```
thejcfactor, line 905: I think search can fall into the same situation query/analytics can if the metadata indicates errors from individual partitions?  Should we also apply `map_search_errors()` type logic like we do for query/analytics?

core/protostellar/component.cxx
avsej: Agreed on both counts, and done.

`search` now builds its response through a `stamped()` helper, as the query overload does, so the
expired-budget path and the unmappable-query refusal both carry `index_name`, `query` and
`client_context_id` instead of returning a bare response. That matters most on exactly this path:
nothing was sent, so there is no gateway response to supply those fields later, and a caller
correlating by index or context id would have had nothing to correlate on.

`an_expired_budget_is_refused_on_the_io_context()` is added to `live_search`, matching the
query/analytics shape — it asserts the unambiguous timeout, that the response still names the index
and the caller's context id, and that the completion arrives on the io context rather than inline
out of `execute()`. Reverting the stamping turns it red.

core/protostellar/component.cxx
avsej: Same treatment, for the same reason: the view overload now builds its responses through a
`stamped()` helper, so the expired-budget path and the `can_encode()` refusal both carry
`design_document_name`, `view_name` and `client_context_id`.

`an_expired_budget_is_refused_on_the_io_context()` is added to `live_view` alongside it. It does not
depend on the gateway serving views — the refusal happens before anything is dispatched — so unlike
the round-trip case it runs rather than skips, which makes it the case that keeps this path honest.

core/protostellar/component.cxx
avsej: Worth separating two things here, because search differs from query/analytics in what the metadata
means rather than in whether it is read.

The partition errors are already carried: `decode()` copies `MetaData.errors` into
`meta.errors` and `error_partition_count` into the metrics, so a caller can see which pindexes
failed and how many.

What is deliberately not done is deriving an error code from them, and that is where search parts
company with query/analytics. There, a metadata error list means the statement failed, which is why
`map_analytics_error()` turns it into a `ctx.ec`. An FTS response reporting failed partitions is a
*successful* query with incomplete results: the HTTP path records the same two fields and leaves
`ctx.ec` unset (`document_search.cxx` — it populates `meta.errors` from `status.errors` and returns
the response with its hits). Applying `map_search_errors()`-style logic over couchbase2 would make it
fail a query that `couchbase://` answers, which is the sort of divergence this series is trying to
avoid.

Since nothing asserted that, the behaviour was unpinned — a test now covers it:
`decode_carries_partial_partition_failures` builds a response with two failed pindexes, and asserts
both error messages and the failed count survive decoding while `ctx.ec` stays unset. Dropping the
error map turns it red.

If the desired semantic is that a partial FTS result should surface as an error, that is worth doing
deliberately and on both transports rather than only on this one, and I would rather file it than
diverge here.

avsej:
Three review points addressed inline, plus the CI failure they sit next to.

**`cng_live_search_test` was failing in CI, and it was the cluster, not the client.** The job
allocates a single node running `kv, n1ql, index` — no search service at all. The case queries an
index that was never created and accepts the gateway's `index_not_found`, but with no fts service
there is nothing to report the index missing, so the request comes back as an internal error and the
assertion fails. Confirmed by pointing the case at a non-existent index on a cluster that *does* run
fts: it passes there. The CI cluster now includes `fts`.

No search quota is raised with it, deliberately: no index is built in CI, so the default is enough.
An index that has to serve queries is a different matter — under the operator's 256 MiB default an
index is accepted and then never builds a usable partition, so every query against it fails with
`pindex not available`. That is a problem for search index management rather than for this PR, and
`test/cng/README.md` says so.

**Stamped responses.** Both `search` and `document_view` now build their responses through a
`stamped()` helper, as the query overload does, so the expired-budget path and the converter refusal
carry the request's identifiers instead of returning a bare response. That matters most precisely
there: nothing was sent, so no gateway response can supply those fields afterwards.
`an_expired_budget_is_refused_on_the_io_context()` is added to both live suites, matching the
query/analytics shape.

**Search partition errors** are already carried — `decode()` copies `MetaData.errors` and
`error_partition_count` — and deriving a `ctx.ec` from them is deliberately not done. A metadata
error list means something different for search than for query/analytics: there it means the
statement failed, here it means the query succeeded with incomplete results, and the HTTP path
records the same two fields and leaves `ctx.ec` unset. Mapping them would make couchbase2 fail a
query that `couchbase://` answers. The behaviour was unpinned, so a test now covers it.

Mutation-tested: reverting either stamping, or dropping the partition-error map, turns the
corresponding new case red.

thejcfactor (approved):

PR #1039, opened 2026-07-28
CXXCBC-900: route bucket management over the couchbase2 transport

avsej:
Bucket management was rejected by the HTTP execute front-door gate on
couchbase2. Wire the six bucket admin operations onto the unary gRPC path via
admin.bucket.v1.

Add a bucket-admin converter mapping core bucket_settings to and from the proto:
create/update encode the mutable settings (ram quota, replicas, flush, max
expiry, durability, eviction, compression, history retention), with bucket type,
replica indexes, storage backend, and conflict resolution set only on create
since they are immutable afterward. Every field the encode side writes is read
back, and the optional ones through their presence bit, so get-modify-update
does not reset a setting the caller never touched. A history retention size the
32-bit core field cannot hold is left unset rather than wrapped. The proto enums
have no `unknown` sentinel, so core `unknown` values fall through to the server
default; memcached buckets have no couchbase2 representation and are reported as
feature_not_available, on the io context like every other completion.
admin.bucket.v1 has no GetBucket RPC, so bucket_get filters the bucket list by
name. The component gains a BucketAdminService stub and six unary execute()
overloads (get_all/get/create/update/drop/flush); cluster_impl routes them.
component_timeouts gains a management entry, taken from the cluster options like
every other service.

FLUSH_DISABLED joins the mapped FAILED_PRECONDITION violations, so flushing a
bucket that has flush turned off reports bucket_not_flushable rather than a
generic server failure.

The gateway (1.2.2) serves admin.bucket.v1. The live suite covers the settings
round trip, an update/flush/drop lifecycle, and the bucket-specific error codes
the gateway's ResourceInfo details select; converter unit tests cover the
mapping and its presence-sensitive fields. A minimum durability level is checked
separately, since a cluster with too few data servers cannot host a bucket that
asks for one, and that must cost only its own case.

The live cases create a bucket of their own, which does not fit beside the test
bucket at the operator's default quota, so CI raises the data service quota
before creating it, and allocates three KV nodes so the durability case runs
there rather than skipping.

The CNG job reports the runner's cpu, memory and disk before allocating, and
runs ctest verbosely: several cluster-only cases skip themselves on a condition
of the cluster, and a green log that does not say which ones ran cannot be told
apart from one that stopped exercising them.

---
Stacked PR **17/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1038 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -776,12 +776,18 @@ public:
       if constexpr (std::is_same_v<Request, operations::query_request> ||
                     std::is_same_v<Request, operations::analytics_request> ||
                     std::is_same_v<Request, operations::search_request> ||
-                    std::is_same_v<Request, operations::document_view_request>) {
+                    std::is_same_v<Request, operations::document_view_request> ||
+                    std::is_same_v<Request, operations::management::bucket_get_all_request> ||
+                    std::is_same_v<Request, operations::management::bucket_get_request> ||
+                    std::is_same_v<Request, operations::management::bucket_create_request> ||
+                    std::is_same_v<Request, operations::management::bucket_update_request> ||
+                    std::is_same_v<Request, operations::management::bucket_drop_request> ||
+                    std::is_same_v<Request, operations::management::bucket_flush_request>) {
         component->execute(std::move(request), std::forward<Handler>(handler));
         return;
       } else {
-        // Management over couchbase2 is not wired yet. Reject cleanly rather than falling through
-        // to the MCBP session manager, which is never bootstrapped on this path.
+        // Remaining management ops over couchbase2 are not wired yet. Reject cleanly rather than
+        // falling through to the MCBP session manager, which is never bootstrapped on this path.
         return handler(
           request.make_response({ errc::common::feature_not_available }, response_type{}));
       }
@@ -944,7 +950,8 @@ public:
             options.query_timeout,
             options.analytics_timeout,
             options.search_timeout,
-            options.view_timeout
+            options.view_timeout,
+            options.management_timeout
           } });
     }
     CB_LOG_INFO(R"(open couchbase2 cluster, id: "{}", endpoint: "{}")", id_, endpoint);
--- a/core/protostellar/bucket_admin_converter.hxx
+++ b/core/protostellar/bucket_admin_converter.hxx
@@ -1,0 +1,301 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates core bucket management settings to and from couchbase.admin.bucket.v1 (unary admin
+// RPCs). The proto enums have no `unknown` sentinel, so the core `unknown` values fall back to the
+// server default on encode. memcached buckets have no couchbase2 representation;
+// apply_create_only_settings() returns false for them so the component surfaces
+// feature_not_available.
+//
+// Every field the encode side writes is read back by decode_bucket(), because get-modify-update is
+// the ordinary way these settings are changed: a field that decodes to a default is sent back as a
+// default on the next update and silently resets the bucket.
+
+#include "core/management/bucket_settings.hxx"
+#include "core/operations/management/bucket_get_all.hxx"
+
+#include <couchbase/durability_level.hxx>
+
+#include <couchbase/admin/bucket/v1/bucket.pb.h>
+#include <couchbase/kv/v1/kv.pb.h>
+
+#include <cstdint>
+#include <limits>
+#include <optional>
+
+namespace couchbase::core::protostellar::bucket_admin
+{
+namespace v1 = ::couchbase::admin::bucket::v1;
+namespace mgmt = ::couchbase::core::management::cluster;
+
+[[nodiscard]] inline auto
+to_proto_durability(couchbase::durability_level level)
+  -> std::optional<::couchbase::kv::v1::DurabilityLevel>
+{
+  switch (level) {
+    case couchbase::durability_level::none:
+      return std::nullopt;
+    case couchbase::durability_level::majority:
+      return ::couchbase::kv::v1::DURABILITY_LEVEL_MAJORITY;
+    case couchbase::durability_level::majority_and_persist_to_active:
+      return ::couchbase::kv::v1::DURABILITY_LEVEL_MAJORITY_AND_PERSIST_TO_ACTIVE;
+    case couchbase::durability_level::persist_to_majority:
+      return ::couchbase::kv::v1::DURABILITY_LEVEL_PERSIST_TO_MAJORITY;
+  }
+  return std::nullopt;
+}
+
+// Populate the bucket-settings fields that are mutable on both Create and Update requests.
+template<typename Request>
+void
+apply_settings(const mgmt::bucket_settings& settings, Request& proto)
+{
+  proto.set_bucket_name(settings.name);
+  // ram_quota_mb == 0 means "unset" in the core settings (create defaults it to 100, update omits
+  // it); only send an explicit value so an unset quota does not overwrite the server-side value.
+  if (settings.ram_quota_mb != 0) {
+    proto.set_ram_quota_mb(settings.ram_quota_mb);
+  }
+
+  // Omitted when the caller expressed no opinion, so the server applies its own default. Sending an
+  // explicit zero instead would create a bucket with no replicas on their behalf.
+  if (settings.num_replicas.has_value()) {
+    proto.set_num_replicas(*settings.num_replicas);
+  }
+  if (settings.flush_enabled.has_value()) {
+    proto.set_flush_enabled(*settings.flush_enabled);
+  }
+  if (settings.max_expiry.has_value()) {
+    proto.set_max_expiry_secs(*settings.max_expiry);
+  }
+  if (settings.minimum_durability_level.has_value()) {
+    if (auto durability = to_proto_durability(*settings.minimum_durability_level);
+        durability.has_value()) {
+      proto.set_minimum_durability_level(*durability);
+    }
+  }
+  switch (settings.eviction_policy) {
+    case mgmt::bucket_eviction_policy::full:
+      proto.set_eviction_mode(v1::EVICTION_MODE_FULL);
+      break;
+    case mgmt::bucket_eviction_policy::value_only:
+      proto.set_eviction_mode(v1::EVICTION_MODE_VALUE_ONLY);
+      break;
+    case mgmt::bucket_eviction_policy::no_eviction:
+      proto.set_eviction_mode(v1::EVICTION_MODE_NONE);
+      break;
+    case mgmt::bucket_eviction_policy::not_recently_used:
+      proto.set_eviction_mode(v1::EVICTION_MODE_NOT_RECENTLY_USED);
+      break;
+    case mgmt::bucket_eviction_policy::unknown:
+      break; // leave server default
+  }
+  switch (settings.compression_mode) {
+    case mgmt::bucket_compression::off:
+      proto.set_compression_mode(v1::COMPRESSION_MODE_OFF);
+      break;
+    case mgmt::bucket_compression::active:
+      proto.set_compression_mode(v1::COMPRESSION_MODE_ACTIVE);
+      break;
+    case mgmt::bucket_compression::passive:
+      proto.set_compression_mode(v1::COMPRESSION_MODE_PASSIVE);
+      break;
+    case mgmt::bucket_compression::unknown:
+      break;
+  }
+  if (settings.history_retention_collection_default.has_value()) {
+    proto.set_history_retention_collection_default(*settings.history_retention_collection_default);
+  }
+  if (settings.history_retention_bytes.has_value()) {
+    proto.set_history_retention_bytes(*settings.history_retention_bytes);
+  }
+  if (settings.history_retention_duration.has_value()) {
+    proto.set_history_retention_duration_secs(*settings.history_retention_duration);
+  }
+}
+
+// Bucket type, replica indexes, storage backend, and conflict resolution are immutable after
+// creation, so they are only set on CreateBucketRequest. Returns false when the bucket type has no
+// couchbase2 representation (memcached).
+[[nodiscard]] inline auto
+apply_create_only_settings(const mgmt::bucket_settings& settings, v1::CreateBucketRequest& proto)
+  -> bool
+{
+  switch (settings.bucket_type) {
+    case mgmt::bucket_type::couchbase:
+      proto.set_bucket_type(v1::BUCKET_TYPE_COUCHBASE);
+      break;
+    case mgmt::bucket_type::unknown:
+      // Leave bucket_type unset so the gateway applies its server-side default rather than
+      // silently forcing a couchbase bucket.
+      break;
+    case mgmt::bucket_type::ephemeral:
+      proto.set_bucket_type(v1::BUCKET_TYPE_EPHEMERAL);
+      break;
+    case mgmt::bucket_type::memcached:
+      return false;
+  }
+  // A bucket must be created with a RAM quota. Core uses ram_quota_mb == 0 to mean "unset" (and
+  // apply_settings() omits it in that case), so mirror the classic create path and fall back to the
+  // historical 100 MiB default rather than sending a zero quota.
+  if (settings.ram_quota_mb == 0) {
+    proto.set_ram_quota_mb(100);
+  }
+  if (settings.replica_indexes.has_value()) {
+    proto.set_replica_indexes(*settings.replica_indexes);
+  }
+  switch (settings.storage_backend) {
+    case mgmt::bucket_storage_backend::couchstore:
+      proto.set_storage_backend(v1::STORAGE_BACKEND_COUCHSTORE);
+      break;
+    case mgmt::bucket_storage_backend::magma:
+      proto.set_storage_backend(v1::STORAGE_BACKEND_MAGMA);
+      break;
+    case mgmt::bucket_storage_backend::unknown:
+      break;
+  }
+  switch (settings.conflict_resolution_type) {
+    case mgmt::bucket_conflict_resolution::timestamp:
+      proto.set_conflict_resolution_type(v1::CONFLICT_RESOLUTION_TYPE_TIMESTAMP);
+      break;
+    case mgmt::bucket_conflict_resolution::sequence_number:
+      proto.set_conflict_resolution_type(v1::CONFLICT_RESOLUTION_TYPE_SEQUENCE_NUMBER);
+      break;
+    case mgmt::bucket_conflict_resolution::custom:
+      proto.set_conflict_resolution_type(v1::CONFLICT_RESOLUTION_TYPE_CUSTOM);
+      break;
+    case mgmt::bucket_conflict_resolution::unknown:
+      break;
+  }
+  return true;
+}
+
+// Decode one ListBucketsResponse bucket back into core settings.
+[[nodiscard]] inline auto
+decode_bucket(const v1::ListBucketsResponse_Bucket& proto) -> mgmt::bucket_settings
+{
+  mgmt::bucket_settings settings;
+  settings.name = proto.bucket_name();
+  settings.ram_quota_mb = proto.ram_quota_mb();
+  switch (proto.bucket_type()) {
+    case v1::BUCKET_TYPE_EPHEMERAL:
+      settings.bucket_type = mgmt::bucket_type::ephemeral;
+      break;
+    default:
+      settings.bucket_type = mgmt::bucket_type::couchbase;
+      break;
+  }
+  settings.num_replicas = proto.num_replicas();
+  settings.flush_enabled = proto.flush_enabled();
+  settings.replica_indexes = proto.replica_indexes();
+  settings.max_expiry = proto.max_expiry_secs();
+  switch (proto.eviction_mode()) {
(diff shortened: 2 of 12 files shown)


core/protostellar/dispatcher.hxx
```diff
@@ -0,0 +1,309 @@
...
+
+  void OnReadDone(bool ok) override
+  {
+    if (!ok) {
+      return; // stream finished; OnDone follows
+    }
+    auto message = std::make_shared<Response>(std::move(current_));
+    asio::post(io_, [this, message]() {
+      on_row_(std::move(*message));
+    });
```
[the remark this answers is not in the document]
avsej: Defensible as written. gRPC serializes a single call's reactions (OnReadDone completes before OnDone runs), so every row task is posted before the OnDone task. The SDK contract is a single-threaded io_context, which runs posted tasks in FIFO order, so the self-delete task always runs last -- after every row handler. `delete this` from the terminal post is the standard self-owning-reactor pattern; the row handlers never dereference a freed reactor.

core/protostellar/bucket_admin_converter.hxx
```diff
@@ -0,0 +1,249 @@
...
+      break;
+    case mgmt::bucket_conflict_resolution::unknown:
+      break;
+  }
+  return true;
+}
+
+// Decode one ListBucketsResponse bucket back into core settings.
+[[nodiscard]] inline auto
+decode_bucket(const v1::ListBucketsResponse_Bucket& proto) -> mgmt::bucket_settings
```
avsej, line 191: Four fields that the encode side writes are not read back here, so bucket settings do not round-trip:

- **`minimum_durability_level`** is never read — every bucket reports "no minimum durability".
- **`storage_backend`** presence is ignored. The field is `optional`, so the value getter returns the zero enum both for an explicit couchstore and for an absent field; a magma bucket created through this same converter reads back as couchstore.
- **`conflict_resolution_type`** is never read — always reports as unknown.
- **`history_retention_bytes`** is narrowed from the proto's `uint64` to `uint32`, so a retention size above 4 GiB wraps.

The failure scenario is the common management pattern: `get_bucket`, modify one field, `update_bucket` with the result. The three unread fields come back as defaults, are sent back as defaults, and the bucket's minimum durability, storage backend and conflict resolution are silently reset. `couchbase://` preserves them.

```cpp
  if (proto.has_minimum_durability_level()) { /* map the enum */ }
  if (proto.has_storage_backend()) {
    settings.storage_backend = proto.storage_backend() == v1::STORAGE_BACKEND_MAGMA
                                 ? mgmt::bucket_storage_backend::magma
                                 : mgmt::bucket_storage_backend::couchstore;
  }
  if (proto.has_conflict_resolution_type()) { /* map the enum */ }
  settings.history_retention_bytes = proto.history_retention_bytes();   // keep 64 bits
```

The most valuable missing test is a round trip: create a bucket with a non-default value for each of these four, read it back, assert equality. That catches all four at once, and it is the shape gocbcoreps' `impl_admin_bucket_test.go` uses.

avsej (commented):
## Review — PR #1039 (`CXXCBC-900`)

This commit routes bucket management over couchbase2. The encode side is thorough: the RAM quota is in the right
unit, every enum the core API exposes has a mapping, and `flush_enabled`, `num_replicas`, `replica_index` and the
history-retention settings are all carried. The comparison worth making is against
gocbcoreps `impl_admin_bucket_v1.go` and its test file, which documents the expected field values, and the encode
side holds up against it.

The decode side does not. Three fields that the encode side sets are not read back, so bucket settings do not
round-trip.

### Strengths

- The RAM quota unit is correct — a frequent source of bugs in this area, and it is right here.
- Enum mappings for bucket type, storage backend, eviction policy, compression mode, conflict resolution and
  minimum durability are exhaustive on the encode side, with no default-fallthrough that would silently send a
  zero value.
- `create` and `update` share their field-population helper rather than duplicating it, so the two cannot drift.

### Findings

|---|---|---|---|---|
| 1 | high | correctness | `decode_bucket()` never reads `minimum_durability_level` | `core/protostellar/bucket_admin_converter.hxx:183` |
| 2 | high | correctness | `decode_bucket()` ignores the presence bit on `storage_backend` and defaults it to couchstore | `core/protostellar/bucket_admin_converter.hxx:183` |
| 3 | medium | correctness | `decode_bucket()` never reads `conflict_resolution_type` | `core/protostellar/bucket_admin_converter.hxx:183` |
| 4 | low | correctness | `history_retention_bytes` is truncated from `uint64` to `uint32` | `core/protostellar/bucket_admin_converter.hxx:183` |

#### 1-4. Bucket settings do not round-trip (high)

```cpp
decode_bucket(const v1::ListBucketsResponse_Bucket& proto) -> mgmt::bucket_settings
```

Four problems in one function, all of the same kind — the encode side writes a field and the decode side does not
read it, so `get_bucket` and `get_all_buckets` report defaults rather than the configured values:

- **`minimum_durability_level` is never read.** Every bucket reports "no minimum durability" regardless of how it
  was created. A caller doing get-modify-update therefore silently clears the setting.
- **`storage_backend` presence is ignored.** The field is `optional`, so the plain value getter returns the zero
  enum both for an explicit couchstore and for an absent field. A magma bucket created through this same converter
  reads back as couchstore.
- **`conflict_resolution_type` is never read**, so it always reports as unknown.
- **`history_retention_bytes` is narrowed** from the proto's `uint64` to a `uint32`, so a retention size above
  4 GiB wraps.

Failure scenario for the first three together, which is the common management pattern: an application calls
`get_bucket`, modifies one field, and calls `update_bucket` with the result. The three unread fields come back as
defaults, are sent back as defaults, and the bucket's minimum durability level, storage backend and conflict
resolution are silently reset. Over `couchbase://` the same sequence preserves them.

```cpp
if (proto.has_minimum_durability_level()) {
  switch (proto.minimum_durability_level()) {
    case ::couchbase::kv::v1::DURABILITY_LEVEL_MAJORITY:
      settings.minimum_durability_level = durability_level::majority;
      break;
    ...
  }
}
if (proto.has_storage_backend()) {
  settings.storage_backend = proto.storage_backend() == v1::STORAGE_BACKEND_MAGMA
                               ? mgmt::bucket_storage_backend::magma
                               : mgmt::bucket_storage_backend::couchstore;
}
if (proto.has_conflict_resolution_type()) {
  settings.conflict_resolution_type = /* map the enum */;
}
settings.history_retention_bytes = proto.history_retention_bytes();   // keep 64 bits
```

The single most valuable missing test is a round trip: create a bucket with a non-default value for each of
these four fields, read it back, and assert equality. That would have caught all four at once, and it is the same
test shape that gocbcoreps' `impl_admin_bucket_test.go` uses.

(inline comment on `core/protostellar/bucket_admin_converter.hxx:183`)

### Notes

- The encode side omits `num_replicas` when the caller did not set it, where gocb always sends an explicit zero.
  Omitting is defensible — it lets the server apply its own default rather than forcing zero replicas — and is
  arguably more correct than the reference behaviour. Worth a comment saying so, since it is a deliberate
  divergence.
- Creating a memcached bucket reports `feature_not_available` where the reference reports `invalid_argument`.
  `invalid_argument` is the better fit: memcached buckets are not a couchbase2 capability gap, they are a bucket
  type the gateway does not accept.
- Eviction policy is not validated against bucket type, so combinations the reference rejects locally are sent to
  the gateway. That is a difference in where the error comes from rather than in the outcome, and it is acceptable
  provided the gateway's error maps to something actionable.

### Verdict

Needs changes. The four decode omissions are one function and one test away from being closed, and until then a
get-modify-update cycle silently discards three bucket settings.

core/protostellar/bucket_admin_converter.hxx
avsej: All four confirmed and fixed in `decode_bucket()`, with two corrections to the suggested patch.

**`minimum_durability_level`** and **`storage_backend`** are now read through their presence bits.
`storage_backend` had a second half to it: the old `default:` arm turned an absent field into an
explicit couchstore, so an absent field now leaves the core `unknown`, which is what the gateway
means by omitting it.

**`conflict_resolution_type`** is now read, but `has_conflict_resolution_type()` does not exist on
this message. Field 12 is a plain `ConflictResolutionType`, unlike the `optional` fields 10 and 11
beside it — the presence accessor exists only on `CreateBucketRequest`. The enum is therefore mapped
unconditionally, and an absent field is indistinguishable from an explicit
`CONFLICT_RESOLUTION_TYPE_TIMESTAMP` (enum 0). A comment states that where it applies.

**`history_retention_bytes`** no longer truncates, though `// keep 64 bits` is not available: the
destination is `std::optional<std::uint32_t>` in the public
`couchbase/management/bucket_settings.hxx`, so widening it is an API change rather than a converter
fix. A value the field cannot hold is left unset instead, which is the behaviour that matters under
the failure scenario described — truncating reports a retention the bucket does not have, and the
next update sends it back and *shrinks* the setting, whereas an empty optional is omitted on update
and the server keeps its own value. Widening the field fixes this path and the HTTP one together
(`bucket_settings_json.hxx` parses straight into `std::uint32_t`) and is filed separately.

The round trip is added as `bucket_settings_round_trip_against_live_gateway`, using values the
gateway would not choose by itself — it defaults conflict resolution to sequence_number and leaves
minimum durability unset — so an unread field cannot pass as a server default. Reverting any part of
the fix turns it red.

avsej:
All four decode omissions are confirmed and fixed, and the round trip they break is now covered by a
live test.

### The four fields

`minimum_durability_level` and `storage_backend` are now read through their presence bits.
`storage_backend` had two halves to the bug: the missing presence check, and a `default:` arm that
turned an absent field into an explicit couchstore. An absent field now leaves the core `unknown`,
which is what the gateway means by omitting it.

`conflict_resolution_type` is now read, with one correction to the suggested patch: this field
carries **no presence bit** on `ListBucketsResponse.Bucket`. It is a plain
`ConflictResolutionType conflict_resolution_type = 12`, unlike fields 10 and 11 beside it, so
`has_conflict_resolution_type()` does not exist on this message — only on `CreateBucketRequest`,
where the field is `optional`. The enum is therefore mapped unconditionally, and an absent field is
indistinguishable from an explicit `CONFLICT_RESOLUTION_TYPE_TIMESTAMP`, enum 0. A comment states
that at the point it applies.

`history_retention_bytes` no longer truncates, but not by keeping 64 bits: the destination is
`std::optional<std::uint32_t>` in `couchbase/management/bucket_settings.hxx`, so widening it is a
public API change rather than a converter fix. A value the core field cannot hold is left unset
instead. That is the choice that matters under the failure scenario in the review — truncating
reports a retention the bucket does not have, and the next update sends the truncated number back and
*shrinks* the setting, whereas an empty optional is omitted on update and the server keeps its own
value. Widening the field fixes this path and the HTTP one together (`bucket_settings_json.hxx`
parses straight into `std::uint32_t` and has the same ceiling) and is proposed as a separate ticket.

### The round trip

`bucket_settings_round_trip_against_live_gateway` creates a bucket with a minimum durability level
and a conflict resolution type the gateway would not choose by itself — it defaults conflict
resolution to sequence_number and leaves minimum durability unset — reads it back, asserts equality,
and drops the bucket from a guard so a failed assertion does not leave one behind.

Two of the four are deliberately not asserted live. `storage_backend` only proves anything for magma,
which the gateway requires a 1024 MB quota for, and history retention a further 2048 MB — more than a
development cluster reliably has spare. Both are covered by converter unit tests against a
constructed response, and the live case says why it stops short. The case also skips rather than
fails when the cluster has no unallocated KV quota, keyed on the gateway's own "ram quota specified
is too large" message, since no client-side refusal produces one.

### Found by borrowing the reference suites

The live coverage was rebuilt around the cases gocbcoreps' `impl_admin_bucket_test.go` runs —
create/update/flush/delete against both their success and their error paths — and one of them
immediately failed for a real reason.

**`FLUSH_DISABLED` was not mapped.** The gateway raises flush-on-a-non-flushable-bucket as
`FAILED_PRECONDITION` with a `PreconditionFailure` violation of that type
(`errorhandler.go: NewBucketFlushDisabledStatus`), and `map_precondition()` knew only the KV
violations, so it fell through to `internal_server_failure` — the caller is told the cluster is
broken, when in fact they can turn flush on. It now maps to `bucket_not_flushable`, covered by a
converter unit case and by the live lifecycle.

The remaining borrowed cases all pass as they stand, which is worth recording because it is the
gateway's `ResourceInfo` details doing the work: creating an existing bucket reports `bucket_exists`
rather than `document_exists`, and update/drop/flush/get against a missing bucket each report
`bucket_not_found` rather than `document_not_found`. Mutating either lookup away turns those
assertions red, so they are pinned rather than incidental.

Two live behaviours needed to be waited for rather than asserted immediately, matching what the
reference does with `require.Eventually`: a create or an update is not visible to the next `get`
straight away, and enabling flush is visible in the bucket list before the server will act on it. Both
are bounded polls, so a permanent failure is still a failure.

### Also fixed, from the review bodies

- The header comment attributed the memcached gate to `apply_settings()`, which returns `void`. The
  gate is in `apply_create_only_settings()`.
- The memcached refusal was delivered inline out of `execute()`, re-entering the caller before it had
  its `pending_call`. It is now posted to the io context like every other completion — the same
  defect class as the search and view refusals in #1038.
- The live list case asserted only that the response was non-empty, which a gateway pointed at a
  different cluster also satisfies. It now names the bucket under test.

### Verified already handled — no change

- The `ram_quota_mb == 0` sentinel, raised three times: `apply_settings()` already sends the field
  only when non-zero and the create path already backfills the 100 MB default, each with a comment.
- `component.hxx` already uses `#pragma push_macro` / `pop_macro` for
  `COUCHBASE_CXX_CLIENT_IGNORE_CORE_DEPRECATIONS`.
- `view_converter.hxx`'s `can_encode()` already rejects `row_callback`.
- `origin.cxx` copy-assignment: the field is copied, and the iteration reset is deliberate and
  documented.

### Not changed, with reasons

**memcached as `invalid_argument` rather than `feature_not_available`.** The references disagree, so
this reads as a decision rather than a defect. The JVM client raises `FeatureNotAvailableException`
with a message naming couchbase2; the .NET client throws `ArgumentOutOfRangeException` from its
bucket-type conversion; gocbcoreps does not handle the case. Whichever is right is worth choosing for
all three clients together, so this is left as it is for now.

**Eviction policy not validated against bucket type** — accepted as noted: the error arrives from the
gateway rather than locally, with the same outcome.

**IPv6 authority bracketing, the missing `server_->Wait()`, and the per-row allocation in the
dispatcher** are all real and all in files owned by other commits in this series; the first two
pre-exist on `main`. They are recorded as follow-ups rather than fixed here.

### Verification

Eleven mutations across the two suites, each reverting one part of the fix — all eleven killed,
including the live ones, which confirms the round-trip case fails rather than skips when a field
stops being read, and that the refusal-delivery assertion catches an inline completion. Full `cng`
suite green against a 1.2.2 gateway, per-commit.

Running the live cases needs more than the operator's default 256 MiB data quota, since a stock
allocation commits most of it to the test bucket. The CNG cluster action now raises it to 512 MiB —
the smallest quota that fits the test bucket plus a 128 MB bucket under update, and nothing against
the runner's 16 GB — and polls until the value is actually applied rather than waiting on a
condition that is already true. Without that, the bucket cases would report the gateway's "ram quota
specified is too large", skip, and leave the suite green while testing less.

The quota cannot be expressed in the cluster definition — the `cao` deployer emits no `cluster:`
block and its definition section has no quota fields — and setting it through `/pools/default` is
reconciled away by the operator within the minute, so a patch is the only route.
`test/cng/README.md` documents that, along with the sequence from an empty machine to a cluster the
whole suite can use. Search is deliberately left alone: CI runs no fts node, and an index under the
256 MiB default is accepted and then never builds a usable partition, which is what search index
management will have to provision for.

core/cluster.cxx
```diff
@@ -936,11 +944,14 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
-                                        options.key_value_timeout,
-                                        options.query_timeout,
-                                        options.analytics_timeout });
+        protostellar::component_config{
+          std::move(channel), origin_.credentials(), protostellar::component_timeouts {
+            options.key_value_timeout,
+            options.query_timeout,
+            options.analytics_timeout,
+            options.search_timeout,
+            options.view_timeout
```
thejcfactor: I think we are missing the mgmt timeout, right?

core/protostellar/component.cxx
```diff
@@ -827,4 +842,368 @@ component::execute(operations::analytics_request request,
...
+auto
+component::execute(
+  operations::management::bucket_get_all_request request,
+  utils::movable_function<void(operations::management::bucket_get_all_response)>&& handler)
+  -> pending_call
+{
+  auto proto = std::make_shared<bucket_admin_v1::ListBucketsRequest>();
+  const auto timeout = request.timeout.value_or(timeouts_.management);
+  if (timeout <= std::chrono::milliseconds::zero()) {
+    return fail_expired_ctx(io_, handler, operations::management::bucket_get_all_response{});
```
thejcfactor: Do we want to use the same `stamped` pattern for the mgmt ops?  I believe all the operations have `std::optional<std::string> client_context_id{};` and the http error context can take the `client_context_id`.

If we do, do we also want to carry the `client_context_id` through w/ the completion callback?

avsej:
# Response (round 2) — PR #1039 (CXXCBC-900: bucket management)

Both points were right, and the second one reached further than this PR. Fixed in the commits that
own the code rather than all at the tip, so each PR in the series stays reviewable on its own.

## 1. The management timeout (`core/cluster.cxx:953`)

> I think we are missing the mgmt timeout, right?

Confirmed. `component_timeouts` gained a `management` field in this commit, and the `cluster_impl`
initialiser was left at five values, so every management operation over couchbase2 used the struct's
default of 75s and a caller's configured `management_timeout` was discarded. `options.management_timeout`
is now passed.

Worth noting what did *not* catch it. The struct exists precisely so that adding a field cannot
shift a positional argument onto the wrong parameter, and that worked — the five values still landed
on the five fields they were written for. Silently taking the default is the failure mode that
survives designated initialisation, and nothing else covered it: the value is only observable as a
deadline on the wire, and every management case in the suite completes well inside 75s either way.

## 2. The `stamped` pattern for management operations (`core/protostellar/component.cxx:988`)

> Do we want to use the same `stamped` pattern for the mgmt ops? I believe all the operations have
> `std::optional<std::string> client_context_id{};` and the http error context can take the
> `client_context_id`.
>
> If we do, do we also want to carry the `client_context_id` through w/ the completion callback?

Yes to both, and it applied to more than this PR: 25 management overloads across #1039, #1040, #1041
and #1042 all built a default-constructed response, so the field the classic path fills
(`io/http_command.hxx:194`) was empty on the couchbase2 path for every one of them. Each commit now
stamps its own operations.

The shape differs slightly from the search and view overloads, which capture a per-operation
`stamped` lambda because each stamps service-specific fields — index and query for search, design
document and view name for views. A management response has only the one field to carry, so a single
helper does it for all of them:

```cpp
template<typename Request>
[[nodiscard]] auto
stamped_management(const Request& request) -> typename Request::response_type
{
  typename Request::response_type response;
  response.ctx.client_context_id = request.client_context_id.value_or(std::string{});
  return response;
}
```

It is stamped on all three kinds of path, since the ones that never reach the server are exactly the
ones a caller cannot correlate any other way: the expired-timeout return, the client-side refusals
(a memcached bucket here, a collection update that cannot be expressed in #1040), and the completion
callback — which is the second half of the question, and yes, the id is carried into it.

One deliberate difference from the classic path: only what the caller supplied is echoed, where
`http_command` substitutes a generated uuid for an absent one. That uuid earns its place there
because it travels with the HTTP request and comes back in the server's logs; couchbase2 sends no
such field, so an id the caller never chose would correlate nothing and only look as though it did.

## How the sweep was done, and what it caught

The edit is four coordinated changes per operation, so it was scripted
(`tools/stamp_mgmt_ctx.rb`) rather than hand-applied 25 times, with the site count taken
independently by `ast-grep` before and after so a pattern that quietly matched less than expected
could not pass as success.

It did quietly match less. Three of the six capture lists in this commit went unedited on the first
pass, because `clang-format` had wrapped `proto` onto its own line and the pattern assumed one line.
The script tallies edits per operation and refuses to write when an operation gains nothing, which
named the three before the compiler saw them.

## Verification

Per commit, against a live gateway — Server 8.1.0-1743, Cloud Native Gateway 1.2.2-123, operator
2.8.0:

| PR | management operations stamped | cng suite |
|---|---|---|
| #1039 | 6 bucket | 27/27 |
| #1040 | 6 scope and collection | 29/29 |
| #1041 | 4 query index | 31/31 |
| #1042 | 9 search index | 33/33 |

The timeout fix itself has no test, and that is worth stating plainly rather than leaving implied.
The only thing an omitted timeout changes is a deadline on the wire, and every management case in
the suite finishes well inside both the correct value and the 75s default, so nothing in the suite
can tell the two apart. Closing that needs a fixture that can set `cluster_options` and execute a
request without the fixture's own 20s default masking the cluster-level one; it is written up as a
follow-up rather than half-built here.

core/cluster.cxx
avsej: Confirmed, and fixed: `options.management_timeout` is now passed, so the entry no longer falls back
to the struct's own default of 75s and a configured management timeout reaches the component.

Worth recording what did not catch it. The struct shape prevents a *different* failure — a new field
shifting a positional argument onto the wrong parameter — and that part worked. An omitted entry
silently taking its default compiles and passes, and nothing observes it: the only symptom is a
deadline on the wire, and every management case in the suite finishes well inside both values. The
fixture change that would give this a test (settable `cluster_options`, and an execute path that does
not inject the fixture's own 20s default over the cluster-level one) is recorded as a follow-up
rather than half-built here.

core/protostellar/component.cxx
avsej: Yes to both, and it reached further than this PR: 25 management overloads across #1039, #1040, #1041
and #1042 all built a default-constructed response, so `ctx.client_context_id` was empty on the
couchbase2 path everywhere the classic path fills it (`io/http_command.hxx:194`). Each commit now
stamps its own operations.

The shape is a single helper rather than a per-operation `stamped` lambda, because a management
response has only the one field to carry where search and view stamp service-specific ones:

```cpp
template<typename Request>
[[nodiscard]] auto
stamped_management(const Request& request) -> typename Request::response_type
{
  typename Request::response_type response;
  response.ctx.client_context_id = request.client_context_id.value_or(std::string{});
  return response;
}
```

And yes to the second half — it is carried into the completion callback, along with the two paths
that never reach the server: the expired-timeout return and the client-side refusals. Those are the
ones with no server response to correlate by, so they are where the id matters most.

One deliberate difference from the classic path: only what the caller supplied is echoed. The uuid
`http_command` generates for an absent id earns its place there because it travels with the request
and comes back in the server's logs; couchbase2 sends no such field, so a generated id would
correlate nothing while looking as though it did.

thejcfactor (approved):

PR #1040, opened 2026-07-28
CXXCBC-900: route scope and collection management over couchbase2

avsej:
Complete the bucket & collection management ticket with the six scope/collection
admin operations over admin.collection.v1 (unary).

scope_get_all decodes ListCollections into a core collections_manifest (per-scope
and per-collection uids are not carried by the proto, so only the manifest uid is
populated). scope_create/drop and collection_create/update/drop are thin unary
calls returning the new manifest uid; collection create/update carry the optional
max_expiry and history-retention settings. The component gains a
CollectionAdminService stub and the six execute() overloads; cluster_impl routes
them.

A collection's max_expiry holds three states in one signed field -- a positive
TTL, 0 to inherit the bucket default, -1 for no expiry -- while max_expiry_secs
is an unsigned optional, so the sentinel travels as the field's presence: no
expiry is an explicit 0 on the wire and inherit is the field left unset. Below -1
the sign carries no state and the request is refused with invalid_argument before
the encode, as the classic path does, since every negative otherwise maps onto
the no-expiry wire form. Update has no encoding left for inherit, because an
unset field there means "leave the collection as it is", so a collection update
asking for it is refused with feature_not_available rather than reported as
applied.

A create or update carrying history retention is routed to the component like any
other. execute_with_bucket_capability_check reads a bucket capability from a
cluster map couchbase2 does not have, and reaching it bootstraps an MCBP session
against a port the gateway does not serve; the server validates the setting
either way.

The live fixtures pin gRPC's process-global callback queue for the lifetime of
the binary. A fixture per case destroys the last channel between cases, and the
teardown that follows races the threads still polling the queue and aborts the
process; the in-process harnesses already hold it open for the same reason.

The CNG cluster's data service quota covers a magma bucket, which the history
retention cases need and the gateway will not create below 1024 MB.

---
Stacked PR **18/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1039 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -782,7 +782,13 @@ public:
                     std::is_same_v<Request, operations::management::bucket_create_request> ||
                     std::is_same_v<Request, operations::management::bucket_update_request> ||
                     std::is_same_v<Request, operations::management::bucket_drop_request> ||
-                    std::is_same_v<Request, operations::management::bucket_flush_request>) {
+                    std::is_same_v<Request, operations::management::bucket_flush_request> ||
+                    std::is_same_v<Request, operations::management::scope_get_all_request> ||
+                    std::is_same_v<Request, operations::management::scope_create_request> ||
+                    std::is_same_v<Request, operations::management::scope_drop_request> ||
+                    std::is_same_v<Request, operations::management::collection_create_request> ||
+                    std::is_same_v<Request, operations::management::collection_update_request> ||
+                    std::is_same_v<Request, operations::management::collection_drop_request>) {
         component->execute(std::move(request), std::forward<Handler>(handler));
         return;
       } else {
@@ -815,6 +821,16 @@ public:
                                             bucket_capability cap,
                                             Handler&& handler)
   {
+    // There is no bucket capability to read over couchbase2, and no MCBP connection to read it
+    // over: open_bucket() below would bootstrap a session against a port the gateway does not
+    // serve, so a request that only wants a capability checked would fail on the check rather
+    // than on what it asked for. The gateway still validates the capability the request depends
+    // on -- history retention needs a magma bucket wherever the request arrives from -- so
+    // routing straight through refuses the same requests, one hop later and in the server's own
+    // words.
+    if (is_protostellar()) {
+      return execute(std::move(request), std::forward<Handler>(handler));
+    }
     auto bucket_name = request.bucket_name;
     return open_bucket(
       bucket_name,
--- a/core/protostellar/collection_admin_converter.hxx
+++ b/core/protostellar/collection_admin_converter.hxx
@@ -1,0 +1,92 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Decodes couchbase.admin.collection.v1 ListCollections into a core collections_manifest, and maps
+// a collection's max_expiry onto the wire in the other direction. The rest of the scope/collection
+// encodes are string setters and live inline in the component; max_expiry does not, because its
+// two halves only agree if they are read together. Per-scope and per-collection uids are not
+// carried by the proto, so only the top-level manifest uid is populated.
+
+#include "core/topology/collections_manifest.hxx"
+
+#include <couchbase/admin/collection/v1/collection.pb.h>
+
+#include <cstdint>
+#include <optional>
+#include <utility>
+
+namespace couchbase::core::protostellar::collection_admin
+{
+namespace v1 = ::couchbase::admin::collection::v1;
+
+// Maps the core max_expiry sentinel onto max_expiry_secs. The proto field is unsigned, so the sign
+// the core struct carries has to become the field's presence: a negative value ("no expiry") is an
+// explicit 0, and 0 has no encoding at all -- nullopt here means leave the field unset. What an
+// unset field then means belongs to the caller, and it differs between the two RPCs: on create it
+// is "inherit the bucket default", which is what 0 asked for, and on update it is "leave the
+// collection as it is", which is not.
+//
+// -1 is the only negative the core struct gives a meaning; the caller refuses anything below it
+// with invalid_argument before reaching here, as the classic path does. Every negative maps to the
+// same wire form, so an unvalidated -5 would silently become "no expiry".
+[[nodiscard]] inline auto
+encode_max_expiry(std::int32_t max_expiry) -> std::optional<std::uint32_t>
+{
+  if (max_expiry < 0) {
+    return 0U;
+  }
+  if (max_expiry == 0) {
+    return {};
+  }
+  return static_cast<std::uint32_t>(max_expiry);
+}
+
+[[nodiscard]] inline auto
+decode_manifest(const v1::ListCollectionsResponse& proto) -> topology::collections_manifest
+{
+  topology::collections_manifest manifest;
+  manifest.uid = proto.manifest_uid();
+  for (const auto& proto_scope : proto.scopes()) {
+    topology::collections_manifest::scope scope;
+    scope.name = proto_scope.name();
+    for (const auto& proto_collection : proto_scope.collections()) {
+      topology::collections_manifest::collection collection;
+      collection.name = proto_collection.name();
+      // max_expiry_secs is an unsigned optional, so the gateway spends the presence bit on the
+      // distinction the core struct makes with a sign: the field left unset is "inherit the bucket
+      // default" (0), an explicit 0 is "no expiry" (-1). The value getter alone collapses the two,
+      // because it returns 0 for an absent field as well.
+      if (!proto_collection.has_max_expiry_secs()) {
+        collection.max_expiry = 0;
+      } else if (proto_collection.max_expiry_secs() == 0) {
+        collection.max_expiry = -1;
+      } else {
+        collection.max_expiry = static_cast<std::int32_t>(proto_collection.max_expiry_secs());
+      }
+      if (proto_collection.has_history_retention_enabled()) {
+        collection.history = proto_collection.history_retention_enabled();
+      }
+      scope.collections.push_back(std::move(collection));
+    }
+    manifest.scopes.push_back(std::move(scope));
+  }
+  return manifest;
+}
+
+} // namespace couchbase::core::protostellar::collection_admin
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -26,6 +26,7 @@
 #include "core/operations/query_response_parsing.hxx"
 #include "core/protostellar/analytics_converter.hxx"
 #include "core/protostellar/bucket_admin_converter.hxx"
+#include "core/protostellar/collection_admin_converter.hxx"
 #include "core/protostellar/credentials.hxx"
 #include "core/protostellar/error_utils.hxx"
 #include "core/protostellar/kv_converter.hxx"
@@ -39,6 +40,7 @@
 #include <couchbase/analytics/v1/analytics.grpc.pb.h>
 
 #include <couchbase/admin/bucket/v1/bucket.grpc.pb.h>
+#include <couchbase/admin/collection/v1/collection.grpc.pb.h>
 #include <couchbase/kv/v1/kv.grpc.pb.h>
 #include <couchbase/search/v1/search.grpc.pb.h>
 #include <couchbase/view/v1/view.grpc.pb.h>
@@ -62,6 +64,7 @@ namespace analytics_v1 = ::couchbase::analytics::v1;
 namespace search_v1 = ::couchbase::search::v1;
 namespace view_v1 = ::couchbase::view::v1;
 namespace bucket_admin_v1 = ::couchbase::admin::bucket::v1;
+namespace collection_admin_v1 = ::couchbase::admin::collection::v1;
 
 // Defined here rather than in the header so the generated gRPC types stay out of every consumer of
 // component.hxx.
@@ -72,6 +75,7 @@ struct component::stubs {
   std::unique_ptr<search_v1::SearchService::Stub> search;
   std::unique_ptr<view_v1::ViewService::Stub> view;
   std::unique_ptr<bucket_admin_v1::BucketAdminService::Stub> bucket_admin;
+  std::unique_ptr<collection_admin_v1::CollectionAdminService::Stub> collection_admin;
 };
 
 namespace
@@ -142,7 +146,8 @@ component::component(asio::io_context& io, component_config config)
              analytics_v1::AnalyticsService::NewStub(config.channel),
              search_v1::SearchService::NewStub(config.channel),
              view_v1::ViewService::NewStub(config.channel),
-             bucket_admin_v1::BucketAdminService::NewStub(config.channel) }) }
+             bucket_admin_v1::BucketAdminService::NewStub(config.channel),
+             collection_admin_v1::CollectionAdminService::NewStub(config.channel) }) }
   , authorization_{ authorization_header(config.credentials) }
   , timeouts_{ config.timeouts }
   // Initialised last (see the declaration order in the header), so the channel can be moved in.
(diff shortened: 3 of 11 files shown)


core/protostellar/component.cxx
```diff
@@ -0,0 +1,1130 @@
...
+  operations::management::collection_create_request request,
+  utils::movable_function<void(operations::management::collection_create_response)>&& handler)
+  -> pending_call
+{
+  auto proto = std::make_shared<collection_admin_v1::CreateCollectionRequest>();
+  proto->set_bucket_name(request.bucket_name);
+  proto->set_scope_name(request.scope_name);
+  proto->set_collection_name(request.collection_name);
+  if (request.max_expiry.has_value()) {
+    proto->set_max_expiry_secs(static_cast<std::uint32_t>(*request.max_expiry));
```
avsej: `max_expiry` is a signed `std::int32_t` with three meanings — positive is a TTL, `0` is "inherit the bucket default", `-1` is "no expiry" — and this cast turns `-1` into `4294967295`.

Scenario: `create_collection` with `max_expiry = -1` creates a no-expiry collection over `couchbase://`; over `couchbase2://` the gateway is sent a maximum expiry of ~136 years, or rejects it. The update path applies the same cast.

```suggestion
      if (*request.max_expiry < 0) {
        // "no expiry" is an explicit 0 on the wire; "inherit the bucket default" is unset.
        proto->set_max_expiry_secs(0);
      } else if (*request.max_expiry > 0) {
        proto->set_max_expiry_secs(static_cast<std::uint32_t>(*request.max_expiry));
      }
```

core/protostellar/collection_admin_converter.hxx
```diff
@@ -0,0 +1,58 @@
...
+{
+  topology::collections_manifest manifest;
+  manifest.uid = proto.manifest_uid();
+  for (const auto& proto_scope : proto.scopes()) {
+    topology::collections_manifest::scope scope;
+    scope.name = proto_scope.name();
+    for (const auto& proto_collection : proto_scope.collections()) {
+      topology::collections_manifest::collection collection;
+      collection.name = proto_collection.name();
+      collection.max_expiry = static_cast<std::int32_t>(proto_collection.max_expiry_secs());
```
avsej: `max_expiry_secs` is `optional uint32`, so this value getter returns `0` both for an explicit zero and for an absent field — but those are different states, and the core struct distinguishes them (`-1` versus `0`).

Consequence: even with the encode side fixed, a collection created with `max_expiry = -1` reads back as `0`, which a caller interprets as "inherits the bucket default".

```suggestion
    if (!proto_collection.has_max_expiry_secs()) {
      collection.max_expiry = 0;    // inherit the bucket default
    } else if (proto_collection.max_expiry_secs() == 0) {
      collection.max_expiry = -1;   // explicitly no expiry
    } else {
      collection.max_expiry = static_cast<std::int32_t>(proto_collection.max_expiry_secs());
    }
```

A round-trip test across all three states would pin this and the encode-side sentinel together.

avsej (commented):
## Review — PR #1040 (`CXXCBC-900`)

This commit routes scope and collection management over couchbase2. The converter is short — 58 lines against
the 249 of the bucket converter in #1039 — and the terseness is mostly justified, because `ListCollections`
returns a nested scope/collection structure with far fewer knobs than a bucket definition. The scope/collection
tree reconstruction is correct and the `_default` scope and collection need no special handling on this path.

The problem is `max_expiry`, which is the one field on a collection that carries a sentinel, and it is
mishandled in both directions.

### Strengths

- The scope and collection tree is reconstructed faithfully from the nested `ListCollectionsResponse`, with no
  flattening or reordering that would surprise a caller comparing against the classic path.
- Create, update and drop for both scopes and collections are all routed, so there is no partially-supported
  management surface where some calls work and others silently fall through.

### Findings

|---|---|---|---|---|
| 1 | high | protocol-parity | `max_expiry == -1` ("no expiry") is sent as 4294967295 instead of the 0 sentinel | `core/protostellar/component.cxx:1030` |
| 2 | high | protocol-parity | The `max_expiry` decode ignores field presence and never inverts the sentinel | `core/protostellar/collection_admin_converter.hxx:47` |

#### 1. `max_expiry == -1` is sent as 4294967295 (high)

```cpp
proto->set_max_expiry_secs(static_cast<std::uint32_t>(*request.max_expiry));
```

`collection_create_request::max_expiry` is a signed `std::int32_t` carrying three distinct meanings: a positive
value is a TTL in seconds, `0` means "inherit the bucket default", and `-1` means "no expiry". The proto field
is `optional uint32`, and the cast converts `-1` to `4294967295`.

Failure scenario: `create_collection` is called with `max_expiry = -1` to create a collection whose documents
never expire. Over `couchbase://` the collection is created with no maximum expiry. Over `couchbase2://` the
gateway receives a maximum expiry of 4294967295 seconds — roughly 136 years — or rejects the value outright,
depending on how it validates. Either way the caller's intent is not what was sent, and the same cast is applied
on the update path.

The sentinel needs translating in both places:

```cpp
if (request.max_expiry.has_value()) {
  // couchbase2 cannot carry a negative max expiry: "no expiry" (-1) is an explicit 0 on the
  // wire, and "inherit the bucket default" (0) is the field left unset.
  if (*request.max_expiry < 0) {
    proto->set_max_expiry_secs(0);
  } else if (*request.max_expiry > 0) {
    proto->set_max_expiry_secs(static_cast<std::uint32_t>(*request.max_expiry));
  }
}
```

(inline comment on `core/protostellar/component.cxx:1030`)

#### 2. The `max_expiry` decode ignores field presence and never inverts the sentinel (high)

```cpp
collection.max_expiry = static_cast<std::int32_t>(proto_collection.max_expiry_secs());
```

`max_expiry_secs` is `optional uint32`, so the value getter returns `0` both when the gateway sent an explicit
zero and when it sent nothing at all. Those are different states — "no expiry" versus "inherit the bucket
default" — and the core struct has distinct representations for them (`-1` and `0`).

The read side therefore cannot round-trip the write side even once finding 1 is fixed: a collection created with
`max_expiry = -1` is reported back as `0`, which a caller reading it will interpret as "inherits the bucket
default".

```cpp
if (!proto_collection.has_max_expiry_secs()) {
  collection.max_expiry = 0;    // inherit the bucket default
} else if (proto_collection.max_expiry_secs() == 0) {
  collection.max_expiry = -1;   // explicitly no expiry
} else {
  collection.max_expiry = static_cast<std::int32_t>(proto_collection.max_expiry_secs());
}
```

A round-trip test over all three states — inherit, no-expiry, and a positive TTL — would pin both findings and
is the single most valuable test this commit is missing.

(inline comment on `core/protostellar/collection_admin_converter.hxx:47`)

### Verdict

Needs changes. Both findings are in the same field, they are a matched encode/decode pair, and together they
mean a collection's expiry policy cannot be set or read correctly over couchbase2.

avsej:
# Response — PR #1040 (CXXCBC-900: scope and collection management)

Both findings are fixed. A third state of the same field, which the transport cannot express at all,
is handled with them. Details below, in the review's numbering.

## 1. `max_expiry == -1` is sent as 4294967295 (high)

> `collection_create_request::max_expiry` is a signed `std::int32_t` carrying three distinct
> meanings: a positive value is a TTL in seconds, `0` means "inherit the bucket default", and `-1`
> means "no expiry". The proto field is `optional uint32`, and the cast converts `-1` to
> `4294967295`.

Confirmed and fixed, on both the create and the update path. The failure scenario needs one
correction.

> Over `couchbase2://` the gateway receives a maximum expiry of 4294967295 seconds — roughly 136
> years — or rejects the value outright, depending on how it validates.

Neither happens with this gateway. `collectionadminserver.go:211-212` takes the `> 0` branch for
4294967295 and computes `int32(*in.MaxExpirySecs)`, and a Go conversion to a narrower integer type
truncates, so `int32(uint32(4294967295))` is `-1` — the value the correct encoding produces. This
was checked against the gateway rather than reasoned about: with the cast put back, the live
round trip still reports `never_expires` as -1 and passes.

For a *negative* max_expiry the encoding as reviewed therefore reaches the right server state, by
way of an unsigned-to-signed wraparound in the gateway's implementation. The observable defect on
the encode side is the other value: an explicit `max_expiry = 0` is written to the wire as 0, which
the gateway reads as no expiry, so a caller asking a collection to inherit the bucket default gets
one whose documents never expire. That inversion is what the live case pins, with a collection
created at an explicit 0 alongside one created with the field unset.

The fix stands either way: 4294967295 is out of contract whatever this gateway does with it, and a
result that depends on a wraparound is not a contract.

The suggested translation was adopted in substance, with one change of placement: it lives in
`collection_admin_converter.hxx` as `encode_max_expiry()` rather than inline at the two call sites.
Three reasons — the encode and the decode of the same sentinel are then next to each other and
cannot drift apart; create and update share one implementation instead of two copies; and a pure
function is directly unit-testable, which the inline form was not.

```cpp
[[nodiscard]] inline auto
encode_max_expiry(std::int32_t max_expiry) -> std::optional<std::uint32_t>
{
  if (max_expiry < 0) {
    return 0U;
  }
  if (max_expiry == 0) {
    return {};
  }
  return static_cast<std::uint32_t>(max_expiry);
}
```

The mapping was checked against the gateway rather than inferred. `collectionadminserver.go:209-216`
reads an unset field as maxTTL 0, an explicit 0 as -1, and a positive value as itself; the same
three-way mapping appears in gocb (`collectionsmgmtprovider_ps.go:100-108`) with the same comment
about protostellar not carrying negative values.

## 2. The `max_expiry` decode ignores field presence and never inverts the sentinel (high)

> `max_expiry_secs` is `optional uint32`, so the value getter returns `0` both when the gateway
> sent an explicit zero and when it sent nothing at all. Those are different states — "no expiry"
> versus "inherit the bucket default" — and the core struct has distinct representations for them
> (`-1` and `0`).

Confirmed and fixed as suggested. `decode_manifest()` now branches on `has_max_expiry_secs()` first,
so an absent field decodes as 0 and an explicit 0 as -1. This matches `collectionadminserver.go:78-83`
on the encode side of the gateway.

> A round-trip test over all three states — inherit, no-expiry, and a positive TTL — would pin both
> findings and is the single most valuable test this commit is missing.

Added, at both levels. `max_expiry_round_trips_through_the_wire_form` composes the encode and the
decode over all three states without a server, and
`collection_max_expiry_round_trips_against_live_gateway` creates four collections against a real
gateway — the three states, plus "inherit the bucket default" a second time as an explicit 0 rather
than an unset field — and asserts the manifest reports each one as created.

## A third state the transport cannot express

`update_collection` with an explicit `max_expiry = 0` has no encoding over couchbase2 at all, and
reports success without applying anything.

The asymmetry is in the gateway. On `CreateCollection` an unset `max_expiry_secs` means "inherit the
bucket default", which is what 0 asks for. On `UpdateCollection` it means "leave the collection as
it is" — `collectionadminserver.go:321-327` passes a nil `MaxTTL` through to
`cbmgmtx.UpdateCollection`, which omits `maxTTL` from the form (gocbcorex `cbmgmtx/mgmt.go:677`), so
the server keeps the current value. The classic path posts `maxTTL=0` and the server resets the collection to the
bucket default, which is what `update_collection_settings::max_expiry` is documented to do.

So the encoding that finding 1's patch produces for that value — leave the field unset — is a
success report for a change that was never applied. It is now refused with
`errc::common::feature_not_available` before dispatch, alongside the other features with no
couchbase2 equivalent. `collection_update_max_expiry_against_live_gateway` asserts both halves: the
refusal, and that the collection still carries the expiry it had.

gocb reaches the same wire form without the refusal, because its `UpdateCollectionSettings.MaxExpiry`
is a plain `time.Duration` where 0 means "not specified" — the C++ core spends an `optional` on the
distinction, so it can tell the two apart and report it.

## Layered expiry

Expiry is set at two levels, not three: a bucket has one and a collection may have its own.
`CreateScopeRequest` carries `bucket_name` and `scope_name` only, and the manifest's `Scope` message
has no expiry field, so there is no scope layer to test.
`collection_max_expiry_is_layered_over_the_bucket_default` covers what is left — a bucket created
with a maximum expiry, and a collection in it that reports 0 rather than the bucket's value, so the
layers are not flattened on the way through the converter.

What the layering does to a stored *document* is not observable over this transport yet: the
effective expiry of a document is read through subdoc (`$document.exptime`), and couchbase2 does not
carry subdoc in this stack.

## Copilot: `streaming_reactor` lifetime (`core/protostellar/dispatcher.hxx:193`)

> `streaming_reactor` posts row handlers that capture `this`, but reactor lifetime is only owned by
> the completion task posted from `OnDone`. If the completion handler is executed (or destroyed)
> before already-posted row handlers, those row handlers can dereference a destroyed reactor (UAF).
> This ordering is not enforced by the current implementation.

Accurate, and addressed later in the stack rather than here. CXXCBC-910 (#1048) makes the reactor
`enable_shared_from_this` and posts each row handler with `self = shared_from_this()`
(`dispatcher.hxx:290`), so a queued row handler holds the reactor alive and cannot be reached by a
completion that runs first. The thread is outdated against that change.

## Copilot: suppressed comments

> The in-process gRPC server is shut down but not waited for. Without `server_->Wait()`, gRPC
> background threads may still be running while the test tears down, which can cause flakiness or
> leaks (the dispatcher test already added `Wait()`).

Carried as a followup rather than changed here. The teardown these tests actually hit is
CXXCBC-919's callback-queue race, which `pin_callback_queue()` addresses; adding `Wait()` on top is
worth doing but belongs with that work rather than in a management-routing commit.

> This test writes fixed filenames in the system temp directory. If tests run concurrently (e.g.,
> `ctest -j`), they can clobber each other's cert/key files and become flaky. Use unique filenames
> (e.g., `std::filesystem::unique_path`).

Also a followup. It is real, and it is in `credentials.cxx`, which this commit does not touch.

## Also in this commit: client_context_id on the management responses

Raised on #1039 and applied across the series, so the six scope and collection operations here carry
it too. Each response now echoes the caller's `client_context_id` into the http error context — on
the completion, on the expired-timeout return, and on the refusal above, which is the path where it
matters most, since a request that never reached the server has nothing else to correlate it by.

## Verification

Run against a live CNG gateway — Server 8.1.0-1743, Cloud Native Gateway 1.2.2-123, operator 2.8.0
— and not only in-process.

- `cng_collection_admin_converter_test`: 4 cases passing.
- `cng_live_collection_admin_test`: 4 cases passing, 0 skipped — including the no-expiry round trip,
  which needs a gateway that accepts maxTTL=-1. No `X-API-Version` header is sent, so the gateway
  resolves the request to `Latest`, and `CollectionNoExpiry` is below it.
- Negative control on the converter: four mutations, one per state of the mapping, all killed.
- Negative control on the live cases: three mutations, all killed — the decode ignoring presence,
  the encode writing an explicit 0 for "inherit", and the update refusal removed.

One mutation is deliberately absent from the live spec: sending -1 as 4294967295 cannot be caught
against a real gateway, for the wraparound reason under finding 1. It is covered by the converter
spec, which asserts the wire form directly.

core/protostellar/component.cxx
avsej: Confirmed and fixed, on the create and the update path both, with one correction to the failure
scenario.

The gateway does not store 4294967295 and does not reject it: `collectionadminserver.go:211-212`
takes the `> 0` branch and computes `int32(*in.MaxExpirySecs)`, and a Go conversion to a narrower
integer truncates, so `int32(uint32(4294967295))` is `-1` — the same value the correct encoding
produces. Checked rather than reasoned about: with the cast put back, the live round trip still
reports the no-expiry collection as -1 and passes.

The value that *is* observably wrong is the other one. An explicit `max_expiry = 0` was written to
the wire as an explicit 0, which the gateway reads as no expiry, so a collection asked to inherit the
bucket default came back never expiring. The live case now creates a collection at an explicit 0
alongside one with the field unset, which is what pins it.

The fix stands either way — 4294967295 is out of contract whatever this gateway does with it — and
the translation now lives in `collection_admin_converter.hxx` as `encode_max_expiry()` rather than
inline, so the encode sits beside the decode it has to agree with and both are unit-testable.

core/protostellar/collection_admin_converter.hxx
avsej: Confirmed and fixed as suggested: `decode_manifest()` branches on `has_max_expiry_secs()` first, so
an absent field decodes as 0 (inherit the bucket default) and an explicit 0 as -1 (no expiry). That
matches what the gateway writes at `collectionadminserver.go:78-83`.

The round-trip test is there at both levels. `max_expiry_round_trips_through_the_wire_form` composes
the encode and the decode over all three states with no server, and
`collection_max_expiry_round_trips_against_live_gateway` creates four collections against a real
gateway — the three states, plus "inherit" a second time as an explicit 0 rather than an unset
field — and asserts the manifest reports each one as created.

Both halves are negative-controlled: a converter spec with four mutations and a live spec with three,
all killed. The one mutation deliberately absent from the live spec is sending -1 as 4294967295,
which cannot be caught against a real gateway for the truncation reason in the sibling thread; the
converter spec asserts the wire form directly instead.

core/protostellar/dispatcher.hxx
```diff
@@ -131,6 +134,79 @@ default_channel_arguments() -> grpc::ChannelArguments;
...
+
+  void OnReadDone(bool ok) override
+  {
+    if (!ok) {
+      return; // stream finished; OnDone follows
+    }
+    auto message = std::make_shared<Response>(std::move(current_));
+    asio::post(io_, [this, message]() {
+      on_row_(std::move(*message));
+    });
```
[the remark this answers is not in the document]
avsej: Accurate, and addressed later in the same series rather than here. CXXCBC-910 (#1048) makes the
reactor `enable_shared_from_this` and posts each row handler with `self = shared_from_this()`
(`dispatcher.hxx:290`), so a queued row handler holds the reactor alive and a completion that runs
first cannot free it underneath one. This thread is outdated against that change.

test/cng/protostellar/live_collection_admin.cxx
```diff
@@ -0,0 +1,375 @@
...
+  static_cast<void>(fixture.execute(std::move(stale)));
+
+  ops::management::bucket_create_request create_bucket{};
+  create_bucket.bucket = settings;
+  const auto bucket_created = fixture.execute(std::move(create_bucket));
+  // A cluster whose whole KV quota is already committed cannot host another bucket. That is a
+  // property of the cluster and not of the client, reported in the server's own words.
+  if (bucket_created.ctx.ec == couchbase::errc::common::invalid_argument &&
+      bucket_created.error_message.find("ram quota") != std::string::npos) {
+    skip("cluster has no spare KV quota for a test bucket (" + bucket_created.error_message + ")");
```
DemetrisChr: I think it would be better if this was treated as a failure, this could lead to this being skipped without us ever noticing, since the test suite would just appear green, and we might miss legitimate bugs due to the silently reduced test coverage. Ideally, the test workflow should be configured in a way that this does not happen.

core/protostellar/collection_admin_converter.hxx
```diff
@@ -0,0 +1,88 @@
...
+// Maps the core max_expiry sentinel onto max_expiry_secs. The proto field is unsigned, so the sign
+// the core struct carries has to become the field's presence: a negative value ("no expiry") is an
+// explicit 0, and 0 has no encoding at all -- nullopt here means leave the field unset. What an
+// unset field then means belongs to the caller, and it differs between the two RPCs: on create it
+// is "inherit the bucket default", which is what 0 asked for, and on update it is "leave the
+// collection as it is", which is not.
+[[nodiscard]] inline auto
+encode_max_expiry(std::int32_t max_expiry) -> std::optional<std::uint32_t>
+{
+  if (max_expiry < 0) {
```
DemetrisChr, line 51: We document this as follows in the Public API:
> Value of 0 sets max_expiry to the bucket-level setting and value of -1 to set it as no-expiry.

I think a value lower than -1 should be rejected with `invalid_argument` to keep it consistent with classic: https://github.com/couchbase/couchbase-cxx-client/blob/acdbe538356ee59d1004976d9c4bbf7df718606f/core/operations/management/collection_create.cxx#L40-L44

core/cluster.cxx
DemetrisChr, line 768: When history retention is enabled, `execute_with_bucket_capability_check` is called first, which attempts to open bucket mcbp connections. `execute_with_bucket_capability_check` probably needs to be updated so it's a noop when the couchbase2 protocol is in use.

test/cng/protostellar/live_collection_admin.cxx
```diff
@@ -0,0 +1,375 @@
...
+      { "collection_update_max_expiry_against_live_gateway",
+        collection_update_max_expiry_against_live_gateway,
+        timeout::integration,
+        test_env::cluster_only },
+      { "collection_max_expiry_is_layered_over_the_bucket_default",
+        collection_max_expiry_is_layered_over_the_bucket_default,
+        timeout::integration,
+        test_env::cluster_only },
+    },
+  };
```
DemetrisChr, line 638: It would be good to add round trip tests for the history retention setting to ensure that it behaves as expected (see comment about `execute_with_bucket_capability_check` above)

core/protostellar/component.cxx
```diff
@@ -1237,4 +1242,271 @@ component::execute(
...
+  // The sentinel translation create does, minus the one value update cannot express. An explicit 0
+  // on the wire already means "no expiry", and an unset field means "leave the collection as it
+  // is" rather than "inherit the bucket default", so nothing is left to carry a max_expiry of 0 --
+  // the reset to the bucket's own expiry that the classic path sends as maxTTL=0. Refuse it:
+  // reporting success for a setting the gateway was never asked to apply would leave the caller's
+  // documents expiring on the old policy.
+  if (request.max_expiry.has_value() && *request.max_expiry == 0) {
+    operations::management::collection_update_response response;
+    response.ctx.client_context_id = client_context_id;
+    response.ctx.ec = errc::common::feature_not_available;
```
DemetrisChr: I've raised this on slack, this looks like an issue with the Protostellar protocol: https://couchbase.slack.com/archives/C04DB7P157T/p1786014177197979

test/cng/protostellar/live_collection_admin.cxx
avsej: Agreed, and done on both halves. The skip is gone: a bucket create that fails now fails the case,
with the server's message carried into the assertion so the log says why. On the workflow side the
CNG cluster action patches the data service quota and fails the job outright if the new value never
applies, so the shortfall this skip existed for is reported once, up front, rather than discovered
by each case in turn.

The same reasoning removed the other silent discards in this file. `manifest_or_skip()` was called
for its side effect and its result thrown away in five places; it is now
`require_default_collection()`, which asserts the fixture's bucket answers with a default scope and
collection. The leftover-cleanup drops asserted nothing either, and now require success or
not-found — any other error is the transport rather than the leftover, and swallowing it made the
next request fail for a reason the run had already seen.

The two that remain are in guard destructors, which run while a failed assertion is unwinding: a
second exception there terminates the process and replaces the failure message with a crash. A drop
that fails is picked up by the next run's cleanup assertion instead.

core/protostellar/collection_admin_converter.hxx
avsej: Confirmed and fixed, on create and update both.

`collection_create.cxx:39-44` and `collection_update.cxx:39-45` refuse `max_expiry < -1` before encoding. couchbase2 did not, and `encode_max_expiry()` maps *every* negative onto the same wire form — so `-2` was sent as the no-expiry sentinel and created a never-expiring collection while reporting success, where the same call over `couchbase://` returns `invalid_argument`.

The refusal sits in the component, ahead of the encode, so the converter's mapping stays total over the values it is now guaranteed; its comment names the precondition rather than leaving the footgun implicit. `collection_max_expiry_below_the_sentinel_is_refused_against_live_gateway` covers both RPCs and asserts the collection was not created and the updated one kept its TTL. Both halves are negative-controlled: with either guard removed the request is sent, succeeds, and the case fails.

core/cluster.cxx
avsej: Confirmed and fixed. The guard went inside `execute_with_bucket_capability_check` rather than at the two call sites in `cluster::execute`, so any later caller is covered by it; `is_protostellar()` is already the marker the compound operations use to avoid the same MCBP bootstrap.

What routing straight through then means was measured rather than assumed. The server does refuse history retention on a couchstore bucket, so nothing is silently accepted — but the two RPCs report that refusal differently, `internal_server_failure` from `CreateCollection` and `invalid_argument` from `UpdateCollection`, because only the update path's error chain in `collectionadminserver.go` tests `cbmgmtx.ErrServerInvalidArg`. Both are asserted as measured in `collection_history_retention_on_a_couchstore_bucket_against_live_gateway`, which is also the negative control for this guard: with it removed the requests never reach the server.

Worth noting for parity: the classic path answers this off the bucket capability with `feature_not_available`, and couchbase2 cannot, because there is no cluster map to read the capability from. That difference is inherent to the transport.

test/cng/protostellar/live_collection_admin.cxx
avsej: Added, in two cases.

`collection_history_retention_round_trips_against_live_gateway` creates a magma bucket, a collection with history on, and one left at the bucket's default, then updates the second one on. The bucket is created with `history_retention_collection_default` false deliberately: a magma bucket otherwise defaults its collections to history *on*, so an assertion of "on" would pass just as well against a client that dropped the field before sending it.

`collection_history_retention_on_a_couchstore_bucket_against_live_gateway` is the other side and the negative control for the `execute_with_bucket_capability_check` guard.

Two things the cases had to be built around, both measured:

- The magma bucket asks for 1024 MB, which is the gateway's own floor (`bucketadminserver.go`), not the server's — `POST /pools/default/buckets` with `storageBackend=magma&ramQuotaMB=100` returns 202 against Server 8.1.0-1743. The CNG cluster action's data quota therefore goes 512 MiB → 1280 MiB: that bucket plus the 100 MB test bucket, with `ctest` running the CNG suites serially so no two test buckets coexist.
- Only "on" is observable. The gateway sets `history_retention_enabled` when the collection has history and leaves it unset otherwise (`collectionadminserver.go:86-88`), so "off" and "the server did not say" arrive identically; the classic path reads ns_server's own `"history": false` and does report it. The case asserts that unobservability explicitly and labels it a tripwire — it fails the day the gateway sends the field for an off collection, which is the day the state becomes readable.

core/protostellar/component.cxx
avsej: Confirmed, and the Slack account of it holds up against the gateway source.

`UpdateCollection` carries `optional uint32 max_expiry_secs`. The gateway reads it at `collectionadminserver.go:322-328` as: absent → `MaxTTL` nil, `0` → `-1`, anything else → itself. `cbmgmtx.UpdateCollectionOptions.MaxTTL` is a `*int32` and `mgmt.go:677-678` posts `maxTTL` only when it is non-nil, so an absent field means nothing is posted and the server keeps the collection's current expiry.

| intent | classic | couchbase2 |
|---|---|---|
| preserve | leave unset | leave unset |
| no expiry | `maxTTL=-1` | `max_expiry_secs=0` |
| reset to the bucket default | `maxTTL=0` | **no encoding** |
| a specific TTL | `maxTTL=<n>` | `max_expiry_secs=<n>` |

`CreateCollection` is unaffected: there an unset field *is* "inherit the bucket default", which is what a core `max_expiry` of 0 asks for. Only on update do the two meanings of "unset" diverge.

This client refuses that one value with `feature_not_available` rather than sending the encoding that would report success without applying anything, and `collection_update_max_expiry_against_live_gateway` asserts both the refusal and that the collection kept the expiry it had. Closing it properly needs a wire change — there is no value of an unsigned field left to mean "reset".

gocb does not surface the gap because its `UpdateCollectionSettings.MaxExpiry` is a plain `time.Duration` where 0 means "not specified", so it cannot tell the two apart to begin with.

avsej:
# Response — PR #1040, round 2

All five points are addressed. Four became code changes; the fifth is confirmed and is a gap in the
protocol rather than in this commit. Two things were measured along the way that are worth stating
separately, at the end.

## 1. The RAM-quota skip (`live_collection_admin.cxx:321`)

> I think it would be better if this was treated as a failure, this could lead to this being
> skipped without us ever noticing, since the test suite would just appear green, and we might miss
> legitimate bugs due to the silently reduced test coverage. Ideally, the test workflow should be
> configured in a way that this does not happen.

Agreed, and done on both halves.

The skip is gone; a bucket create that fails now fails the case, and the server's message is
carried into the assertion so the log says why. The workflow side was already partly there — the
CNG cluster action patches the data service quota and fails the job if the new value never applies,
rather than letting the cases discover the shortfall one at a time — and this round raises that
quota again for the reason under point 4.

The same reasoning removed a second silent discard in the file. Every case began with
`static_cast<void>(manifest_or_skip(fixture))`, which threw away a manifest that had something to
say; it is now `require_default_collection()`, which asserts the fixture's bucket answers with a
default scope and collection. A bucket that is missing, empty, or not the one the suite was pointed
at used to surface several requests later as whatever the first create happened to report. The
leftover-cleanup drops were discarding their result too, and now assert that the answer is either
success or not-found — anything else is the transport rather than the leftover, and swallowing it
made the *next* request fail for a reason the run had already seen and thrown away.

Two `static_cast<void>` calls remain, both in guard destructors. They run while a failed assertion
is unwinding, and a second exception out of a destructor terminates the process, which would
replace the failure message with a crash. A drop that fails there is picked up by the next run,
where the cleanup assertion above reports it.

## 2. Values below -1 (`collection_admin_converter.hxx:47`)

> I think a value lower than -1 should be rejected with `invalid_argument` to keep it consistent
> with classic

Confirmed and fixed on both RPCs. `collection_create.cxx:39-44` and `collection_update.cxx:39-45`
refuse `max_expiry < -1` before encoding, and couchbase2 did not: `encode_max_expiry()` maps every
negative onto the same wire form, so `-2` was sent as the no-expiry sentinel and created a
never-expiring collection while reporting success. The same call over `couchbase://` returns
`invalid_argument`.

The refusal sits in the component, before the encode, so the converter's mapping stays total over
the values it is now guaranteed. `collection_max_expiry_below_the_sentinel_is_refused_against_live_gateway`
covers both RPCs and asserts the collection was not created and the updated one was left alone.

## 3. `execute_with_bucket_capability_check` (`cluster.cxx:768`)

> When history retention is enabled, `execute_with_bucket_capability_check` is called first, which
> attempts to open bucket mcbp connections. `execute_with_bucket_capability_check` probably needs
> to be updated so it's a noop when the couchbase2 protocol is in use.

Confirmed and fixed. The guard went inside `execute_with_bucket_capability_check` rather than at
the two call sites in `cluster::execute`, so any later caller gets it too — `is_protostellar()` is
already the marker the compound operations use for the same reason.

What routing straight through means was measured rather than assumed. The server does still refuse
history retention on a couchstore bucket, so the request is not silently accepted — but see the
first note at the end for what the caller is told.

## 4. Round trips for history retention (`live_collection_admin.cxx:372`)

> It would be good to add round trip tests for the history retention setting to ensure that it
> behaves as expected

Added, in two cases.

`collection_history_retention_round_trips_against_live_gateway` creates a magma bucket, a collection
with history on, and a collection left at the bucket's default, then updates the second one on. The
bucket is created with `history_retention_collection_default` false on purpose: a magma bucket
otherwise defaults its collections to history *on*, and an assertion of "on" would then pass just as
well against a client that dropped the field before sending it.

`collection_history_retention_on_a_couchstore_bucket_against_live_gateway` is the other side, and it
is the one that pins the fix in point 3: it sends the same requests to a bucket that cannot carry
history and asserts they come back refused by the server. Without the guard they never reach the
server at all.

Two costs came with this, both measured:

- The magma bucket has to ask for 1024 MB. That is the gateway's own floor
  (`bucketadminserver.go`), not the server's — `POST /pools/default/buckets` with
  `storageBackend=magma&ramQuotaMB=100` returns 202 against Server 8.1.0-1743. The CNG cluster
  action's data service quota therefore goes from 512 MiB to 1280 MiB, which is that bucket plus
  the 100 MB test bucket. `ctest` runs the CNG suites serially, so no two test buckets are ever
  allocated at once. Three KV nodes reserve it each, 3.8 GiB of the runner's 15; the run this was
  sized against reported 13 Gi available before the cluster started.
- Only "on" is observable. The gateway populates `history_retention_enabled` when the collection
  has history and leaves it unset otherwise (`collectionadminserver.go:86-88`), so "off" and "the
  server did not say" arrive identically. The classic path reads ns_server's own `"history": false`
  and reports it. The case asserts the unobservability explicitly, labelled as a tripwire: it fails
  the day the gateway starts sending the field for an off collection, which is the day the state
  becomes readable.

## 5. The Protostellar max-expiry gap (`component.cxx:1423`)

> I've raised this on slack, this looks like an issue with the Protostellar protocol

Confirmed, and the Slack account of it is exactly right. Tracing it through:

`UpdateCollection` carries `optional uint32 max_expiry_secs`, and the gateway reads it at
`collectionadminserver.go:322-328` as: absent → `MaxTTL` nil; `0` → `-1`; anything else → itself.
`cbmgmtx.UpdateCollectionOptions.MaxTTL` is a `*int32` and `mgmt.go:677-678` adds `maxTTL` to the
form only when it is non-nil, so an absent `max_expiry_secs` means the field is never posted and the
server leaves the collection's expiry as it is.

So of the four states, three have an encoding and the third does not:

| intent | classic | couchbase2 |
|---|---|---|
| preserve | leave unset | leave unset |
| no expiry | `maxTTL=-1` | `max_expiry_secs=0` |
| reset to the bucket default | `maxTTL=0` | **no encoding** |
| a specific TTL | `maxTTL=<n>` | `max_expiry_secs=<n>` |

`CreateCollection` does not have the problem, because there an unset field *is* "inherit the bucket
default" — which is what a core `max_expiry` of 0 asks for. It is only on update that the two
meanings of "unset" diverge.

The client refuses that one value with `feature_not_available` rather than sending the encoding that
would report success without applying anything, and
`collection_update_max_expiry_against_live_gateway` asserts both the refusal and that the collection
kept the expiry it had. That is the most this side can do; closing it needs a wire change, since
there is no value of an unsigned field left to mean "reset".

gocb does not surface the gap because its `UpdateCollectionSettings.MaxExpiry` is a plain
`time.Duration` where 0 means "not specified" — it cannot tell the two apart to begin with. The C++
core spends an `optional` on the distinction, which is why it can report it.

## Also measured

Two things that are not in the review but came out of testing it, both on the gateway side rather
than here:

**The same server refusal reports differently on the two RPCs.** History retention on a couchstore
bucket comes back as `internal_server_failure` from `CreateCollection` and `invalid_argument` from
`UpdateCollection`. The cause is in `collectionadminserver.go`: the update path's error chain tests
`cbmgmtx.ErrServerInvalidArg` and maps it to an invalid-argument status, and the create path's chain
does not, so the same ns_server 400 falls through to a generic status there. Both codes are asserted
as they are, not as they ought to be, so the cases fail if that is evened out.

Related to it: the classic path answers this request `feature_not_available` off the bucket
capability, and couchbase2 cannot, since there is no cluster map to read the capability from. The
difference is inherent to the transport, not to the mapping.

**The `windows-11-arm` build job.** It is failing, and not because of this stack: the same job fails
on `main` and on every branch in the series. Its log ends part-way through a compile with no
diagnostic and the build step never reaches a conclusion, which is a runner abort rather than a
compile error.

## Verification

Against a live CNG gateway — Server 8.1.0-1743, Cloud Native Gateway 1.2.2-123, operator 2.8.0 —
and not only in-process.

DemetrisChr (approved):

PR #1041, opened 2026-07-28
CXXCBC-901: route query index management over the couchbase2 transport

avsej:
Query index management was rejected by the HTTP execute front-door gate on
couchbase2. Wire it onto the unary gRPC path via admin.query.v1.

The converter owns the translation in both directions. GetAllIndexes decodes
into core query_index structs -- the proto IndexType/IndexState enums have no
`unknown` sentinel and map to the lowercase strings the SDK uses. Create and
drop pick the primary or the secondary RPC from the core is_primary flag, and a
conditional secondary index is reported as feature_not_available, since the
proto CreateIndex has no condition field and encoding one would create an index
over the whole collection under the name the caller asked for.

Index keys are escaped into exactly one N1QL identifier each. The gateway builds
the CREATE INDEX statement and splices the field list into it verbatim, so a key
that leaves its own quoting adds terms to the statement rather than naming a
field. The escape is the one the query parser reads back -- a backslash escapes
the character after it -- so a backtick and a backslash each arrive prefixed by
one. A key that is already one complete identifier is passed through, which is
the form GetAllIndexes reports keys in, after being checked to end where it
claims to.

build_deferred_indexes() issues the single BuildDeferredIndexes request on
protostellar origins rather than the get-deferred + build pair, which has no
couchbase2 encoding: the RPC takes a keyspace and builds every deferred index in
it, and only the composite knows that its own list is that set. RFC 77 lists the
operation among the single ones rather than the composite ones, so it carries no
sub-operation spans there; the classic path keeps both.

Routing is declared by detecting the component's execute() overload for a
request type rather than by a separate list, so an overload the list does not
mention cannot go unreachable.

---
Stacked PR **19/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1040 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -306,6 +306,29 @@ is_feature_supported(const operations::management::search_index_upsert_request& 
 }
 } // namespace
 
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+// Whether the couchbase2 component can carry this request, answered by asking whether it has an
+// execute() overload for it.
+//
+// The alternative -- a hand-maintained list of request types -- has to agree with the overload set
+// and nothing checks that it does. A type in the list with no overload fails to compile, so that
+// half is safe; an overload with no entry in the list is silently unreachable, and that is how
+// query_index_build_deferred_request came to be wired to a gateway RPC that no caller could
+// reach. Detecting the overload makes the list *be* the overload set.
+template<typename Request, typename = void>
+struct component_routes : std::false_type {
+};
+
+template<typename Request>
+struct component_routes<
+  Request,
+  std::void_t<decltype(std::declval<protostellar::component&>().execute(
+    std::declval<Request>(),
+    std::declval<utils::movable_function<void(typename Request::response_type)>>()))>>
+  : std::true_type {
+};
+#endif
+
 class cluster_impl : public std::enable_shared_from_this<cluster_impl>
 {
 public:
@@ -773,22 +796,7 @@ public:
     }
 #ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
     if (auto component = protostellar_component(); component) {
-      if constexpr (std::is_same_v<Request, operations::query_request> ||
-                    std::is_same_v<Request, operations::analytics_request> ||
-                    std::is_same_v<Request, operations::search_request> ||
-                    std::is_same_v<Request, operations::document_view_request> ||
-                    std::is_same_v<Request, operations::management::bucket_get_all_request> ||
-                    std::is_same_v<Request, operations::management::bucket_get_request> ||
-                    std::is_same_v<Request, operations::management::bucket_create_request> ||
-                    std::is_same_v<Request, operations::management::bucket_update_request> ||
-                    std::is_same_v<Request, operations::management::bucket_drop_request> ||
-                    std::is_same_v<Request, operations::management::bucket_flush_request> ||
-                    std::is_same_v<Request, operations::management::scope_get_all_request> ||
-                    std::is_same_v<Request, operations::management::scope_create_request> ||
-                    std::is_same_v<Request, operations::management::scope_drop_request> ||
-                    std::is_same_v<Request, operations::management::collection_create_request> ||
-                    std::is_same_v<Request, operations::management::collection_update_request> ||
-                    std::is_same_v<Request, operations::management::collection_drop_request>) {
+      if constexpr (component_routes<Request>::value) {
         component->execute(std::move(request), std::forward<Handler>(handler));
         return;
       } else {
--- a/core/impl/query_index_manager.cxx
+++ b/core/impl/query_index_manager.cxx
@@ -432,6 +432,39 @@ public:
                                     collection_name,
                                     options.parent_span);
 
+    // A closed cluster is the only thing origin() reports an error for, and it cannot be asked
+    // which transport it is on. Report it here rather than routing on a state that could not be
+    // read: the composite below would open a sub-operation span for a request it never sends.
+    const auto [origin_ec, origin] = core_.origin();
+    if (origin_ec) {
+      top_obs_rec->finish(origin_ec);
+      return handler(couchbase::error{ origin_ec });
+    }
+
+    // Over couchbase2 the gateway does server-side what the two round trips below do here:
+    // BuildDeferredIndexes takes the keyspace and builds every deferred index in it. RFC 77 lists
+    // it among the single operations rather than the composite ones -- only WatchIndexes is
+    // composite there -- so it is dispatched as one request and carries no sub-operation spans.
+    if (origin.uses_protostellar()) {
+      core::operations::management::query_index_build_deferred_request request{
+        bucket_name,
+        scope_name.empty() ? std::optional<std::string>{}
+                           : std::optional<std::string>{ scope_name },
+        collection_name.empty() ? std::optional<std::string>{}
+                                : std::optional<std::string>{ collection_name },
+        {},
+        {},
+        timeout,
+        top_obs_rec->operation_span(),
+      };
+      return core_.execute(
+        std::move(request),
+        [top_obs_rec = std::move(top_obs_rec), handler = std::move(handler)](const auto& resp) {
+          top_obs_rec->finish(resp.ctx.retry_attempts, resp.ctx.ec);
+          handler(core::impl::make_error(resp.ctx));
+        });
+    }
+
     auto get_all_deferred_obs_rec = create_suboperation_observability_recorder(
       core::tracing::operation::mgr_query_get_all_deferred_indexes,
       bucket_name,
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -31,6 +31,7 @@
 #include "core/protostellar/error_utils.hxx"
 #include "core/protostellar/kv_converter.hxx"
 #include "core/protostellar/query_converter.hxx"
+#include "core/protostellar/query_index_admin_converter.hxx"
 #include "core/protostellar/search_converter.hxx"
 #include "core/protostellar/view_converter.hxx"
 
@@ -41,6 +42,7 @@
 
 #include <couchbase/admin/bucket/v1/bucket.grpc.pb.h>
 #include <couchbase/admin/collection/v1/collection.grpc.pb.h>
+#include <couchbase/admin/query/v1/query.grpc.pb.h>
 #include <couchbase/kv/v1/kv.grpc.pb.h>
 #include <couchbase/search/v1/search.grpc.pb.h>
 #include <couchbase/view/v1/view.grpc.pb.h>
@@ -65,6 +67,7 @@ namespace search_v1 = ::couchbase::search::v1;
 namespace view_v1 = ::couchbase::view::v1;
 namespace bucket_admin_v1 = ::couchbase::admin::bucket::v1;
 namespace collection_admin_v1 = ::couchbase::admin::collection::v1;
+namespace query_admin_v1 = ::couchbase::admin::query::v1;
 
 // Defined here rather than in the header so the generated gRPC types stay out of every consumer of
 // component.hxx.
@@ -76,6 +79,7 @@ struct component::stubs {
   std::unique_ptr<view_v1::ViewService::Stub> view;
   std::unique_ptr<bucket_admin_v1::BucketAdminService::Stub> bucket_admin;
   std::unique_ptr<collection_admin_v1::CollectionAdminService::Stub> collection_admin;
+  std::unique_ptr<query_admin_v1::QueryAdminService::Stub> query_admin;
 };
 
 namespace
@@ -147,7 +151,8 @@ component::component(asio::io_context& io, component_config config)
              search_v1::SearchService::NewStub(config.channel),
              view_v1::ViewService::NewStub(config.channel),
              bucket_admin_v1::BucketAdminService::NewStub(config.channel),
-             collection_admin_v1::CollectionAdminService::NewStub(config.channel) }) }
+             collection_admin_v1::CollectionAdminService::NewStub(config.channel),
+             query_admin_v1::QueryAdminService::NewStub(config.channel) }) }
   , authorization_{ authorization_header(config.credentials) }
   , timeouts_{ config.timeouts }
   // Initialised last (see the declaration order in the header), so the channel can be moved in.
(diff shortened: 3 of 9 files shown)


core/cluster.cxx
```diff
@@ -735,6 +757,39 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+                    std::is_same_v<Request, operations::management::scope_create_request> ||
+                    std::is_same_v<Request, operations::management::scope_drop_request> ||
+                    std::is_same_v<Request, operations::management::collection_create_request> ||
+                    std::is_same_v<Request, operations::management::collection_update_request> ||
+                    std::is_same_v<Request, operations::management::collection_drop_request> ||
+                    std::is_same_v<Request, operations::management::query_index_get_all_request> ||
+                    std::is_same_v<Request, operations::management::query_index_create_request> ||
+                    std::is_same_v<Request, operations::management::query_index_drop_request> ||
+                    std::is_same_v<Request,
+                                   operations::management::query_index_build_deferred_request>) {
```
avsej: This request type is routed here and has a `component::execute` overload — but **no caller in the SDK constructs it**, so both are dead code.

The public `query_index_manager::build_deferred_indexes()` (`core/impl/query_index_manager.cxx:420`) is a composite that issues `query_index_get_all_deferred_request` followed by `query_index_build_request`. Neither is in this allowlist and neither has a component overload, so the operation fails with `feature_not_available` at its first step over couchbase2 — even though the gateway exposes exactly the RPC needed and this commit wired it.

Either route the two request types the composite issues, or have the composite use `query_index_build_deferred_request` on couchbase2 — the second is closer to the gateway's model, since `BuildDeferredIndexes` does server-side what the composite does in two round trips.

The underlying cause is that routing must be declared in two unlinked places: this `is_same_v` chain and the `component::execute` overload set, with no compile-time relationship between them. A detection trait would make the allowlist *be* the overload set:

```cpp
template<typename Request, typename = void>
struct component_routes : std::false_type {};

template<typename Request>
struct component_routes<Request, std::void_t<decltype(std::declval<component&>().execute(
  std::declval<Request>(),
  std::declval<utils::movable_function<void(typename Request::response_type)>>()))>>
  : std::true_type {};
```

That catches the dead half. The other half — an operation the public API issues that nothing routes — needs a test that walks the public management API over a couchbase2 cluster.

core/protostellar/query_index_admin_converter.hxx
```diff
@@ -0,0 +1,97 @@
...
+// IndexState enums have no `unknown` sentinel; they map to the lowercase strings the rest of the
+// SDK uses.
+
+#include "couchbase/management/query_index.hxx"
+
+#include <couchbase/admin/query/v1/query.pb.h>
+
+#include <string>
+
+namespace couchbase::core::protostellar::query_index_admin
```
avsej, line 39: The accompanying test covers decoding only. Nothing exercises the encode side — the primary/secondary branch, index-key quoting, `num_replicas`, `deferred`, collection qualification, or the condition guard that reports unmappable. A regression in any of those reaches the gateway silently, and the live suite skips by default.

The most valuable case is key quoting with an identifier that needs escaping, since a mistake there produces a syntactically valid but wrong index. Asserting the condition guard in the same test also pins the one deliberate `feature_not_available` this converter emits.

Separately, RFC 77's Tracing section specifies the Protostellar RPC name for management `db.operation` (`CreateIndex`, `DropIndex`, `GetAllIndexes`, `BuildDeferredIndexes`), and that `WatchIndexes` carries no `db.operation` on its top-level span but contains `manager_query_wait_for_index_online` children. These spans keep the classic `manager_query_*` identifiers. Tracing is marked volatile in the RFC so this is not urgent, but it is cheaper now than retrofitting across nine services.

avsej (commented):
## Review — PR #1041 (`CXXCBC-901`)

This commit routes query index management over couchbase2: create, drop, get-all and build-deferred. The
converter handles index-key quoting, the primary/secondary distinction, deferred build and `num_replicas`, and the
condition field is correctly reported as unmappable rather than silently dropped — the couchbase2
`CreateIndexRequest` has no equivalent, and the comment says so.

The routing has a gap that makes one of the four operations unreachable from the public API while a different
request type is wired but never constructed.

### Findings

|---|---|---|---|---|
| 1 | high | correctness | `build_deferred_indexes()` fails over couchbase2, while the `BuildDeferredIndexes` RPC that is wired is dead code | `core/cluster.cxx:782` |
| 2 | medium | test-coverage | The encode side has no test at all | `core/protostellar/query_index_admin_converter.hxx:31` |
| 3 | low | spec-compliance | Management spans keep the classic `db.operation` identifier rather than the Protostellar RPC name | `core/protostellar/query_index_admin_converter.hxx:31` |

#### 1. A matched pair of routing defects around deferred index builds (high)

```cpp
operations::management::query_index_build_deferred_request>) {
```

`query_index_build_deferred_request` is in the couchbase2 allowlist and has a `component::execute` overload — but
no caller in the SDK constructs it. It is dead code.

Meanwhile the public `query_index_manager::build_deferred_indexes()`
(`core/impl/query_index_manager.cxx:420`) is a composite operation that issues
`query_index_get_all_deferred_request` and then `query_index_build_request`. Neither of those appears in the
allowlist, and neither has a component overload, so the operation fails with `feature_not_available` at its first
step over couchbase2.

Failure scenario: an application creates deferred indexes and calls `build_deferred_indexes()` to bring them
online. Over `couchbase://` this works. Over `couchbase2://` it returns `feature_not_available`, even though the
gateway exposes exactly the RPC needed and this commit routed it — just for a request type nothing produces.

The fix is either to route the two request types the public composite actually issues, or to make the composite
use `query_index_build_deferred_request` when the transport is couchbase2. The second is closer to the gateway's
model, since `BuildDeferredIndexes` does server-side what the composite does with two round trips.

The underlying cause is worth addressing too: routing has to be declared in two unlinked places — the
`component::execute` overload set and a hand-written `std::is_same_v` chain in `core/cluster.cxx` — with no
compile-time relationship between them, and no relationship to the operations the public API issues. A static
assertion tying the allowlist to the overload set would have caught the dead half:

```cpp
// component.hxx, after the class definition
template<typename Request, typename = void>
struct component_routes : std::false_type {};

template<typename Request>
struct component_routes<Request,
  std::void_t<decltype(std::declval<component&>().execute(
    std::declval<Request>(), std::declval<utils::movable_function<void(typename Request::response_type)>>()))>>
  : std::true_type {};
```

with `cluster.cxx` using `component_routes_v<Request>` in place of the `is_same_v` chain. That makes the allowlist
*be* the overload set. The other half — an operation the public API issues but nothing routes — needs the routed
surface to be checked against the operations the public API actually constructs, which a test that walks the
public management API over a couchbase2 cluster would cover.

(inline comment on `core/cluster.cxx:782`)

#### 2. The encode side has no test (medium)

`test/cng/protostellar/query_index_admin_converter.cxx` covers decoding. Nothing exercises the encode side: the
primary/secondary branch, index-key quoting, `num_replicas`, `deferred`, the collection qualification, or the
condition guard that returns unmappable. A regression in any of those reaches the gateway silently, and the live
suite skips by default.

The highest-value case is index-key quoting with an identifier that needs escaping, since that is where a
mistake produces a syntactically valid but wrong index:

```cpp
void
create_index_quotes_keys_and_reports_unmappable_conditions()
{
  operations::management::query_index_create_request request{};
  request.index_name = "ix";
  request.keys = { "`odd name`", "meta().id" };
  request.deferred = true;
  request.num_replicas = 2;
  const auto proto = query_index_admin::encode(request);
  assert_true(proto.has_value(), "a conditionless index is mappable");
  assert_eq(proto->fields_size(), 2, "both keys are carried");
  assert_true(proto->deferred(), "deferred build is requested");
  assert_eq(proto->num_replicas(), 2U, "replica count is carried");

  request.condition = "type = 'x'";
  assert_false(query_index_admin::encode(request).has_value(),
               "a conditional index has no couchbase2 mapping and must be reported unmappable");
}
```

#### 3. Management spans keep the classic `db.operation` identifier (low)

RFC 77's Tracing section specifies that management operations use the Protostellar RPC name for `db.operation` —
`BuildDeferredIndexes`, `CreateIndex`, `DropIndex`, `GetAllIndexes` — and that `WatchIndexes`, being a composite,
carries no `db.operation` on its top-level span but contains `manager_query_wait_for_index_online` children. The
spans here retain the classic `manager_query_*` identifiers.

Tracing is explicitly marked volatile in the RFC, so this is not urgent, but it is the kind of thing that is
cheaper to do while the operations are being added than to retrofit across nine services later.

### Verdict

Needs changes. Finding 1 means one of the four operations this commit exists to route is unreachable, while the
RPC that was wired for it cannot be reached either.

core/cluster.cxx
avsej: Confirmed and fixed, by the second route.

The first does not have a correct encoding: `query_index_build_request` names a list of indexes and
`BuildDeferredIndexesRequest` carries a keyspace and no name list, so routing it would build every
deferred index in the keyspace whenever a caller named a subset. Only the composite knows that its
own list *is* that set, so the branch belongs there —
`build_deferred_indexes()` now issues `query_index_build_deferred_request` on protostellar origins
and keeps the two round trips elsewhere. RFC 77 lists the operation among the single ones and names
`WatchIndexes` as the only composite, so over couchbase2 it is one span with no sub-operation
children; the classic path keeps both sub-recorders and their metrics.

The detection trait is adopted as written. The list and the overload set were compared before the
swap — twenty entries each, identical — so nothing routed today changes.

The other half is covered by
`build_deferred_indexes_through_the_public_api_against_live_gateway`, which connects with
`couchbase::cluster::connect()` and drives `cluster.query_indexes()` end to end. Disabling the
branch turns it red.

One qualification on the finding: `test_integration_management.cxx:2792` and `:3336` do construct
`query_index_build_deferred_request` directly, so it is reachable from the core API. Nothing in the
public API reached it, which is the half that mattered.

core/protostellar/query_index_admin_converter.hxx
avsej: Confirmed, and the quoting the finding asks for a test of did not exist.

The encode wrote `proto->add_fields(key)` for each key as supplied, and the gateway splices the field
list into its statement verbatim (`strings.Join(opts.Fields, ",")` in gocbcorex `cbqueryx`, reached
from `queryadminserver.go:311`). The classic path backticks each key unless it is backticked already
(`query_index_create.cxx:60-65`), and its integration test pins that with
`keys = {"field", "field2", "two words"}` asserting ``index_key[2] == "`two words`"``
(`test_integration_management.cxx:2554`). So `create_index(bucket, name, {"two words"})` created an
index over classic and was a syntax error over couchbase2.

Escaping is more than backticks, and the rule was taken from the parser: `n1ql.nex:40` accepts both
`` \` `` and `` `` `` inside a quoted identifier, but `ProcessEscapeSequences` (`util.go:91`) only
unescapes the backslash form — doubling parses and names a different field — and it rejects any
unrecognised `\X`, so a raw backslash has to be doubled too or the statement does not parse. A key
already forming one complete identifier is passed through, since that is the form `GetAllIndexes`
reports, after being checked to end where it claims to.

Measured against a live gateway with the key ``a`,`b``: wrapped in backticks without escaping, the
statement is valid and creates an index whose `index_key` reads ``["`a`", "`b`"]`` — one key became
two. Escaped, it is one identifier naming a field with a backtick in it, which the parser reads
(``SELECT 1 AS `a\`b` `` returns ``a`b``) but the indexer refuses to index, so nothing is created.
Both outcomes are asserted.

The encodes moved into the converter as pure functions, which is what made any of this reachable
from a test: 11 unit cases and 7 live ones, with 9 and 8 mutations killed respectively. The unit
case does not grade the converter against itself — it re-implements the parser's identifier rule
from `n1ql.nex`, rebuilds the comma-joined field list, and counts the identifiers a parser would
find in it.

The condition guard is pinned at both levels, including the live assertion that no index is created.

The RFC 77 `db.operation` mapping is **not** done here. It covers four managers and roughly forty
call sites across #1039, #1040, #1041 and #1042, three of which are merged, so doing it per-PR would
leave the SDK reporting RPC names for some management operations and classic names for others. It
wants one commit over the whole management surface with a live case on the recording tracer already
in `live_observability.cxx`. Carried as a followup with that scope written down.

Separately: the classic path has the same defect and is not fixed here. `query_index_create.cxx`
interpolates the index name and each key, and `utils::build_keyspace()` the bucket, scope and
collection, with plain backticks and no escaping.

avsej:
Both findings are fixed. The second one turned out to be a defect rather than a coverage gap: the
quoting it asks for a test of did not exist, and writing the test the review names is what showed
that. Details below, in the review's numbering.

## 1. `query_index_build_deferred_request` is routed but unreachable (`core/cluster.cxx:796`)

> This request type is routed here and has a `component::execute` overload — but **no caller in the SDK constructs it**, so both are dead code.

Confirmed, with one qualification: `test_integration_management.cxx:2792` and `:3336` construct it
directly, so it is reachable from the core API. No path through the public API reaches it, which is
the half that matters.

> The public `query_index_manager::build_deferred_indexes()` (`core/impl/query_index_manager.cxx:420`) is a composite that issues `query_index_get_all_deferred_request` followed by `query_index_build_request`. Neither is in this allowlist and neither has a component overload, so the operation fails with `feature_not_available` at its first step over couchbase2 — even though the gateway exposes exactly the RPC needed and this commit wired it.

Confirmed and fixed.

> Either route the two request types the composite issues, or have the composite use `query_index_build_deferred_request` on couchbase2 — the second is closer to the gateway's model, since `BuildDeferredIndexes` does server-side what the composite does in two round trips.

The second, and not only for closeness to the gateway's model: the first does not have a correct
encoding. `query_index_build_request` names a list of indexes to build, and
`BuildDeferredIndexesRequest` carries a keyspace and no name list — it builds every deferred index in
it. Routing that request type would silently build more than the caller asked for whenever the list
was a subset, and only the composite knows that its own list *is* every deferred index in the
keyspace.

So `build_deferred_indexes()` now issues the single request on protostellar origins and keeps the two
round trips elsewhere:

```cpp
if (const auto [origin_ec, origin] = core_.origin(); !origin_ec && origin.uses_protostellar()) {
  core::operations::management::query_index_build_deferred_request request{ ... };
  return core_.execute(std::move(request), ...);
}
```

This is also the shape RFC 77 asks for. Its management table lists `BuildDeferredIndexes` among the
single operations and names `WatchIndexes` as the only composite one, so over couchbase2 the
operation is one span with no sub-operation children. The classic path keeps both sub-operation
recorders and their metrics, unchanged.

> A detection trait would make the allowlist *be* the overload set

Adopted as written, under the same `#ifdef` as the component. The list and the overload set were
compared before the swap — twenty entries each, identical — so the replacement changes nothing about
what is routed today, and a future overload is routed by existing.

It also removes the list from the commits above: #1042 no longer adds nine entries for the search
index request types, since the overloads it adds are what route them. Its live search index cases
pass unchanged, which is the evidence that they still reach the component.

> The other half — an operation the public API issues that nothing routes — needs a test that walks the public management API over a couchbase2 cluster.

`build_deferred_indexes_through_the_public_api_against_live_gateway` connects with
`couchbase::cluster::connect()`, creates a deferred index through `cluster.query_indexes()`, and
builds it. It is the public path end to end rather than a core request, because the hole was
invisible from the core request that was already wired. Disabling the branch above turns it red.

## 2. The encode side is untested (`core/protostellar/query_index_admin_converter.hxx:31`)

> The accompanying test covers decoding only. Nothing exercises the encode side — the primary/secondary branch, index-key quoting, `num_replicas`, `deferred`, collection qualification, or the condition guard that reports unmappable.

Confirmed. The create, drop and build encodes moved out of `component.cxx` into the converter as pure
functions, which is what makes them reachable from a test at all; the four query index overloads in
`component.cxx` now call them and assign no proto field of their own. The suite is 12 unit cases and
7 live ones.

> The most valuable case is key quoting with an identifier that needs escaping, since a mistake there produces a syntactically valid but wrong index.

There was no quoting. The encode wrote `proto->add_fields(key)` for each key as supplied, and the
gateway splices the field list into the statement verbatim — `strings.Join(opts.Fields, ",")` in
gocbcorex's `cbqueryx`, reached from `queryadminserver.go:311`. The classic path backticks each key
unless it is backticked already (`query_index_create.cxx:60-65`), and its own integration test pins
that contract with `keys = {"field", "field2", "two words"}` asserting `index_key[2] == "`two
words`"` (`test_integration_management.cxx:2554`). So `create_index(bucket, name, {"two words"})`
created an index over classic and was a syntax error over couchbase2. gocb has the same asymmetry —
`queryindexprovider_core.go:50` quotes, `queryindexprovider_ps.go:104` passes the slice through.

Escaping the keys is more than adding backticks, and the rule was taken from the query parser rather
than inferred:

- `parser/n1ql/n1ql.nex:40` accepts both `` \` `` and `` `` `` inside a quoted identifier, but
  `ProcessEscapeSequences` (`parser/n1ql/util.go:91`) only unescapes the backslash form and copies a
  doubled backtick verbatim. Doubling parses and names a different field.
- The same function rejects any unrecognised `\X`, so a raw backslash has to be doubled or the whole
  statement fails to parse — and a key ending in a backslash otherwise escapes its own closing
  backtick and swallows the separator after it.

That is the rule the gateway itself uses for the names it escapes (`cbqueryx/utils.go:8`). A key that
is already one complete escaped identifier is passed through, because that is the form
`GetAllIndexes` reports keys in and the classic path accepts — but "already" is checked rather than
assumed, since a key that merely starts and ends with a backtick may close its identifier in the
middle.

Measured against the cluster below, with the key `` a`,`b ``:

- wrapped in backticks without escaping, the statement is valid and creates an index whose
  `index_key` reads ``["`a`", "`b`"]`` — one key became two.
- escaped, it is one identifier naming a field with a backtick in it. The parser reads that (``SELECT
  1 AS `a\`b` `` returns the field ``a`b``), but the indexer refuses to index it, so the create is
  rejected and no index appears.

Both outcomes are asserted, since which one a server gives is the server's business and neither is
what broken escaping produces. The refusal arrives as `internal_server_failure` through the
management RPC and as `invalid_argument` through the query service for the same statement; the case
does not assert on either, only that the request was not refused by the client.

> Asserting the condition guard in the same test also pins the one deliberate `feature_not_available` this converter emits.

Done at both levels. `a_conditional_secondary_index_cannot_be_encoded` covers the three states of the
guard, and `a_conditional_secondary_index_is_refused_over_couchbase2` asserts the refusal against the
gateway and that no index was created. The live case asserts rather than skips: the service is known
to be served by the time it runs, so `feature_not_available` there is the client declining to send.

> Separately, RFC 77's Tracing section specifies the Protostellar RPC name for management `db.operation` … Tracing is marked volatile in the RFC so this is not urgent, but it is cheaper now than retrofitting across nine services.

Not done here, and not deferred silently — the reasoning is that doing it per-PR is worse than doing
it late. The mapping covers four managers and roughly forty call sites across #1039, #1040, #1041 and
#1042, three of which are merged, so converting one service at a time leaves the SDK reporting RPC
names for some management operations and classic names for others. It wants one commit over the whole
management surface, with a live case built on the recording tracer already in `live_observability.cxx`
to assert the attribute per service. Carried as a followup with that scope written down.

## The classic path has the same defect, and is not fixed here

`query_index_create_request::encode_to()` builds its statement with `fmt::format` and plain
backticks: the index name at `query_index_create.cxx:69-79`, each key at `:60-65`, and the keyspace
via `utils::build_keyspace()` (`core/utils/keyspace.hxx:43-55`), which wraps bucket, scope and
collection the same way. None of them escape their contents, so a backtick in any of those values
closes its identifier and the remainder becomes statement text. The `WHERE` clause is interpolated
raw, which is by design — it is an expression.

It is left alone here because it is the classic transport, it is not what this commit routes, and
changing how the SDK builds N1QL wants its own review. Raising it rather than filing it quietly
because the couchbase2 path is now the stricter of the two.

## Copilot: `index_type_to_string()` falls through to "gsi"

> `index_type_to_string()` maps any unrecognized/unspecified proto enum value to "gsi" via the `default:` case, which can silently misreport the server-provided index type. This is especially risky if the proto adds new enum values in the future or if the field is left at its default value.

Applied, for the first of the two reasons. `INDEX_TYPE_GSI` is now its own case and the default
returns `"unknown"`, matching what `index_state_to_string()` already did. The schema has no third
value today, so nothing decodes differently — this is about what a schema bump would decode to.

The second reason does not hold: `INDEX_TYPE_VIEW` is 0, so an unset field decodes as `"view"`, and
a proto3 non-optional enum cannot tell an unset field from an explicit one. The gateway sets the
field on every index it reports.

## Also worth recording

`BuildDeferredIndexesResponse` returns the list of indexes it will build, and
`query_index_build_deferred_response` has no field to put it in, so it is discarded. The public
`build_deferred_indexes()` returns only an error, so nothing is lost to a caller today; gocb's
`BuildDeferredIndexes` does return the names.

## Verification

Run against a live CNG gateway — Server 8.1.0-1743, Cloud Native Gateway 1.2.2-123, operator 2.8.0 —
and not only in-process.

- `cng_query_index_admin_converter_test`: 12 cases passing.
- `cng_live_query_index_admin_test`: 7 cases passing, 0 skipped.
- Negative control on the converter: 9 mutations, all killed. They cover no quoting, quoting without
  escaping, the backslash left unescaped, the backtick doubled instead of escaped, a pre-quoted key
  trusted without checking, an escaped closing backtick accepted as a token, the condition guard
  inverted, an empty scope sent as present, and an unnamed primary drop carrying an empty name.
- Negative control on the live cases: 8 mutations, all killed — including the two-key breakout above
  and the build-deferred routing, which reverts to `feature_not_available` when the branch is
  disabled.

The escaping assertions do not grade the converter against itself: the unit case re-implements the
parser's identifier rule from `n1ql.nex`, reconstructs the comma-joined field list the gateway
builds, and counts the identifiers a parser would find in it. One token per key is the property; a
key that adds a term changes the count.

core/impl/query_index_manager.cxx
```diff
@@ -432,6 +432,30 @@ class query_index_manager_impl : public std::enable_shared_from_this<query_index
                                     collection_name,
                                     options.parent_span);
 
+    // Over couchbase2 the gateway does server-side what the two round trips below do here:
+    // BuildDeferredIndexes takes the keyspace and builds every deferred index in it. RFC 77 lists
+    // it among the single operations rather than the composite ones -- only WatchIndexes is
+    // composite there -- so it is dispatched as one request and carries no sub-operation spans.
+    if (const auto [origin_ec, origin] = core_.origin(); !origin_ec && origin.uses_protostellar()) {
```
DemetrisChr: We are not handling the case where `origin_ec` is set, it falls back to the get_all/build logic below. Do we need to return it in that case?

core/protostellar/query_index_admin_converter.hxx
```diff
@@ -0,0 +1,287 @@
...
+// leaving the backslash bare makes the parser reject the whole statement. Anything else inside the
+// quotes -- spaces, commas, parentheses, newlines -- is part of the identifier and needs no escape,
+// which is why "two words" is a key and not a syntax error.
+//
+// A key that is already one complete identifier is passed through, because that is the form
+// GetAllIndexes reports keys in and the classic query path accepts. "Already" is verified rather
+// than assumed: is_escaped_identifier() is what separates a caller's pre-quoted key from a key
+// carrying quotes that end early.
+[[nodiscard]] inline auto
+encode_index_key(const std::string& key) -> std::string
```
DemetrisChr, line 97: We already have escaping logic for fields in classic. Would it be possible to reuse that?

core/protostellar/component.cxx
```diff
@@ -1509,4 +1514,202 @@ component::execute(
...
+  operations::management::query_index_create_request request,
+  utils::movable_function<void(operations::management::query_index_create_response)>&& handler)
+  -> pending_call
+{
+  const auto client_context_id = request.client_context_id.value_or(std::string{});
+  if (!query_index_admin::can_encode(request)) {
+    operations::management::query_index_create_response response;
+    response.ctx.client_context_id = client_context_id;
+    response.ctx.ec = errc::common::feature_not_available;
+    handler(std::move(response));
```
DemetrisChr: In other operations `asio::post` was used in situations like this, with the justification:
```
// Posted rather than invoked here, for the reason fail_expired gives: completions arrive on the
// io context, never inline out of execute().
```

Do we need the same here?

core/protostellar/query_index_admin_converter.hxx
```diff
@@ -0,0 +1,287 @@
...
+  if (!proto.scope_name().empty()) {
+    index.scope_name = proto.scope_name();
+  }
+  if (!proto.collection_name().empty()) {
+    index.collection_name = proto.collection_name();
+  }
+  for (const auto& field : proto.fields()) {
+    index.index_key.push_back(field);
+  }
+  if (!proto.condition().empty()) {
```
DemetrisChr: `condition` is an optional proto field. We should use `has_condition()` instead of `empty()`

core/protostellar/query_index_admin_converter.hxx
```diff
@@ -0,0 +1,287 @@
...
+    index.condition = proto.condition();
+  }
+  if (!proto.partition().empty()) {
```
DemetrisChr: `partition` is an optional proto field. We should use `has_partition()` instead of `empty()`

core/protostellar/query_index_admin_converter.hxx
avsej: no we do not. the classic will be using this later.

core/impl/query_index_manager.cxx
avsej: Yes. Changed: the error is returned from here.

`cluster_impl::origin()` reports exactly one error, `network::cluster_closed`, and returns a default-constructed origin alongside it. The condition as written could not misroute — a default origin is not protostellar — but it decided routing from a value it had just been told it could not read.

What the fall-through produced was measured rather than assumed: the composite's first request goes through the HTTP `cluster_impl::execute`, which answers `stopped_` with the same `network::cluster_closed`. The error code a caller sees is unchanged by this fix.

Two things do change. The composite no longer opens a `manager_query_get_all_deferred_indexes` sub-operation span for a request that is never sent, so a closed cluster no longer produces a span tree describing work that did not happen. And the error is delivered without the query error context the composite's response carried; on a closed cluster that context holds only the error code, since nothing was encoded or sent.

No test accompanies this one. The user-visible outcome is identical before and after, so an assertion on it would pass either way, and the span-shape difference has no harness at this commit. Saying so rather than adding a check that cannot fail.

core/protostellar/component.cxx
avsej: Yes. Changed: the refusal is posted, carrying the same comment as the three siblings (`query::can_encode`, `analytics::can_encode`, `view::can_encode`). The response is now built by `stamped_management()`, so it carries the caller's `client_context_id` the same way the completion path does.

The property is asserted rather than left to convention. A new case, `protostellar_query_index_admin_component`, calls `execute()` against an `io_context` it has not started, asserts nothing has completed, then runs the context and asserts the completion arrives exactly once with `feature_not_available` and the caller's `client_context_id`. It fails on the previous code at the first assertion. No server is stood up, since the refusal is decided before anything is dispatched.

core/protostellar/query_index_admin_converter.hxx
avsej: Changed to `has_condition()`.

`condition` is field 9 in `GetAllIndexesResponse.Index` and carries presence, so reading emptiness reported an index whose clause the server sent as empty as an index with no clause at all. `scope_name` and `collection_name` in the same message do not carry presence, so those stay on `empty()` — there an empty name is the only form "not collection-qualified" can arrive in. That distinction is now stated at the decode site.

A converter case covers both this and `partition`: an `Index` with each explicitly set to the empty string decodes to two present-and-empty optionals, and one with neither set decodes to two `nullopt`s. It fails on the previous code.

core/protostellar/query_index_admin_converter.hxx
avsej: Changed to `has_partition()`, alongside `condition` above. `partition` is field 10 in `GetAllIndexesResponse.Index` and carries presence; the converter case added for `condition` asserts both.

avsej:
Round 2 addressed. Four of the five threads are code changes; the fifth was answered in-thread.

**`origin_ec` on `build_deferred_indexes()` (thread 8).** The error is now returned from there. `cluster_impl::origin()` reports exactly one error, `network::cluster_closed`, and returns a default-constructed origin with it, so the condition as written could not misroute — but it decided routing from a value it had just been told it could not read. What the fall-through produced was measured: the composite's first request goes through the HTTP `cluster_impl::execute`, which answers `stopped_` with the same `network::cluster_closed`, so the error code a caller sees is unchanged. What changes is that a closed cluster no longer opens a `manager_query_get_all_deferred_indexes` sub-operation span for a request that is never sent, and that the error arrives without the query error context the composite carried — on a closed cluster that context holds only the error code. No test accompanies it: the user-visible outcome is identical before and after, so an assertion on it would pass either way.

**The `can_encode` refusal was delivered inline (thread 10).** It is now posted, with the same comment as the three sibling guards, and the response is built by `stamped_management()` so it carries the caller's `client_context_id` the way the completion path does. A new case, `protostellar_query_index_admin_component`, calls `execute()` against an `io_context` it has not started, asserts nothing has completed, then runs the context and asserts the completion arrives exactly once with `feature_not_available` and the caller's `client_context_id`.

**`condition` and `partition` read by presence (threads 11 and 12).** Both are `optional string` in `GetAllIndexesResponse.Index`, so reading emptiness reported an index whose clause the server sent as empty as an index with no clause at all. `scope_name` and `collection_name` in the same message carry no presence and stay on `empty()`; the distinction is now stated at the decode site. A converter case asserts both directions.

**Reusing the classic escaping (thread 9).** Answered in-thread.

Verified at this commit: 32/32 `ctest -L cng` against Server 8.1.0-1743 / CNG 1.2.2-123 over `couchbase2://`, including the live query-index suite; the converter mutation sweep kills 11/11 and the new component sweep 2/2, with no mutation left unapplied; clang-format clean on every changed file.

core/impl/query_index_manager.cxx
```diff
@@ -432,6 +432,39 @@ class query_index_manager_impl : public std::enable_shared_from_this<query_index
                                     collection_name,
                                     options.parent_span);
 
+    // A closed cluster is the only thing origin() reports an error for, and it cannot be asked
+    // which transport it is on. Report it here rather than routing on a state that could not be
+    // read: the composite below would open a sub-operation span for a request it never sends.
+    const auto [origin_ec, origin] = core_.origin();
+    if (origin_ec) {
+      top_obs_rec->finish(origin_ec);
+      return handler(couchbase::error{ origin_ec });
```
DemetrisChr, line 441: Do we need `asio::post` here as well?

avsej: No — posting here would replace the error with a hang, and this is the one place in the function where that is true.

`cluster::close()` runs `io_.stop()` and then joins the io thread before it returns (`core/impl/public_cluster.cxx`, `do_close()`). By the time `origin()` can report `cluster_closed`, nothing is running that io_context, so a completion posted to it is never executed: the handler never fires and a `std::future` caller waits forever. The destructor comment on `watch_context` already guards the same hazard from the other direction — "so a std::future caller never observes broken_promise and a callback caller never hangs".

That was measured rather than reasoned about. Patching this early return into `asio::post(core_.io_context(), …)` and running the new case below makes it fail on the first assertion: the handler does not arrive within five seconds. Reverting it makes it pass.

The rationale quoted from the component does not carry over. There it is about the `pending_call` that `execute()` returns — an inline completion re-enters the caller before it holds the handle. This overload returns `void`, so there is no handle to get ahead of, and the three sibling `core_.origin()` guards in `core/impl/collection.cxx` deliver inline for the same reason.

Nothing changes for callers either: before this PR a closed cluster fell through to the composite, whose first request is answered by `cluster_impl::execute`'s `stopped_` branch — also inline.

Added `build_deferred_indexes_on_a_closed_cluster_answers_on_the_calling_thread`, which closes the cluster, calls the callback overload and asserts the handler arrives within a bounded wait. The bound is deliberate: it keeps a dropped completion a failing assertion instead of a hung suite. A mutation that drops the completion — the outcome posting produces — kills it.

thejcfactor (approved):

PR #1042, opened 2026-07-28
CXXCBC-901: route search index management over couchbase2

avsej:
Complete the index management ticket with the search index admin operations over
admin.search.v1 (unary).

Core search::index stores params/plan_params/source_params as JSON-object
strings; the proto Index models them as map<string,bytes> keyed by the object's
top-level members. The converter explodes each object into the map on encode and
rebuilds it on decode, and reports failure rather than dropping what it cannot
represent: a definition that lost a member would be created, or handed back,
short of what the caller wrote, and the loss would survive a get-modify-upsert.

upsert selects its RPC on the uuid -- a definition carrying one is an update and
goes to UpdateIndex, one without is a create and goes to CreateIndex. That is
the line the classic path draws and the line FTS draws behind the gateway. get,
get_all, drop, analyze_document and get_documents_count map directly; the
control operations select between paired RPCs from their bool
(ingest->Pause/Resume, query->Allow/Disallow, plan_freeze->Freeze/Unfreeze).
get_stats has no admin.search.v1 RPC and returns feature_not_available.

The operations the classic path validates locally reject an empty index name
without a round trip, a GetIndex response carrying no index is index_not_found,
and a successful call reports status "ok" -- the field the classic path fills
from FTS's own reply and core-API consumers read.

context, on the completion and on the paths that never reach the server, which

The CNG CI cluster raises the search service quota alongside the data one: these cases
build real fulltext indexes, where before nothing was built on that service.

---
Stacked PR **20/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1041 — merge the series bottom-up.

### DIFF
--- a/core/protostellar/component.cxx
+++ b/core/protostellar/component.cxx
@@ -33,6 +33,7 @@
 #include "core/protostellar/query_converter.hxx"
 #include "core/protostellar/query_index_admin_converter.hxx"
 #include "core/protostellar/search_converter.hxx"
+#include "core/protostellar/search_index_admin_converter.hxx"
 #include "core/protostellar/view_converter.hxx"
 
 #include <couchbase/error_codes.hxx>
@@ -43,6 +44,7 @@
 #include <couchbase/admin/bucket/v1/bucket.grpc.pb.h>
 #include <couchbase/admin/collection/v1/collection.grpc.pb.h>
 #include <couchbase/admin/query/v1/query.grpc.pb.h>
+#include <couchbase/admin/search/v1/search.grpc.pb.h>
 #include <couchbase/kv/v1/kv.grpc.pb.h>
 #include <couchbase/search/v1/search.grpc.pb.h>
 #include <couchbase/view/v1/view.grpc.pb.h>
@@ -68,6 +70,7 @@ namespace view_v1 = ::couchbase::view::v1;
 namespace bucket_admin_v1 = ::couchbase::admin::bucket::v1;
 namespace collection_admin_v1 = ::couchbase::admin::collection::v1;
 namespace query_admin_v1 = ::couchbase::admin::query::v1;
+namespace search_admin_v1 = ::couchbase::admin::search::v1;
 
 // Defined here rather than in the header so the generated gRPC types stay out of every consumer of
 // component.hxx.
@@ -80,6 +83,7 @@ struct component::stubs {
   std::unique_ptr<bucket_admin_v1::BucketAdminService::Stub> bucket_admin;
   std::unique_ptr<collection_admin_v1::CollectionAdminService::Stub> collection_admin;
   std::unique_ptr<query_admin_v1::QueryAdminService::Stub> query_admin;
+  std::unique_ptr<search_admin_v1::SearchAdminService::Stub> search_admin;
 };
 
 namespace
@@ -108,6 +112,21 @@ fail_expired(asio::io_context& io, const Request& request, Handler& handler) -> 
   return {};
 }
 
+// Deliver a prepared response carrying ec without sending anything. Posted rather than invoked:
+// completions arrive on the io context, never inline out of execute(), which would re-enter the
+// caller before it holds its pending_call.
+template<typename Response, typename Handler>
+auto
+fail_ctx(asio::io_context& io, Handler& handler, Response response, std::error_code ec)
+  -> pending_call
+{
+  response.ctx.ec = ec;
+  asio::post(io, [handler = std::move(handler), response = std::move(response)]() mutable {
+    handler(std::move(response));
+  });
+  return {};
+}
+
 // The same rule for the services whose responses carry their own error context rather than a
 // key_value one: they resolve a timeout exactly as the KV overloads do, so a non-positive one
 // would reach the dispatcher and produce a call with no deadline there too. The response the
@@ -116,11 +135,7 @@ template<typename Response, typename Handler>
 auto
 fail_expired_ctx(asio::io_context& io, Handler& handler, Response response) -> pending_call
 {
-  response.ctx.ec = errc::common::unambiguous_timeout;
-  asio::post(io, [handler = std::move(handler), response = std::move(response)]() mutable {
-    handler(std::move(response));
-  });
-  return {};
+  return fail_ctx(io, handler, std::move(response), errc::common::unambiguous_timeout);
 }
 
 // A management response carrying the caller's client_context_id, which is what correlates it with
@@ -140,6 +155,20 @@ stamped_management(const Request& request) -> typename Request::response_type
   response.ctx.client_context_id = request.client_context_id.value_or(std::string{});
   return response;
 }
+
+// FTS answers each of the search index management endpoints with {"status":"ok"} on success, and
+// the classic path copies that string into the response, where core-API consumers read it.
+// admin.search.v1 replies with empty messages, so a successful RPC is the same signal and carries
+// the same meaning. A failed one leaves the field empty rather than inventing a status the server
+// never reported; the error code is what describes it.
+template<typename Response>
+void
+set_ok_status(Response& response)
+{
+  if (!response.ctx.ec) {
+    response.status = "ok";
+  }
+}
 } // namespace
 
 component::component(asio::io_context& io, component_config config)
@@ -152,7 +181,8 @@ component::component(asio::io_context& io, component_config config)
              view_v1::ViewService::NewStub(config.channel),
              bucket_admin_v1::BucketAdminService::NewStub(config.channel),
              collection_admin_v1::CollectionAdminService::NewStub(config.channel),
-             query_admin_v1::QueryAdminService::NewStub(config.channel) }) }
+             query_admin_v1::QueryAdminService::NewStub(config.channel),
+             search_admin_v1::SearchAdminService::NewStub(config.channel) }) }
   , authorization_{ authorization_header(config.credentials) }
   , timeouts_{ config.timeouts }
   // Initialised last (see the declaration order in the header), so the channel can be moved in.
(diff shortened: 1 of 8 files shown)


avsej (commented):
## Review — PR #1042 (`CXXCBC-901`)

This commit routes the search index management surface over couchbase2 — twelve operations covering create, get,
get-all, drop, analyze-document, indexed-document-count and the six control operations. The breadth is good and
the control operations are all wired rather than a subset.

Two problems, both about the free-form JSON that search index definitions are made of.

### Findings

|---|---|---|---|---|
| 1 | high | correctness | Upsert always calls `CreateIndex`, so updating an existing index fails | `core/protostellar/component.cxx:1382` |
| 2 | high | correctness | Unparseable or non-object `params` / `plan_params` / `source_params` are silently discarded | `core/protostellar/search_index_admin_converter.hxx:21` |

#### 1. Upsert always calls `CreateIndex` (high)

```cpp
auto proto = std::make_shared<search_admin_v1::CreateIndexRequest>();
```

`search_index_upsert_request` maps unconditionally to `CreateIndexRequest`. The search admin service has a
dedicated `UpdateIndex` RPC (`admin/search/v1/search.proto:30`), and it is never called.

Failure scenario: an application calls `upsert_index` with a definition for an index that already exists — which
is what "upsert" means, and the normal way to change an analyzer or a type mapping. Over `couchbase://` the index
is updated. Over `couchbase2://` the gateway receives `CreateIndex` for an existing name and answers
`ALREADY_EXISTS`, which maps to `index_exists`. The update is impossible through this transport.

gocb, couchbase-jvm-clients and couchbase-net-client all use `UpdateIndex` for this case. The fix is to select the
RPC on whether the caller supplied a UUID — which is how the classic path distinguishes create from update — and
to carry `Index.uuid` through so the gateway can detect a conflicting concurrent update:

```cpp
if (request.index.uuid.empty()) {
  // no uuid: this is a create
  ... CreateIndexRequest ...
} else {
  auto proto = std::make_shared<search_admin_v1::UpdateIndexRequest>();
  search_index_admin::apply_index(request.index, *proto);
  proto->set_uuid(request.index.uuid);
  ... UpdateIndex ...
}
```

No test covers the update path, which is why this was not caught: the converter test exercises `params` only, and
the live suite skips by default.

(inline comment on `core/protostellar/component.cxx:1382`)

#### 2. Unparseable JSON parameters are silently discarded (high)

```cpp
// bit: core stores params/plan_params/source_params as single JSON-object strings, while the proto
```

`params`, `plan_params` and `source_params` are free-form JSON objects in the core representation and maps in the
proto, so the converter parses the string and copies the members across. When the parse fails, or when the parsed
value is not an object, the members are dropped and the operation proceeds — reporting success.

Failure scenario, and it is the common one: get-modify-upsert. An application calls `get_index`, edits one field
of the returned definition, and upserts it. If any of the three parameter blobs does not round-trip — because it
is not a flat object, or because a member value fails to re-parse — those members are silently dropped, and the
index is recreated with part of its definition missing. A type mapping or an analyzer configuration disappears,
the call reports success, and the index quietly stops matching what it used to.

The same helper is used in both directions, so the loss can happen on the read as well as the write.

Failing closed is the right behaviour, and it matches what the rest of this series does on the request side:

```cpp
if (!parsed_ok || !value.is_object()) {
  // A definition we cannot represent must not be silently truncated: the caller would get a
  // success for an index that is missing part of its configuration.
  return false;   // reported as invalid_argument by the caller
}
```

The most valuable missing test is a full round trip that asserts equality of all three parameter blobs, plus a
negative case with a non-object payload asserting that it is rejected rather than dropped. The current converter
test asserts `params` only and would survive crossing `plan_params` with `source_params`.

(inline comment on `core/protostellar/search_index_admin_converter.hxx:21`)

### Notes

- None of the twelve responses populate `status`, which the classic path always sets and which core-API consumers
  read. Worth a single shared helper rather than twelve additions.
- `GetIndexResponse` without an `index` field yields success plus a default-constructed index. Treating a missing
  index as `index_not_found` is more useful than an empty definition.
- The `AnalyzeDocument` response body is parsed without a `try`/`catch`. A non-array payload from the gateway
  therefore throws out of an Asio completion handler, which terminates rather than failing the operation. The
  parse should be guarded and mapped to a protocol error.
- The empty-index-name validation that the classic operations perform locally is skipped on this path, so an empty
  name becomes a gateway round trip returning a KV-shaped error code rather than `invalid_argument`.

### Verdict

Needs changes. Finding 1 makes `upsert_index` unable to update, and finding 2 can silently drop part of an index
definition during the standard get-modify-upsert cycle.

core/protostellar/component.cxx
```diff
@@ -0,0 +1,1797 @@
...
+    });
+}
+
+auto
+component::execute(
+  operations::management::search_index_upsert_request request,
+  utils::movable_function<void(operations::management::search_index_upsert_response)>&& handler)
+  -> pending_call
+{
+  auto proto = std::make_shared<search_admin_v1::CreateIndexRequest>();
```
avsej: `search_index_upsert_request` maps unconditionally to `CreateIndexRequest`, and the dedicated `UpdateIndex` RPC (`admin/search/v1/search.proto:30`) is never called.

Scenario: `upsert_index` with a definition for an index that already exists — the normal way to change an analyzer or type mapping. Over `couchbase://` the index is updated; over `couchbase2://` the gateway answers `ALREADY_EXISTS` → `index_exists`, so updating is impossible through this transport. gocb, couchbase-jvm-clients and couchbase-net-client all use `UpdateIndex` here.

Select the RPC on whether a UUID was supplied — how the classic path distinguishes create from update — and carry `Index.uuid` so the gateway can detect a conflicting concurrent update:

```cpp
if (request.index.uuid.empty()) {
  // create
} else {
  auto proto = std::make_shared<search_admin_v1::UpdateIndexRequest>();
  search_index_admin::apply_index(request.index, *proto);
  proto->set_uuid(request.index.uuid);
}
```

No test covers the update path, which is why this was not caught: the converter test exercises `params` only and the live suite skips by default.

core/protostellar/search_index_admin_converter.hxx
```diff
@@ -0,0 +1,126 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+// Translates core FTS index settings to and from couchbase.admin.search.v1. The one non-trivial
+// bit: core stores params/plan_params/source_params as single JSON-object strings, while the proto
```
avsej, line 21: When the parse of `params` / `plan_params` / `source_params` fails, or the parsed value is not an object, the members are dropped and the operation proceeds — reporting success.

The failing case is the common one: get-modify-upsert. `get_index`, edit one field, upsert. If any of the three blobs does not round-trip, those members are silently dropped and the index is recreated with part of its definition missing — a type mapping or analyzer configuration disappears, the call succeeds, and the index quietly stops matching what it did before. The same helper runs in both directions, so the loss can occur on the read as well.

Fail closed, matching what the rest of this series does on the request side:

```cpp
  if (!parsed_ok || !value.is_object()) {
    // A definition that cannot be represented must not be silently truncated: the caller would
    // get a success for an index missing part of its configuration.
    return false;   // surfaced as invalid_argument
  }
```

The most valuable missing test is a round trip asserting equality of all three blobs, plus a negative case with a non-object payload asserting rejection. The current test asserts `params` only and would survive crossing `plan_params` with `source_params`.

Four smaller items in this file and its component overloads: none of the twelve responses populate `status` (classic always does); a `GetIndexResponse` with no `index` yields success plus a default-constructed index rather than `index_not_found`; the `AnalyzeDocument` body is parsed with no `try`/`catch`, so a non-array payload throws out of an Asio handler and terminates; and the local empty-index-name validation the classic operations perform is skipped, so an empty name becomes a gateway round trip with a KV-shaped error code.

avsej:
Correcting the earlier position in this comment: these operations should get test coverage **before**
this merges, rather than as follow-up work.

The reason is ordering. The search query path merges first, and this PR is what makes that path
properly testable: with index management routed over couchbase2, a test can create an index, query
it and drop it on its own, instead of depending on an index provisioned out of band and on the
partition happening to be ready. Merging these operations untested spends exactly the leverage they
provide.

The gap, so the work has a definite scope rather than "add tests":

Nine operations are routed. What ships with them is four converter unit cases — the params map
explode/rebuild, and index decode — plus one live case that lists indexes. So `get_all` is the only
operation exercised end to end; `upsert`, `get`, `drop`, `analyze_document`,
`get_documents_count` and the three control operations have no coverage.

**The control operations are the priority.** Each selects between a pair of RPCs from a bool —
ingest to Pause/Resume, query to Allow/Disallow, plan_freeze to Freeze/Unfreeze. That is six RPC
selections, and an inverted pair fails silently: the call succeeds and does the opposite of what was
asked. Nothing in the current tests would tell the two apart.

**Create versus update.** `upsert` maps onto CreateIndex and carries the existing uuid as
`prev_index_uuid`, so a single path serves both. Neither direction is covered, nor is the behaviour
when the uuid is absent or stale.

**`get_stats`** returns `feature_not_available` because admin.search.v1 has no RPC for it. No case
asserts that, so the deliberate refusal is indistinguishable from the operation simply not being
routed — the same failure mode that makes a skip-on-error-code test unfalsifiable.

A single live lifecycle case — create, get, documents count, each control pair, drop — would cover
most of this in one pass, with the converter unit tests extended to pin the RPC chosen for each
control bool. That same case is what later gives the search query tests a real index to query, so it
is worth building here rather than deferring.

core/protostellar/component.cxx
avsej: Confirmed and fixed, with a root cause one layer below the one described — which makes it worse, not better: `prev_index_uuid` is not the wrong channel for the uuid, it is a channel that ends nowhere.

The gateway does carry it. `CreateIndex` copies `in.PrevIndexUuid` into `cbsearchx.Index.PrevIndexUUID` (`searchadminserver.go:60`) and `encodeIndex` writes it into the PUT body as `prevIndexUUID` (`cbsearchx/search.go:672`). FTS then reads that name only through `req.FormValue("prevIndexUUID")` (`cbgt/rest/rest_create_index.go:288`), and `FormValue` parses a query string and a form-encoded body — never a JSON one, and the request is `application/json`. `cbgt.IndexDef` has no `prevIndexUUID` member either (`cbgt/defs.go:36-50`). The field is dropped, `prevIndexUUID` falls back to `indexDef.UUID` — the body's `uuid` member — and an upsert that supplied a uuid reached FTS looking exactly like one that had not.

That is why the classic path sends `uuid` and never `prevIndexUUID` (`search_index_upsert.cxx:53-54`): the body's `uuid` **is** the update channel.

The suggested rule is adopted as written. `UpdateIndexRequest.Index.uuid` becomes `cbsearchx.Index.UUID` becomes the body's `uuid` becomes FTS's `prevIndexUUID`, which is matched against the current definition (`cbgt/manager_api.go:278-288`) — so the concurrency check is the one that now runs. gocb draws the same line in the same place (`searchindexprovider_ps.go:148`), and its create path sets no `prev_index_uuid` at all. Neither does this one now: a definition carrying a uuid is an update, and an update is not a `CreateIndexRequest`.

Two live cases, because the fix has two ways to be wrong. `an_existing_index_is_updated_through_upsert_against_live_gateway` is the cycle described — create, `get_index`, change `plan_params`, upsert what came back — and fails with `index_exists` against the previous encoding. `an_upsert_without_a_uuid_for_an_existing_name_is_refused_against_live_gateway` pins the other direction: routing every upsert through `UpdateIndex` would turn a create of an existing name into a silent overwrite, and it must stay `index_exists`. A component case asserts which RPC each one actually reaches, against a server that records the method name.

core/protostellar/search_index_admin_converter.hxx
avsej: Confirmed and fixed, in both directions. `fill_params` returns false for a non-empty blob that does not parse or does not parse to an object; `params_to_json` returns nothing when a map member is not JSON; `apply_index`/`decode_index` propagate that. An encode failure is `invalid_argument` before anything is sent, a decode failure is `parsing_failure`. `get_all` fails the whole listing rather than returning it one index short — a list missing an entry reads as "that index does not exist", which is the same defect wearing different clothes. gocb fails closed both ways too (`serializeBytesMap`/`deserializeBytesMap`, `searchindexprovider_ps.go:539-566`).

An empty blob is still the unset convention and still contributes nothing; only content that cannot be represented is refused.

The crossing is what the test payloads are chosen for: the three blobs hold three different values, so a converter that swapped two still round-trips each against itself and fails the comparison. The negative cases cover an unparseable blob and a non-object one (array, string, number, null) in each of the three positions, on both the create and update encoders. Against the server the same property is a different claim — FTS augments a definition with its own defaults, so nothing round-trips byte-for-byte and only each blob's own probe member can be looked for, in its own blob and in the two it must not appear in.

On the four smaller items:

`status` is populated through one shared helper. FTS answers these endpoints with `{"status":"ok"}` and the classic path copies that string; `admin.search.v1` replies with empty messages, so a successful RPC is the same signal. A failed call leaves the field empty rather than reporting a status the server never sent. `AnalyzeDocumentResponse` is the one message carrying FTS's own status, and that is preferred over the inferred one. One divergence stays: the classic path also fills `status` from a 400 body, so it can report `"fail"` next to an error code; nothing equivalent crosses this schema.

A `GetIndexResponse` with no index is `index_not_found`.

The `AnalyzeDocument` item does not hold. The completion assigns `response.analysis = resp.analyzed()` and parses nothing — `analysis` is a `std::string` and `analyzed` is already the raw JSON of FTS's `analyzed` member, which is what the classic path re-serialises into the same field. The same line is at `component.cxx:1560` in the reviewed commit `d49e1cf`, so nothing changed since. There is no parse to guard here. What the item sits next to is real and is fixed: `resp.status()` was being dropped.

The empty-name check is added for the seven operations the classic path guards, and deliberately not for `get_documents_count`, whose classic `encode_to` has no such check either — adding one there would be a new asymmetry pointing the other way. The stated consequence is not what happens on this gateway, though: it answers an empty name with `InvalidArgument` itself (`errorhandler.go:468`), so the reported code was already `invalid_argument`. What the guard changes is that nothing is sent, and that the code no longer depends on the gateway agreeing.

That distinction is why there is a new component suite. The empty-name case was first written against the live gateway, and mutation testing showed it had no teeth — removing the guard from `upsert`, and again from `control_ingest`, left every live case passing. The `index_not_found` mutation could not be evaluated at all, since a real gateway does not return a `GetIndexResponse` without an index. Those properties moved to an in-process server that records what it was asked, where the assertion is the call list rather than the error code. It also covers the six control operations, which the live suite cannot: a gateway accepts a resume of an index that was never paused and this transport cannot read the state back, so a bool wired to the wrong branch is invisible end to end — but not to a recorded method name. The live case kept its assertion and lost its claim, and is now named for what it pins.

avsej:
Both findings are fixed. Three of the four notes are fixed; the third does not hold against the
code, and what it points at is fixed for a different reason. Details below, in the review's
numbering.

## 1. Upsert always calls `CreateIndex` (`core/protostellar/component.cxx:1382`)

> `search_index_upsert_request` maps unconditionally to `CreateIndexRequest`. The search admin service has a dedicated `UpdateIndex` RPC (`admin/search/v1/search.proto:30`), and it is never called.

Confirmed and fixed. The root cause is one layer further down than the review states, and it makes
the case stronger rather than weaker: `prev_index_uuid` is not merely the wrong channel for the
uuid, it is a channel that ends nowhere.

The gateway does carry it — `CreateIndex` copies `in.PrevIndexUuid` into
`cbsearchx.Index.PrevIndexUUID` (`searchadminserver.go:60`), and `encodeIndex` writes it into the PUT body as
`prevIndexUUID` (`cbsearchx/search.go:672`). FTS then reads that name only through
`req.FormValue("prevIndexUUID")` (`cbgt/rest/rest_create_index.go:288`), and `FormValue` parses a
query string and a form-encoded body — never a JSON one, and the request is `application/json`.
`cbgt.IndexDef` has no `prevIndexUUID` member either (`cbgt/defs.go:36-50`). So the field is
dropped, `prevIndexUUID` falls back to `indexDef.UUID` — the body's `uuid` member — and an upsert
that supplied a uuid arrived at FTS looking exactly like one that had not.

That is also why the classic path sends `uuid` and never `prevIndexUUID`
(`search_index_upsert.cxx:53-54`): the body's `uuid` **is** the update channel.

> Select the RPC on whether a UUID was supplied — which is how the classic path distinguishes create from update — and to carry `Index.uuid` through so the gateway can detect a conflicting concurrent update

Adopted as written. `UpdateIndexRequest.Index.uuid` becomes `cbsearchx.Index.UUID` becomes the
body's `uuid` becomes FTS's `prevIndexUUID`, which is checked against the current definition
(`cbgt/manager_api.go:278-288`), so the concurrency check the review asks for is the one that now
runs. gocb draws the same line in the same place — `searchindexprovider_ps.go:148` branches on
`indexDefinition.UUID == ""` and its create path sets no `prev_index_uuid` at all.

`prev_index_uuid` is therefore never set on this path now, which is a deliberate consequence: a
definition carrying a uuid is an update, and an update is not a `CreateIndexRequest`.

> No test covers the update path

`an_existing_index_is_updated_through_upsert_against_live_gateway` is the cycle the review
describes — create, `get_index`, change `plan_params`, upsert what came back — and asserts the
change reached the server and the uuid moved on. It fails with `index_exists` against the previous
encoding.

The mirror case matters as much: routing *every* upsert through `UpdateIndex` would turn a create
of an existing name into a silent overwrite.
`an_upsert_without_a_uuid_for_an_existing_name_is_refused_against_live_gateway` pins that it stays
`index_exists`.

## 2. Unparseable JSON parameters are silently discarded (`search_index_admin_converter.hxx:21`)

> When the parse fails, or when the parsed value is not an object, the members are dropped and the operation proceeds — reporting success.

Confirmed and fixed, in both directions. `fill_params` now returns false for a non-empty blob that
does not parse or does not parse to an object, `params_to_json` returns nothing when a map member
is not JSON, and `apply_index`/`decode_index` propagate that. An encode failure is reported as
`invalid_argument` before anything is sent; a decode failure as `parsing_failure`.

> The same helper is used in both directions, so the loss can happen on the read as well as the write.

Which is the half that a request-side-only fix would leave open, so the read side fails closed too.
`get_all` fails the whole listing rather than returning it one index short — a list missing an entry
reads as "that index does not exist", which is the same defect wearing different clothes. gocb does
the same in both directions (`serializeBytesMap`/`deserializeBytesMap`,
`searchindexprovider_ps.go:539-566` — both `return nil, err`).

An empty blob is still the unset convention and still contributes nothing; only a blob with content
that cannot be represented is refused.

> The most valuable missing test is a full round trip that asserts equality of all three parameter blobs, plus a negative case with a non-object payload asserting that it is rejected rather than dropped. The current converter test asserts `params` only and would survive crossing `plan_params` with `source_params`.

Both, and the crossing is what the payloads are chosen for: the three blobs hold three different
values, so a converter that swapped two of them still round-trips each against itself and fails the
comparison. `all_three_parameter_blobs_survive_the_round_trip` and
`a_definition_survives_the_full_encode_decode_cycle` cover the proto; the negative cases cover an
unparseable blob and a non-object one (array, string, number, null) in each of the three positions,
on both the create and the update encoders.

The same property is checked against the server, where it is a different claim: FTS augments a
definition with its own defaults, so nothing round-trips byte-for-byte and only each blob's own
probe member can be looked for — in its own blob, and in the two it must not appear in.
`the_three_parameter_blobs_stay_separate_against_live_gateway` does that.

## Notes

> None of the twelve responses populate `status`, which the classic path always sets and which core-API consumers read. Worth a single shared helper rather than twelve additions.

Done, as one helper. FTS answers these endpoints with `{"status":"ok"}` and the classic path copies
that string; `admin.search.v1` replies with empty messages, so a successful RPC is the same signal
and `"ok"` is what it means. A failed call leaves the field empty rather than reporting a status the
server never sent — the error code is what describes it. `AnalyzeDocumentResponse` is the one
message that carries FTS's own status, and that is preferred over the inferred one.

One divergence stays: the classic path also fills `status` from a 400 body, so it can report
`"fail"` alongside an error code. Nothing equivalent crosses `admin.search.v1`.

> `GetIndexResponse` without an `index` field yields success plus a default-constructed index. Treating a missing index as `index_not_found` is more useful than an empty definition.

Done.

> The `AnalyzeDocument` response body is parsed without a `try`/`catch`. A non-array payload from the gateway therefore throws out of an Asio completion handler, which terminates rather than failing the operation.

This one does not hold. The completion assigns `response.analysis = resp.analyzed()` and parses
nothing — `analysis` is a `std::string` and `AnalyzeDocumentResponse.analyzed` is already the raw
JSON of FTS's `analyzed` member, which is what the classic path re-serialises into the same field.
The same line is at `component.cxx:1560` in the reviewed commit `d49e1cf`, so this is not something
that changed since. There is no parse on this path to guard, and no throw to escape a handler.

What the note is next to is real and is fixed: `resp.status()` was dropped, which is the `status`
note applied to the one response that carries a server-sent one.

> The empty-index-name validation that the classic operations perform locally is skipped on this path, so an empty name becomes a gateway round trip returning a KV-shaped error code rather than `invalid_argument`.

Done for the seven operations the classic path guards, and deliberately not for
`get_documents_count`, whose classic `encode_to` has no such check either
(`search_index_get_documents_count.cxx`) — adding one there would be a new asymmetry pointing the
other way.

The stated consequence is not what happens on this gateway, though. It answers an empty name with
`InvalidArgument` itself (`errorhandler.go:468`), so the reported code was already
`invalid_argument` and the guard does not change it here. What the guard does change is that
nothing is sent, and that the code no longer depends on the gateway agreeing. That distinction is
what the new component suite exists to assert; see below.

## The tests, and a round of them that had no teeth

Three suites now: 11 converter cases, 9 component cases against an in-process `admin.search.v1`
server, 11 live cases. Each has a mutation spec, and the middle suite exists because the first
sweep found two assertions that could not fail.

`an_empty_index_name_is_refused_before_it_is_sent` and the `index_not_found` case were written
against the live gateway. Removing the empty-name guard from `upsert`, and again from
`control_ingest`, left every live case passing — because the gateway returns `invalid_argument` for
an empty name too, so the error code cannot distinguish a local refusal from a round trip. The
`index_not_found` mutation could not even be evaluated: a real gateway does not return a
`GetIndexResponse` without an index.

So those properties moved to a server that records what it was asked. The component suite asserts
the *call list*: an empty name leaves it empty, an unrepresentable definition leaves it empty, and
`upsert` appears in it as `CreateIndex` or `UpdateIndex` according to the uuid. It also covers the
six control operations, which the live suite cannot: a gateway accepts a resume of an index that
was never paused and this transport cannot read the state back, so a bool wired to the wrong branch
is invisible end to end — but not to a recorded method name.

The live case kept its assertion and lost its claim: it is now
`an_empty_index_name_is_reported_as_invalid_argument_against_live_gateway`, which is what it pins.

## Verification

Against a live CNG gateway — Server 8.1.0-1743, Cloud Native Gateway 1.2.2-123, operator 2.8.0 —
and not only in-process.

- `cng_search_index_admin_converter_test`: 11 cases passing; 8 mutations, all killed.
- `cng_search_index_admin_component_test`: 9 cases passing; 10 mutations, all killed.
- `cng_live_search_index_admin_test`: 11 cases passing, 0 skipped; 6 mutations, all killed.

The mutations cover: the pre-review encoding (uuid as `prev_index_uuid`), upsert forced down each
branch, the empty-name guard removed from two different overloads, an unrepresentable definition
sent anyway, a missing index reported as success, an undecodable one handed back empty, each of the
three blobs dropped or crossed, `status` never set and `status` set on a failure, a control bool
inverted, and an unrouted management operation refused with a different code.

One limit worth stating: nothing here asserts the *effect* of the six control operations. The state
they change is not on `Index` and no `admin.search.v1` RPC reports it, so what is checked is that
each bool reaches its own RPC and that the call succeeds.

## Also worth recording

`gocb`'s protostellar create path assigns the source uuid to the source *name*
(`searchindexprovider_ps.go:214-215`: `if index.SourceUUID != "" { req.SourceName = &index.SourceUUID }`),
overwriting the bucket name it set three lines earlier. This SDK sets `source_uuid` from
`source_uuid`; the divergence is noted only so the two are not reconciled in the wrong direction
later.

The RFC 77 tracing mapping for management `db.operation` is still outstanding across the four
management services and is carried as a followup from #1041, not re-litigated here.

avsej:
The coverage this asks for is in. Answering it point by point, since the scope was given as a definite list rather than "add tests".

> Nine operations are routed. What ships with them is four converter unit cases — the params map explode/rebuild, and index decode — plus one live case that lists indexes. So `get_all` is the only operation exercised end to end

Now three suites: 11 converter cases, 9 component cases against an in-process `admin.search.v1` server, and 11 live cases. Every routed operation is exercised end to end, and each suite has a mutation spec — 8, 10 and 6 mutations respectively, all killed.

> **The control operations are the priority.** … That is six RPC selections, and an inverted pair fails silently: the call succeeds and does the opposite of what was asked. Nothing in the current tests would tell the two apart.

Nothing in a *live* test can tell them apart either, which is why this needed a different kind of test rather than a bigger live one. A gateway accepts a resume of an index that was never paused, and no `admin.search.v1` RPC reports ingest, querying or plan-freeze state — `Index` carries name, type, uuid, the three parameter maps and the three source fields, and nothing else. So the effect is unobservable over this transport, and a live case can only assert that the call succeeded.

The in-process server records the method name instead. `the_control_operations_select_their_paired_rpc` drives all six and asserts the call list is exactly `PauseIndexIngest`, `ResumeIndexIngest`, `AllowIndexQuerying`, `DisallowIndexQuerying`, `FreezeIndexPlan`, `UnfreezeIndexPlan` in turn. Inverting one bool is a mutation in the spec, and it is killed.

That the effect cannot be read back at all is filed as a followup against the schema rather than worked around.

> **Create versus update.** `upsert` maps onto CreateIndex and carries the existing uuid as `prev_index_uuid`, so a single path serves both. Neither direction is covered

It did not serve both — that is the defect the review round found, and it is fixed. `prev_index_uuid` never reaches FTS: the gateway writes it into the PUT body, but cbgt reads that name only through `req.FormValue`, which does not parse a JSON body, and `cbgt.IndexDef` has no such member. Every upsert was a create, and updating an index was impossible over `couchbase2://`.

`upsert` now selects `CreateIndex` or `UpdateIndex` on whether the definition carries a uuid, which is the line the classic path draws and the line FTS draws. Covered in both places: the component suite asserts which RPC each takes and that the uuid reaches `Index.uuid`, and two live cases cover the two ways it can be wrong — an update that must succeed, and a create of an existing name that must stay `index_exists`. Both directions are mutations, both killed.

> **`get_stats`** returns `feature_not_available` because admin.search.v1 has no RPC for it. No case asserts that, so the deliberate refusal is indistinguishable from the operation simply not being routed — the same failure mode that makes a skip-on-error-code test unfalsifiable.

Asserted, and the point about falsifiability is what shapes the case. `get_stats_is_refused_over_couchbase2_against_live_gateway` lists indexes first — which succeeds, and so establishes that the gateway serves `admin.search.v1` — and only then asserts `feature_not_available` for `get_stats`. Without that first call the assertion could not fail, since a gateway with no search admin service at all would produce the same code. A mutation changing the refusal to a different code is killed.

> A single live lifecycle case — create, get, documents count, each control pair, drop — would cover most of this in one pass

That case exists. Alongside it: the update cycle, the create-of-an-existing-name refusal, the three parameter blobs checked against a real definition for crossing, a definition that cannot be represented being refused before it is sent, an empty index name, a missing index reporting `index_not_found`, `analyze_document`, and `get_stats`.

> That same case is what later gives the search query tests a real index to query, so it is worth building here rather than deferring.

The lifecycle case creates an index on the test bucket and drops it through an RAII guard, so it leaves nothing behind. Worth flagging for whoever picks up the search query tests: the CNG CI cluster patched only the data service quota and left search at the operator default, and `create-cng-cluster` said in as many words that this was fine because no index was being built and that index management would have to provision for one. It now raises search to 1 GiB alongside the data quota. Whether the default would in fact have sufficed was not measured — the quota was raised because the cases changed from querying a service to building on it, not because a failure was observed.

One coverage limit stated plainly rather than left implicit: the six control operations are asserted to reach their own RPC and to succeed. What they actually do to the index is not asserted anywhere, because this transport cannot read it back.

thejcfactor (approved):

PR #1043, opened 2026-07-28
CXXCBC-896: retry couchbase2 KV operations via the SDK retry strategy

avsej:
couchbase2 operations executed once and returned immediately, ignoring the transport-agnostic retry machinery every request already carries. Feed the KV path into it.

error_utils gains retry_reason_for(error_code): a transport UNAVAILABLE (mapped to temporary_failure) becomes service_not_available; everything else -- timeouts, cas mismatch, auth, not-found -- is terminal. cluster_impl wraps the KV protostellar dispatch in execute_protostellar: on a retryable error it consults request.retries' strategy (falling back to the origin default) and, if it says retry, re-issues after the backoff. Query/analytics/search/view/management carry no retry_context and continue to dispatch directly.

Every operation is bounded by a deadline resolved before the first attempt -- the request's own timeout, or the cluster's key_value_timeout when it carries none -- and each attempt is issued against what is left of that deadline rather than a fresh full-length one. The fallback is what bounds the loop at all: the public API leaves the per-operation timeout unset unless the caller asks for one, and the strategy never refuses a retry for service_not_available. Idempotency is enforced by the strategy, so ambiguous mutations are not replayed. A backoff that would land past the deadline is capped to it rather than refusing the retry, so the operation spends its whole budget and ends on the deadline as a timeout instead of on the last attempt's transport error. The attempt scheduled at the deadline receives a non-positive remainder, which the component answers as a timeout carrying the accumulated attempts and reasons. That timeout is the ambiguous one for a non-idempotent operation already dispatched at least once, and the unambiguous one otherwise -- the rule mcbp_command::cancel() applies on the classic path.

The request is held by shared_ptr and re-issued from it, and component::execute takes it by const reference, so an operation no longer copies its whole request -- for a mutation, the document body -- before every dispatch.

make_error_context takes the retry attempt count and reason set rather than pinning them at zero, sourced from the request the loop carries. A retried KV operation reports both in its error context, as RFC 77 requires and as the classic transport already does; the other services still report zero over couchbase2 (CXXCBC-921).

--- Stacked PR **21/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1042 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -20,6 +20,11 @@
 #ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
 #include "core/protostellar/component.hxx"
 #include "core/protostellar/credentials.hxx"
+#include "core/protostellar/error_utils.hxx"
+
+#include <couchbase/retry_strategy.hxx>
+
+#include <asio/steady_timer.hpp>
 #endif
 
 #define COUCHBASE_CXX_CLIENT_IGNORE_CORE_DEPRECATIONS
@@ -304,6 +309,23 @@ is_feature_supported(const operations::management::search_index_upsert_request& 
 {
   return !request.index.is_vector_index() || capabilities.supports_vector_search();
 }
+
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+// Extract the error_code from a response's error context. KV responses carry a
+// key_value_error_context (ec() accessor); the query/analytics/search/view/management responses
+// carry a plain error_context::* struct with an `ec` field.
+inline auto
+protostellar_response_ec(const key_value_error_context& ctx) -> std::error_code
+{
+  return ctx.ec();
+}
+template<typename Context>
+inline auto
+protostellar_response_ec(const Context& ctx) -> std::error_code
+{
+  return ctx.ec;
+}
+#endif
 } // namespace
 
 #ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
@@ -754,7 +776,7 @@ public:
 #ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
     if (auto component = protostellar_component(); component) {
       if constexpr (protostellar::component_supports_v<Request>) {
-        component->execute(std::move(request), std::forward<Handler>(handler));
+        execute_protostellar(std::move(request), std::forward<Handler>(handler));
         return;
       } else {
         return handler(request.make_response(
@@ -797,7 +819,9 @@ public:
 #ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
     if (auto component = protostellar_component(); component) {
       if constexpr (component_routes<Request>::value) {
-        component->execute(std::move(request), std::forward<Handler>(handler));
+        // Query/analytics/search/view/management requests do not carry a retry_context, so they
+        // dispatch directly; transport-level retry applies to the KV path (which does).
+        component->execute(request, std::forward<Handler>(handler));
         return;
       } else {
         // Remaining management ops over couchbase2 are not wired yet. Reject cleanly rather than
@@ -1007,6 +1031,116 @@ public:
     return false;
 #endif
   }
+
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+  // Execute a couchbase2 operation through the component with transport-level retry. Retryable
+  // errors (see protostellar::retry_reason_for) consult the request's retry strategy (falling back
+  // to the origin default), and on a positive decision the request is re-issued after a backoff.
+  // Idempotency is enforced by the strategy, so ambiguous mutations are not replayed.
+  //
+  // The deadline is resolved here, before the first attempt, and every attempt is issued against
+  // it. Falling back to the cluster's KV default when the request carries no timeout is what bounds
+  // the loop at all: the public API leaves the per-operation timeout unset unless the caller asks
+  // for one, and the strategy never refuses a retry for service_not_available, so a request issued
+  // with default options against an unavailable gateway would otherwise retry until the process
+  // ended -- never completing, and never reporting an error either.
+  template<class Request, class Handler>
+  void execute_protostellar(Request request, Handler&& handler)
+  {
+    // The fallback below is the KV default, so routing anything else here would bound it with a
+    // budget belonging to another service. Non-KV requests carry no retry_context and have their
+    // own path.
+    static_assert(protostellar::component_supports_v<Request>,
+                  "execute_protostellar resolves the key-value timeout");
+    const auto deadline = std::chrono::steady_clock::now() +
+                          request.timeout.value_or(origin_.options().key_value_timeout);
+    // Held by shared_ptr and re-issued from there. The request owns the document body for a
+    // mutation, so copying it to keep a re-issue possible charged every operation on the data plane
+    // for a retry that in normal operation never happens.
+    dispatch_protostellar(
+      std::make_shared<Request>(std::move(request)), std::forward<Handler>(handler), deadline);
+  }
+
+  // One attempt of the above, re-entered from the backoff timer for each retry.
+  template<class Request, class Handler>
+  void dispatch_protostellar(std::shared_ptr<Request> request,
+                             Handler&& handler,
+                             std::chrono::steady_clock::time_point deadline)
+  {
+    // Each attempt gets what is LEFT of the operation's budget rather than a fresh full-length one.
+    // Without this an attempt started just inside the deadline would run for the whole timeout
+    // again, so a retried operation could take roughly twice its nominal budget. Narrowing keeps
+    // the request carrying a duration -- which is all the component accepts -- while making that
+    // duration mean "until the operation's deadline". A non-positive remainder is not
+    // special-cased, and is in fact how the loop ends: the component rejects an exhausted budget
+    // as a timeout instead of dispatching a call with no deadline.
+    request->timeout = std::chrono::duration_cast<std::chrono::milliseconds>(
+      deadline - std::chrono::steady_clock::now());
+    // Re-read through the locked accessor on every attempt: a retry re-enters this
+    // function from a timer, long after the caller checked.
+    auto component = protostellar_component();
+    if (!component) {
+      return handler(
+        request->make_response(make_key_value_error_context(errc::network::cluster_closed,
+                                                            request->id,
+                                                            request->retries.retry_attempts(),
+                                                            request->retries.retry_reasons()),
+                               typename Request::encoded_response_type{}));
+    }
+    component->execute(
+      *request,
+      [self = shared_from_this(), request, handler = std::forward<Handler>(handler), deadline](
+        auto response) mutable {
+        if (!self->stopped_) {
+          if (auto reason = protostellar::retry_reason_for(protostellar_response_ec(response.ctx));
+              reason.has_value()) {
+            auto strategy = request->retries.strategy()
+                              ? request->retries.strategy()
+                              : self->origin_.options().default_retry_strategy_;
+            if (strategy) {
+              const auto action = strategy->retry_after(request->retries, reason.value());
+              // The wait is capped at the deadline rather than the retry being refused for landing
+              // past it. Refusing would end the operation early on the last attempt's transport
+              // error -- temporary_failure for an unreachable gateway -- while the caller's budget
+              // was still running, so the same condition that times out over the classic transport
+              // would report a different code here. Capping lets the operation use its whole budget
+              // and end on the deadline: the attempt scheduled at it gets a non-positive remainder
+              // above, which the component answers as a timeout carrying the accumulated attempts
+              // and reasons. That is terminal, so the loop cannot spin on it -- retry_reason_for
+              // does not classify a timeout as retryable. Both other implementations of this loop
+              // cap the same way (io/retry_orchestrator.hxx cap_duration, and Java's
+              // RetryOrchestratorProtostellar.capDuration).
+              const auto when =
+                std::min(std::chrono::steady_clock::now() + action.duration(), deadline);
+              if (action.need_to_retry()) {
+                // Recorded before the re-issue, so the next attempt carries the accumulated count
+                // and the error context it builds reports the retries that preceded it.
+                request->retries.record_retry_attempt(reason.value());
+                auto timer = std::make_shared<asio::steady_timer>(self->ctx_);
+                timer->expires_at(when);
+                timer->async_wait([self,
+                                   request,
+                                   handler = std::move(handler),
+                                   deadline,
+                                   response = std::move(response),
+                                   timer](const std::error_code& timer_ec) mutable {
+                  if (timer_ec) {
+                    // timer cancelled (e.g. cluster closed): deliver the last retryable response
+                    // rather than dropping the caller's completion, which would hang it.
+                    handler(std::move(response));
+                    return;
+                  }
+                  self->dispatch_protostellar(std::move(request), std::move(handler), deadline);
+                });
+                return;
+              }
+            }
+          }
+        }
+        handler(std::move(response));
+      });
+  }
+#endif
 
   void do_open(utils::movable_function<void(std::error_code)> handler)
   {
--- a/core/protostellar/error_utils.cxx
+++ b/core/protostellar/error_utils.cxx
@@ -241,8 +241,28 @@ error_message(const grpc::Status& status) -> std::string
 }
 
 auto
-make_error_context(const grpc::Status& status, const document_id& id, operation_kind kind)
-  -> key_value_error_context
+retry_reason_for(std::error_code ec) -> std::optional<couchbase::retry_reason>
+{
+  // UNAVAILABLE maps to temporary_failure (error_utils map_status_code): the call did not reach a
+  // healthy server, so it is safe to retry (even for mutations — the strategy's
+  // allows_non_idempotent_retry(service_not_available) permits it). Everything else is terminal at
+  // the transport: timeouts consumed the budget, cas_mismatch/auth/not_found are semantic.
+  if (ec == errc::common::temporary_failure) {
+    return couchbase::retry_reason::service_not_available;
+  }
+  // A locked document (FAILED_PRECONDITION "LOCKED") is retried with backoff per RFC 77.
+  if (ec == errc::key_value::document_locked) {
+    return couchbase::retry_reason::key_value_locked;
+  }
+  return std::nullopt;
+}
+
+auto
+make_error_context(const grpc::Status& status,
+                   const document_id& id,
+                   operation_kind kind,
+                   std::size_t retry_attempts,
+                   std::set<couchbase::retry_reason> retry_reasons) -> key_value_error_context
 {
   const auto ec = map_status(status, kind);
   std::optional<key_value_extended_error_info> extended{};
@@ -256,8 +276,8 @@ make_error_context(const grpc::Status& status, const document_id& id, operation_
     ec,
     {},
     {},
-    0,
-    {},
+    retry_attempts,
+    std::move(retry_reasons),
     id.key(),
     id.bucket(),
     id.scope(),
--- a/test/cng/protostellar/kv_retry.cxx
+++ b/test/cng/protostellar/kv_retry.cxx
@@ -1,0 +1,549 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Transport-level retry for KV couchbase2 operations (CXXCBC-896), specifically the property that a
+// retried operation stays inside its overall time budget. Drives a real core::cluster against a
+// localhost KvService so the whole path -- routing, retry decision, backoff, re-dispatch and the
+// per-attempt deadline -- is exercised together; the budget is a property of their interaction and
+// is not observable from any one of them. Env-agnostic: local server, no external gateway.
+
+#define COUCHBASE_CXX_CLIENT_IGNORE_CORE_DEPRECATIONS
+
+#include "framework/test_runner.hxx"
+
+#include "core/operations.hxx"
+
+#include "core/cluster.hxx"
+
+#include "core/cluster_credentials.hxx"
+#include "core/document_id.hxx"
+#include "core/origin.hxx"
+#include "core/utils/connection_string.hxx"
+
+#include <couchbase/best_effort_retry_strategy.hxx>
+#include <couchbase/error_codes.hxx>
+#include <couchbase/retry_reason.hxx>
+
+#include <couchbase/kv/v1/kv.grpc.pb.h>
+
+#include <asio/executor_work_guard.hpp>
+#include <asio/io_context.hpp>
+
+#include <grpcpp/security/server_credentials.h>
+#include <grpcpp/server_builder.h>
+
+#include <atomic>
+#include <chrono>
+#include <cstddef>
+#include <future>
+#include <string>
+#include <thread>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+namespace ops = ::couchbase::core::operations;
+namespace v1 = ::couchbase::kv::v1;
+using ::couchbase::core::cluster_credentials;
+using ::couchbase::core::document_id;
+using ::couchbase::core::origin;
+using ::couchbase::core::utils::parse_connection_string;
+using namespace std::chrono_literals;
+
+// Answers UNAVAILABLE (retryable: service_not_available) immediately for the first two dispatches,
+// then stops answering. The pause is what exposes the budget: by the time the third attempt is
+// issued most of the operation's time is already spent on backoff, so the deadline that attempt
+// receives is the whole question.
+class stalling_kv_service final : public v1::KvService::Service
+{
+public:
+  std::atomic<int> calls{ 0 };
+
+  auto Get(grpc::ServerContext* context,
+           const v1::GetRequest* /* request */,
+           v1::GetResponse* /* response */) -> grpc::Status override
+  {
+    if (++calls <= 2) {
+      return { grpc::StatusCode::UNAVAILABLE, "try again" };
+    }
+    // Outlasts any budget this test grants, so the client's deadline is what ends the call -- but
+    // polled rather than slept through, so the handler returns once the client gives up instead of
+    // running on into the suite's teardown and holding the server past Shutdown().
+    const auto give_up_by = std::chrono::steady_clock::now() + 5s;
+    while (!context->IsCancelled() && std::chrono::steady_clock::now() < give_up_by) {
+      std::this_thread::sleep_for(5ms);
+    }
+    return { grpc::StatusCode::UNAVAILABLE, "still unavailable" };
+  }
+};
+
+// Never answers anything but UNAVAILABLE, and answers it immediately. The retry strategy always
+// permits a retry for service_not_available, so nothing here ever ends the loop from below: what
+// ends it is the operation's own deadline.
+class unavailable_kv_service final : public v1::KvService::Service
+{
+public:
+  std::atomic<int> calls{ 0 };
+
+  auto Get(grpc::ServerContext* /* context */,
+           const v1::GetRequest* /* request */,
+           v1::GetResponse* /* response */) -> grpc::Status override
+  {
+    ++calls;
+    return { grpc::StatusCode::UNAVAILABLE, "never available" };
+  }
+
+  // An unimplemented RPC answers UNIMPLEMENTED, which is terminal rather than retryable, so a
+  // mutation against a Get-only service would end on its first attempt without ever entering the
+  // loop this file exists to test -- and the dispatch counter, incremented only in Get, would sit
+  // at zero.
+  auto Upsert(grpc::ServerContext* /* context */,
+              const v1::UpsertRequest* /* request */,
+              v1::UpsertResponse* /* response */) -> grpc::Status override
+  {
+    ++calls;
+    return { grpc::StatusCode::UNAVAILABLE, "never available" };
+  }
+};
+
+// Fails a fixed number of dispatches with UNAVAILABLE and then serves the document, so the number
+// of retries the operation performs is decided by the service rather than by timing.
+class flaky_kv_service final : public v1::KvService::Service
+{
+public:
+  explicit flaky_kv_service(int failures)
+    : failures_{ failures }
+  {
+  }
+
+  std::atomic<int> calls{ 0 };
+
+  auto Get(grpc::ServerContext* /* context */,
+           const v1::GetRequest* /* request */,
+           v1::GetResponse* response) -> grpc::Status override
+  {
+    if (++calls <= failures_) {
+      return { grpc::StatusCode::UNAVAILABLE, "not yet" };
+    }
+    response->set_content_uncompressed("{}");
+    response->set_cas(42);
+    return grpc::Status::OK;
+  }
+
+private:
+  const int failures_;
+};
+
+// A retried operation must finish at its deadline, not at "deadline + one more full timeout".
+//
+// The budget is 400ms and the backoff is a flat 150ms, so the third attempt starts around 300ms in
+// with roughly 100ms of budget left, against a server that has stopped answering. Given the
+// remaining budget the operation ends at ~400ms; given a fresh full-length one it ends at ~700ms.
+// The assertion is placed between those two outcomes, so it fails if an attempt is ever handed more
+// time than the operation has left.
+void
+a_retried_kv_operation_stays_within_its_budget()
+{
+  int port = 0;
+  stalling_kv_service service;
+  grpc::ServerBuilder builder;
+  builder.RegisterService(&service);
+  builder.AddListeningPort("127.0.0.1:0", grpc::InsecureServerCredentials(), &port);
+  auto server = builder.BuildAndStart();
+
+  auto parsed = parse_connection_string("couchbase2://127.0.0.1:" + std::to_string(port));
+  origin cluster_origin{ cluster_credentials{}, parsed }; // no credentials -> plaintext allowed
+  cluster_origin.options().enable_tls = false;
+  cluster_origin.options().default_retry_strategy_ =
+    couchbase::make_best_effort_retry_strategy([](std::size_t) {
+      return 150ms;
+    });
+
+  asio::io_context io;
+  auto guard = asio::make_work_guard(io);
+  std::thread io_thread([&io]() {
+    io.run();
+  });
+  couchbase::core::cluster cluster{ io };
+
+  std::promise<std::error_code> opened;
+  cluster.open(cluster_origin, [&opened](std::error_code ec) {
+    opened.set_value(ec);
+  });
+  assert_false(static_cast<bool>(opened.get_future().get()), "open(couchbase2://) succeeds");
+
+  ops::get_request request;
+  request.id = document_id{ "b", "s", "c", "k" };
+  request.timeout = 400ms;
+
+  const auto started = std::chrono::steady_clock::now();
+  std::promise<ops::get_response> done;
+  cluster.execute(std::move(request), [&done](ops::get_response response) {
+    done.set_value(std::move(response));
+  });
+  const auto result = done.get_future().get();
+  const auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
+    std::chrono::steady_clock::now() - started);
+
+  // Tear down before asserting. A failed assertion throws, and unwinding past a still-joinable
+  // std::thread calls std::terminate -- which would replace the assertion's message with a bare
+  // "terminate called", hiding exactly the failure this case exists to report.
+  std::promise<void> closed;
+  cluster.close([&closed]() {
+    closed.set_value();
+  });
+  closed.get_future().get();
+  guard.reset();
+  io_thread.join();
+  server->Shutdown(std::chrono::system_clock::now());
+
+  // The call count is reported, not just tested: the two ways this can fail -- the first attempt
+  // never reaching the service, or a retry not being issued -- are indistinguishable from a bare
+  // "expected >= 2", and this case previously failed about one run in seven on a loaded machine
+  // with no way to tell which had happened.
+  assert_true(service.calls.load() >= 2,
+              "the operation was retried at least once (dispatches reaching the service: " +
+                std::to_string(service.calls.load()) + ")");
+  // A timeout, not the last attempt's transport error: the operation is entitled to its whole
+  // budget, so it ends on its deadline rather than on whichever UNAVAILABLE arrived last. The read
+  // is idempotent, so the timeout is the unambiguous one.
+  assert_true(result.ctx.ec() == couchbase::errc::common::unambiguous_timeout,
+              "the operation ends on its deadline, reported as a timeout, got: " +
+                result.ctx.ec().message());
+  assert_true(elapsed < 600ms,
+              "a retried operation finishes at its deadline, not a full timeout past it");
+}
+
+// A request issued through the public API carries no timeout unless the caller asked for one, and
+// the strategy never refuses a retry for service_not_available. The bound therefore has to come
+// from the cluster's own key_value_timeout; without that fallback the loop has nothing to stop it
+// and the caller's handler is never invoked -- not a timeout, not an error, no completion at all.
+//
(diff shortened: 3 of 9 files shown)


cmake/Protostellar.cmake
```diff
@@ -0,0 +1,181 @@
...
+  get_filename_component(_rel_dir "${rel}" DIRECTORY)
+  get_filename_component(_rel_we "${rel}" NAME_WE)
+  file(MAKE_DIRECTORY "${_ps_out}/${_rel_dir}")
+  set(_cc "${_ps_out}/${_rel_dir}/${_rel_we}.pb.cc")
+  set(_h "${_ps_out}/${_rel_dir}/${_rel_we}.pb.h")
+  add_custom_command(
+    OUTPUT "${_cc}" "${_h}"
+    COMMAND protobuf::protoc
+    ARGS --cpp_out "${_ps_out}" ${_ps_import_flags} "${base}/${rel}"
+         --experimental_allow_proto3_optional
```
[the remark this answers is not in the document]
avsej: Addressed: dropped the legacy `ARGS` keyword from both add_custom_command(OUTPUT) blocks; the protoc flags now follow COMMAND directly. (For the record, add_custom_command does accept `ARGS` as a no-op legacy separator -- which is why configuration succeeded before -- but the keyword-less form is the modern spelling.)

core/protostellar/search_converter.hxx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Translate the opaque core query JSON into the structured Query oneof, for the trivial forms only.
+// Returns false (leaving proto.query unset) when the query shape is not one we map yet.
+[[nodiscard]] inline auto
+apply_query(const operations::search_request& request, v1::SearchQueryRequest& proto) -> bool
+{
+  tao::json::value parsed;
+  try {
+    parsed = utils::json::parse(request.query);
```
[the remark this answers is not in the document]
avsej: Not a defect: core/utils/json.hxx declares a dedicated `parse(const json_string&)` overload, so `utils::json::parse(request.query)` binds to it directly with no std::string conversion. It compiles and links cleanly -- the search_converter unit test builds and passes.

core/cluster.cxx
```diff
@@ -825,6 +913,144 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+  // carries a timeout, by an overall deadline stamped at the first attempt so they never outlive
+  // the operation budget. Idempotency is enforced by the strategy, so ambiguous mutations are not
+  // replayed.
+  template<class Request, class Handler>
+  void execute_protostellar(Request request,
+                            Handler&& handler,
+                            std::optional<std::chrono::steady_clock::time_point> deadline = {})
+  {
+    if (!deadline.has_value() && request.timeout.has_value()) {
+      deadline = std::chrono::steady_clock::now() + request.timeout.value();
```
avsej: **Blocker: the retry loop is unbounded on the default path.**

The deadline is stamped only when `request.timeout` is engaged, and the public API does not default it — `core/impl/collection.cxx:845` passes `options.timeout` straight through, so it is `std::nullopt` unless the caller sets a per-operation timeout. The component supplies a default *per attempt*, but nothing bounds the loop.

The strategy does not bound it either: `best_effort_retry_strategy::retry_after` (`core/impl/best_effort_retry_strategy.cxx:97-100`) returns a backoff whenever `request.idempotent() || allows_non_idempotent_retry(reason)`, and `service_not_available` permits non-idempotent retry — so it never returns `do_not_retry` for this reason, and there is no attempt cap.

Scenario: `collection.upsert("k", doc)` with default options against a gateway that is down. Every attempt fails fast with `UNAVAILABLE` → `temporary_failure` → `service_not_available`; the deadline check is skipped; the strategy always says retry. The handler is never invoked — no timeout, no error, no completion.

```suggestion
    if (!deadline.has_value()) {
      // Never leave the loop unbounded: fall back to the same per-service default the component
      // applies per attempt, so a request without an explicit timeout still has a budget.
      deadline = std::chrono::steady_clock::now() +
                 request.timeout.value_or(default_kv_timeout_for(origin_.options()));
    }
```

The `!deadline.has_value() ||` escape in the retry condition below should then be dropped, so a missing deadline can never mean "retry forever".

avsej (commented):
## Review — PR #1043 (`CXXCBC-896`)

This commit adds transport-level retry for couchbase2 KV operations by wrapping `component::execute` in
`cluster_impl::execute_protostellar`, consulting the request's retry strategy on a retryable error and
re-issuing after a backoff. The integration with the SDK's existing `retry_strategy` and `retry_reason`
machinery is the right approach, and the timer-cancelled path correctly delivers the last response rather than
dropping the caller's completion — a failure mode that is easy to introduce here and easy to miss.

Three defects in the loop mean it does not currently bound or report retries correctly, and one of them prevents
the operation from ever completing on the default configuration.

### Strengths

- Reusing `retry_strategy::retry_after` and `retry_reason` rather than inventing a parallel backoff keeps
  couchbase2 consistent with the classic transport and honours a user-supplied strategy.
- The timer-cancelled branch delivers the last retryable response instead of returning silently, so closing the
  cluster during a backoff does not hang the caller. That is the correct choice and it is commented.
- `retry_reason_for` is a narrow, testable classifier rather than logic embedded in the loop, and its mapping of
  a locked document to `key_value_locked` matches RFC 77 and both couchbase-jvm-clients and gocb — .NET gets
  this wrong (CNG-5).

### Findings

|---|---|---|---|---|
| 1 | blocker | correctness | The retry loop is unbounded in wall-clock time whenever `request.timeout` is unset, which is the default | `core/cluster.cxx:1008` |
| 2 | high | correctness | The backoff timer re-enters `execute_protostellar` without re-checking `protostellar_` or `stopped_` | `core/cluster.cxx:1036` |
| 3 | high | correctness | Each retried attempt receives a fresh full-length deadline rather than the remaining budget | `core/cluster.cxx:1030` |
| 4 | medium | observability | The error context always reports zero retry attempts and no retry reasons | `core/protostellar/error_utils.cxx:228` |
| 5 | medium | efficiency | Every operation deep-copies its whole request, including the document body, on a path that almost never retries | `core/cluster.cxx:1010` |
| 6 | medium | test-coverage | The retry loop itself has no test; only the classification predicate does | `core/cluster.cxx:1010` |

#### 1. The retry loop is unbounded when `request.timeout` is unset (blocker)

```cpp
deadline = std::chrono::steady_clock::now() + request.timeout.value();
```

The overall deadline is stamped only when `request.timeout.has_value()`. The public API does not default it:
`core/impl/collection.cxx:845` passes `options.timeout` — an engaged-or-not `std::optional` — straight through,
so `request.timeout` is `std::nullopt` unless the caller sets a per-operation timeout. The component supplies a
default *per attempt* via `resolve_kv_timeout`, but nothing bounds the loop.

The strategy does not bound it either. `best_effort_retry_strategy::retry_after`
(`core/impl/best_effort_retry_strategy.cxx:97-100`) returns a backoff whenever
`request.idempotent() || allows_non_idempotent_retry(reason)`, and `service_not_available` permits
non-idempotent retry, so it never returns `do_not_retry` for this reason. There is no attempt cap.

Failure scenario: `collection.upsert("k", doc)` with default options against a gateway that is down. Each
attempt fails fast with `UNAVAILABLE`, which maps to `temporary_failure`, which `retry_reason_for` classifies as
`service_not_available`; `deadline` is `nullopt` so the bound is skipped; the strategy always says retry. The
operation retries forever and the caller's handler is never invoked — no timeout, no error, no completion.

Resolving the effective timeout before the first attempt, and removing the "no deadline means no bound" escape,
fixes it:

```cpp
if (!deadline.has_value()) {
  // Never leave the loop unbounded: fall back to the same per-service default the component
  // applies per attempt, so a request without an explicit timeout still has a budget.
  deadline = std::chrono::steady_clock::now() +
             request.timeout.value_or(default_timeout_for<Request>(origin_.options()));
}
...
if (action.need_to_retry() && when < deadline.value()) {
```

(inline comment on `core/cluster.cxx:1008`)

#### 2. The backoff timer does not re-check `protostellar_` or `stopped_` (high)

```cpp
if (timer_ec) {
```

When the timer fires normally the handler calls `self->execute_protostellar(...)` directly, which dereferences
`protostellar_`. If `cluster::close()` ran during the backoff, that member has been reset — so this is a null
dereference. Note that the near-identical helper added for the non-KV path in #1050 does carry the guard, sixty
lines away in the same file, which makes this an oversight rather than a deliberate difference:

```cpp
if (timer_ec || !self->protostellar_ || self->stopped_) {
  handler(std::move(response));
  return;
}
```

Beyond backporting the guard, the two loops are close enough that folding them into one function would remove
the class of problem. Cancelling outstanding backoff timers from `close()` would also deliver the completion
promptly rather than at the end of a backoff the result of which is already known.

(inline comment on `core/cluster.cxx:1036`)

#### 3. Each retried attempt gets a fresh full-length deadline (high)

```cpp
timer->async_wait([self,
```

The re-dispatch passes the original request through unchanged, so `component::execute` resolves the timeout
again and `dispatcher::unary` stamps `now + timeout` on the new attempt. The overall `deadline` check only
decides *whether* to start another attempt; it does not shrink that attempt's budget. An operation whose first
attempt consumes most of its timeout can therefore start a second attempt with a full budget and run for close
to twice the nominal timeout.

Shrinking the per-attempt budget to what is left is the fix, and it composes with finding 1:

```cpp
const auto remaining = std::chrono::duration_cast<std::chrono::milliseconds>(
  deadline.value() - std::chrono::steady_clock::now());
retry_state.timeout = remaining;
```

(inline comment on `core/cluster.cxx:1030`)

#### 4. The error context always reports zero retry attempts (medium)

```cpp
if (ec == errc::common::temporary_failure) {
```

The loop calls `retry_state.retries.record_retry_attempt(reason)`, so the count is accumulated — but the
response handed to the caller was built by `make_error_context`, which constructs the context with
`retry_attempts` fixed at `0` and `retry_reasons` empty, and the loop never writes the accumulated state back.
A user retried twenty-nine times sees `retried: 0` and no reasons, and `last_dispatched_to` is empty even though
the gateway endpoint is known.

RFC 77's ErrorContext section treats these as part of the contract precisely because they are what let someone
diagnose a connection problem; its worked example shows `"retried":29,"retryReasons":["SERVICE_NOT_AVAILABLE"]`.
The classic transport populates them. This is also what #1045's observability parity claim and #1057's
`SDK_COUCHBASE2_OBSERVABILITY` capability depend on.

The loop is the only place that holds the state, so it is the place to stamp it before invoking the handler.

(inline comment on `core/protostellar/error_utils.cxx:228`)

#### 5. Every operation deep-copies its whole request (medium)

```cpp
auto retry_state = request; // preserved for a possible re-issue (carries request.retries)
```

This runs unconditionally, before dispatch, on every couchbase2 operation. For a KV mutation the request holds
the document body, so each `upsert` copies the entire value — on the data-plane hot path, to support a retry that
in normal operation never happens.

Holding the request in a `shared_ptr` and re-issuing from that removes both the copy and the need for a separate
`retry_state`:

```cpp
auto pending = std::make_shared<Request>(std::move(request));
... component->execute(*pending, [pending, ...](auto response) mutable { ... });
```

That needs `component::execute` to accept a reference rather than taking by value, which is a small change and
also removes the redundant copy already flagged by cppcheck's `passedByValue` on the `execute` overloads.

(inline comment on `core/cluster.cxx:1010`)

#### 6. The retry loop has no test (medium)

`test/cng/protostellar/retry.cxx` tests `retry_reason_for`, which is the classification predicate — not the loop.
Nothing exercises a request that actually fails and is re-issued, so none of findings 1 to 4 would be caught,
and neither would a regression that stopped retrying altogether.

The in-process KV service used by `test/cng/protostellar/component.cxx` already has the machinery: a key that
makes the service answer `UNAVAILABLE` a fixed number of times before succeeding is enough to pin the loop's
observable behaviour.

```cpp
void
kv_retries_until_the_deadline_then_reports_the_attempts()
{
  in_process_server server{ /* fail the first two attempts with UNAVAILABLE */ };
  ...
  request.timeout = 500ms;                       // the loop must respect this
  cluster.execute(std::move(request), [&](auto response) { result = std::move(response); });
  io.run();

  assert_false(static_cast<bool>(result->ctx.ec()), "the third attempt succeeds");
  assert_eq(result->ctx.retry_attempts(), 2U, "the error context reports the retries");
}
```

A companion case with a server that always fails, asserting the handler runs once and within the timeout, would
close finding 1 permanently.

### Note on idempotency

Retrying a non-idempotent mutation on `UNAVAILABLE` is worth stating explicitly, because it looks alarming.
`allows_non_idempotent_retry(service_not_available)` permits it, and this matches gocb and RFC 77's error table,
which specifies `UNAVAILABLE` → retry with `SERVICE_NOT_AVAILABLE` and no idempotency caveat. The reasoning is
that `UNAVAILABLE` means the call did not reach a healthy server. That reasoning is not airtight — gRPC also
reports `UNAVAILABLE` for a connection reset after the request was written — but it is the cross-SDK consensus
and should not be changed unilaterally here. If it is to be tightened, `socket_closed_while_in_flight` is the
retry reason that already carries the right semantics, and the change belongs in a cross-SDK discussion.

### Verdict

Blocked. Finding 1 means a couchbase2 operation issued with default options against an unavailable gateway never
completes, which is a hang rather than an error. Findings 2 and 3 are in the same twenty lines.

core/cluster.cxx
```diff
@@ -825,6 +913,144 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+                retry_state.retries.record_retry_attempt(reason.value());
+                auto timer = std::make_shared<asio::steady_timer>(self->ctx_);
+                timer->expires_at(when);
+                timer->async_wait([self,
+                                   retry_state = std::move(retry_state),
+                                   handler = std::move(handler),
+                                   deadline,
+                                   response = std::move(response),
+                                   timer](const std::error_code& timer_ec) mutable {
+                  if (timer_ec) {
```
avsej, line 1127: When the timer fires normally this falls through to `self->execute_protostellar(...)`, which dereferences `protostellar_`. If `cluster::close()` ran during the backoff that member has been reset, so this is a null dereference.

The non-KV twin added in #1050 has the guard, sixty lines below in this same file — which makes this an oversight rather than a deliberate difference:

```suggestion
                  if (timer_ec || !self->protostellar_ || self->stopped_) {
```

Two follow-ups worth considering: the two loops are close enough to fold into one function, which removes the possibility of them diverging again; and cancelling outstanding backoff timers from `close()` would deliver the completion promptly instead of at the end of a backoff whose outcome is already decided.

core/cluster.cxx
```diff
@@ -825,6 +913,144 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+                              ? retry_state.retries.strategy()
+                              : self->origin_.options().default_retry_strategy_;
+            if (strategy) {
+              const auto action = strategy->retry_after(retry_state.retries, reason.value());
+              const auto when = std::chrono::steady_clock::now() + action.duration();
+              if (action.need_to_retry() && (!deadline.has_value() || when < deadline.value())) {
+                retry_state.retries.record_retry_attempt(reason.value());
+                auto timer = std::make_shared<asio::steady_timer>(self->ctx_);
+                timer->expires_at(when);
+                timer->async_wait([self,
```
avsej, line 1121: The re-dispatch passes the request through unchanged, so `component::execute` resolves the timeout again and `dispatcher::unary` stamps `now + timeout` on the new attempt. The overall `deadline` only gates *whether* another attempt starts; it does not shrink that attempt's budget.

Consequence: an operation whose first attempt consumes most of its timeout can start a second with a full budget and run for close to twice the nominal timeout.

Shrink the per-attempt budget to what is left before re-issuing:

```cpp
const auto remaining = std::chrono::duration_cast<std::chrono::milliseconds>(
  deadline.value() - std::chrono::steady_clock::now());
retry_state.timeout = remaining;
```

core/cluster.cxx
```diff
@@ -825,6 +913,144 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+  // replayed.
+  template<class Request, class Handler>
+  void execute_protostellar(Request request,
+                            Handler&& handler,
+                            std::optional<std::chrono::steady_clock::time_point> deadline = {})
+  {
+    if (!deadline.has_value() && request.timeout.has_value()) {
+      deadline = std::chrono::steady_clock::now() + request.timeout.value();
+    }
+    auto retry_state = request; // preserved for a possible re-issue (carries request.retries)
```
avsej: This copy is unconditional and happens before dispatch on every couchbase2 operation. For a KV mutation the request holds the document body, so each `upsert` deep-copies the entire value on the data-plane hot path, to support a retry that in normal operation never happens.

Holding the request in a `shared_ptr` and re-issuing from that removes both the copy and the separate `retry_state`:

```cpp
auto pending = std::make_shared<Request>(std::move(request));
```

That needs `component::execute` to take a reference rather than by value — a small change that also removes the redundant copy cppcheck already reports as `passedByValue` on those overloads.

Separately: nothing tests this loop. `test/cng/protostellar/retry.cxx` covers `retry_reason_for`, the classification predicate, not the re-issue. A case using the in-process KV service to fail twice with `UNAVAILABLE` and then succeed — asserting the operation succeeds and `ctx.retry_attempts() == 2` — would pin the loop's observable behaviour and catch the deadline and retry-reporting defects above.

core/protostellar/error_utils.cxx
```diff
@@ -218,6 +218,23 @@ error_message(const grpc::Status& status) -> std::string
...
 }
 
+auto
+retry_reason_for(std::error_code ec) -> std::optional<couchbase::retry_reason>
+{
+  // UNAVAILABLE maps to temporary_failure (error_utils map_status_code): the call did not reach a
+  // healthy server, so it is safe to retry (even for mutations — the strategy's
+  // allows_non_idempotent_retry(service_not_available) permits it). Everything else is terminal at
+  // the transport: timeouts consumed the budget, cas_mismatch/auth/not_found are semantic.
+  if (ec == errc::common::temporary_failure) {
```
avsej, line 250: The classification itself is right, and matching a locked document to `key_value_locked` follows RFC 77 and both couchbase-jvm-clients and gocb (.NET gets this wrong — CNG-5).

The gap is downstream: the loop accumulates attempts via `retry_state.retries.record_retry_attempt(reason)`, but the response delivered to the caller comes from `make_error_context`, which builds the context with `retry_attempts` fixed at `0` and `retry_reasons` empty — and the loop never writes the accumulated state back. So a user retried twenty-nine times sees `retried: 0`, and `last_dispatched_to` is empty although the gateway endpoint is known.

RFC 77's ErrorContext section treats these as part of the contract, and its worked example shows `"retried":29,"retryReasons":["SERVICE_NOT_AVAILABLE"]`. The classic transport populates them, and #1045's observability parity claim and #1057's `SDK_COUCHBASE2_OBSERVABILITY` capability both depend on it.

The retry loop is the only place holding the state, so it should stamp it onto `response.ctx` before invoking the handler.

core/cluster.cxx
avsej: Confirmed and fixed.

The deadline is now resolved before the first attempt, from `request.timeout.value_or(origin_.options().key_value_timeout)`, and the `!deadline.has_value() ||` escape in the retry condition is gone — `deadline` is a plain `time_point` that every attempt is measured against, so a missing per-operation timeout can no longer mean "no bound".

The fallback is the cluster's own `key_value_timeout` rather than a constant because that is the value the component already applies per attempt. The bound on the operation and the bound on each of its attempts now come from the same place.

Covered by `an_operation_without_a_timeout_is_bounded_by_the_cluster_default`, which issues a `get_request` with no timeout against a service answering `UNAVAILABLE` immediately and forever, with `key_value_timeout` set to 400ms. It asserts the completion arrives, that it arrives inside the bound, and that a retry happened first. The wait is bounded, so the previous behaviour — no completion at all — fails the case instead of hanging the suite. A mutation restoring an effectively unbounded fallback kills it.

core/cluster.cxx
avsej: This one does not hold against the branch as it stands, and the suggested guard would make the read less safe rather than more.

The re-entry does not dereference `protostellar_`. It calls `protostellar_component()`, which takes `protostellar_mutex_` and returns a copy of the `shared_ptr`, and the null case is answered two lines later with `cluster_closed`. The accessor exists for exactly this reason and says so at its definition. `!self->protostellar_` would read that member outside the mutex, on the io thread, while `close()` may be writing it — which is the race the accessor was introduced to remove.

On the premise: the non-KV twin in #1050 does carry that guard, but that is a reason to look at the twin rather than to copy it here. `execute_protostellar_http` reads `self->protostellar_` unguarded in its timer handler, and it also still stamps its deadline only when the request carries a timeout, with the same `!deadline.has_value() ||` escape in its retry condition — so it has the unbounded-loop defect this review opens with, for query, search, analytics and view. Raised there rather than fixed here, since it is that commit's contract.

Measured rather than argued: `closing_the_cluster_during_a_backoff_still_answers` closes the cluster while an operation is provably inside its backoff (it waits for the first attempt to have failed) and asserts the caller is still answered, with a bounded wait so a dropped completion fails the case instead of hanging it. It passes as written.

The completion arrives when the backoff expires and the re-entry finds no component, which is the second point raised: the outcome is decided at `close()` but not delivered until the timer fires. Cancelling outstanding timers from `close()` would deliver it promptly, but that is a change to `close()` rather than to this loop, it changes what a caller observes on a closed cluster, and it is not needed for correctness — so it is left out of this commit rather than folded in silently.

core/cluster.cxx
avsej: Already fixed on the branch, before this round — the review was submitted against an earlier state of it. Each attempt has its budget narrowed to what is left of the operation's deadline before dispatch, and the reasoning, including the "roughly twice its nominal budget" consequence, is recorded at the narrowing site.

It is covered by `a_retried_kv_operation_stays_within_its_budget`, which places its assertion between the two outcomes: a 400ms budget with a flat 150ms backoff ends at ~400ms given the remaining budget and at ~700ms given a fresh one, and the case asserts under 600ms. A mutation that still computes the remainder but no longer applies it kills the case, so the assertion is known to be able to fail.

Nothing changed here in this round beyond the loop being split into an entry point and a per-attempt function, which is what removed the `std::optional` deadline.

core/cluster.cxx
avsej: Fixed as suggested. The request is held in a `shared_ptr` and re-issued from it, the separate `retry_state` is gone, and `component::execute` takes the request by `const&`.

The signature change went to all 44 `execute` overloads rather than only the fifteen the retry loop can re-issue, so the component presents one convention instead of two. It is a removal of copies and not a transfer of them: nothing in `component.cxx` moves out of the request, every overload copies what it needs (the encoded proto, the id, now the retry counters) synchronously before dispatch, and no completion holds a reference to it. The `passedByValue` reports go with it.

Net effect on the data plane is one fewer full copy of the document body per operation. The request is now moved once into the `shared_ptr` and then encoded into the proto; before, it was copied wholesale for `retry_state`, moved into `execute`, and then encoded.

On the second point: the file is `test/cng/protostellar/kv_retry.cxx`, and since this review it has covered the loop rather than the predicate — `a_retried_kv_operation_stays_within_its_budget` drives a real `core::cluster` against a localhost `KvService`. This round adds the cases the suggestion asks for: one asserting the exact retry count and reason reported to the caller, one asserting that an operation with no timeout of its own still terminates, and one closing the cluster mid-backoff. Six mutations cover them.

core/protostellar/error_utils.cxx
avsej: Fixed for KV here. This is CXXCBC-921, which reaches further than this commit; the split is recorded on the ticket.

`make_error_context` now takes the attempt count and the reason set as parameters instead of the two pinned literals, and the KV overloads source them from `request.retries` — the model `core/error_context/key_value.hxx:92` already uses on the classic path, so both transports derive the field the same way.

Nothing needed writing back. The loop records the attempt on the request *before* re-issuing it, so the request each attempt carries already holds the accumulated state, and the context that attempt builds reports it.

Three places report it, not one: the completion of every KV operation; the compressed-content rejection, which replaces the context wholesale and would otherwise discard what the attempt had accumulated; and `fail_expired`, the exhausted-budget path — the outcome the retry loop actually produces when it runs out of time, and so the one report where a non-zero count is the norm.

Covered by `a_retried_operation_reports_its_attempts_and_reasons`, which uses a service that fails a fixed number of dispatches, so the expected count is decided by the service and not by timing. Two mutations — the count pinned back to zero, the reason set discarded — kill it separately, so neither half of the assertion is decorative.

`live_observability.cxx` carried a comment saying the value was always zero because `make_error_context` pinned it. That is no longer true for KV, so the comment was corrected rather than left contradicting the code. Its assertion still checks only that the attribute is present: the upsert it drives succeeds first time, so the only value it can carry is zero, and asserting that would pass just as well with the plumbing removed.

What is left on CXXCBC-921: the query, search, analytics and view contexts declare the same two fields and no couchbase2 path assigns them, so they publish their defaults; `last_dispatched_to` is still empty although the endpoint is known; and the span attribute is still unasserted for want of an operation that actually retries. The non-KV contexts belong with the non-KV retry loop in #1050, and none of it is reachable from this commit.

test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  component comp{ io, channel, credentials, 20000ms };
```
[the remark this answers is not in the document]
avsej: Already addressed; this thread is stale rather than open.

Both sites build a `component_config` by named-field assignment and pass that, so the timeout lands in `component_timeouts::key_value` rather than being passed as a bare duration. The file compiles and is built and run by `ctest -L cng` on every commit of this stack.

avsej:
Round addressed. Four of the six findings are fixed here; two were already fixed on the branch and are answered as such rather than re-fixed, since the review was submitted against an earlier state of it.

**Finding 1 (blocker) — the unbounded loop.** The deadline is resolved before the first attempt from `request.timeout.value_or(origin_.options().key_value_timeout)`, and the `!deadline.has_value() ||` escape is gone, so a request issued with default options can no longer retry without a bound. The fallback is the cluster's own KV timeout, which is the value the component already applies per attempt, so the operation and its attempts are bounded from the same place.

**Finding 5 — the per-operation deep copy.** The request is held in a `shared_ptr` and re-issued from it; `component::execute` now takes it by `const&` across all 44 overloads. Nothing moves out of the request anywhere in `component.cxx`, so this removes a copy rather than relocating one: one fewer full copy of the document body per KV operation.

**Finding 4 — retry attempts reported as zero.** Fixed for KV. `make_error_context` takes the count and reasons as parameters instead of pinning them, sourced from `request.retries` the way the classic path already does. This is CXXCBC-921 and the rest of that ticket — the query, search, analytics and view contexts, and `last_dispatched_to` — stays with it; the split is recorded there.

**Finding 6 — no test for the loop.** `kv_retry.cxx` gains three cases: an operation with no timeout of its own still terminates, a retried operation reports the right count and reason, and closing the cluster mid-backoff still answers the caller. Six mutations in the negative-control spec kill them.

**Findings 2 and 3 were already fixed.** Finding 3 (fresh full-length deadline per attempt) has been narrowed to the remaining budget for some time and carries its own test. Finding 2 (the timer not re-checking `protostellar_`) does not hold: the re-entry goes through `protostellar_component()`, which reads the member under the mutex and answers `cluster_closed` when it is empty. Adding `!self->protostellar_` would read that member outside the lock that exists to make the read safe. That is now measured rather than argued — a case closes the cluster mid-backoff and asserts the caller is still answered.

One thing found while rebasing the series onto this: `execute_protostellar_http` in #1050 has the same unbounded-loop defect — the deadline stamped only when the request carries a timeout, and the same `!deadline.has_value() ||` escape — for query, search, analytics and view, along with the same unconditional request copy and an unguarded read of `protostellar_` in its timer handler. That is CXXCBC-906's contract rather than this one's, so it is raised there rather than fixed here. It does bear on finding 2 though: the twin was cited as evidence that the missing guard here was an oversight, and it is the other way round — the twin is the one reading the member outside the lock.

One suggestion is deliberately not taken: cancelling outstanding backoff timers from `close()`. It would deliver the completion promptly instead of at the end of a backoff whose outcome is already decided, but it changes `close()` and what a caller observes on a closed cluster, and it is not needed for correctness. Left out rather than folded in silently.

The idempotency note is agreed with and nothing changed there.

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. The paths were traced by hand and compared against `core/io/retry_orchestrator.hxx` and Java's `RetryOrchestratorProtostellar`. I have also built the branch and run the `cng` suite locally against a CAO cluster (`CB_SANITIZER=thread`, macOS/arm64), which is where the flake noted under the first heading comes from — GitHub Actions is degraded on our side, so there is no CI signal on this PR to lean on. Please treat the reasoning as a starting point for discussion, not ground truth.

The loop reads well, and the four cases in `kv_retry.cxx` pin the properties that matter (budget, bound, bookkeeping, close-during-backoff) rather than the shape of the code. One thing about how the loop *ends*, though.

## The loop gives up before its deadline, and reports the transport error rather than a timeout

`dispatch_protostellar`'s retry gate is

```cpp
if (action.need_to_retry() && when < deadline) {
```

so when the backoff would land past the deadline, no further attempt is made and the loop delivers the **last attempt's response**. For the condition this PR is built around — gateway unreachable, `UNAVAILABLE` → `temporary_failure` — that is what the caller ends up with: `errc::common::temporary_failure`, delivered *before* the deadline they asked for.

With stock defaults (`key_value_timeout` 2.5s, `best_effort` + `controlled_backoff` = 1, 10, 50, 100, 500, then 1000ms), attempts start at ~0/1/11/61/161/661/1661ms. The attempt at 1661ms fails, `when` = 2661 > 2500, and the operation completes at **~1.66s with `temporary_failure`** — roughly 840ms of the caller's budget unused.

Both existing implementations do the opposite: they never refuse a retry for being close to the deadline, they *shrink the wait* and let the deadline itself end the operation.

- classic: `core/io/retry_orchestrator.hxx` → `priv::cap_duration()` trims the backoff so the retry lands at the deadline, then the request's own deadline completes it as `unambiguous_timeout` / `ambiguous_timeout`.
- Java Protostellar: `RetryOrchestratorProtostellar.capDuration()`, with the comment saying it outright — *"there is intentionally no timeout handling here - that is left to GRPC to raise."*

So over classic an unreachable node yields a timeout, and over `couchbase2` the same condition yields `temporary_failure`, which is a transport-visible difference for anyone branching on timeouts.

The machinery for the other behaviour is already here: cap the wait at the deadline instead of refusing it, and the per-attempt narrowing this commit already does hands the final attempt a non-positive budget — which `component::execute`'s expired-budget guard turns into `unambiguous_timeout` **with** the accumulated attempts and reasons stamped, via the `fail_expired` path this same commit extended. The terminal error becomes a timeout carrying correct retry bookkeeping, and the operation uses the budget it was given.

One consequence to decide on if that changes: `fail_expired` justifies `unambiguous_timeout` for every operation on the grounds that nothing was sent. That is exact for a first attempt, but after a retry a mutation *has* been dispatched at least once, and the classic path would call that `ambiguous_timeout`.

**A side effect worth fixing either way: `a_retried_kv_operation_stays_within_its_budget` is flaky.** It failed once in seven runs of `cng_kv_retry_test` on a tsan build (macOS/arm64):

```
"a_retried_kv_operation_stays_within_its_budget" FAILED:
  test/cng/protostellar/kv_retry.cxx:198: the operation was retried at least once
```

That assertion is `service.calls.load() >= 2`, and it is the first one in the case, so the run tells us only that fewer than two dispatches reached `stalling_kv_service` — not which error the operation ended on.

What makes the case timing-dependent is the gate. It grants a 400ms budget against a flat 150ms backoff, so `when < deadline` holds only while the operation is inside its first **250ms**; once past that, a retryable `UNAVAILABLE` is not retried. Two ways that trips at ~1-in-7, and I can't tell them apart from the failure alone:

- the first attempt's round trip lands after 250ms, so the retry after it is refused; or
- the first attempt fails locally before it reaches the server (channel still connecting → `UNAVAILABLE`, `calls` stays 0), the retry at 150ms is the first dispatch the service counts, and *its* backoff would land at ~450ms, past the deadline.

Either way the retry budget runs out about 150ms early and the caller gets `temporary_failure` rather than a timeout. Under cap-the-backoff semantics the wait shrinks instead of the retry being refused, and neither path can arise — the case stops depending on first-attempt latency at all.

Two small things that would make this diagnosable if you'd rather keep the current semantics: assert the terminal error code (see the note below) and report `calls` in the failure message, so a CI flake says which of the two happened. A 400ms budget with a 250ms effective window is tight enough that I'd expect this on loaded runners, not just under a sanitizer.

**Test note, same finding:** `an_operation_without_a_timeout_is_bounded_by_the_cluster_default` is precisely this path (unavailable-forever service, 400ms budget, 50ms flat backoff → gives up at ~350ms), but it asserts only that a completion arrives and that it arrives under 2s. Asserting the terminal error code there would pin whichever answer you land on.

## Secondary: always-retry reasons now go through the strategy

Both reasons `retry_reason_for` can return are `always_retry` ones (`core/impl/retry_reason.cxx`: `service_not_available` and `key_value_locked` are both in that list). Classic's `maybe_retry` short-circuits on `always_retry(reason)` and retries with `controlled_backoff` **without consulting the strategy at all**; Java does the same (`if (reason.alwaysRetry())`). This loop always asks the strategy, so a caller-supplied strategy that declines — `fail_fast_retry_strategy`, say — changes `couchbase2` behaviour relative to classic for the same server condition. With the default best-effort strategy the two agree, which is why nothing in the suite sees it.

Not obviously the wrong call — it arguably respects the user's strategy more — but it is a deliberate divergence from the list that exists to take that decision away from the strategy, so it seems worth either matching or documenting.

Nothing else: the `const&` conversion checked out (no overload moves out of the request, and every lambda that outlives `execute()` captures by copy — the by-reference `stamped` captures are all consumed synchronously), the idempotency claim holds (`retry_context<true>` / `<false>` per operation, so the strategy really does see it), retry state is recorded before re-issue so each attempt's context reports the retries that preceded it, and the close-during-backoff path delivers exactly once on either side of the close.

test/cng/protostellar/kv_retry.cxx
```diff
@@ -0,0 +1,436 @@
...
+           const v1::GetRequest* /* request */,
+           v1::GetResponse* /* response */) -> grpc::Status override
+  {
+    if (++calls <= 2) {
+      return { grpc::StatusCode::UNAVAILABLE, "try again" };
+    }
+    // Far longer than any budget this test grants, so the client's deadline is what ends the call.
+    std::this_thread::sleep_for(5s);
+    return { grpc::StatusCode::UNAVAILABLE, "still unavailable" };
+  }
```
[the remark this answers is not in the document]
avsej: Fixed. The fixed sleep is now a poll on cancellation, so the handler returns as soon as the client's deadline expires rather than running on into the suite's teardown:

```cpp
const auto give_up_by = std::chrono::steady_clock::now() + 5s;
while (!context->IsCancelled() && std::chrono::steady_clock::now() < give_up_by) {
  std::this_thread::sleep_for(5ms);
}
```

The upper bound is kept so the handler still terminates if cancellation never arrives, rather than trading a fixed sleep for a potential hang.

This is adjacent to a flake reported separately on this PR for the case that drives this service, though it is not its cause — that one was the retry gate ending the operation early, and is fixed in the same round. Holding the RPC handler past `Shutdown()` is a real second-order effect worth removing on its own terms.

avsej:
Round addressed. The main finding held and is fixed; the flake and both test notes go with it. The secondary finding does not hold — the two sources it cites both say the opposite of the claim, and that is set out below in full because it is a factual point about the codebase rather than a judgement call.

## The loop gave up before its deadline — fixed

> so when the backoff would land past the deadline, no further attempt is made and the loop delivers the **last attempt's response**

Confirmed, and the arithmetic is right. The wait is now capped at the deadline rather than the retry being refused for landing past it:

```cpp
const auto when = std::min(std::chrono::steady_clock::now() + action.duration(), deadline);
if (action.need_to_retry()) {
```

The attempt scheduled at the deadline then receives a non-positive remainder from the per-attempt narrowing, which the component answers as a timeout carrying the accumulated attempts and reasons. That terminal state cannot re-enter the loop, because `retry_reason_for` does not classify a timeout as retryable — so capping ends the operation rather than extending it.

Both precedents were checked rather than taken on trust, and both hold: `priv::cap_duration` in `core/io/retry_orchestrator.hxx` trims the backoff exactly this way, and `RetryOrchestratorProtostellar.capDuration` does the same on Java's Protostellar path, applied in both `maybeRetry` and `retryWithDuration`.

> One consequence to decide on if that changes: `fail_expired` justifies `unambiguous_timeout` for every operation on the grounds that nothing was sent. That is exact for a first attempt, but after a retry a mutation *has* been dispatched at least once, and the classic path would call that `ambiguous_timeout`.

Taken, and this was the more important half — it is the difference between telling a caller a write provably never landed and telling it the write may have been applied. `fail_expired` now decides it the way `mcbp_command::cancel()` does:

```cpp
const auto ec = (request.retries.idempotent() || request.retries.retry_attempts() == 0)
                  ? errc::common::unambiguous_timeout  // safe to retry, or never sent to the server
                  : errc::common::ambiguous_timeout;   // non-idempotent and dispatched at least once
```

The comment above it justified the old blanket answer on "nothing is sent", which the change makes false — that attempt sends nothing, but its predecessors did — so it was rewritten rather than left contradicting the code beneath it. This also lines up with what a dispatched attempt already reported: `map_status_code` keys `DEADLINE_EXCEEDED` on `operation_kind`, so a mutation that timed out on the wire was already ambiguous, and only the expired-budget path disagreed.

`a_retried_mutation_that_runs_out_of_budget_is_ambiguous` pins it, and a mutation forcing the condition true kills it.

## The flake — fixed by the same change, and the diagnostics added anyway

> **A side effect worth fixing either way: `a_retried_kv_operation_stays_within_its_budget` is flaky.** It failed once in seven runs

The mechanism given is right: the gate made the effective retry window 250ms of a 400ms budget, so the case depended on first-attempt latency. With the wait capped instead of refused, neither of the two paths described can arise — no retry is ever declined for proximity to the deadline.

> Two small things that would make this diagnosable if you'd rather keep the current semantics: assert the terminal error code (see the note below) and report `calls` in the failure message

Both done regardless, since they are what makes the next failure legible rather than a bare `expected >= 2`. That paid for itself immediately: with the count reported, a new case failed with `dispatches: 0`, which named the cause at once — the in-process service implemented only `Get`, so a mutation was answered `UNIMPLEMENTED`, which is terminal rather than retryable, and never entered the loop at all. A bare count assertion would have looked like a retry bug.

> **Test note, same finding:** `an_operation_without_a_timeout_is_bounded_by_the_cluster_default` … asserts only that a completion arrives and that it arrives under 2s. Asserting the terminal error code there would pin whichever answer you land on.

Added there too. Both cases now assert the timeout code rather than merely that some completion arrived, which is what distinguishes ending on the deadline from ending on the last attempt's `temporary_failure`. A mutation restoring the old gate kills them.

On evidence for the flake specifically: 15 consecutive runs pass here, but that is a weak signal — this is an unloaded Linux box with no sanitizer, not the tsan/macOS/arm64 configuration where it was seen. The argument for it being fixed is the mechanism, not the run count.

## Copilot: the stalling service ignored cancellation — fixed

The 5s `sleep_for` is now a poll on `ServerContext::IsCancelled()`, so the handler returns when the client gives up instead of running on into teardown and holding the server past `Shutdown()`. That is also a plausible contributor to the tail latency of that case.

## Secondary: always-retry reasons — this does not hold

> Both reasons `retry_reason_for` can return are `always_retry` ones (`core/impl/retry_reason.cxx`: `service_not_available` and `key_value_locked` are both in that list).

They appear in that switch, but on the arm that returns **false**. `always_retry()` is a single switch whose first group — nineteen reasons including `service_not_available`, `key_value_locked` and, confusingly, `do_not_retry` — falls through to `return false;`, and only `key_value_not_my_vbucket`, `key_value_collection_outdated` and `views_no_active_partition` return `true`. Linked against the library and called directly:

```
service_not_available    -> 0
key_value_locked         -> 0
key_value_not_my_vbucket -> 1
```

So classic's `maybe_retry` does **not** short-circuit for either reason this loop can produce; it falls through to exactly the same "consult the strategy" path, and reaches `cap_duration` there. This loop is doing what classic does.

> Java does the same (`if (reason.alwaysRetry())`).

Java agrees with C++ on the values. `RetryReason`'s constructor is `RetryReason(boolean allowsNonIdempotentRetry, boolean alwaysRetry)` — the always-retry flag is the **second** argument — and the entries are `SERVICE_NOT_AVAILABLE(true, false)` and `KV_LOCKED(true, false)`. Both are `allowsNonIdempotentRetry = true`, `alwaysRetry = false`. `KV_NOT_MY_VBUCKET(true, true)` is the contrast.

So there is no divergence to match or document, and a `fail_fast_retry_strategy` declines these reasons on the classic transport too. Nothing changed for this finding.

(One genuine C++/Java difference turned up while checking: `key_value_collection_outdated` is always-retry in C++ and `alwaysRetry = false` in Java. It is unrelated to this loop — no path here produces that reason — and is not touched.)

## Verification

- 36/36 `ctest -L cng` at this commit against a live gateway, nothing skipped.
- 8/8 negative-control mutations killed, up from six: the two new ones restore the old gate and force the ambiguity decision, so both new properties can fail.
- `kv_retry.cxx` run 15× consecutively without failure, with the caveat above about the configuration.
- The stack was re-walked from this commit up; every commit above it builds and passes.

thejcfactor (approved):

thejcfactor (approved):

PR #1044, opened 2026-07-28
CXXCBC-903: document running the CNG tests over couchbase2

avsej:
Add `test/cng/README.md` describing how to bring up a local Cloud Native Gateway with k3d and
cbdinocluster and run the CNG (`couchbase2://`) tests against it: the environment gates
(env-agnostic vs cluster-only), the one-time Kubernetes setup, a cluster definition that enables the
gateway, obtaining the `couchbase2://` connection string, the `tls_verify=none` requirement for
local dev gateways, the build/run commands, and the `TEST_CB2_*` variables that override the
credentials and bucket the cluster-only tests use.

The cluster definition pins release versions (`8.0.0` / `2.8.0` / `1.2.1`) rather than build
numbers. cbdinocluster maps a version carrying a build number to `ghcr.io/cb-vanilla/*`, which needs
a Couchbase-organization token, while a plain release maps to the public `couchbase/*` images — so
the pins are what makes the walkthrough followable from outside the organization. They match
`.github/actions/create-cng-cluster/action.yml`, which CI runs on every push.

---

**This pull request is now based directly on `main` and no longer part of a stack.** It was
reordered to the bottom of the CNG series so it can merge on its own: the diff is one file,
`test/cng/README.md`, and it depends on nothing else in the series. The six resolved review threads
on code files date from when this branch still carried the commits beneath it; those files are no
longer in the diff.

### DIFF
--- a/test/cng/README.md
+++ b/test/cng/README.md
@@ -1,0 +1,211 @@
+# CNG (`couchbase2://`) tests
+
+These tests exercise the Protostellar / Cloud Native Gateway (CNG) transport. They are built
+with the hand-rolled harness under `framework/` (not Catch2) and are enabled by the
+`COUCHBASE_CXX_CLIENT_BUILD_CNG_TESTS` CMake option.
+
+Each test declares an environment gate:
+
+- **env-agnostic** — runs anywhere; uses in-process or loopback gRPC servers. No cluster needed.
+- **cluster-only** — needs a reachable `couchbase2://` gateway; skipped when
+  `TEST_CONNECTION_STRING` is unset (the harness exits `77`, which ctest reports as *Skipped*).
+
+Getting the cluster-only tests to run means standing up a Couchbase cluster whose Cloud Native
+Gateway is deployed, and pointing the tests at its `couchbase2://` endpoint. The rest of this
+document shows how to do that locally with `cbdinocluster` and `k3d`.
+
+## Prerequisites
+
+Linux with:
+
+- **Docker** — the container runtime everything else builds on.
+- **[k3d](https://k3d.io)** — runs a lightweight Kubernetes (k3s) cluster inside Docker. CNG is
+  deployed by the Couchbase Autonomous Operator, which needs Kubernetes.
+- **[cbdinocluster](https://github.com/couchbaselabs/cbdinocluster)** — allocates Couchbase
+  clusters for testing. Its `cao` deployer installs the operator into your k3d cluster and, when
+  the cluster definition asks for it, the Cloud Native Gateway.
+- A C++ toolchain plus **gRPC and protobuf ≥ 3.15**, required by the couchbase2 transport. System
+  packages (`grpc-devel` / `libgrpc++-dev` and `protobuf-devel` / `libprotobuf-dev`) are only a
+  convenience: when `find_package(gRPC)` does not find them the build fetches gRPC and its protobuf
+  submodule and builds them in-tree, which needs no packages but makes the first build considerably
+  longer. Building with `-stdlib=libc++` always fetches from source, because distribution packages
+  are compiled against libstdc++ and would fail to link.
+
+## 1. One-time Kubernetes setup
+
+Create a local Kubernetes cluster with k3d and register it with cbdinocluster. Do this once per
+machine (k3d clusters do not survive a reboot — see the note at the end to restart one):
+
+```console
+$ k3d cluster create mycluster
+$ cbdinocluster init --auto --kube-config ~/.kube/config
+```
+
+`cbdinocluster init` records the k3d context so the `cao` deployer can use it.
+
+`--auto` takes every default and asks nothing, which is what CI does and what you want here.
+Without it the command is an interactive questionnaire covering each provider it supports — Docker,
+Kubernetes, AWS, Azure, GCP, Capella, DNS and GitHub — and only two answers matter for this
+walkthrough: **Kubernetes must be enabled**, and the kubeconfig it uses must be the one `k3d cluster
+create` wrote. Declining Kubernetes, or pointing it at a different kubeconfig, leaves the `cao`
+deployer unusable and the failure does not surface until `allocate` several steps later. The cloud
+providers are unused here and can all be declined.
+
+There is no need to name a context. `cbdinocluster` accepts `--k8s-context`, but its handler reads
+a differently named key and the value is discarded, so it always falls back to the kubeconfig's
+current context — which `k3d cluster create` has just set to the new cluster.
+
+Confirm the Kubernetes side is healthy — this should list `cao` among the deployers with no
+namespace error:
+
+```console
+$ cbdinocluster ps
+```
+
+## 2. Define a cluster with the Cloud Native Gateway
+
+The gateway is deployed by the operator and is requested **per cluster** in the definition file.
+A definition without a `cao.gateway-version` yields a working cluster with **no** `couchbase2://`
+endpoint. Write `cluster-def.yaml`:
+
+```yaml
+nodes:
+  - count: 1
+    version: 8.0.0
+    services: [kv]
+  - count: 1
+    version: 8.0.0
+    services: [kv, n1ql, index]
+  - count: 1
+    version: 8.0.0
+    services: [kv, fts]
+  - count: 1
+    version: 8.0.0
+    services: [kv, cbas]
+cao:
+  operator-version: "2.8.0"
+  gateway-version: "1.2.1"   # presence of this line is what enables CNG
+```
+
+**Use release versions, not build numbers.** cbdinocluster maps a version carrying a build number
+(`8.1.0-1743`, `1.2.2-123`) to `ghcr.io/cb-vanilla/*`, which requires a Couchbase-organization token;
+a plain release version maps to the public `couchbase/*` images on Docker Hub and needs no
+credentials. CI pins release versions for exactly this reason, so the definition above is what an
+outside contributor can actually pull.
+
+There is deliberately no `docker:` memory block here: the `cao` deployer ignores it, so service
+quotas stay at the server defaults regardless. That is also why the bucket below is created with a
+small explicit quota. The four nodes cover every service the suite can touch; if you only need KV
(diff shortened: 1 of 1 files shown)


avsej (commented):
## Review — PR #1044 (`CXXCBC-903`)

This commit adds `test/cng/README.md`, documenting how to stand up a gateway with cbdinocluster and how to build
and run the CNG suite. The documentation was checked against the code rather than read on its own, and it holds
up: both CMake options it names exist (`COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` and
`COUCHBASE_CXX_CLIENT_BUILD_CNG_TESTS`), the ctest label `cng` matches `test/cng/CMakeLists.txt`, the
`TEST_CONNECTION_STRING` variable matches what the harness reads, and the exit-77-means-Skipped behaviour is
described accurately. It does not contradict `AGENTS.md` or the CI workflow, and it uses `couchbase2` rather than
`Protostellar` in the user-facing text, which is what RFC 77's Naming section requires.

One gap.

### Findings

|---|---|---|---|---|
| 1 | medium | docs | The four `TEST_CB2_*` override variables the live tests read are not documented | `test/cng/README.md:141` |

#### 1. The `TEST_CB2_*` override variables are not documented (medium)

```
Default credentials are `Administrator` / `password` and the bucket is `default`. A single test
```

The defaults stated here are correct, but the variables that override them are not named anywhere in the
document. The live tests read four:

| Variable | Default |
|---|---|
| `TEST_CB2_USERNAME` | `Administrator` |
| `TEST_CB2_PASSWORD` | `password` |
| `TEST_CB2_BUCKET` | `default` |
| `TEST_CB2_SEARCH_INDEX` | `cng-index` |

The consequence is that anyone pointing the suite at a gateway that is not a freshly created cbdinocluster — a
shared environment, or a cluster with different credentials — cannot discover from the README how to configure it,
and has to read `test/cng/framework/` to find out. Documenting the defaults without documenting the knobs is the
one thing a README of this kind most needs to get right.

`TEST_CB2_SEARCH_INDEX` deserves a sentence of its own: the live search tests expect an FTS index named
`cng-index` to exist, and the document's cbdinocluster walkthrough does not create one. Without it those tests do
not exercise what they appear to, so a reader following the instructions exactly gets less coverage than they
would reasonably assume.

```markdown
### Overriding the defaults

The live suites read these in addition to `TEST_CONNECTION_STRING`:

| Variable | Default | Notes |
|---|---|---|
| `TEST_CB2_USERNAME` | `Administrator` | |
| `TEST_CB2_PASSWORD` | `password` | |
| `TEST_CB2_BUCKET` | `default` | must exist on the cluster |
| `TEST_CB2_SEARCH_INDEX` | `cng-index` | the live search tests expect this FTS index to exist; the walkthrough above does not create it |
```

(inline comment on `test/cng/README.md:141`)

### Verdict

Ready with nits. The document is accurate as far as it goes; finding 1 is an addition rather than a correction.

test/cng/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+To include the cluster-only tests, export the endpoint (with `tls_verify=none` for a local
+gateway) first:
+
+```console
+$ export TEST_CONNECTION_STRING="couchbase2://<host>:<port>?tls_verify=none"
+$ ctest --test-dir build -L cng --output-on-failure
+```
+
+Default credentials are `Administrator` / `password` and the bucket is `default`. A single test
```
avsej: These defaults are correct, but the variables that override them are not named anywhere in the document. The live suites read four beyond `TEST_CONNECTION_STRING`:

| Variable | Default |
|---|---|
| `TEST_CB2_USERNAME` | `Administrator` |
| `TEST_CB2_PASSWORD` | `password` |
| `TEST_CB2_BUCKET` | `default` |
| `TEST_CB2_SEARCH_INDEX` | `cng-index` |

Anyone pointing the suite at anything other than a freshly created cbdinocluster — a shared environment, or a cluster with different credentials — has to read `test/cng/framework/` to find these. Documenting the defaults without the knobs is the thing a README of this kind most needs to avoid.

`TEST_CB2_SEARCH_INDEX` deserves an explicit note: the live search tests expect an FTS index named `cng-index`, and the cbdinocluster walkthrough above does not create one — so a reader following the instructions exactly gets less coverage than they would assume.

avsej: Accepted. The variables are now documented, and checking them turned up two further problems in the same document.

> These defaults are correct, but the variables that override them are not named anywhere in the document. The live suites read four beyond `TEST_CONNECTION_STRING`

Confirmed as exactly four, with those exact defaults — a sweep for `TEST_` across `test/cng/` returns `TEST_CONNECTION_STRING` and those four and nothing else. An **Overriding the defaults** section was added, stating that `TEST_CONNECTION_STRING` alone decides *whether* the cluster-only tests run while the rest only change *what* they run against, with a worked example of pointing the suite at a shared cluster. The bucket row records that it must already exist, since the tests do not create it.

One correction to the finding, caused by this pull request having been reordered to sit directly on `main`. At that position only three of the four are read: `live_kv.cxx` and `live_connect.cxx` are the only cluster-only cases present, and `live_search.cxx` — the sole reader of `TEST_CB2_SEARCH_INDEX` — first appears in the search/views commit. Documenting all four here would name a variable no test in the tree reads. The fourth row and its note were therefore added in that commit instead, so each commit's README is true at the point it merges. The full table lands as soon as search does.

> `TEST_CB2_SEARCH_INDEX` deserves an explicit note: the live search tests expect an FTS index named `cng-index`, and the cbdinocluster walkthrough above does not create one — so a reader following the instructions exactly gets less coverage than they would assume.

The conclusion holds and the note was written, with one detail sharpened: the case does not *require* the index. It skips only on `feature_not_available` and asserts only that the gateway serves `search.v1`, so a missing index returns an index error and the transport round-trip is still proven. What is lost without an index is hits, so the effect is a weaker check rather than a failure or a skip. The note says that, and says how to get the stronger version.

Two further problems, found while verifying the above.

**The cluster definition pinned images that need credentials.** The example specified server `8.1.0-1743` and gateway `1.2.2-123`. A version carrying a build number maps to `ghcr.io/cb-vanilla/*`, which requires a Couchbase-organization token; a plain release version maps to the public `couchbase/*` images and needs none. The document's whole purpose is to be followable, and as written it was not followable by anyone outside the organization. It now pins `8.0.0` / `2.8.0` / `1.2.1`, matching what `.github/actions/create-cng-cluster/action.yml` uses, with the reason stated so the pins are not silently "modernised" back later.

**A paragraph described coverage that does not exist.** The document instructed the reader to create `myscope` / `mycollection` because "some tests expect non-default collections". No test does — those two names appear nowhere in the tree except that paragraph, and every cluster-only case addresses `_default` / `_default`. The paragraph now states that the bucket is all the tests need, and keeps the commands only as an explicitly optional aid for manual work.

**Prerequisites overstated system packages.** They were listed as required. `ThirdPartyDependencies.cmake` includes `GrpcProtobuf.cmake` whenever `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` is set, and that module fetches gRPC and its protobuf submodule when `find_package(gRPC)` fails — so the packages are a build-time convenience, not a gate. The entry now says so, and adds that `-stdlib=libc++` always fetches from source because distribution packages are built against libstdc++. This also answers a suppressed reviewer comment that raised the same point.

Also from the suppressed set, on the diff containing unrelated code: that is resolved rather than answered. This branch now contains `main` plus the single documentation commit, so the diff is one file and the pull request body no longer describes a stack.

The `docker:` memory block was dropped from the example. The `cao` deployer ignores it, so it implied quotas that were never applied; the document now says quotas stay at server defaults, which is why the bucket is created with an explicit small one.

Not verified: the walkthrough was not executed end to end against a fresh k3d cluster in this round. The version pins and registry behaviour come from the CI action that runs on every push, and the `cbdinocluster` invocations are unchanged from the reviewed version apart from the collections paragraph, but a manual run is what would make this "verified" rather than "consistent with CI".

test/cng/README.md
```diff
@@ -0,0 +1,197 @@
...
+  are compiled against libstdc++ and would fail to link.
+
+## 1. One-time Kubernetes setup
+
+Create a local Kubernetes cluster with k3d and register it with cbdinocluster. Do this once per
+machine (k3d clusters do not survive a reboot — see the note at the end to restart one):
+
+```console
+$ k3d cluster create mycluster
+$ cbdinocluster init --kube-config ~/.kube/config
```
thejcfactor: Should we mention what input is required in the interactive init w.r.t. kubernetes?  If they don't get that right, future steps will fail.

avsej: > Should we mention what input is required in the interactive init w.r.t. kubernetes?  If they don't get that right, future steps will fail.

Agreed, and the failure mode is the bad kind: nothing complains at `init` time, and the first symptom is `allocate` failing several steps later with an error that does not point back here.

The walkthrough now runs `cbdinocluster init --auto`, which takes every default and asks nothing. That is what the CI action does (`.github/actions/create-cng-cluster/action.yml`), so it is the path with the most mileage on it.

For anyone who does run it interactively, the surrounding text now says which answers matter. `init` walks through every provider the tool supports — Docker, Kubernetes, AWS, Azure, GCP, Capella, DNS and GitHub — and only two of those answers affect this walkthrough: **Kubernetes has to be enabled**, and the kubeconfig has to be the one `k3d cluster create` just wrote. Declining Kubernetes, or pointing it at a different kubeconfig, is exactly the case you describe. The cloud providers are unused here and can all be declined.

One related thing that was worth stating explicitly, since it looks like something you would need to get right and is not: there is no need to name a context. `cbdinocluster` accepts `--k8s-context`, but its handler reads a differently named key, so the value is discarded and it always falls back to the kubeconfig's current context — which `k3d cluster create` has just set to the new cluster. That behaviour was already documented in a comment in the CI action; it now appears in the README too, so a reader does not go looking for a flag to set.

Separately, and not from this thread: the `gateway-otlp-endpoint` key was dropped from the list of `cao` options. OpenTelemetry is not part of what this document is for, and listing the key invited configuring it as though the walkthrough depended on it.

thejcfactor (approved):

PR #1045, opened 2026-07-28
CXXCBC-902: verify couchbase2 observability via the transport-agnostic recorder

avsej:
Operation spans and the db.client.operation.duration latency metric are emitted
by the public-layer observability recorder before transport selection, so a
couchbase2 operation already flows through the configured tracer/meter with no
transport-specific code -- the ticket's "skip if op-level tracing already wraps
the call" condition. Add a live test that proves it end to end: a KV upsert over
couchbase2:// through the public cluster API, with a capturing tracer and meter,
must start at least one span and record at least one latency value. The
per-dispatch child span that the MCBP transport emits is not produced on
couchbase2; the operation-level span and latency metric are, which is what this
test asserts.

---
Stacked PR **23/29** in the CNG (`couchbase2://`) transport series. This branch includes every commit from PRs 01–23; the change specific to this PR is its top commit. Depends on #1043 — merge the series bottom-up.

### DIFF
--- a/test/cng/protostellar/live_observability.cxx
+++ b/test/cng/protostellar/live_observability.cxx
@@ -1,0 +1,337 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Observability over couchbase2 (CXXCBC-902). Operation spans and the operation-latency metric
+// are emitted at the public layer (observability_recorder) BEFORE transport selection, so a
+// couchbase2 operation gets them through the existing tracer/meter with no transport-specific code.
+// This test proves that end to end via the public couchbase::cluster API, and it asserts the SHAPE
+// of what was emitted rather than merely that something was: the exact span sequence, the absence
+// of a dispatch_to_server span (RFC 77 leaves dispatch to the gRPC layer's own span), the operation
+// and service attributes, and the metric those land on. A count of spans is satisfied by any span
+// with any name carrying any attributes, so it cannot see a lost request_encoding child, a classic
+// operation identifier leaking through, or latency filed under the wrong service.
+//
+// It is also the only case in the suite that drives the public API, which makes it the guard for
+// everything below observability -- it is what caught setup_observability() going missing from
+// open_protostellar(). cluster_only, and the CI `cng` job runs it on every pull request.
+
+#include "framework/test_runner.hxx"
+
+#include <couchbase/cluster.hxx>
+#include <couchbase/codec/tao_json_serializer.hxx>
+#include <couchbase/metrics/meter.hxx>
+#include <couchbase/tracing/request_tracer.hxx>
+
+#include <algorithm>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <vector>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+// Records what observability actually emitted -- the name of every span, its parent, the tags put
+// on it, and the name and tags of every metric recorded -- rather than only how many there were.
+//
+// A count is satisfied by any span with any name carrying any attributes, so it cannot see the
+// failures that matter here: a KV operation losing its request_encoding child, a span carrying the
+// classic operation identifier instead of the Protostellar one, or a latency metric filed under the
+// wrong service. Those are the regressions this test exists to catch, and each of them keeps the
+// counters happily non-zero.
+struct emitted {
+  mutable std::mutex mutex;
+  std::vector<std::string> span_names;
+  std::map<std::string, std::map<std::string, std::string>> span_tags;
+  std::vector<std::string> metric_names;
+  std::vector<std::map<std::string, std::string>> metric_tags;
+
+  void note_span(const std::string& name)
+  {
+    const std::scoped_lock lock{ mutex };
+    span_names.push_back(name);
+  }
+
+  void note_tag(const std::string& span, const std::string& key, const std::string& value)
+  {
+    const std::scoped_lock lock{ mutex };
+    span_tags[span][key] = value;
+  }
+
+  void note_metric(const std::string& name, const std::map<std::string, std::string>& tags)
+  {
+    const std::scoped_lock lock{ mutex };
+    metric_names.push_back(name);
+    metric_tags.push_back(tags);
+  }
+
+  [[nodiscard]] auto has_span(const std::string& name) const -> bool
+  {
+    const std::scoped_lock lock{ mutex };
+    return std::find(span_names.begin(), span_names.end(), name) != span_names.end();
+  }
+
+  [[nodiscard]] auto tag(const std::string& span, const std::string& key) const -> std::string
+  {
+    const std::scoped_lock lock{ mutex };
+    const auto s = span_tags.find(span);
+    if (s == span_tags.end()) {
+      return {};
+    }
+    const auto t = s->second.find(key);
+    return t == s->second.end() ? std::string{} : t->second;
+  }
+
+  [[nodiscard]] auto metric_tag(std::size_t index, const std::string& key) const -> std::string
+  {
+    const std::scoped_lock lock{ mutex };
+    if (index >= metric_tags.size()) {
+      return {};
+    }
+    const auto t = metric_tags[index].find(key);
+    return t == metric_tags[index].end() ? std::string{} : t->second;
+  }
+
+  [[nodiscard]] auto describe() const -> std::string
+  {
+    const std::scoped_lock lock{ mutex };
+    std::string out{ "spans:" };
+    for (const auto& name : span_names) {
+      out += " " + name;
+    }
+    for (const auto& [s, tags] : span_tags) {
+      out += " [" + s + ":";
+      for (const auto& [k, v] : tags) {
+        out += " " + k + "=" + v;
+      }
+      out += "]";
+    }
+    out += " | metrics:";
+    for (std::size_t i = 0; i < metric_names.size(); ++i) {
+      out += " " + metric_names[i] + "{";
+      for (const auto& [k, v] : metric_tags[i]) {
+        out += k + "=" + v + ",";
+      }
+      out += "}";
+    }
+    return out;
+  }
+};
+
+class recording_span : public couchbase::tracing::request_span
+{
+public:
+  recording_span(std::string name,
+                 std::shared_ptr<couchbase::tracing::request_span> parent,
+                 std::shared_ptr<emitted> sink)
+    : request_span(std::move(name), std::move(parent))
+    , sink_{ std::move(sink) }
+  {
+    sink_->note_span(this->name());
+  }
+  void add_tag(const std::string& tag_name, std::uint64_t value) override
+  {
+    sink_->note_tag(name(), tag_name, std::to_string(value));
+  }
+  void add_tag(const std::string& tag_name, const std::string& value) override
+  {
+    sink_->note_tag(name(), tag_name, value);
+  }
+  void end() override
+  {
+  }
+
+private:
+  std::shared_ptr<emitted> sink_;
+};
+
+class recording_tracer : public couchbase::tracing::request_tracer
+{
+public:
+  explicit recording_tracer(std::shared_ptr<emitted> sink)
+    : sink_{ std::move(sink) }
+  {
+  }
+  auto start_span(std::string name, std::shared_ptr<couchbase::tracing::request_span> parent = {})
+    -> std::shared_ptr<couchbase::tracing::request_span> override
+  {
+    return std::make_shared<recording_span>(std::move(name), std::move(parent), sink_);
+  }
+
+private:
+  std::shared_ptr<emitted> sink_;
+};
+
+class recording_recorder : public couchbase::metrics::value_recorder
+{
+public:
+  recording_recorder(std::string name,
+                     std::map<std::string, std::string> tags,
+                     std::shared_ptr<emitted> sink)
+    : name_{ std::move(name) }
+    , tags_{ std::move(tags) }
+    , sink_{ std::move(sink) }
+  {
+  }
+  void record_value(std::int64_t /* value */) override
+  {
+    sink_->note_metric(name_, tags_);
+  }
+
+private:
+  std::string name_;
+  std::map<std::string, std::string> tags_;
+  std::shared_ptr<emitted> sink_;
+};
+
+class recording_meter : public couchbase::metrics::meter
+{
+public:
+  explicit recording_meter(std::shared_ptr<emitted> sink)
+    : sink_{ std::move(sink) }
+  {
+  }
+  auto get_value_recorder(const std::string& name, const std::map<std::string, std::string>& tags)
+    -> std::shared_ptr<couchbase::metrics::value_recorder> override
+  {
+    return std::make_shared<recording_recorder>(name, tags, sink_);
+  }
+
+private:
+  std::shared_ptr<emitted> sink_;
+};
+
+// safe_getenv rather than std::getenv: MSVC deprecates getenv (C4996) and these builds treat
+// warnings as errors, so a raw call fails the Windows legs outright (C2220) while leaving the
+// Linux and macOS legs green.
+auto
+env_or(const char* name, const char* fallback) -> std::string
+{
+  return safe_getenv(name).value_or(fallback);
+}
+
+void
+kv_op_emits_span_and_metric_over_couchbase2()
+{
+  const auto connstr = safe_getenv("TEST_CONNECTION_STRING");
+  if (!connstr.has_value()) {
+    skip("TEST_CONNECTION_STRING is not set");
+  }
+  const auto& connection_string = connstr.value();
+  if (connection_string.rfind("couchbase2://", 0) != 0) {
+    skip("TEST_CONNECTION_STRING is not a couchbase2:// endpoint");
+  }
+
+  auto observed = std::make_shared<emitted>();
+
+  couchbase::cluster_options options{ env_or("TEST_CB2_USERNAME", "Administrator"),
+                                      env_or("TEST_CB2_PASSWORD", "password") };
+  options.security().tls_verify(couchbase::tls_verify_mode::none);
+  options.tracing().tracer(std::make_shared<recording_tracer>(observed));
+  options.metrics().meter(std::make_shared<recording_meter>(observed));
+
+  auto [connect_err, cluster] = couchbase::cluster::connect(connection_string, options).get();
+  assert_false(connect_err.ec().operator bool(), "connect(couchbase2://) succeeds");
+
+  auto collection =
+    cluster.bucket(env_or("TEST_CB2_BUCKET", "default")).scope("_default").collection("_default");
+  auto [upsert_err, upsert_result] = collection.upsert("cng-observability-key", true, {}).get();
+  assert_false(upsert_err.ec().operator bool(), "upsert over couchbase2 succeeds");
+  static_cast<void>(upsert_result); // only the error and the emitted spans/metrics matter here
+
+  cluster.close().get();
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto seen = observed->describe();
+
+  // The shape, not the count. Each assertion below names a regression it would catch.
+
+  // A KV operation keeps the classic top-level span and its request_encoding child. Losing the
+  // child -- the likeliest casualty of transport-specific code creeping into the public layer --
+  // leaves the counters non-zero and would pass a "something was emitted" check.
+  assert_true(observed->span_names == std::vector<std::string>{ "upsert", "request_encoding" },
+              "KV emits exactly the upsert span and its request_encoding child; " + seen);
+
+  // RFC 77: couchbase2 must NOT emit dispatch_to_server -- the gRPC library's own span replaces it.
+  // Reintroducing it would double-count dispatch in every trace.
+  assert_false(observed->has_span("dispatch_to_server"),
+               "couchbase2 leaves dispatch_to_server to the gRPC layer; " + seen);
+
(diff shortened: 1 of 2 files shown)


test/cng/protostellar/live_observability.cxx
```diff
@@ -0,0 +1,171 @@
...
+
+  auto collection =
+    cluster.bucket(env_or("TEST_CB2_BUCKET", "default")).scope("_default").collection("_default");
+  auto [upsert_err, upsert_result] = collection.upsert("cng-observability-key", true, {}).get();
+  assert_false(upsert_err.ec().operator bool(), "upsert over couchbase2 succeeds");
+  static_cast<void>(upsert_result); // only the error and the emitted spans/metrics matter here
+
+  cluster.close().get();
+
+  assert_true(spans->load() > 0, "the operation started at least one span via the tracer");
```
avsej: The verification for this ticket is two non-zero counters, which passes for any span with any name and any attributes. It cannot detect the failures that matter: a missing `request_encoding` span on the KV path, a `dispatch_to_server` span that RFC 77 says should be replaced by the gRPC library's span, a management `db.operation` still carrying the classic identifier instead of the Protostellar RPC name, or metrics tagged with the wrong service.

Asserting the shape makes it a regression test:

```cpp
  assert_true(recorded_span_names(*recorder) ==
                std::vector<std::string>{ "upsert", "request_encoding" },
              "KV keeps the classic top-level span and its request_encoding child");
  assert_eq(attribute(*recorder, "upsert", "db.couchbase.service"), "kv", "service attribute");
```

Two further points:

- The case is `test_env::cluster_only`, so it skips on every pull request. The argument for this test is that the recorder is transport-independent, which means it can be `test_env::agnostic` and run against the in-process gRPC server the other component tests already use — turning it into a gate that runs on every change.
- Observability parity does not currently hold for retries: `make_error_context` fixes `retry_attempts` at `0` and `retry_reasons` empty, and the retry loops in #1043/#1050 never write the accumulated state back, so `retry_count` is always zero. RFC 77's ErrorContext example shows `"retried":29` for exactly this reason, and the classic transport populates it.

avsej (commented):
## Review — PR #1045 (`CXXCBC-902`)

This commit adds `test/cng/protostellar/live_observability.cxx` to verify that the transport-agnostic
observability recorder is exercised on the couchbase2 path. The premise is sound: the recorder sits above
transport selection, so a couchbase2 operation should produce the same spans and metrics as a classic one
without any transport-specific wiring. Confirming that with a test rather than assuming it is the right instinct.

The test as written does not confirm much, and the claim of observability parity does not hold for one
dimension.

### Findings

|---|---|---|---|---|
| 1 | medium | test-coverage | The verification is two `> 0` counters, and it is gated behind a live gateway although the path is transport-independent | `test/cng/protostellar/live_observability.cxx:150` |
| 2 | medium | correctness | Retry attempts and reasons never reach the error context, so the retry dimension of observability parity is not achieved | `test/cng/protostellar/live_observability.cxx:150` |

#### 1. The verification is two `> 0` counters, behind a live-gateway gate (medium)

```cpp
assert_true(spans->load() > 0, "the operation started at least one span via the tracer");
```

The whole verification of the ticket is that two counters are non-zero. That will pass for any span with any
name and any attributes, so it cannot detect the failures that would actually matter: a missing
`request_encoding` span on the KV path, a `dispatch_to_server` span that should have been replaced by the gRPC
library's own span, a management `db.operation` still carrying the classic identifier rather than the
Protostellar RPC name, or metrics tagged with the wrong service.

Asserting the shape is not much more work and is what makes the test a regression test:

```cpp
assert_true(recorded_span_names(*recorder) == std::vector<std::string>{ "upsert", "request_encoding" },
            "KV keeps the classic top-level span and its request_encoding child");
assert_eq(attribute(*recorder, "upsert", "db.system"), "couchbase", "db.system attribute");
assert_eq(attribute(*recorder, "upsert", "db.couchbase.service"), "kv", "service attribute");
```

Separately, the case is `test_env::cluster_only`, so it skips on every pull request and only runs in the job
that stands up a real gateway. Since the argument for this test is precisely that the recorder is independent of
transport, the case can be `test_env::agnostic` and run against the in-process gRPC server the other component
tests already use. That converts it from a gateway-only check into a gate that runs on every change.

(inline comment on `test/cng/protostellar/live_observability.cxx:150`)

#### 2. The retry dimension of observability parity is not achieved (medium)

`make_error_context` in `core/protostellar/error_utils.cxx` constructs the context with `retry_attempts` fixed
at `0` and `retry_reasons` empty, and the retry loops added in #1043 and #1050 never write the accumulated state
back onto the response. So an operation that was retried twenty-nine times reports zero retries, and the
observability `retry_count` tag derived from that field is always zero.

RFC 77's ErrorContext section treats this as part of the contract, and its worked example shows
`"retried":29,"retryReasons":["SERVICE_NOT_AVAILABLE"]` precisely because those fields are what let a user
diagnose a connection problem. The classic transport populates them.

This is not this commit's defect, but it is the commit that asserts observability parity, so the assertion is
currently broader than what the code delivers. The narrow fix belongs in the retry wrappers, which are the only
place that holds the state:

```cpp
// in execute_protostellar / execute_protostellar_http, before invoking the caller's handler
protostellar_set_retry_info(response.ctx, retry_state.retries);
```

A test that drives one retryable failure and then asserts `ctx.retry_attempts() == 1` would keep it fixed.

### Verdict

Needs changes — modest ones. Making the case transport-agnostic and asserting span shape rather than a count
turns this from a smoke test into the regression gate the ticket is asking for.

test/cng/protostellar/live_observability.cxx
avsej: > The verification for this ticket is two non-zero counters, which passes for any span with any name and any attributes. It cannot detect the failures that matter: a missing `request_encoding` span on the KV path, a `dispatch_to_server` span that RFC 77 says should be replaced by the gRPC library's span, a management `db.operation` still carrying the classic identifier instead of the Protostellar RPC name, or metrics tagged with the wrong service.

Accepted. The tracer and meter now record what was emitted — every span name, the tags on each, and the name and tags of each metric — and the case asserts the shape.

The expected shape was **observed before it was asserted**, rather than transcribed from the suggestion, and that turned out to matter: the attribute is `couchbase.service` on both the span and the metric, not `db.couchbase.service`. What a KV upsert over couchbase2 actually emits:

```
spans: upsert request_encoding
  [upsert: couchbase.collection.name=_default couchbase.retries=0 couchbase.scope.name=_default
           couchbase.service=kv db.namespace=default db.operation.name=upsert db.system.name=couchbase]
  [request_encoding: db.system.name=couchbase]
metrics: db.client.operation.duration{__unit=s, couchbase.service=kv, db.operation.name=upsert,
         db.namespace=default, db.system.name=couchbase, couchbase.scope.name=_default, ...}
```

So the assertions are: the span sequence is exactly `upsert` then `request_encoding`; `dispatch_to_server` is **absent**, which is the RFC 77 point stated as a negative so reintroducing it fails; the operation span carries `db.operation.name=upsert`, `couchbase.service=kv`, `db.system.name=couchbase` and the bucket in `db.namespace`; and exactly one `db.client.operation.duration` metric is recorded, tagged with the service and the operation. Failure messages print the entire observed shape, so a break shows what changed rather than only that something did.

Negative-controlled: renaming the child span from `request_encoding` to `dispatch_to_server` in `observability_recorder::create_request_encoding_span()` turns the case red with the full shape in the message. (First attempt at that control did not compile, and the stale binary "passed" — worth noting since a failed build leaves the previous binary in place.)

> The case is `test_env::cluster_only`, so it skips on every pull request.

The conclusion is right that an agnostic version would be better, but this particular premise does not hold, and it changes the priority.

`tests.yml` triggers on `pull_request`, and the `cng` job has **no `if:` gate** — it stands up a real Cloud Native Gateway via cbdinocluster's `cao` deployer, exports `TEST_CONNECTION_STRING` from it, and runs `ctest --label-regex cng`. So `cluster_only` cases do gate every pull request today; they skip in the `unit` job and run in the `cng` one.

That is not an argument against your point, only against its urgency. The `cng` job is slow (k3d plus operator plus server, 40-minute timeout) and independently flaky, so an env-agnostic version that also runs in the fast `unit` job is worth having.

It is not free, though, and the obstacle is not effort. Driving the **public** API is the whole value of this case — it is the only one that reaches `observability_recorder` — and that requires a real listening endpoint: `parse_connection_string()` sets `tls = true` for `couchbase2://` unconditionally, and plaintext with credentials is refused on purpose by the fail-closed guard in `make_channel_credentials()`. There are no test certificates in the tree and nothing uses `SslServerCredentials`. So it needs either a committed self-signed cert and key, or one generated at test start. Relaxing the client to permit plaintext would trade a deliberate production guard for test convenience, which is not a trade worth making. Left for a follow-up rather than done here, since the `cng` job already runs this.

Concretely, the value of it running is not hypothetical. This case caught a real crash: `setup_observability()` had been deleted from `open_protostellar()`, leaving `tracer_` null so that `observability_recorder`'s `tracer_.lock()->create_span(...)` dereferenced an empty `weak_ptr` and segfaulted on every operation issued through the public API over couchbase2. The commit that dropped the line has been fixed, and this case has been moved to sit directly above #1034 so that per-commit verification catches that class of regression at the commit that causes it rather than nine commits later — verified by deleting the line again and watching this case fail at the right commit.

> Observability parity does not currently hold for retries: `make_error_context` fixes `retry_attempts` at `0` and `retry_reasons` empty, and the retry loops in #1043/#1050 never write the accumulated state back, so `retry_count` is always zero.

Confirmed, and now visible in the output above as `couchbase.retries=0`. `make_error_context()` builds the context with a positional aggregate initialiser whose fifth and sixth arguments — `retry_attempts` and the reason set per `key_value_error_context`'s constructor — are hard-coded `0` and `{}`.

Filed as **CXXCBC-921**. It is not fixed here because `make_error_context` and the retry loops belong to #1043 and #1050; this commit only reads the attribute.

One deliberate choice follows from it: the case asserts that `couchbase.retries` is *present* but does **not** assert its value, with a comment saying why. Asserting `"0"` would lock the defect in as expected behaviour. When CXXCBC-921 is fixed, the assertion to add is an operation that actually retries — a locked document is convenient, since `document_locked` maps to `retry_reason::key_value_locked` — checking a non-zero count and the reason.

Worth recording alongside it: fifteen unlabelled positional arguments is how those two fields came to be pinned without anyone noticing, and it is the same hazard that bit `component_config` when it gained a member. Named-field assignment is in the suggested approach on the ticket.

avsej:
Reposting this visibly. It was originally answered in the inline thread on `live_observability.cxx`, but that file was rewritten by this very change before the reply went up, so GitHub marked the thread outdated and collapsed it — the reply is there but hidden behind "Resolved conversations". My sequencing error; the content is unchanged.

---

> The verification for this ticket is two non-zero counters, which passes for any span with any name and any attributes. It cannot detect the failures that matter: a missing `request_encoding` span on the KV path, a `dispatch_to_server` span that RFC 77 says should be replaced by the gRPC library's span, a management `db.operation` still carrying the classic identifier instead of the Protostellar RPC name, or metrics tagged with the wrong service.

Accepted. The tracer and meter now record what was emitted — every span name, the tags on each, and the name and tags of each metric — and the case asserts the shape.

The expected shape was **observed before it was asserted**, rather than transcribed from the suggestion, and that turned out to matter: the attribute is `couchbase.service` on both the span and the metric, not `db.couchbase.service`. What a KV upsert over couchbase2 actually emits:

```
spans: upsert request_encoding
  [upsert: couchbase.collection.name=_default couchbase.retries=0 couchbase.scope.name=_default
           couchbase.service=kv db.namespace=default db.operation.name=upsert db.system.name=couchbase]
  [request_encoding: db.system.name=couchbase]
metrics: db.client.operation.duration{__unit=s, couchbase.service=kv, db.operation.name=upsert,
         db.namespace=default, db.system.name=couchbase, couchbase.scope.name=_default, ...}
```

So the assertions are: the span sequence is exactly `upsert` then `request_encoding`; `dispatch_to_server` is **absent**, which is the RFC 77 point stated as a negative so reintroducing it fails; the operation span carries `db.operation.name=upsert`, `couchbase.service=kv`, `db.system.name=couchbase` and the bucket in `db.namespace`; and exactly one `db.client.operation.duration` metric is recorded, tagged with the service and the operation. Failure messages print the entire observed shape, so a break shows what changed rather than only that something did.

Negative-controlled: renaming the child span from `request_encoding` to `dispatch_to_server` in `observability_recorder::create_request_encoding_span()` turns the case red with the full shape in the message. (First attempt at that control did not compile, and the stale binary "passed" — worth noting since a failed build leaves the previous binary in place.)

> The case is `test_env::cluster_only`, so it skips on every pull request.

The conclusion is right that an agnostic version would be better, but this particular premise does not hold, and it changes the priority.

`tests.yml` triggers on `pull_request`, and the `cng` job has **no `if:` gate** — it stands up a real Cloud Native Gateway via cbdinocluster's `cao` deployer, exports `TEST_CONNECTION_STRING` from it, and runs `ctest --label-regex cng`. So `cluster_only` cases do gate every pull request today; they skip in the `unit` job and run in the `cng` one.

That is not an argument against your point, only against its urgency. The `cng` job is slow (k3d plus operator plus server, 40-minute timeout) and independently flaky, so an env-agnostic version that also runs in the fast `unit` job is worth having.

It is not free, though, and the obstacle is not effort. Driving the **public** API is the whole value of this case — it is the only one that reaches `observability_recorder` — and that requires a real listening endpoint: `parse_connection_string()` sets `tls = true` for `couchbase2://` unconditionally, and plaintext with credentials is refused on purpose by the fail-closed guard in `make_channel_credentials()`. There are no test certificates in the tree and nothing uses `SslServerCredentials`. So it needs either a committed self-signed cert and key, or one generated at test start. Relaxing the client to permit plaintext would trade a deliberate production guard for test convenience, which is not a trade worth making. Left for a follow-up rather than done here, since the `cng` job already runs this.

Concretely, the value of it running is not hypothetical. This case caught a real crash: `setup_observability()` had been deleted from `open_protostellar()`, leaving `tracer_` null so that `observability_recorder`'s `tracer_.lock()->create_span(...)` dereferenced an empty `weak_ptr` and segfaulted on every operation issued through the public API over couchbase2. The commit that dropped the line has been fixed, and this case has been moved to sit directly above #1034 so that per-commit verification catches that class of regression at the commit that causes it rather than nine commits later — verified by deleting the line again and watching this case fail at the right commit.

> Observability parity does not currently hold for retries: `make_error_context` fixes `retry_attempts` at `0` and `retry_reasons` empty, and the retry loops in #1043/#1050 never write the accumulated state back, so `retry_count` is always zero.

Confirmed, and now visible in the output above as `couchbase.retries=0`. `make_error_context()` builds the context with a positional aggregate initialiser whose fifth and sixth arguments — `retry_attempts` and the reason set per `key_value_error_context`'s constructor — are hard-coded `0` and `{}`.

Filed as **CXXCBC-921**. It is not fixed here because `make_error_context` and the retry loops belong to #1043 and #1050; this commit only reads the attribute.

One deliberate choice follows from it: the case asserts that `couchbase.retries` is *present* but does **not** assert its value, with a comment saying why. Asserting `"0"` would lock the defect in as expected behaviour. When CXXCBC-921 is fixed, the assertion to add is an operation that actually retries — a locked document is convenient, since `document_locked` maps to `retry_reason::key_value_locked` — checking a non-zero count and the reason.

Worth recording alongside it: fifteen unlabelled positional arguments is how those two fields came to be pinned without anyone noticing, and it is the same hazard that bit `component_config` when it gained a member. Named-field assignment is in the suggested approach on the ticket.

test/cng/protostellar/live_observability.cxx
```diff
@@ -0,0 +1,321 @@
...
+              "KV emits exactly the upsert span and its request_encoding child; " + seen);
+
+  // RFC 77: couchbase2 must NOT emit dispatch_to_server -- the gRPC library's own span replaces it.
+  // Reintroducing it would double-count dispatch in every trace.
+  assert_false(observed->has_span("dispatch_to_server"),
+               "couchbase2 leaves dispatch_to_server to the gRPC layer; " + seen);
+
+  // The operation identifier must be the Protostellar RPC name. A management or KV path still
+  // reporting the classic identifier here is the failure this pins.
+  assert_eq(observed->tag("upsert", "db.operation.name"),
```
DemetrisChr, line 278: > The operation identifier must be the Protostellar RPC name

I should update the RFC to mention this, but this is no longer the case with the stable semantic conventions. `db.operation.name` now always matches the span name, for both classic and Protostellar. We should update the implementation if that's not the case.

N.B. Observability in Protostellar is still volatile across all SDKs, we should make sure we mention that in the release notes.

test/cng/protostellar/live_observability.cxx
```diff
@@ -0,0 +1,321 @@
...
+            std::string{ "upsert" },
+            "the span carries the Protostellar operation name");
+  assert_eq(observed->tag("upsert", "couchbase.service"),
+            std::string{ "kv" },
+            "the span is attributed to the kv service, not a transport-specific one");
+  assert_eq(observed->tag("upsert", "db.system.name"), std::string{ "couchbase" }, "system tag");
+  assert_eq(observed->tag("upsert", "db.namespace"), bucket, "the span names the bucket");
```
DemetrisChr, line 285: `couchbase.collection.name` and `couchbase.scope.name` should also be present

test/cng/protostellar/live_observability.cxx
```diff
@@ -0,0 +1,321 @@
...
+  // Latency lands on the transport-agnostic metric, tagged for the right service and operation --
+  // metrics filed under the wrong service are invisible until someone reads a dashboard.
+  assert_true(observed->metric_names == std::vector<std::string>{ "db.client.operation.duration" },
+              "exactly one operation-duration metric is recorded; " + seen);
+  assert_eq(observed->metric_tag(0, "couchbase.service"),
+            std::string{ "kv" },
+            "the metric is tagged with the kv service");
+  assert_eq(observed->metric_tag(0, "db.operation.name"),
+            std::string{ "upsert" },
+            "the metric is tagged with the operation");
```
DemetrisChr, line 305: `couchbase.collection.name` and `couchbase.scope.name` should also be present.

avsej:
All three are addressed. The short version: the implementation already did the right thing on all
three counts, and the test was not checking any of it -- so these were live gaps in coverage, and
one comment that stated the wrong rule in prose while the assertion below it did the right thing.

### 1. `db.operation.name` and the RPC name

> > The operation identifier must be the Protostellar RPC name
>
> I should update the RFC to mention this, but this is no longer the case with the stable semantic conventions. `db.operation.name` now always matches the span name, for both classic and Protostellar. We should update the implementation if that's not the case.

The comment was wrong and has been rewritten. The assertion underneath it was already comparing
`db.operation.name` against `"upsert"` -- the span name, not the RPC name -- so the prose and the
code disagreed, and the prose was the part that was out of date.

The implementation needs no change: `db.operation.name` is set once, from the operation name, in
the transport-agnostic recorder (`core/impl/observability_recorder.cxx`), so there is no
couchbase2-specific path that could report `/couchbase.kv.v1.KvService/Upsert` here. Verified
against a live gateway rather than by reading: the assertion passes.

It now compares against the span name itself instead of a repeated literal, so the two cannot
drift apart while both still look correct in isolation:

```cpp
assert_eq(observed->tag("upsert", "db.operation.name"),
          observed->span_names.at(0),
          "db.operation.name matches the span name");
```

The span-name list is pinned exactly (`{ "upsert", "request_encoding" }`) by an assertion above it,
so this is anchored rather than self-referential.

### 2 and 3. `couchbase.collection.name` and `couchbase.scope.name`

> `couchbase.collection.name` and `couchbase.scope.name` should also be present

Both are present, on the span and on the metric, and both are now asserted -- four assertions,
since the two threads land on the two different sinks and each needed its own.

They were already being emitted: the recorder is built in the public layer
(`core/impl/collection.cxx` sets bucket, scope and collection on it), which is above the transport,
so couchbase2 inherits them. That is precisely why they were worth pinning here -- a couchbase2
path that built its own recorder, or that stopped at the bucket, would have satisfied every other
assertion in this case while losing two thirds of the keyspace. The test asserted `db.namespace`
and nothing further.

Confirmed against a live gateway: all four new assertions pass, and they are not trivially
satisfiable, since a missing attribute reads as an empty string and would fail against
`"_default"`.

### The release-notes point

> N.B. Observability in Protostellar is still volatile across all SDKs, we should make sure we mention that in the release notes.

Noted, and it cannot be actioned in this pull request: this repository has no `CHANGELOG` file
(checked at the repository root), so there is nowhere in-tree for it to go. It needs to be raised
wherever the release notes are actually authored, and it is recorded on the handoff notes for this
series so it is not lost between now and the ship-enable work.

The RFC update is yours to make; nothing here depends on it, since the client already follows the
stable-conventions behaviour.

test/cng/protostellar/live_observability.cxx
avsej: Addressed — answered in full at https://github.com/couchbase/couchbase-cxx-client/pull/1045#issuecomment-5193570377. In short: the implementation already behaved correctly on all three points; the assertions were missing and one comment stated the opposite rule in prose. Four assertions added and the comment corrected, verified against a live gateway.

test/cng/protostellar/live_observability.cxx
avsej: Addressed — answered in full at https://github.com/couchbase/couchbase-cxx-client/pull/1045#issuecomment-5193570377. In short: the implementation already behaved correctly on all three points; the assertions were missing and one comment stated the opposite rule in prose. Four assertions added and the comment corrected, verified against a live gateway.

test/cng/protostellar/live_observability.cxx
avsej: Addressed — answered in full at https://github.com/couchbase/couchbase-cxx-client/pull/1045#issuecomment-5193570377. In short: the implementation already behaved correctly on all three points; the assertions were missing and one comment stated the opposite rule in prose. Four assertions added and the comment corrected, verified against a live gateway.

test/cng/protostellar/live_observability.cxx
DemetrisChr:
```
  // An earlier reading of RFC 77 had
  // couchbase2 reporting the Protostellar RPC name here (`/couchbase.kv.v1.KvService/Upsert`); the
  // stable semantic conventions do not, so a path reporting the RPC name -- or the classic
  // identifier where the span says otherwise -- is the failure this pins.
```
I think this part of the comment is not necessary, it probably just adds confusion for someone reading this code.

`/couchbase.kv.v1.KvService/Upsert` is also not what the legacy conventions had specified, just `Upsert`

DemetrisChr (approved):

test/cng/protostellar/live_observability.cxx
avsej: Removed — the comment is now one sentence: the rule, plus why the comparison is written against the span name rather than a literal. The `/couchbase.kv.v1.KvService/Upsert` claim was wrong and went with it; the legacy conventions specified `Upsert`. The adjacent scope/collection comment was trimmed the same way. Pushed as `efe405f`.

DemetrisChr (approved):

PR #1046, opened 2026-07-28
CXXCBC-904: honor kv_durable_timeout for durable KV mutations over couchbase2

avsej:
Durable KV mutations were bounded by the standard key_value timeout (2.5s) rather than the dedicated kv_durable_timeout (10s), so a synchronous-durability write timed out on the read budget.

Add key_value_durable to component_timeouts and select it, through resolve_kv_timeout, for a request that asks for a durability level. The retry loop resolves the operation's deadline by the same rule rather than a copy of it: it narrows the request timeout on every attempt, so the component is always handed an engaged value and applying the rule only there would leave it unreachable. The cluster's timeouts are assigned per field, so a member left unpopulated is a visible omission rather than one covered by the struct's own default.

No lower bound is imposed on an explicit request timeout. The MCBP path raises a durable operation to a 1500ms floor, which it needs because it has no durable default at all; neither RFC 59 nor RFC 77 states a floor.

--- Stacked PR **24/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1043 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -325,6 +325,24 @@ protostellar_response_ec(const Context& ctx) -> std::error_code
 {
   return ctx.ec;
 }
+
+// The cluster's configured timeouts, in the form the component and the retry loop both read them
+// from. Assigned per field rather than brace-initialised positionally: every member has the same
+// type, so a mis-ordering is not a compile error, and an omitted one silently takes the struct's
+// own default instead of the value the caller configured.
+inline auto
+make_component_timeouts(const cluster_options& options) -> protostellar::component_timeouts
+{
+  protostellar::component_timeouts timeouts{};
+  timeouts.key_value = options.key_value_timeout;
+  timeouts.key_value_durable = options.key_value_durable_timeout;
+  timeouts.query = options.query_timeout;
+  timeouts.analytics = options.analytics_timeout;
+  timeouts.search = options.search_timeout;
+  timeouts.view = options.view_timeout;
+  timeouts.management = options.management_timeout;
+  return timeouts;
+}
 #endif
 } // namespace
 
@@ -993,14 +1011,7 @@ public:
       protostellar_ = std::make_shared<protostellar::component>(
         ctx_,
         protostellar::component_config{
-          std::move(channel), origin_.credentials(), protostellar::component_timeouts {
-            options.key_value_timeout,
-            options.query_timeout,
-            options.analytics_timeout,
-            options.search_timeout,
-            options.view_timeout,
-            options.management_timeout
-          } });
+          std::move(channel), origin_.credentials(), make_component_timeouts(options) });
     }
     CB_LOG_INFO(R"(open couchbase2 cluster, id: "{}", endpoint: "{}")", id_, endpoint);
     return handler({});
@@ -1044,16 +1055,26 @@ public:
   // for one, and the strategy never refuses a retry for service_not_available, so a request issued
   // with default options against an unavailable gateway would otherwise retry until the process
   // ended -- never completing, and never reporting an error either.
+  //
+  // Which default applies is resolve_kv_timeout's rule, not a second copy of it: a durable mutation
+  // is bounded by KvDurableTimeout and everything else by the standard KV timeout. It has to be
+  // applied here, because dispatch_protostellar narrows request->timeout on every attempt including
+  // the first -- so the component is always handed an engaged timeout and its own fallback never
+  // selects for a request that arrives through this loop. Resolving with the standard default here
+  // would cap a durable mutation at the budget the durable default exists to replace.
   template<class Request, class Handler>
   void execute_protostellar(Request request, Handler&& handler)
   {
-    // The fallback below is the KV default, so routing anything else here would bound it with a
+    // The fallback below is a KV default, so routing anything else here would bound it with a
     // budget belonging to another service. Non-KV requests carry no retry_context and have their
     // own path.
     static_assert(protostellar::component_supports_v<Request>,
                   "execute_protostellar resolves the key-value timeout");
-    const auto deadline = std::chrono::steady_clock::now() +
-                          request.timeout.value_or(origin_.options().key_value_timeout);
+    const auto deadline =
+      std::chrono::steady_clock::now() +
+      protostellar::resolve_kv_timeout(request.timeout,
+                                       protostellar::requested_durability(request),
+                                       make_component_timeouts(origin_.options()));
     // Held by shared_ptr and re-issued from there. The request owns the document body for a
     // mutation, so copying it to keep a re-issue possible charged every operation on the data plane
     // for a retry that in normal operation never happens.
--- a/core/protostellar/component.hxx
+++ b/core/protostellar/component.hxx
@@ -30,6 +30,7 @@
 // context. cluster_impl routes KV ops here when the cluster was opened with couchbase2://.
 
 #include "core/cluster_credentials.hxx"
+#include "core/io/mcbp_traits.hxx"
 #include "core/operations/document_analytics.hxx"
 #include "core/operations/document_append.hxx"
 #include "core/operations/document_decrement.hxx"
@@ -98,6 +99,7 @@ struct component_timeouts {
   std::chrono::milliseconds search{ timeout_defaults::search_timeout };
   std::chrono::milliseconds view{ timeout_defaults::view_timeout };
   std::chrono::milliseconds management{ timeout_defaults::management_timeout };
+  std::chrono::milliseconds key_value_durable{ timeout_defaults::key_value_durable_timeout };
 };
 
 // Construction parameters for `component`, grouped so that adding one is a source-compatible
@@ -146,8 +148,40 @@ inline constexpr bool component_supports_v<operations::append_request> = true;
 template<>
 inline constexpr bool component_supports_v<operations::prepend_request> = true;
 
-// Default per-service operation timeouts applied when a request does not carry its own. Grouped in
-// a struct so adding a service does not grow the component constructor's positional argument list.
+// Resolves the effective timeout for a KV operation: the request's own timeout if set, else the
+// per-service default — the durable default for a synchronous-durability mutation (RFC 77
+// KvDurableTimeout), the standard KV default otherwise. Free function so it is unit-testable.
+//
+// No lower bound is imposed on an explicit request timeout. The MCBP path raises a durable
+// operation's timeout to a 1500ms floor, which it needs because it has no durable default at all
+// and a durable write there inherits the 2.5s KV budget; resolving KvDurableTimeout is the
+// RFC-specified mechanism for that, and neither RFC 59 nor RFC 77 states a floor. A caller who
+// names a timeout gets the one they named.
+[[nodiscard]] inline auto
+resolve_kv_timeout(std::optional<std::chrono::milliseconds> request_timeout,
+                   couchbase::durability_level level,
+                   const component_timeouts& timeouts) -> std::chrono::milliseconds
+{
+  const auto fallback =
+    (level == couchbase::durability_level::none) ? timeouts.key_value : timeouts.key_value_durable;
+  return request_timeout.value_or(fallback);
+}
+
+// The durability level a request asks for, or `none` for a request type that cannot carry one.
+// Lets a caller generic over the request type apply the rule above without knowing which types
+// have the member; `supports_durability_v` is the same trait the MCBP path keys its own durability
+// handling on, so the two transports agree on which operations are durable.
+template<class Request>
+[[nodiscard]] auto
+requested_durability(const Request& request) -> couchbase::durability_level
+{
+  if constexpr (io::mcbp_traits::supports_durability_v<Request>) {
+    return request.durability_level;
+  } else {
+    static_cast<void>(request);
+    return couchbase::durability_level::none;
+  }
+}
 
 class component
 {
--- a/test/cng/protostellar/live_kv.cxx
+++ b/test/cng/protostellar/live_kv.cxx
@@ -35,13 +35,17 @@
 #include "core/utils/binary.hxx"
 #include "core/utils/connection_string.hxx"
 
+#include <couchbase/durability_level.hxx>
 #include <couchbase/error_codes.hxx>
 
 #include <asio/executor_work_guard.hpp>
 #include <asio/io_context.hpp>
 
+#include <array>
+#include <cstddef>
 #include <cstdlib>
 #include <string>
+#include <utility>
 
 namespace couchbase::cng::test
 {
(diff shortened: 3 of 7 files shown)


test/cng/framework/test_framework.hxx
```diff
@@ -0,0 +1,259 @@
...
+// to compile.
+#if defined(__has_builtin)
+#define COUCHBASE_CNG_HAS_LOCATION_BUILTINS                                                        \
+  (__has_builtin(__builtin_FILE) && __has_builtin(__builtin_LINE) &&                               \
+   __has_builtin(__builtin_FUNCTION))
+#elif defined(__GNUC__) || defined(__clang__)
+#define COUCHBASE_CNG_HAS_LOCATION_BUILTINS 1
+#else
+#define COUCHBASE_CNG_HAS_LOCATION_BUILTINS 0
+#endif
```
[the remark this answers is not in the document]
avsej: Addressed: the builtins are now used only when __has_builtin confirms all three, or -- where __has_builtin itself is unavailable -- only under Clang; every other compiler (incl. GCC without __has_builtin) falls back to the unknown-location implementation.

core/cluster.cxx
```diff
@@ -825,6 +913,145 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+      CB_LOG_ERROR(R"([{}]: failed to establish couchbase2 channel: {})", id_, e.what());
+      stopped_ = true;
+      work_.reset();
+      return handler(errc::common::invalid_argument);
+    }
+    protostellar_ = std::make_shared<protostellar::component>(
+      ctx_,
+      std::move(channel),
+      origin_.credentials(),
+      protostellar::component_timeouts{ options.key_value_timeout,
```
avsej: `component_timeouts` is an aggregate of seven identically-typed `std::chrono::milliseconds` members and this is its only construction site, initialised positionally. `key_value_durable` was inserted as the *second* member, which silently changes the meaning of every following initialiser.

Two things make that riskier than a normal reordering: the members are all the same type, so a mis-ordering or omission is not a compile error; and an omitted member default-initialises to `0ms`, which `dispatcher::unary` treats as **no deadline at all** rather than as a default — so the failure mode is an unbounded call.

The struct's comment says it exists "so adding a service does not grow the component constructor's positional argument list", which is right — but constructing it positionally gives that benefit back. C++17 has no designated initialisers, so per-field assignment is the option that cannot be silently mis-ordered:

```cpp
protostellar::component_timeouts timeouts{};
timeouts.key_value = options.key_value_timeout;
timeouts.key_value_durable = options.key_value_durable_timeout;
timeouts.query = options.query_timeout;
// ...
```

Separately: the durable default resolved here bounds each *attempt*, not the operation. `execute_protostellar` stamps its overall deadline only from `request.timeout`, which is `std::nullopt` by default — so a durable mutation with default options retries without limit, now at 10s per attempt rather than 2.5s. The unbounded-retry defect is reported on #1043; the fix is for the retry loop to call `resolve_kv_timeout` rather than reading the optional directly.

core/protostellar/component.hxx
```diff
@@ -0,0 +1,355 @@
...
+  std::chrono::milliseconds search{};
+  std::chrono::milliseconds view{};
+  std::chrono::milliseconds management{};
+};
+
+// Resolves the effective timeout for a KV operation: the request's own timeout if set, else the
+// per-service default — the durable default for a synchronous-durability mutation (RFC 77
+// KvDurableTimeout), the standard KV default otherwise. Free function so it is unit-testable.
+[[nodiscard]] inline auto
+resolve_kv_timeout(std::optional<std::chrono::milliseconds> request_timeout,
```
avsej, line 161: Making this a free function so it can be unit-tested directly is the right structure.

The test that accompanies it, though, calls it with each durability level and asserts the returned value — which pins a three-line function rather than the wiring that can actually regress: whether all eight mutation paths call it, whether reads avoid it, and whether the resolved value reaches the gRPC deadline. A case against the in-process server issuing a durable mutation with no explicit timeout and asserting the observed deadline would cover that.

Two boundary cases are also missing: `majority_and_persist_to_active`, and the negative case that a read never picks the durable budget.

One behavioural note worth a comment in the code: the classic MCBP path in this repository does not apply `kv_durable_timeout` to KV operations, so after this commit the same durable write gets 10s over `couchbase2://` and 2.5s over `couchbase://`. This commit is the one that matches RFC 77, so the divergence is in the right direction — but recording it prevents the next reader concluding one of the two is a bug. The classic path also enforces a 1500ms floor on durable timeouts that this path does not; whether that applies here is worth deciding explicitly.

avsej (commented):
## Review — PR #1046 (`CXXCBC-904`)

This commit makes durable KV mutations use `kv_durable_timeout` instead of `kv_timeout`, via a new
`resolve_kv_timeout` free function that picks the durable default when a durability level other than `none` is
requested and otherwise the standard one. Factoring it out as a free function so it can be unit-tested directly is
the right call, and the defaults match RFC 77's settings table (2.5s and 10s). An explicit per-request timeout
still wins over both, which matches the classic behaviour.

Two observations.

### Strengths

- `resolve_kv_timeout` is a pure function of its inputs, which makes the selection rule testable without standing
  up a component. That is a better structure than resolving the timeout inline at each of the mutation call sites.
- The rule is keyed on the durability level actually requested rather than on whether the operation is a mutation,
  so a non-durable `upsert` correctly keeps the 2.5s budget.

### Findings

|---|---|---|---|---|
| 1 | medium | correctness | The resolved durable timeout does not bound the retry loop, so a durable mutation with default options can retry without limit | `core/cluster.cxx:971` |
| 2 | medium | api-design | A field was inserted into the middle of a positionally-initialised aggregate, where an omitted member means "no deadline" | `core/cluster.cxx:971` |
| 3 | low | test-coverage | The test restates the helper rather than pinning the wiring that can regress | `core/protostellar/component.hxx:150` |

#### 1. The durable timeout does not bound the retry loop (medium)

`resolve_kv_timeout` is consulted inside `component::execute`, which sets the *per-attempt* gRPC deadline. The
retry loop in `cluster_impl::execute_protostellar` (#1043) stamps its overall deadline only from
`request.timeout`, which is `std::nullopt` unless the caller set one explicitly — so the default this commit
introduces bounds each attempt but not the operation.

Consequence: a durable `upsert` issued with default options against an unavailable gateway retries indefinitely,
now with a 10-second budget per attempt instead of 2.5, so it also consumes four times as long per round. The
underlying unbounded-retry defect is reported on #1043; this commit is worth mentioning it on because it is the
commit that establishes where the default lives, and the fix is to have the retry loop call the same
`resolve_kv_timeout` rather than reading the optional directly.

#### 2. A field inserted mid-aggregate, where omission means "no deadline" (medium)

```cpp
protostellar::component_timeouts{ options.key_value_timeout,
```

`component_timeouts` is an aggregate of seven identically-typed `std::chrono::milliseconds` members, and this is
its only construction site — a positional brace-init list. `key_value_durable` was inserted as the second member,
which silently changes the meaning of every following initialiser.

Two things make that riskier than usual. The members are all the same type, so a mis-ordering or an omission is
not a compile error. And an omitted member default-initialises to `0ms`, which `dispatcher::unary` treats as *no
deadline at all* rather than as a default — so the failure mode of getting this wrong is an unbounded call, not a
short timeout.

The struct's own comment says it exists "so adding a service does not grow the component constructor's positional
argument list", which is exactly right — but constructing it positionally gives that benefit up. C++17 has no
designated initialisers, so the options are a small builder, or explicit per-field assignment:

```cpp
protostellar::component_timeouts timeouts{};
timeouts.key_value = options.key_value_timeout;
timeouts.key_value_durable = options.key_value_durable_timeout;
timeouts.query = options.query_timeout;
...
```

which cannot be silently mis-ordered and makes an omission visible.

(inline comment on `core/cluster.cxx:971`)

#### 3. The test restates the helper (low)

`test/cng/protostellar/component_timeouts.cxx` calls `resolve_kv_timeout` directly with each durability level and
asserts the value it returns. That pins the function, which is worth having — but the function is three lines and
is not where a regression would occur. What can regress is the wiring: whether all eight mutation paths actually
call it, whether reads do not, and whether the resolved value reaches the gRPC deadline.

A test against the in-process server that issues a durable mutation with no explicit timeout and asserts the
observed deadline would cover the part that can break. Two boundary cases are also missing:
`majority_and_persist_to_active`, and the negative case that a read never picks the durable budget.

(inline comment on `core/protostellar/component.hxx:150`)

### Note

The classic MCBP transport in this repository does not apply `kv_durable_timeout` to KV operations, so after this
commit the same durable write gets 10 seconds over `couchbase2://` and 2.5 seconds over `couchbase://`. This
commit is the one that matches RFC 77, so the divergence is in the right direction — but it is a divergence
within one library, and a comment recording it would save the next reader from concluding one of the two is a bug.
The classic path also enforces a 1500ms floor on durable timeouts that this path does not; whether that floor
should apply here is worth deciding explicitly rather than by omission.

### Verdict

Ready with nits. The selection rule is correct and well-placed; the findings are about the boundary between this
commit and the retry loop, and about how the aggregate is constructed.

core/cluster.cxx
avsej: Both halves of this were live, and both turned out to be worse than described.

**The aggregate.** `key_value_durable` is declared *last*, after `management`, and every member has a default member initialiser — so nothing was shifted and no value landed in the wrong field. What had already happened is the other failure the same shape allows: this construction site passed **six** initialisers for a **seven**-member struct, so `key_value_durable` was never populated from the options and silently took the struct's own default. The two properties named here are exactly what hid it — the members share a type, so a short list is not a compile error, and the value it fell back to (the RFC default, 10s) is plausible enough that no assertion on a stock configuration would notice. What it cost is the configured value: `key_value_durable_timeout` is settable through `timeout_options` and the connection string, and is raised to 20s by the `wan_development` profile, and none of that reached this transport.

Now assigned per field, in one function both users read:

```cpp
inline auto
make_component_timeouts(const cluster_options& options) -> protostellar::component_timeouts
{
  protostellar::component_timeouts timeouts{};
  timeouts.key_value = options.key_value_timeout;
  timeouts.key_value_durable = options.key_value_durable_timeout;
  ...
}
```

**The per-attempt budget.** The unbounded-loop half was fixed on #1043, so the loop is bounded now — but it was bounded by `key_value_timeout`, and the durable default was not merely applied at the wrong level. `dispatch_protostellar` assigns `request->timeout` from the remaining budget on *every* attempt including the first, so `component::execute` is always handed an engaged optional and its durable branch never selects. Every KV operation from the public API arrives that way, so on the production path the durable default reached nothing at all.

Fixed as suggested — the loop resolves through the same `resolve_kv_timeout` rather than a second copy of the rule, with a new `requested_durability` helper supplying the level for a request type that may not have one. It keys on `io::mcbp_traits::supports_durability_v`, the same trait the MCBP path uses, which covers exactly the nine types with a `durability_level` member; the eight overloads resolving durably in `component.cxx` are those nine minus `mutate_in`, which is not routed over couchbase2 — so the two answers cannot diverge.

Two of the six negative-control mutations cover these separately, and the new wiring case asserts a lower and an upper bound on the deadline the server observes, one for each of the two defects.

core/protostellar/component.hxx
avsej: Agreed, and this is where both defects on this PR were hiding — neither was reachable by any assertion on the function itself.

A localhost `KvService` now records `ServerContext::deadline()` for each call, which is the only observable that names which budget was resolved. The cluster is configured with `key_value_timeout = 400ms` and `key_value_durable_timeout = 3000ms`, both away from their stock defaults, so the value that comes back says not only which timeout was chosen but whether it came from the caller's configuration at all.

`a_durable_mutation_without_a_timeout_gets_the_configured_durable_budget` asserts two bounds, each catching a different defect: below 1500ms the operation was bounded by the standard KV budget, above 3000ms the durable budget came from the built-in default rather than the configured value.

Both missing boundary cases are added. `majority_and_persist_to_active` joins the unit case, and the read appears at two levels — `a_read_is_never_given_the_durable_budget` drives a real `get` and asserts the observed deadline is the standard budget, and a unit case asserts `requested_durability` answers `none` for a request type that has no `durability_level` member at all.

On the note about the MCBP path: the divergence is confirmed — `key_value_durable_timeout` is read on the classic side only by the transactions layer, and no MCBP KV path consults it.

The floor is now a decision rather than an omission, recorded at `resolve_kv_timeout`: no floor here. That floor exists on the MCBP path because it has no durable default, so a durable write inherits the 2.5s KV budget and the floor is what keeps a short timeout from making durability unachievable. Resolving `KvDurableTimeout` addresses the same problem by the mechanism the RFC specifies, and neither RFC 59 nor RFC 77 states a floor.

The sibling implementations were then read rather than assumed, and they support that: Java (`CoreProtostellarUtil.kvDurableTimeout`) and .NET (`Stellar/KeyValue/StellarCollection.cs`) both select the durable default for any level other than none and apply no floor. Go (`gocb/kvopmanager_ps.go`) is the outlier on both axes — it selects the durable default only for levels stronger than `MAJORITY`, and coerces a sub-1500ms explicit timeout up to that floor. That is a cross-SDK question rather than this PR's; a ticket is being filed.

The cases asserting an explicit sub-floor timeout survives are gocb's own rows for this rule, inverted, and enumerated across all three durable levels rather than sampled — a floor applied only to the persistence levels, which is the shape of the disagreement, would slip past a majority-only case. A mutation introduces the floor in Go's shape, so "no floor" can fail rather than merely being unimplemented.

That is reasoning from the four implementations and the RFCs. What a durable write actually does at a 100ms budget against a real gateway was not measured.

test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  component comp{ io, channel, credentials, 20000ms };
+
```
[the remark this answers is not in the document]
avsej: This does not hold at head, and had not by the time it was filed against a later push.

`component`'s constructor takes two arguments, `(asio::io_context&, component_config)` — there is no fourth parameter — and this test builds the config by named-field assignment:

```cpp
component_config config;
config.channel = channel;
config.timeouts = { 20000ms };
component comp{ io, config };
```

`config.timeouts = { 20000ms }` is brace-initialisation of `component_timeouts`, whose first member is `key_value`, so it binds and compiles. The suite builds and runs at this commit.

One real observation adjacent to it, since this commit adds a seventh member to that struct: the assignment supplies one initialiser for seven members, so everything after `key_value` — including the `key_value_durable` added here — takes the struct's default rather than the 20s this test asks for. That is harmless as the file stands, because no case in it issues a durable operation, and it predates this commit: the same line supplied one of six before. It is left alone rather than folded in, since the shape belongs to the commit that wrote it. It is worth knowing about before a durable case is ever added here, which would silently get 10s instead of 20s.

The same hazard in the production construction site was real and is fixed on this commit — see the top-level comment.

avsej:
Round addressed. All three findings were live in substance, and two of them were understated: the mechanism each named had changed since the review, and in both cases what the code actually did was worse than what was described.

**Finding 1 — the durable timeout and the retry loop.** The unbounded half has since been fixed on #1043, so the loop no longer retries without a bound. What remained was worse than "bounds each attempt but not the operation": `dispatch_protostellar` assigns `request->timeout` from the remaining budget on *every* attempt including the first, so by the time `component::execute` calls `resolve_kv_timeout` the optional is always engaged and the durable branch never selects. Every KV operation issued through the public API arrives that way, so the durable default reached nothing at all — and the budget it was overridden by came from `key_value_timeout`, the 2.5s read budget this commit exists to replace. On the production path the commit's stated behaviour did not happen.

Fixed as suggested: `execute_protostellar` resolves its deadline through the same `resolve_kv_timeout` rather than a second copy of the rule. A new `requested_durability` helper supplies the level for a request type that may not carry one, keyed on `io::mcbp_traits::supports_durability_v` — the same trait the MCBP path uses, so the two transports agree on which operations are durable. It covers exactly the nine request types with a `durability_level` member, and the eight overloads resolving durably in `component.cxx` are those nine minus `mutate_in`, which is not routed over couchbase2 at all, so the client-side and component-side answers cannot diverge.

**Finding 2 — the aggregate.** The stated mechanism does not hold: `key_value_durable` is declared *last*, after `management`, and every member carries a default member initialiser, so no following initialiser was shifted. The hazard named had already fired by the other route the same shape allows — the construction site passed **six** initialisers for a **seven**-member struct, so `key_value_durable` was never populated from the cluster options and took the struct's own default. The two properties called out are exactly what hid it: the members share a type, so the short list was not a compile error, and the value it fell back to (the RFC default, 10s) is plausible enough that no assertion on a stock configuration would notice. What it cost was the configured value — `key_value_durable_timeout` is settable through `timeout_options` and the connection string and is raised to 20s by the `wan_development` profile, and none of that reached this transport.

Now assigned per field, in one function that both the component construction and the retry loop read, so the loop applies the rule to the same values the component was built with rather than to a second mapping that could drift from it.

To be precise about scope: that is the production construction site. `component_timeouts` is also brace-initialised in `test/cng/protostellar/live_kv.cxx`, supplying one initialiser for seven members, so everything after `key_value` — including the member added here — takes the struct's default rather than the value that test asks for. Harmless as that file stands, since no case in it issues a durable operation, and it predates this commit: the same line supplied one of six before. Left alone rather than folded in, but worth knowing before a durable case is added there.

**Finding 3 — the test.** Agreed, and both defects above were in exactly that gap; neither was reachable by any assertion on the function. A localhost `KvService` now records `ServerContext::deadline()` for each call, which is the only observable that names which budget was resolved. The cluster is configured with `key_value_timeout = 400ms` and `key_value_durable_timeout = 3000ms`, both away from their stock defaults, so the value that comes back says not just which timeout was chosen but whether it came from the caller's configuration at all. The durable case asserts two bounds and each catches one defect: below 1500ms the operation was bounded by the standard budget, above 3000ms the durable budget came from the built-in default rather than the configured one.

Both missing boundary cases are added. `majority_and_persist_to_active` joins the unit case, and the read appears at two levels — one case drives a real `get` and asserts the observed deadline is the standard budget, and a unit case asserts `requested_durability` answers `none` for a request type with no `durability_level` member. Seven negative-control mutations kill these separately, including the opposite regression of widening *every* operation to the durable budget.

Coverage was widened past the findings in two further places, both prompted by reading the other SDKs:

- **The durability mapping had no negative control at all**, and `majority_and_persist_to_active` was the one level `kv_converter` never asserted. The proto enum has no `NONE` member and starts at `MAJORITY = 0`, so the three durable levels occupy 0, 1, 2 in declaration order — a level shifted by one is a value the gateway *accepts*, so the write succeeds and the caller silently gets weaker or stronger guarantees than were asked for. That is now its own case with a mutation mapping a level onto its neighbour, which kills it.
- **No live case issued a durable mutation at all.** One now exercises all three levels against a real gateway. Its limit is worth stating plainly: it cannot pin *which* budget was resolved, since a gateway answers a durable write identically either way — that stays in-process. What it does catch is a level the gateway refuses, which no unit test can. That was confirmed rather than assumed: sending an out-of-range level makes it fail with `invalid_argument`, while the neighbour-mapping mutation above passes it, which is why both venues are needed.

**The note — divergence and the floor.** The divergence is confirmed: `key_value_durable_timeout` is read on the classic side only by the transactions layer, and no MCBP KV path consults it.

The floor is now a decision rather than an omission, recorded at `resolve_kv_timeout`: no floor here. That floor exists on the MCBP path because it has no durable default at all — a durable write there inherits the 2.5s KV budget, so the floor is what keeps a short timeout from making durability unachievable. Resolving `KvDurableTimeout` addresses the same problem by the mechanism the RFC specifies, and neither RFC 59 nor RFC 77 states a floor; adding one would silently raise a timeout the caller named.

The other implementations of this protocol were checked rather than assumed, and they mostly agree with that:

| SDK | levels selecting the durable default | floor on an explicit timeout |
|---|---|---|
| Java, `CoreProtostellarUtil.kvDurableTimeout` | any level other than none | none |
| .NET, `Stellar/KeyValue/StellarCollection.cs` | any level other than none | none |
| Go, `gocb/kvopmanager_ps.go` | only levels **stronger than** `MAJORITY` | **1500ms, coerced up** |
| this commit | any level other than none | none |

So Go is the outlier on both axes, and a plain `majority` write gets 2.5s there against 10s everywhere else. Go is self-consistent about it — its classic path draws the same line — which suggests a deliberate reading that plain `majority` waits only on replica acknowledgement while the stronger levels wait on disk. RFC 59 defines KVDurableTimeout as covering "synchronous durability", which includes plain `majority`, so this is worth settling across SDKs rather than per SDK; a ticket is being filed and no C++ change is proposed for it here.

Both properties are asserted and both have negative controls, including one that introduces the floor in the shape Go uses — so "no floor" is falsifiable rather than merely unimplemented. The rows asserting a sub-floor explicit timeout survives at each durability level are taken from gocb's own table for this rule, inverted: it expects coercion at every level, this expects none.

That is reasoning from the four implementations and the RFCs. What a durable write actually does at a 100ms budget against a real gateway was not measured.

One thing changed beyond the findings: a comment above `class component` duplicated the `component_timeouts` doc comment and no longer described anything beneath it, this commit having inserted `resolve_kv_timeout` between the two so that it read as that function's documentation. Removed.

The commit message has been corrected as well: it claimed the field was populated in `open_protostellar`, which is the part that was not happening.

thejcfactor (approved):

PR #1047, opened 2026-07-28
CXXCBC-905: compress KV values over couchbase2

avsej:
The couchbase2 KV path neither requested nor produced snappy-compressed content and read only the
uncompressed field, so a compressed response would have decoded to an empty value (silent data
loss) and large values were always sent raw. RFC 77 makes compression default-on, matching the
Java/Go/.NET SDKs.

Thread the cluster compression settings into the converter: send COMPRESSION_ENABLED_OPTIONAL on
every read that can carry a document body unless the user disabled compression, snappy-compress
Upsert/Insert/Replace when the value is at least min_size and shrinks to at most min_ratio of its
original, and snappy-decode a content_compressed response. COMPRESSION_ENABLED_ALWAYS is never
sent.

A payload that does not decode, or that declares more than the transport's maximum message size,
fails the operation rather than yielding a value. Empty bytes with a valid CAS are
indistinguishable from an empty document, and sizing an allocation from an unvalidated length
header is a denial of service the process cannot survive.

---
Stacked PR **25/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1046 — merge the series bottom-up.

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -1010,8 +1010,10 @@ public:
       const std::scoped_lock lock(protostellar_mutex_);
       protostellar_ = std::make_shared<protostellar::component>(
         ctx_,
-        protostellar::component_config{
-          std::move(channel), origin_.credentials(), make_component_timeouts(options) });
+        protostellar::component_config{ std::move(channel),
+                                        origin_.credentials(),
+                                        make_component_timeouts(options),
+                                        make_compression_settings(options) });
     }
     CB_LOG_INFO(R"(open couchbase2 cluster, id: "{}", endpoint: "{}")", id_, endpoint);
     return handler({});
--- a/core/protostellar/component.hxx
+++ b/core/protostellar/component.hxx
@@ -75,6 +75,7 @@
 #include "core/operations/management/search_index_get_all.hxx"
 #include "core/operations/management/search_index_get_documents_count.hxx"
 #include "core/operations/management/search_index_upsert.hxx"
+#include "core/protostellar/compression_settings.hxx"
 #include "core/protostellar/dispatcher.hxx"
 #include "core/timeout_defaults.hxx"
 #include "core/utils/movable_function.hxx"
@@ -111,6 +112,7 @@ struct component_config {
   std::shared_ptr<grpc::Channel> channel{};
   cluster_credentials credentials{};
   component_timeouts timeouts{};
+  kv::compression_settings compression{};
 };
 
 // The core KV request types the component can execute over couchbase2. cluster_impl routes only
@@ -374,6 +376,7 @@ private:
   std::unique_ptr<stubs> stubs_;
   std::string authorization_;
   component_timeouts timeouts_;
+  kv::compression_settings compression_;
   // Declared last, so it is destroyed first: ~dispatcher() cancels and drains the calls issued
   // through the stubs above, which must therefore still be alive while it runs.
   dispatcher dispatcher_;
--- a/core/protostellar/kv_converter.cxx
+++ b/core/protostellar/kv_converter.cxx
@@ -1,0 +1,70 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// snappy compression for the couchbase2 KV converter, kept out of kv_converter.hxx so the header
+// (included by white-box tests) does not leak a dependency on <snappy.h> / the snappy library.
+
+#include "core/protostellar/kv_converter.hxx"
+
+#include "core/protostellar/dispatcher.hxx"
+
+#include <snappy.h>
+
+#include <cstddef>
+
+namespace couchbase::core::protostellar::kv
+{
+auto
+maybe_compress(const std::vector<std::byte>& value, const compression_settings& compression)
+  -> std::optional<std::string>
+{
+  if (!compression.enabled || value.size() < compression.min_size) {
+    return std::nullopt;
+  }
+  const auto raw = utils::to_string(value);
+  std::string compressed;
+  snappy::Compress(raw.data(), raw.size(), &compressed);
+  // Only worth sending compressed when it shrinks to at most min_ratio of the original; a value
+  // landing exactly on the ratio is sent compressed, as it is in the other SDKs.
+  if (static_cast<double>(compressed.size()) <=
+      static_cast<double>(raw.size()) * compression.min_ratio) {
+    return compressed;
+  }
+  return std::nullopt;
+}
+
+auto
+snappy_uncompress(const std::string& compressed) -> std::optional<std::string>
+{
+  // snappy::Uncompress resizes its output to the length declared in the frame header before it
+  // validates the body, so a few bytes can ask for a multi-gigabyte allocation. Read the declared
+  // length first and refuse anything the transport could not have carried: the resulting
+  // std::bad_alloc would be thrown inside a gRPC completion and take the process with it.
+  std::size_t length{};
+  if (!snappy::GetUncompressedLength(compressed.data(), compressed.size(), &length)) {
+    return std::nullopt;
+  }
+  if (length > max_receive_message_size) {
+    return std::nullopt;
+  }
+  std::string uncompressed;
+  if (snappy::Uncompress(compressed.data(), compressed.size(), &uncompressed)) {
+    return uncompressed;
+  }
+  return std::nullopt;
+}
+} // namespace couchbase::core::protostellar::kv
--- a/core/protostellar/kv_converter.hxx
+++ b/core/protostellar/kv_converter.hxx
@@ -24,6 +24,7 @@
 
 #include "core/document_id.hxx"
 #include "core/error_context/key_value.hxx"
+#include "core/logger/logger.hxx"
 #include "core/operations/document_append.hxx"
 #include "core/operations/document_decrement.hxx"
 #include "core/operations/document_exists.hxx"
@@ -39,6 +40,7 @@
 #include "core/operations/document_touch.hxx"
 #include "core/operations/document_unlock.hxx"
 #include "core/operations/document_upsert.hxx"
+#include "core/protostellar/compression_settings.hxx"
 #include "core/utils/binary.hxx"
 
 #include <couchbase/cas.hxx>
@@ -139,11 +141,9 @@ to_core_token(const v1::MutationToken& token) -> couchbase::mutation_token
 }
 
 // Every response that carries a document body puts it in a content oneof, and the generated
-// accessor for the inactive arm returns the empty-string singleton. Reading content_uncompressed()
-// while the compressed arm is active would therefore produce an empty value with a valid CAS and
-// no error -- indistinguishable from an empty document. Compression negotiation lands in
-// CXXCBC-905; until then every body-bearing decoder refuses the compressed arm instead of losing
-// the value, which is the same fail-closed rule the request side applies.
+// accessor for the inactive arm returns the empty-string singleton. The active arm therefore has to
+// be tested before reading either one, or a compressed payload decodes to an empty value with a
+// valid CAS -- indistinguishable from an empty document.
 template<typename Proto>
 [[nodiscard]] auto
 content_is_compressed(const Proto& proto) -> bool
@@ -151,13 +151,77 @@ content_is_compressed(const Proto& proto) -> bool
   return proto.content_case() == Proto::kContentCompressed;
 }
 
+// ── Compression (RFC 77) ───────────────────────────────────────────────────────
+
+// snappy lives in kv_converter.cxx (compiled into the client library) so <snappy.h> stays out of
+// this header, which white-box tests include — they then need neither snappy's headers nor its
+// symbols. maybe_compress returns the compressed bytes when compression is enabled and the value
+// clears the size/ratio threshold, else std::nullopt ("store uncompressed"); snappy_uncompress
+// returns std::nullopt on malformed input and on a frame declaring more than the transport's
+// maximum message size, which no legitimate value can exceed.
+[[nodiscard]] auto
+maybe_compress(const std::vector<std::byte>& value, const compression_settings& compression)
+  -> std::optional<std::string>;
+[[nodiscard]] auto
+snappy_uncompress(const std::string& compressed) -> std::optional<std::string>;
+
+// Signals the gateway that it may return snappy-compressed content (RFC 77: send
+// COMPRESSION_ENABLED_OPTIONAL when enabled; leave the field unset when the user disabled
+// compression; never COMPRESSION_ENABLED_ALWAYS).
+template<typename Proto>
+inline void
+set_read_compression(Proto& proto, const compression_settings& compression)
+{
+  if (compression.enabled) {
+    proto.set_compression(v1::COMPRESSION_ENABLED_OPTIONAL);
+  }
+}
+
+// Fills a write request's content oneof, snappy-compressing when maybe_compress deems it
+// worthwhile.
+template<typename Proto>
+inline void
+set_write_content(Proto& proto,
+                  const std::vector<std::byte>& value,
+                  const compression_settings& compression)
+{
+  if (auto compressed = maybe_compress(value, compression)) {
+    proto.set_content_compressed(std::move(*compressed));
+  } else {
+    proto.set_content_uncompressed(utils::to_string(value));
+  }
+}
+
+// Reads a response's content oneof, snappy-decoding a compressed payload. Returns nullopt when the
+// compressed arm does not decode: handing back empty bytes instead would produce a value that no
+// caller can distinguish from an empty document, which is the same silent data-loss trap that made
+// the unguarded oneof read wrong before compression landed. A gateway emitting an undecodable
+// snappy frame is a server-side fault, so the callers surface it as internal_server_failure rather
+// than inventing a value.
+template<typename Proto>
+inline auto
+decode_content(const Proto& proto) -> std::optional<std::vector<std::byte>>
+{
+  if (content_is_compressed(proto)) {
+    if (auto uncompressed = snappy_uncompress(proto.content_compressed())) {
+      return utils::to_binary(*uncompressed);
+    }
+    CB_LOG_DEBUG("failed to decompress snappy payload in response (compressed_size={})",
+                 proto.content_compressed().size());
+    return std::nullopt;
+  }
+  return utils::to_binary(proto.content_uncompressed());
+}
+
 // ── Get ───────────────────────────────────────────────────────────────────────
 
 inline auto
-encode(const operations::get_request& request) -> v1::GetRequest
+encode(const operations::get_request& request, const compression_settings& compression)
+  -> v1::GetRequest
 {
   v1::GetRequest proto;
   set_location(proto, request.id);
+  set_read_compression(proto, compression);
   return proto;
 }
 
@@ -166,21 +230,24 @@ decode(const v1::GetResponse& proto, key_value_error_context ctx) -> operations:
 {
   operations::get_response response;
   response.ctx = std::move(ctx);
-  if (content_is_compressed(proto)) {
-    response.ctx.override_ec(errc::common::feature_not_available);
+  auto value = decode_content(proto);
+  if (!value) {
+    response.ctx.override_ec(errc::common::internal_server_failure);
     return response;
   }
-  response.value = utils::to_binary(proto.content_uncompressed());
+  response.value = std::move(*value);
   response.cas = couchbase::cas{ proto.cas() };
   response.flags = proto.content_flags();
   return response;
 }
 
 inline auto
-encode(const operations::get_projected_request& request) -> v1::GetRequest
+encode(const operations::get_projected_request& request, const compression_settings& compression)
+  -> v1::GetRequest
 {
   v1::GetRequest proto;
   set_location(proto, request.id);
+  set_read_compression(proto, compression);
   for (const auto& path : request.projections) {
     proto.add_project(path);
   }
@@ -194,11 +261,12 @@ decode(const v1::GetResponse& proto,
 {
   operations::get_projected_response response;
   response.ctx = std::move(ctx);
-  if (content_is_compressed(proto)) {
-    response.ctx.override_ec(errc::common::feature_not_available);
+  auto value = decode_content(proto);
+  if (!value) {
+    response.ctx.override_ec(errc::common::internal_server_failure);
     return response;
   }
-  response.value = utils::to_binary(proto.content_uncompressed());
+  response.value = std::move(*value);
   response.cas = couchbase::cas{ proto.cas() };
   response.flags = proto.content_flags();
   if (proto.has_expiry()) {
@@ -223,11 +291,12 @@ decode_mutation(const Proto& proto, key_value_error_context ctx) -> Response
 }
 
 inline auto
-encode(const operations::upsert_request& request) -> v1::UpsertRequest
+encode(const operations::upsert_request& request, const compression_settings& compression)
+  -> v1::UpsertRequest
 {
   v1::UpsertRequest proto;
   set_location(proto, request.id);
-  proto.set_content_uncompressed(utils::to_string(request.value));
+  set_write_content(proto, request.value, compression);
   proto.set_content_flags(request.flags);
   // preserve_expiry is expressed by omitting the oneof, NOT through the proto's
   // preserve_expiry_on_existing flag. The gateway rejects that flag outright whenever the oneof is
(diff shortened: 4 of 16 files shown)


test/cng/protostellar/live_kv.cxx
```diff
@@ -0,0 +1,308 @@
...
+  credentials.password = env_or("TEST_CB2_PASSWORD", "password");
+
+  const auto bucket = env_or("TEST_CB2_BUCKET", "default");
+  const auto id = document_id{ bucket, "_default", "_default", "cng-live-kv-key" };
+  const std::string document = "{\"cng\":true}";
+
+  asio::io_context io;
+  auto work = asio::make_work_guard(io);
+  auto channel = ps::make_channel(endpoint, options, credentials);
+  component comp{ io, channel, credentials, 20000ms };
```
avsej: **Build blocker.** The fifth required parameter `kv::compression_settings compression` is added to `component::component`, and this call site (also `:183`) still passes four arguments. With `CB_COUCHBASE2=1`, `cng_live_kv_test` fails to compile:

```
test/cng/protostellar/live_kv.cxx:100:53: error: no matching function for call to
  'component::component(<brace-enclosed initializer list>)'
core/protostellar/component.hxx:164:3: note: candidate expects 5 arguments, 4 provided
```

For accuracy: this is the third commit to break the same untouched call site. It was written in #1032/#1033; #1036 added a fifth parameter, #1038 replaced the timeouts with the `component_timeouts` aggregate so `20000ms` stopped converting even at arity four, and this commit adds a fifth again. The `build` check passes on #1034/#1035 and fails from #1036 upward on every platform leg — twelve consecutive pull requests.

Immediate fix at both sites (C++17, so no designated initialisers):

```suggestion
  component comp{ io, channel, credentials, component_timeouts{ 20000ms, 20000ms }, {} };
```

The durable fix is on the constructor: five positional parameters with no defaults make every addition a silent breaking change, and the first call site to break is in a test only a feature-gated build compiles. A `component_config` struct with named fields removes the class of failure.

core/protostellar/kv_converter.hxx
```diff
@@ -0,0 +1,561 @@
...
+  }
+}
+
+// Reads a response's content oneof, snappy-decoding a compressed payload. A valid compressed
+// payload always decodes back to its original bytes (never an empty-but-successful value -- the
+// pre-fix silent-data-loss trap); only malformed snappy from the trusted gateway falls back to
+// empty.
+template<typename Proto>
+inline auto
+decode_content(const Proto& proto) -> std::vector<std::byte>
```
avsej: **Blocker: a decode failure becomes an empty document with a success status.**

When the compressed arm is active and `snappy::Uncompress` fails, this returns an empty vector, the error context is untouched, and the caller sees success with a valid CAS, correct flags and no value. That is the silent-data-loss shape this commit exists to remove, reintroduced on the error path.

Read-modify-write scenario: `get` returns an empty value with success on a corrupted payload, the application mutates that nothing, and `replace` succeeds because the CAS is still valid — the original document is gone, with no error at any point.

Make the failure representable:

```cpp
template<typename Proto>
inline auto
decode_content(const Proto& proto) -> std::optional<std::vector<std::byte>>
{
  if (proto.has_content_compressed()) {
    auto uncompressed = snappy_uncompress(proto.content_compressed());
    if (!uncompressed) {
      return std::nullopt;
    }
    return to_bytes(*uncompressed);
  }
  return to_bytes(proto.content_uncompressed());
}
```

and have each of the three `decode` overloads set `ctx.ec` (`errc::network::protocol_error`) on `std::nullopt`. This is the convention `can_encode` already applies on the request side.

Related: `get_projected` is the one read path not converted here — it neither calls the read-compression helper nor uses `decode_content`. Since `GetRequest.compression` is `optional` with `COMPRESSION_ENABLED_OPTIONAL` as its zero value, a gateway using the plain value getter may compress that response anyway. gocb sets `Project` and `Compression` on the same request and decompresses unconditionally (`kvprovider_ps.go:491-497`).

core/protostellar/kv_converter.cxx
```diff
@@ -0,0 +1,55 @@
...
+    return compressed;
+  }
+  return std::nullopt;
+}
+
+auto
+snappy_uncompress(const std::string& compressed) -> std::optional<std::string>
+{
+  std::string uncompressed;
+  if (snappy::Uncompress(compressed.data(), compressed.size(), &uncompressed)) {
```
avsej, line 65: `snappy::Uncompress` resizes its output to the length varint in the payload header **before** validating the body, so a few bytes can declare a multi-gigabyte uncompressed length and cause an allocation of that size. The resulting `std::bad_alloc` is thrown inside an Asio completion handler and terminates the process rather than failing the operation.

The threat model — a buggy or compromised gateway — is one a client library should defend against. `snappy::GetUncompressedLength` reads the header without allocating:

```cpp
  std::size_t length{};
  if (!snappy::GetUncompressedLength(compressed.data(), compressed.size(), &length)) {
    return std::nullopt;
  }
  // The channel already caps messages at 25 MiB (RFC 77); an uncompressed value cannot
  // legitimately exceed the maximum document size either.
  if (length > max_uncompressed_value_size) {
    return std::nullopt;
  }
```

avsej (commented):
## Review — PR #1047 (`CXXCBC-905`)

This commit adds Snappy compression for KV values over couchbase2: `COMPRESSION_ENABLED_OPTIONAL` on the read
requests, `content_compressed` on the write requests, and a `decode_content` helper that handles the response
oneof. The thresholds match the reference implementations exactly — 32 bytes and a 0.83 ratio, the same values as
couchbase-jvm-clients `CompressionConfig.DEFAULT_MIN_SIZE`/`DEFAULT_MIN_RATIO` and gocb's compressor — and
`COMPRESSION_ENABLED_ALWAYS` is correctly never sent. RFC 77's Compression section is the part of the
specification all three shipping SDKs agree on, and this implementation lands on the same side of it.

Two things block it: the commit does not compile, and the error path it adds destroys data silently.

### Strengths

- The read/write split matches RFC 77 exactly: `COMPRESSION_ENABLED_OPTIONAL` on `GetRequest`,
  `GetAndLockRequest` and `GetAndTouchRequest`, `content_compressed` on `InsertRequest`, `UpsertRequest` and
  `ReplaceRequest`, and the field left entirely unset when the user has disabled compression.
- Both the size and ratio thresholds are honoured, with the same defaults as couchbase-jvm-clients and gocb.
  Honouring only `min_size` is a common shortcut and it was not taken here.
- `decode_content` centralises the response-side oneof handling, which is what makes the remaining gap a
  one-place fix rather than a per-operation one.

### Findings

|---|---|---|---|---|
| 1 | blocker | build | A fifth required constructor parameter is added without updating two existing call sites | `test/cng/protostellar/live_kv.cxx:100` |
| 2 | blocker | correctness | A malformed compressed payload decodes to an empty document with a success status | `core/protostellar/kv_converter.hxx:191` |
| 3 | high | security | Decompression is unbounded: the declared length is allocated before the payload is validated | `core/protostellar/kv_converter.cxx:50` |
| 4 | medium | correctness | `get_projected` neither negotiates compression nor decodes it | `core/protostellar/kv_converter.hxx:191` |

#### 1. A fifth required constructor parameter without updating two call sites (blocker)

```cpp
component comp{ io, channel, credentials, 20000ms };
```

`component::component` gains a fifth required parameter, `kv::compression_settings compression`, and
`test/cng/protostellar/live_kv.cxx:100` and `:183` still pass four arguments. With `CB_COUCHBASE2=1` — which is
what CI uses — `cng_live_kv_test` fails to compile:

```
test/cng/protostellar/live_kv.cxx:100:53: error: no matching function for call to
  'couchbase::core::protostellar::component::component(<brace-enclosed initializer list>)'
core/protostellar/component.hxx:164:3: note: candidate expects 5 arguments, 4 provided
```

Worth being precise about the history, because it is not solely this commit's regression. That call site was
written in #1032/#1033 and has never been touched, while the constructor changed under it three times: #1036
added a fifth parameter, #1038 replaced the timeout parameters with the `component_timeouts` aggregate — at
which point `20000ms` stopped converting even though the arity was back to four — and this commit adds a fifth
parameter again. The `build` check passes on #1034 and #1035 and fails on #1036 and every pull request above it,
on every platform leg. So the series has been red for twelve consecutive pull requests, and this commit is the
third cause rather than the first.

The immediate fix is at both call sites (C++17, so no designated initialisers):

```cpp
component comp{ io, channel, credentials, component_timeouts{ 20000ms, 20000ms }, {} };
```

The design fix matters more. A five-parameter positional constructor with no defaults makes every added
parameter a silent breaking change to every call site, and the one call site that breaks first is in a test that
only a feature-gated build compiles. A configuration struct removes the whole class:

```cpp
struct component_config {
  std::shared_ptr<grpc::Channel> channel;
  cluster_credentials credentials;
  component_timeouts timeouts{};
  kv::compression_settings compression{};
};

component(asio::io_context& io, component_config config);
```

(inline comment on `test/cng/protostellar/live_kv.cxx:100`)

#### 2. A malformed compressed payload decodes to an empty document with success (blocker)

```cpp
decode_content(const Proto& proto) -> std::vector<std::byte>
```

When the compressed arm of the oneof is active and `snappy::Uncompress` fails, the helper returns an empty
vector and the caller has no way to distinguish that from a genuinely empty document. The error context is not
touched, so the response reports success with a valid CAS, correct flags, and no value.

This is the same failure mode the commit exists to remove — the header comment even describes the previous
behaviour as a silent-data-loss trap — reintroduced on the error path rather than the unsupported-feature path.

Failure scenario, read-modify-write: an application does `get`, mutates the returned document, and `replace`s it
with the original CAS. A corrupted or truncated compressed payload from the gateway yields an empty value with
success, the application applies its mutation to nothing, and the `replace` succeeds because the CAS is still
valid. The stored document is now whatever the mutation produces from an empty input, and the original content is
gone. No error is reported at any point.

The helper needs to be able to fail:

```cpp
template<typename Proto>
inline auto
decode_content(const Proto& proto) -> std::optional<std::vector<std::byte>>
{
  if (proto.has_content_compressed()) {
    auto uncompressed = snappy_uncompress(proto.content_compressed());
    if (!uncompressed) {
      return std::nullopt;      // caller sets ctx.ec
    }
    return to_bytes(*uncompressed);
  }
  return to_bytes(proto.content_uncompressed());
}
```

with each of the three `decode` overloads setting the context error — `errc::network::protocol_error` or
`errc::common::internal_server_failure` — when it gets `std::nullopt`. Failing closed here matches the
convention `can_encode` already applies on the request side.

(inline comment on `core/protostellar/kv_converter.hxx:191`)

#### 3. Decompression is unbounded (high)

```cpp
if (snappy::Uncompress(compressed.data(), compressed.size(), &uncompressed)) {
```

`snappy::Uncompress` resizes its output to the length varint in the payload header before validating the body, so
a small response can declare a very large uncompressed length and cause an allocation of that size. A handful of
bytes can request several gigabytes. The resulting `std::bad_alloc` is thrown inside an Asio completion handler,
which is a `noexcept` context in this codebase's handler chain, so the process terminates rather than the
operation failing.

The threat model here is a buggy or compromised gateway, which is worth defending against for a client library.
`snappy::GetUncompressedLength` reads the header without allocating, so the bound can be checked first:

```cpp
auto
snappy_uncompress(const std::string& compressed) -> std::optional<std::string>
{
  std::size_t length{};
  if (!snappy::GetUncompressedLength(compressed.data(), compressed.size(), &length)) {
    return std::nullopt;
  }
  // The channel already refuses messages above 25 MiB (RFC 77); an uncompressed value cannot
  // legitimately exceed the maximum document size either.
  if (length > max_uncompressed_value_size) {
    return std::nullopt;
  }
  std::string uncompressed;
  if (!snappy::Uncompress(compressed.data(), compressed.size(), &uncompressed)) {
    return std::nullopt;
  }
  return uncompressed;
}
```

(inline comment on `core/protostellar/kv_converter.cxx:50`)

#### 4. `get_projected` neither negotiates compression nor decodes it (medium)

`encode(get_projected_request)` does not call the read-compression helper that `encode(get_request)` uses, and
the corresponding decode still reads `content_uncompressed` directly rather than going through `decode_content`.
`get_projected` is the one read path in the file that this commit did not convert.

That matters because `GetRequest.compression` is `optional CompressionEnabled` whose zero value is
`COMPRESSION_ENABLED_OPTIONAL`, so a gateway reading it with the plain value getter rather than a presence check
sees `OPTIONAL` for an unset field and may return `content_compressed` anyway. The projected read then produces an
empty value with a success status — finding 2, on a path that has not even opted in. gocb sets `Project` and
`Compression` on the same `GetRequest` and decompresses unconditionally (`kvprovider_ps.go:491-497`);
couchbase-jvm-clients routes every Get response through the same `convertContent` (`CoreProtostellarKeyValueResponses.java:111-123`).

### Verdict

Blocked. The commit does not compile, and finding 2 turns a transport-level decode failure into silent data loss
on a read-modify-write cycle — which is the specific outcome this commit's own comment says it exists to remove.

test/cng/protostellar/live_kv.cxx
avsej: This does not hold against the current head. The constructor is already `component(asio::io_context&, component_config)`, `component_config` already carries `compression` as a named field, and both call sites already assign by name rather than positionally:

```cpp
component_config config;
config.channel = channel;
config.timeouts = { 20000ms };
component comp{ io, config };
```

So the design fix asked for here is the one in place. No change was made.

The class of failure it removes did fire again, though, one level up: the couchbase2 cluster built `component_config` with three initialisers for its four fields, so `compression` silently took the struct's own default and `cluster_options.enable_compression` never reached the transport. That is fixed by a named helper beside `make_component_timeouts`.

core/protostellar/kv_converter.hxx
avsej: Already addressed at the current head, in the shape proposed: `decode_content` returns `std::optional<std::vector<std::byte>>` and each decoder reports `errc::common::internal_server_failure` on `std::nullopt` rather than handing back a value. The read-modify-write framing is now the stated reason in the code comment and the commit message, and a mutation returning an empty vector instead of `std::nullopt` is killed by `malformed_compressed_content_fails_closed`.

The related point is taken and `get_projected` now negotiates compression and decodes through `decode_content` like every other read, with the `feature_not_available` refusal removed from both the converter and the component.

The mechanism given for it does not hold, though. The reference gateway takes `compression` as a pointer and branches on presence, not on its value:

```go
	if compressMode != nil {
		switch *compressMode {
		case kv_v1.CompressionEnabled_COMPRESSION_ENABLED_OPTIONAL:
```

With the field unset it takes the `else` branch and uncompresses before replying, so an unset field was not read as `OPTIONAL`. What justifies the change is the citation: projections and compression travel on one `GetRequest`, and gocb, the Java SDK and the .NET SDK all set them together. Refusing a compressed answer outright made this the one read the client could not complete.

core/protostellar/kv_converter.cxx
avsej: Fixed as suggested. `snappy_uncompress` now reads the declared length with `snappy::GetUncompressedLength`, which does not allocate, and refuses anything above the transport's maximum message size before inflating.

The bound is the 25 MiB receive limit RFC 77 requires. It was a bare literal in `default_channel_arguments()` and is now the named `max_receive_message_size` that both sites share: a second copy of the number would drift from the limit it is meant to express.

`an_oversized_declared_length_is_refused` covers it, built from a *valid* frame just over the cap rather than a crafted header — a malformed header is refused with or without the bound, so that assertion would have passed for the wrong reason. The matching mutation raises the bound out of reach and is killed.

avsej:
## Response — PR #1047 (`CXXCBC-905`)

Every point is answered below. Two of the four findings were already addressed at the head the review names and are reported as such rather than claimed as new work; the other two are fixed here, together with two defects found while checking them.

### 1. A fifth required constructor parameter without updating two call sites

> `component::component` gains a fifth required parameter, `kv::compression_settings compression`, and `test/cng/protostellar/live_kv.cxx:100` and `:183` still pass four arguments.

> The design fix matters more. A five-parameter positional constructor with no defaults makes every added parameter a silent breaking change to every call site […] A configuration struct removes the whole class

This does not hold against the current head. The constructor is already `component(asio::io_context&, component_config)`, `component_config` already carries `compression` as a named field, and both call sites already assign by name:

```cpp
component_config config;
config.channel = channel;
config.timeouts = { 20000ms };
component comp{ io, config };
```

The suggested design is the one in place, arrived at for the reason given. No change was needed.

### 2. A malformed compressed payload decodes to an empty document with success

> When the compressed arm of the oneof is active and `snappy::Uncompress` fails, the helper returns an empty vector and the caller has no way to distinguish that from a genuinely empty document.

> The helper needs to be able to fail

Also already addressed at the current head, in the shape proposed: `decode_content` returns `std::optional<std::vector<std::byte>>`, and each decoder reports `errc::common::internal_server_failure` on `std::nullopt` rather than handing back a value.

The read-modify-write scenario is the right way to frame the severity, and it is now the stated reason in both the code comment and the commit message. A negative control was added for it: a mutation returning an empty vector instead of `std::nullopt` is killed by `malformed_compressed_content_fails_closed`.

### 3. Decompression is unbounded

> `snappy::Uncompress` resizes its output to the length varint in the payload header before validating the body, so a small response can declare a very large uncompressed length and cause an allocation of that size.

Fixed. `snappy_uncompress` now reads the declared length with `snappy::GetUncompressedLength`, which does not allocate, and refuses anything above the transport's maximum message size before inflating:

```cpp
std::size_t length{};
if (!snappy::GetUncompressedLength(compressed.data(), compressed.size(), &length)) {
  return std::nullopt;
}
if (length > max_receive_message_size) {
  return std::nullopt;
}
```

The consequence given for the throw was checked rather than taken on trust: the io thread is `std::thread{ [&io] { io.run(); } }` with no catch, so an exception escaping a completion escapes the thread function and terminates the process. That is what the code comment now states.

The bound is the 25 MiB receive limit RFC 77 requires, which was a bare literal in `default_channel_arguments()` and is now the named `max_receive_message_size` both sites share — a second copy of the number would drift from the limit it is supposed to mean.

The test is `an_oversized_declared_length_is_refused`, and it is deliberately built from a *valid* frame just over the cap rather than a crafted header: a malformed header is refused with or without the bound, so the assertion would pass for the wrong reason. The corresponding mutation raises the bound out of reach and is killed.

### 4. `get_projected` neither negotiates compression nor decodes it

> `encode(get_projected_request)` does not call the read-compression helper that `encode(get_request)` uses, and the corresponding decode still reads `content_uncompressed` directly rather than going through `decode_content`.

Fixed: the projected read now negotiates compression and decodes through `decode_content` like every other read, and the `feature_not_available` refusal is gone from both the converter and the component.

The stated mechanism does not hold, though, and the reason to make the change is a different one.

> `GetRequest.compression` is `optional CompressionEnabled` whose zero value is `COMPRESSION_ENABLED_OPTIONAL`, so a gateway reading it with the plain value getter rather than a presence check sees `OPTIONAL` for an unset field and may return `content_compressed` anyway.

The reference gateway takes the field as a pointer and branches on presence, so an unset field is not read as `OPTIONAL`:

```go
func (h CompressHandler) MaybeCompressContent(
	in []byte, datatype memdx.DatatypeFlag, compressMode *kv_v1.CompressionEnabled,
) (bool, []byte, *status.Status) {
	if compressMode != nil {
		switch *compressMode {
		case kv_v1.CompressionEnabled_COMPRESSION_ENABLED_OPTIONAL:
```

With the field unset it takes the `else` branch and *uncompresses* before replying. So the previous code was not at risk of the finding-2 outcome on this path.

What justifies the change is the citation: projections and compression travel on the same `GetRequest`, and all three shipping SDKs put them there together — gocb has one `Get` carrying both `Project` and `Compression` and decompresses unconditionally (`kvprovider_ps.go:466-500`); the Java SDK sets compression before it adds the projections, on the same builder (`CoreProtostellarKeyValueRequests.java:99-105`); the .NET SDK sets `CompressionEnabled.Optional` on its projected read too (`StellarCollection.cs:205-231`). Refusing a compressed answer with `feature_not_available` made this the one read the client could not complete, and the gateway serves projections from its ordinary `Get` handler, so nothing about the path makes it unable to carry one.

### Two defects found while checking the above

**The user's compression setting never reached the transport.** `component_config` has carried a `compression` field, but the couchbase2 cluster built one with three initialisers for a four-field aggregate, so the fourth silently took the struct's own default. `cluster_options.enable_compression` — which the connection string and `compression().enabled(false)` both feed — was ignored, and every couchbase2 cluster compressed regardless. It is now mapped by a named helper beside `make_component_timeouts`, for the same reason that one exists.

**The ratio threshold was exclusive.** A value compressing to exactly `min_ratio` of its original was sent raw. All three references accept it, each by rejecting only a ratio strictly above the threshold: the Java SDK in `MemcacheProtocol.tryCompression`, the .NET SDK in `OperationCompressor.Compress`, gocb in `compressor.go`. The comparison is now `<=`. The size threshold needed no change — the Java and .NET SDKs both compress at `>= minSize`, as this does; gocb is the lone outlier with `> minSize`.

### Coverage

The unit suite gained boundary and per-path rows rather than more of the same: a value of exactly `min_size`, a ratio landing exactly on `min_ratio` (derived from the fixture's own compressed size over a power-of-two original, so the comparison is exact in binary floating point and can tell the inclusive bound from the exclusive one), an incompressible value that must go out raw, compression opt-in and decode on all four read paths, and the oversized frame above. The component suite gained `get_projected_negotiates_and_decodes_compression`, which asserts what left the client and what it made of the compressed reply on one call. Nine mutations across the two suites, all killed.

`live_kv.cxx` gained `compressed_values_round_trip_against_live_gateway`, and it was checked for falsifiability rather than assumed: truncating the frame the client sends makes a live gateway reject the write with `invalid_argument`, which is the property no in-process test can establish — that what the client produces is a snappy frame the gateway accepts and stores faithfully.

Two limits of that case are worth stating rather than leaving implied. Against a live gateway a read never came back on the compressed arm, including a read of a document the client had just stored compressed; both halves of the read-side decoding were confirmed unexercised live by breaking `decode_content` and watching the case still pass. Read-side decompression is therefore covered against an in-process server only. And the mapping from `cluster_options` into `component_config` — the defect above — has no test: it is reachable only by opening a cluster, and which arm of the oneof went out is not observable through any API a test can reach. The same is true of `make_component_timeouts` beside it.

core/protostellar/component.hxx
```diff
@@ -111,6 +112,7 @@ struct component_config {
   std::shared_ptr<grpc::Channel> channel{};
   cluster_credentials credentials{};
   component_timeouts timeouts{};
+  kv::compression_settings compression{};
```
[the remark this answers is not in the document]
avsej: The enabled flag is threaded at the head this review ran against. `cluster.cxx` builds the config with four fields, the fourth from a named helper beside `make_component_timeouts`:

```cpp
protostellar::component_config{ std::move(channel),
                                origin_.credentials(),
                                make_component_timeouts(options),
                                make_compression_settings(options) };
```

That was the same defect this comment describes — three initialisers for a four-field aggregate, so `compression` silently took the struct's own default and `enabled(false)` was ignored — and it is fixed in this push.

The thresholds are a separate matter and are deliberately left on the converter's defaults. They are dropped a layer higher than this file: `core::cluster_options` carries `enable_compression` and nothing else, and `public_cluster.cxx` maps only `opts.compression.enabled` into it, so `min_size`/`min_ratio` never reach core on any transport. Carrying them would be a change to the core options struct that the classic MCBP path shares, which is outside this ticket. The defaults here are the same 32 bytes and 0.83 the Java, .NET and Go SDKs use, so a caller who does not override them sees identical behaviour either way.

core/protostellar/kv_converter.cxx
```diff
@@ -0,0 +1,55 @@
...
```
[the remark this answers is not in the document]
avsej: This is what the code at the head this review ran against already does — the declared length is parsed first, checked against the transport maximum, and only then decompressed:

```cpp
std::size_t length{};
if (!snappy::GetUncompressedLength(compressed.data(), compressed.size(), &length)) {
  return std::nullopt;
}
if (length > max_receive_message_size) {
  return std::nullopt;
}
std::string uncompressed;
if (snappy::Uncompress(compressed.data(), compressed.size(), &uncompressed)) {
```

`an_oversized_declared_length_is_refused` covers it, built from a valid frame just over the cap rather than a crafted header, since a malformed header is refused with or without the bound. No further change was made.

core/protostellar/component.hxx
```diff
@@ -76,6 +76,7 @@
 #include "core/operations/management/search_index_get_documents_count.hxx"
 #include "core/operations/management/search_index_upsert.hxx"
 #include "core/protostellar/dispatcher.hxx"
+#include "core/protostellar/kv_converter.hxx"
```
DemetrisChr: This will include the generated gRPC KV stubs, which violates the following which is mentioned below:
```
  // The generated gRPC stubs are held behind an opaque pointer so the generated protobuf and gRPC
  // surface stays out of every translation unit that includes this header -- none of it appears in
  // the interface, and the stub count grows with every service the transport learns to speak.
```

Perhaps we should move the `compression_settings` struct to this header file, to avoid the need for the `kv_converter.hxx` include?

core/protostellar/kv_converter.hxx
```diff
@@ -139,25 +139,94 @@ to_core_token(const v1::MutationToken& token) -> couchbase::mutation_token
...
+// than inventing a value.
+template<typename Proto>
+inline auto
+decode_content(const Proto& proto) -> std::optional<std::vector<std::byte>>
+{
+  if (content_is_compressed(proto)) {
+    if (auto uncompressed = snappy_uncompress(proto.content_compressed())) {
+      return utils::to_binary(*uncompressed);
+    }
+    return std::nullopt;
```
DemetrisChr, line 211: Should we log that there was an uncompression failure? Possibly at debug-level. Otherwise, the user only gets an `internal_server_failure` which might make it hard to debug if such an issue ever occurs.

core/cluster.cxx
```diff
@@ -343,6 +343,15 @@ make_component_timeouts(const cluster_options& options) -> protostellar::compone
   timeouts.management = options.management_timeout;
   return timeouts;
 }
+
+// Core carries only the on/off switch, so min_size and min_ratio keep the converter's defaults.
```
DemetrisChr: This is true, but we do have them in the Public API's `compression_options`, so this will silently ignore `min_size` and `min_ratio`. Although, that issue currently exists in classic as well.

Should we pass them on to the core now, or do you prefer to handle that at a later stage across both protostellar and classic?

core/protostellar/kv_converter.hxx
avsej: Added `CB_LOG_DEBUG("failed to decompress snappy payload in response (compressed_size={})", proto.content_compressed().size());` inside `decode_content` when `snappy_uncompress` returns `std::nullopt`.

core/cluster.cxx
avsej: Propagated `min_size` and `min_ratio` from `couchbase::compression_options` in `public_cluster.cxx` through `core::cluster_options` into `make_compression_settings`. Also decoupled `compression_settings` into `core/protostellar/compression_settings.hxx` so `component.hxx` stays isolated from Protobuf headers.

DemetrisChr (approved):

PR #1049, opened 2026-07-28
[WIP] CXXCBC-908: wait_until_ready over couchbase2 via gRPC channel readiness

avsej:
wait_until_ready had no couchbase2 branch: the generic MCBP ping-poll finds no
sessions on a couchbase2 cluster, so every call ran to the deadline and returned
unambiguous_timeout with no cause.

Branch core::impl::wait_until_ready on cluster::is_protostellar(). The couchbase2
path polls the gRPC channel connectivity state -- GetState(true) on a 50ms asio
timer until the channel reaches READY (success) or the timeout elapses
(unambiguous_timeout). No health RPC is sent; this mirrors Java and Go, and
GetState forces a connect on the lazily-dialed channel. WaitForConnected/
WaitForStateChange are avoided because they block, and the SDK runs a single
io_context thread. With one channel per cluster, online/degraded both wait for
READY, service_types and any bucket name have nothing to probe and are ignored,
and offline is rejected with invalid_argument as the classic path does.

The poll is a self-owning operation on the component; if the io_context is torn
down mid-wait it delivers cluster_closed rather than dropping the handler, so a
future-based caller never observes broken_promise -- matching the classic
operation's contract.

---
Stacked PR **27/30** in the CNG (`couchbase2://`) transport series. The change specific to this PR is its top commit; the commits beneath it belong to earlier PRs in the series. Depends on #1048 — merge the series bottom-up.

### DIFF
--- a/core/impl/wait_until_ready.cxx
+++ b/core/impl/wait_until_ready.cxx
@@ -337,6 +337,19 @@ wait_until_ready(core::cluster core,
                  std::set<service_type> services,
                  utils::movable_function<void(std::error_code)> handler)
 {
+  if (core.is_protostellar()) {
+    // couchbase2 exposes a single gRPC channel; readiness is that channel reaching a connected
+    // state, not a per-service/per-bucket ping. So service_types and bucket_name have nothing to
+    // probe and are ignored (mirrors Java/Go). offline is rejected exactly as the classic op does
+    // below, so behavior is uniform across transports; every other desired_state waits for READY.
+    if (desired_state == couchbase::cluster_state::offline) {
+      return asio::post(core.io_context(), [handler = std::move(handler)]() mutable {
+        handler(errc::common::invalid_argument);
+      });
+    }
+    return core.protostellar_wait_until_ready(timeout, std::move(handler));
+  }
+
   auto op = std::make_shared<wait_until_ready_operation>(std::move(core),
                                                          std::move(bucket_name),
                                                          timeout,
--- a/cmake/Protostellar.cmake
+++ b/cmake/Protostellar.cmake
@@ -242,4 +242,5 @@ set(couchbase_cxx_protostellar_TRANSPORT_FILES
     ${PROJECT_SOURCE_DIR}/core/protostellar/dispatcher.cxx
     ${PROJECT_SOURCE_DIR}/core/protostellar/credentials.cxx
     ${PROJECT_SOURCE_DIR}/core/protostellar/error_utils.cxx
+    ${PROJECT_SOURCE_DIR}/core/protostellar/kv_converter.cxx
     ${PROJECT_SOURCE_DIR}/core/protostellar/component.cxx)
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -343,6 +343,15 @@ make_component_timeouts(const cluster_options& options) -> protostellar::compone
   timeouts.management = options.management_timeout;
   return timeouts;
 }
+
+// Core carries only the on/off switch, so min_size and min_ratio keep the converter's defaults.
+inline auto
+make_compression_settings(const cluster_options& options) -> protostellar::kv::compression_settings
+{
+  protostellar::kv::compression_settings compression{};
+  compression.enabled = options.enable_compression;
+  return compression;
+}
 #endif
 } // namespace
 
@@ -1010,8 +1019,10 @@ public:
       const std::scoped_lock lock(protostellar_mutex_);
       protostellar_ = std::make_shared<protostellar::component>(
         ctx_,
-        protostellar::component_config{
-          std::move(channel), origin_.credentials(), make_component_timeouts(options) });
+        protostellar::component_config{ std::move(channel),
+                                        origin_.credentials(),
+                                        make_component_timeouts(options),
+                                        make_compression_settings(options) });
     }
     CB_LOG_INFO(R"(open couchbase2 cluster, id: "{}", endpoint: "{}")", id_, endpoint);
     return handler({});
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -1054,6 +1054,22 @@ public:
 #endif
   }
 
+  // couchbase2 wait_until_ready: there is no MCBP bootstrap to ping, so readiness is the gRPC
+  // channel's connectivity state. Delegates to the component's channel-state poll. Callers gate on
+  // is_protostellar(); the fail-closed branch only guards a misuse / non-couchbase2 build.
+  void protostellar_wait_until_ready(std::chrono::milliseconds timeout,
+                                     utils::movable_function<void(std::error_code)>&& handler)
+  {
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+    if (auto component = protostellar_component(); component) {
+      return component->wait_until_ready(timeout, std::move(handler));
+    }
+#else
+    (void)timeout;
+#endif
+    return handler(errc::common::feature_not_available);
+  }
+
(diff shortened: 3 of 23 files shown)


core/impl/wait_until_ready.cxx
```diff
@@ -337,6 +337,19 @@ wait_until_ready(core::cluster core,
...
                  utils::movable_function<void(std::error_code)> handler)
 {
+  if (core.is_protostellar()) {
+    // couchbase2 exposes a single gRPC channel; readiness is that channel reaching a connected
+    // state, not a per-service/per-bucket ping. So service_types and bucket_name have nothing to
+    // probe and are ignored (mirrors Java/Go). offline is rejected exactly as the classic op does
+    // below, so behavior is uniform across transports; every other desired_state waits for READY.
+    if (desired_state == couchbase::cluster_state::offline) {
+      return asio::post(core.io_context(), [handler = std::move(handler)]() mutable {
+        handler(errc::common::invalid_argument);
```
avsej, line 347: RFC 77 states that a supplied desired `ClusterState` is silently ignored, because the connectivity signal lacks the granularity to honour it. This rejects `offline` with `invalid_argument` instead.

The comment's reasoning — that the classic operation rejects `offline` the same way, so behaviour is uniform across transports within this library — is defensible, and arguably more useful to a caller than conforming to a clause the RFC itself derives from a lack of granularity. The audit also records that gocb honours the desired state and couchbase-jvm-clients partially honours it, so the SDKs do not agree here either.

Recommendation: keep the behaviour, and extend the comment to say it deliberately diverges from the RFC, referencing CNG-1's desired-state sub-issue. The thing to avoid is each SDK settling this differently without it being visible.

The same applies to the mechanism itself. Polling the connectivity state looks like a violation of the WaitUntilReady section, but CNG-1 records that no shipping SDK uses the mandated `Health.Check` RPC — couchbase-jvm-clients polls `managedChannel.getState()`, gocb polls `conn.GetState()`, couchbase-net-client throws `FeatureNotAvailable` — and the decision is unresolved. This matches the de-facto consensus and should stay; a CNG-1 reference in the comment would record that.

The real consequence of not sending an RPC is worth tracking: with no request there is no underlying gRPC exception to record, so the `lastException` diagnostic the RFC's WaitUntilReady goals depend on cannot be produced. A timeout tells the user the channel never became ready, but not whether the hostname is wrong, the certificate is untrusted, or the gateway is down.

Separately, and worth documenting user-facing rather than only in a comment: success here does not imply usable credentials, since channel readiness is transport-level. An application using `wait_until_ready` as a startup gate will now start successfully with wrong credentials and fail on first operation. RFC 77 lists this as an explicit non-goal so the behaviour is conformant, but it differs from the classic transport.

avsej (commented):
## Review — PR #1049 (`CXXCBC-908`)

This commit implements `wait_until_ready` over couchbase2 by polling the gRPC channel's connectivity state until
it reaches `READY` or the timeout expires, with `cluster::wait_until_ready` and `bucket::wait_until_ready` sharing
one implementation. The code is well commented and the comments explain the deliberate choices rather than just
the mechanism, which is the right standard for a divergence like this one.

### On the mechanism, since it looks like a specification violation and is not

RFC 77's WaitUntilReady section mandates the gRPC `Health.Check` RPC and explicitly rules out relying on the
connectivity-state enum, on the grounds that the enum cannot explain *why* a connection is failing. Polling the
state therefore looks like a straightforward violation.

It is not one to act on. The RFC 77 implementation audit records (CNG-1, High priority, unresolved) that **no
shipping SDK** uses the health RPC: couchbase-jvm-clients polls `managedChannel.getState()`,
gocb polls `conn.GetState()` — there is no `grpc/health` import anywhere in gocbcoreps — and
couchbase-net-client throws `FeatureNotAvailable`. The decision between ratifying connectivity-state polling and
requiring the health RPC has not been made. This commit matches the de-facto consensus of the two SDKs that
implement the call at all, which is the right place to be while CNG-1 is open.

Worth adding to the comment: a reference to CNG-1 and a note that this deviates from the RFC as written, so the
next reader does not have to rediscover that the divergence is deliberate.

The real cost of not sending an RPC is the one the RFC cares about: without a request there is no underlying gRPC
exception to record, so the `lastException` / `lastError` diagnostic that the RFC's own WaitUntilReady goals and
Testing subsection depend on cannot be produced. A timeout here therefore tells the user that the channel did not
become ready, but not why — not whether the hostname is wrong, the certificate is untrusted, or the gateway is
down. That is the gap to track against CNG-1 rather than to fix by switching mechanism unilaterally.

### Strengths

- `cluster::wait_until_ready` and `bucket::wait_until_ready` share one implementation, which RFC 77 requires
  explicitly.
- `service_types` and `bucket_name` are ignored with a comment saying why, matching both the RFC and the two
  reference implementations.
- The polling class is a `shared_ptr`-owned state machine rather than a loop capturing raw state, so the deadline
  and the completion cannot outlive their own storage.

### Findings

|---|---|---|---|---|
| 1 | medium | spec-compliance | A requested `cluster_state::offline` is rejected where the specification says the desired state is ignored | `core/impl/wait_until_ready.cxx:347` |
| 2 | medium | api-surface | Success does not imply the credentials are usable, which differs from the classic transport | `core/impl/wait_until_ready.cxx:347` |

#### 1. `cluster_state::offline` is rejected (medium)

```cpp
handler(errc::common::invalid_argument);
```

RFC 77 states that a supplied desired `ClusterState` is silently ignored, because the connectivity signal does not
have the granularity to honour it. This path rejects `offline` with `invalid_argument` instead.

The comment gives the reasoning — the classic operation rejects `offline` the same way, so behaviour is uniform
across transports within this library — and that is a defensible position. Uniformity with the sibling transport
is arguably more valuable to a user than conformance with a clause the RFC itself describes as a consequence of
lacking granularity, and the audit records that gocb honours the desired state while couchbase-jvm-clients
partially honours it, so the SDKs do not agree here either.

The recommendation is to keep the behaviour and make the divergence explicit in the comment, referencing CNG-1's
sub-issue on desired-state handling. What should not happen is for this to be settled differently in each SDK
without anyone noticing, which is the current situation.

#### 2. Success does not imply usable credentials (medium)

Channel readiness is a transport-level property, so `wait_until_ready` returns success for any reachable endpoint
whose TLS handshake completes — including one where the configured username and password are wrong. The first
actual operation then fails with an authentication error.

RFC 77 lists this as an explicit non-goal, so the behaviour is conformant, and the health RPC would not change it
either since the RFC also says no `Authorization` header is sent with it. But it is a real behavioural difference
from the classic transport, where `wait_until_ready` does surface bad credentials, and an application that uses
`wait_until_ready` as a startup gate will now start successfully and fail on first use.

That difference belongs in the user-facing documentation for `couchbase2://` rather than only in a source comment,
since it changes what a startup check means.

(inline comment on `core/impl/wait_until_ready.cxx:347`)

### Note on the poll loop

The poll interval is a fixed constant and the loop re-arms a timer rather than spinning, so it cannot busy-wait,
and the deadline is checked once per iteration. `TRANSIENT_FAILURE` is treated as not-yet-ready rather than as a
terminal failure, which is correct — the channel's own reconnect backoff drives recovery, and failing fast on a
transient state would defeat the purpose of the call.

One thing worth confirming: this commit touches ten files including the public cluster and bucket, so it is worth
checking that the classic `wait_until_ready` path is byte-for-byte unaffected. The diff suggests it is — the
couchbase2 branch is an early return above the existing operation — but that is the kind of change where a
regression would be quiet.

### Verdict

Ready with nits. The mechanism is the right one to ship while CNG-1 is unresolved; both findings are about
recording deliberate divergences so they are not mistaken for oversights later.

PR #1052, opened 2026-07-30
CXXCBC-485: document the streaming query/analytics API with runnable examples

avsej:
### What changes

Adds five examples in a new `test_integration_examples_streaming` target, referenced from the public headers as Doxygen `@snippet`s (`cluster`, `scope`, `query_row`, `analytics_row`, `query_stream_result`, `analytics_stream_result`). No production code changes.

| Example | Covers |
|---|---|
| blocking drain | three-state `next()`, `signature()` before drain, custom document type, metadata + metrics |
| scope-level iterator | range-`for`; a prefix read timed against the buffered path over the same result |
| non-blocking pull loop | callback `next()` chained from completion handlers |
| analytics streaming | the analytics equivalents |
| error handling | both failure channels, partial success, sticky terminal, `meta_data()` after teardown, timeout classification |

They run as integration tests, so the documented code and its printed output are verified against a live cluster.

The error-handling example shows a populated `couchbase::error::ctx()`, which is what #1054 makes possible.

Two behaviours the examples pin down explicitly:

- The timeout example sets `.readonly(true)`. `query_options` defaults it to `false` and read-only-ness is not inferred from statement text, so an unflagged `SELECT` times out as `ambiguous_timeout`.
- A missing keyspace maps to `index_failure` (server error 12003).

### Verification

`test_integration_examples_streaming` passes against Couchbase Server 7.1.6: 5 cases, 27 assertions. Doxygen adds no warnings, and every added snippet renders into the generated HTML. `bin/check-clang-format` is clean against the parent commit, and the changed translation units compile under GCC 13.3 with the CI warning set. macOS, Windows/MSVC and the sanitizer legs are covered by CI only.

#1054, which this depended on, is merged, so the diff here is now the examples commit alone.

### DIFF
--- a/couchbase/analytics_row.hxx
+++ b/couchbase/analytics_row.hxx
@@ -67,6 +67,8 @@ public:
    * analytics_row, and decoding empty/malformed content throws the serializer's parse exception —
    * gate on the three-state next() contract (check the error, then the optional) before decoding.
    *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx example-analytics-stream
+   *
    * @tparam Serializer the serializer to use (defaults to @ref codec::tao_json_serializer)
    * @tparam Document the document type to decode into (defaults to the serializer's document_type)
    * @return the decoded document
--- a/couchbase/analytics_stream_result.hxx
+++ b/couchbase/analytics_stream_result.hxx
@@ -49,6 +49,8 @@ using analytics_row_handler = std::function<void(error, std::optional<analytics_
  * (or @ref cancel() called) before @ref meta_data() resolves. Only a single @ref next()
  * call may be outstanding at a time.
  *
+ * @snippet{trimleft} test_integration_examples_streaming.cxx example-analytics-stream
+ *
  * @note The future-returning overloads (`next()`, `meta_data()`) and the eager @ref iterator
  * block the calling thread until the result is ready. They must not be called from within an SDK
  * completion handler (i.e. the library's I/O thread) — doing so blocks that thread against itself
@@ -112,6 +114,11 @@ public:
 
   /**
    * Returns the analytics signature captured from the response metadata, if present.
+   *
+   * The signature is part of the response preamble, so — unlike @ref meta_data() — it is available
+   * as soon as the stream has started, without draining it first:
+   *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx example-analytics-stream
    *
    * @return optional binary JSON signature
    *
--- a/couchbase/cluster.hxx
+++ b/couchbase/cluster.hxx
@@ -233,6 +233,11 @@ public:
    * request falls back to the buffered @ref query() path and its rows are replayed through the
    * returned handle.
    *
+   * The handler runs on the library's I/O thread, so the stream must be drained with the callback
+   * @ref query_stream_result#next() overload rather than the blocking ones:
+   *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx example-query-stream-async
+   *
    * @param statement the N1QL query statement.
    * @param options options to customize the query request.
    * @param handler the handler that implements @ref query_stream_handler
@@ -246,6 +251,21 @@ public:
 
   /**
    * Performs a streaming query against the query (N1QL) services.
+   *
+   * Rows are pulled from the returned handle one at a time, so a result of any size costs the same
+   * bounded amount of memory:
+   *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx example-query-stream
+   *
+   * The document type the rows are decoded into:
+   *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx streaming-product
+   *
+   * A failure surfaces either here, before the stream starts, or as the terminal of
+   * @ref query_stream_result#next() — possibly after rows have already been consumed, which a
+   * buffered @ref query() cannot produce. Both channels have to be handled:
+   *
+   * @snippet{trimleft} test_integration_examples_streaming.cxx example-query-stream-errors
    *
    * @param statement the N1QL query statement.
    * @param options options to customize the query request.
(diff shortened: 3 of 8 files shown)


thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. Findings were verified by reading the actual implementation (`core/query_stream_component.cxx`, `core/impl/public_query_stream_result.cxx`, `core/row_streamer.cxx`, etc.) rather than compiled/run against a live cluster; please treat the reasoning as a starting point for discussion, not ground truth.

---

Nice PR — traced the async pull loop, the `cancel()`-then-drain path, and the timeout/metrics claims against the actual core implementation and everything checks out (including the trickier bits: no reference cycle in `row_counter`'s `shared_from_this()` capture, the recursive `pull()` genuinely doesn't grow the stack since asio never completes the channel inline, and `cancel()`-then-`next()` is a real mutex-protected guarantee, not a lucky race). Two small things worth fixing before merge, since the PR's whole point is *runnable* examples:

1. **`example-query-stream` and `example-query-stream-iterator` won't compile if extracted verbatim.** Both snippets format a `std::chrono::milliseconds` with `fmt::println` (test/test_integration_examples_streaming.cxx:191-193, 314-317, 341-343), but the chrono formatter (`<spdlog/fmt/bundled/chrono.h>`) is only pulled in at file scope (line 26) — outside every `//! [...]` marker. Since each snippet deliberately lists its own include block, copying just the tagged region into a fresh TU hits fmt's `static_assert`. Worth adding the chrono include to those two blocks.

2. **`example-query-stream-iterator` part [1] can print a nondeterministic row count.** It runs `scope.query_stream(statement)` with default `scan_consistency` (`not_bounded`) right after `seed_products()` upserts the 10 documents (test/test_integration_examples_streaming.cxx:267-273), so index catch-up timing decides whether all 10 rows show up. The recorded sample output claims all ten. `example-query-stream` a few hundred lines up already uses `.scan_consistency(request_plus)` for exactly this reason — might be worth matching here too.

Everything else I found was minor doc-phrasing precision (nothing that affects correctness) — happy to go through those separately if useful, but didn't want to pile on a stack review with nits that don't change anything.

avsej:
Both points are correct and both are fixed in `e6c43b80`. The first turned out to be broader than
reported.

## 1. Snippets do not compile when extracted verbatim

> **`example-query-stream` and `example-query-stream-iterator` won't compile if extracted verbatim.** Both snippets format a `std::chrono::milliseconds` with `fmt::println` (test/test_integration_examples_streaming.cxx:191-193, 314-317, 341-343), but the chrono formatter (`<spdlog/fmt/bundled/chrono.h>`) is only pulled in at file scope (line 26) — outside every `//! [...]` marker. Since each snippet deliberately lists its own include block, copying just the tagged region into a fresh TU hits fmt's `static_assert`. Worth adding the chrono include to those two blocks.

Confirmed, and the diagnosis generalises: the same file-scope reasoning applies to fmt itself. No
snippet included any fmt header, and `couchbase/fmt/error.hxx` specializes `fmt::formatter` without
including one, so an extracted translation unit failed on `'fmt' has not been declared` before ever
reaching the chrono `static_assert`. That affected **all five** snippets, not only the two that
format a duration; the file compiles as a whole because `test_helper_integration.hxx` at line 18
pulls fmt in transitively, which is exactly the scoping problem described.

Include order matters for the same reason, so the blocks now follow the layout already used by the
standalone programs in `examples/` (`bulk_transactional_get_replace.cxx`, `minimal.cxx`): the
`spdlog/fmt/bundled/*` headers first, then the `couchbase/fmt/*.hxx` formatter specializations in
their own group, with a comment recording why the order is load-bearing.

- all five snippets: `#include <spdlog/fmt/bundled/format.h>`
- `example-query-stream` and `example-query-stream-iterator`: also
  `#include <spdlog/fmt/bundled/chrono.h>`

This was verified rather than reasoned about. A harness extracts each `//! [tag]` region into its own
translation unit — adding the `streaming-product` snippet only to the two examples that decode into
that type, since those are the two the documentation shows it with — and syntax-checks each with the
build's real flags. Before the change all five failed; after it all five compile:

```
=== OK  example-analytics-stream
=== OK  example-query-stream
=== OK  example-query-stream-async
=== OK  example-query-stream-errors
=== OK  example-query-stream-iterator
```

The harness is not part of the PR. It could be added as a test so the property is enforced rather
than spot-checked, which would have caught this at the time; that seems worth doing separately rather
than growing this PR.

## 2. Nondeterministic row count in `example-query-stream-iterator` part [1]

> **`example-query-stream-iterator` part [1] can print a nondeterministic row count.** It runs `scope.query_stream(statement)` with default `scan_consistency` (`not_bounded`) right after `seed_products()` upserts the 10 documents (test/test_integration_examples_streaming.cxx:267-273), so index catch-up timing decides whether all 10 rows show up. The recorded sample output claims all ten. `example-query-stream` a few hundred lines up already uses `.scan_consistency(request_plus)` for exactly this reason — might be worth matching here too.

Confirmed and matched, with the reason stated inline so the next reader does not have to rediscover
it:

```cpp
    // request_plus, because the documents were written moments ago: with the default
    // (not_bounded) the query runs against whatever the index has caught up with, so a freshly
    // written document may or may not be visible and the row count would vary between runs.
    const auto options =
      couchbase::query_options{}.scan_consistency(couchbase::query_scan_consistency::request_plus);
    auto [err, result] = scope.query_stream(statement, options).get();
```

The recorded sample output, which lists all ten rows, is now what the example actually guarantees.

## Doc-phrasing notes

Please do send them. The examples are the documentation, so precision in the prose is the deliverable
rather than a nit, and a separate pass over them is welcome.

## Verification of `e6c43b80`

- Each snippet extracted and syntax-checked standalone: five of five compile, as above.
- `test_integration_examples_streaming` 5 cases / 27 assertions, `test_integration_query` 28 / 16221,
  `test_integration_analytics` 13 / 1229 — all pass against Couchbase Server 7.1.6.
- `git-clang-format` reports no changes against the base; the changed translation units compile under
  GCC 13.3 with the CI warning set.
- Not covered locally: 7.0.5, macOS, Windows/MSVC and the sanitizer legs. Those are proven by CI only.

thejcfactor (approved):

PR #1054, opened 2026-07-30
CXXCBC-485: report an error context for streaming query and analytics failures

avsej:
**Stack 1 of 2.** Followed by #1052. Rebased on `main` after #1053 merged; this PR now contains only its own commit.

### What changes

Streaming failures, at dispatch and at the stream terminal alike, were reported as a bare error code in a generic message: `couchbase::error::ctx()` was empty, where the buffered path supplies the service's error code and message, the statement and encoded parameters, the HTTP status and body, and the node that answered.

- `query_stream_component` / `analytics_stream_component` build an `error_context` from the request before dispatch and fill in response-side fields as they arrive.
- `query_stream` / `analytics_stream` expose `error_details()` — the service-reported diagnostics parsed out of the preamble and, once reached, the trailer.
- The context travels with the result handle, so `internal_*_stream_result` reports a mid-stream terminal against it. A cancellation has no service-reported error, so its context describes only the request.
- Stream diagnostics augment the context rather than replace it: only fields the stream actually reported are stamped. The prepared-statement path needs that, because it reaches the streaming handle by replaying an already-buffered response and so starts out with a fully populated context the replay has nothing to add to.
- `io::http_streaming_response` captures the session's host, port and local/remote addresses at construction, exposed via `http_response::dispatch_info()`; the streaming layer previously reported no endpoint at all.
- Both core-internal streaming handler signatures (cluster entry points and components) carry `error_context::query` / `error_context::analytics` instead of `std::error_code`.

The public API is unchanged.

### Tests

- `integration: streaming query reports the same error context as buffered query`
- `integration: streaming analytics reports the same error context as buffered analytics`
- `integration: a streaming query terminal error carries the request context`
- `integration: a streaming prepared statement keeps the buffered response diagnostics` — verified to fail if the stamping is made unconditional

### Verification

Against a 6-node 8.0.1 cluster (kv/n1ql/index/fts/cbas): `test_integration_query`, `test_integration_analytics`, and the streaming unit suites pass. `bin/check-clang-format` clean against the parent commit.

### DIFF
--- a/core/analytics_stream.cxx
+++ b/core/analytics_stream.cxx
@@ -54,6 +54,28 @@ normalize_stream_error(std::error_code ec) -> std::error_code
 }
 } // namespace
 
+namespace
+{
+// Lift the service-reported diagnostics out of a parsed preamble/trailer. Called for both sections,
+// so whichever one carried the failure is what the error context ends up describing.
+auto
+extract_error_details(const operations::analytics_response::analytics_meta_data& meta,
+                      std::string raw) -> stream_error_details
+{
+  stream_error_details details{};
+  details.client_context_id = meta.client_context_id;
+  // Retained whether or not the section reports an error: it becomes error_context::http_body,
+  // which the buffered path fills for every response, and even an error-free envelope names the
+  // request the service logged.
+  details.raw = std::move(raw);
+  if (!meta.errors.empty()) {
+    details.first_error_code = meta.errors.front().code;
+    details.first_error_message = meta.errors.front().message;
+  }
+  return details;
+}
+} // namespace
+
 class analytics_stream_impl : public std::enable_shared_from_this<analytics_stream_impl>
 {
 public:
@@ -73,11 +95,18 @@ public:
       try {
         meta = operations::parse_analytics_meta(utils::json::parse(preamble));
       } catch (const std::exception&) {
+        {
+          // Keep the unparseable preamble: it is the only diagnostic there is for a malformed
+          // response, and the buffered path reports the raw body in the same situation.
+          const std::scoped_lock lock{ self->mutex_ };
+          self->error_details_.raw = std::move(preamble);
+        }
         return on_ready(errc::common::parsing_failure);
       }
       {
         const std::scoped_lock lock{ self->mutex_ };
         self->signature_ = meta.signature;
+        self->error_details_ = extract_error_details(meta, std::move(preamble));
       }
       // An upfront error is rare for the streaming path (status/errors usually arrive in the
       // trailer). The preamble status is typically "running", so only surface an early error when
@@ -130,9 +159,14 @@ public:
           auto meta = operations::parse_analytics_meta(utils::json::parse(raw_meta.value()));
           terminal_ec = operations::map_analytics_error(meta);
           const std::scoped_lock lock{ self->mutex_ };
+          // The trailer is where a request that failed part-way reports itself, so its diagnostics
+          // supersede the preamble's for the terminal error.
+          self->error_details_ = extract_error_details(meta, std::move(raw_meta.value()));
           self->meta_data_ = std::move(meta);
         } catch (const std::exception&) {
           terminal_ec = errc::common::parsing_failure;
+          const std::scoped_lock lock{ self->mutex_ };
+          self->error_details_.raw = std::move(raw_meta.value());
         }
       }
       {
@@ -157,6 +191,12 @@ public:
     return meta_data_;
   }
 
+  [[nodiscard]] auto error_details() const -> stream_error_details
+  {
+    const std::scoped_lock lock{ mutex_ };
+    return error_details_;
+  }
+
   void cancel()
   {
     streamer_.cancel();
@@ -167,6 +207,7 @@ private:
   mutable std::mutex mutex_{};
   std::optional<std::string> signature_{};
   std::optional<operations::analytics_response::analytics_meta_data> meta_data_{};
+  stream_error_details error_details_{};
   // Terminal-sticky state (guarded by mutex_): once next_row reports the terminal, it is remembered
   // so every later pull re-delivers it instead of parking on the drained channel.
   bool terminal_reached_{ false };
@@ -197,6 +238,12 @@ auto
 analytics_stream::signature() const -> std::optional<std::string>
 {
   return impl_->signature();
+}
+
+auto
+analytics_stream::error_details() const -> stream_error_details
+{
+  return impl_->error_details();
 }
 
 auto
--- a/core/analytics_stream.hxx
+++ b/core/analytics_stream.hxx
@@ -19,6 +19,7 @@
 
 #include "core/operations/document_analytics.hxx"
 #include "row_streamer.hxx"
+#include "stream_error_details.hxx"
 #include "utils/movable_function.hxx"
 
 #include <memory>
@@ -62,6 +63,13 @@ public:
    * the response carried an upfront error (or the preamble failed to parse).
    */
   void start(utils::movable_function<void(std::error_code early_error)>&& on_ready);
+
+  /**
+   * Service-reported diagnostics from the most recently parsed JSON section — the preamble after
+   * start(), the trailer once the terminal has been reached. The dispatch layer stamps these into
+   * the error context so a streaming failure reports the same detail as a buffered one.
+   */
+  [[nodiscard]] auto error_details() const -> stream_error_details;
 
   /**
    * Retrieves the next row. A populated row is delivered with a falsy error_code. An empty row
--- a/core/analytics_stream_component.cxx
+++ b/core/analytics_stream_component.cxx
@@ -110,13 +110,28 @@ public:
 
     http_request http_req{ service_type::analytics, "POST" };
     http_req.path = "/query/service";
+
+    // Everything about the request that an error context reports, known before dispatch. The
+    // response-side fields (HTTP status, the service's first error, the endpoint) are filled in as
+    // they become available, so a failure at any stage carries as much detail as the buffered path.
+    // Held behind a shared_ptr, not by value: it carries the encoded request body in
+    // `parameters`, and both the dispatch callback and the synchronous dispatch-failure branch
+    // below need it, so sharing avoids deep-copying that body on every streaming request.
+    auto base_ctx = std::make_shared<error_context::analytics>();
+    base_ctx->client_context_id = client_context_id;
+    base_ctx->statement = request.statement;
+    base_ctx->method = http_req.method;
+    base_ctx->path = http_req.path;
+
     try {
       http_req.body = build_streaming_analytics_body(request, client_context_id, timeout);
     } catch (const std::exception&) {
       // A caller-supplied raw/parameter value that is not valid JSON fails encoding; surface it as
       // invalid_argument rather than letting the exception escape the public API.
-      return handler({}, errc::common::invalid_argument);
+      base_ctx->ec = errc::common::invalid_argument;
+      return handler({}, std::move(*base_ctx));
     }
+    base_ctx->parameters = http_req.body;
     http_req.timeout = timeout;
     http_req.client_context_id = client_context_id;
     http_req.is_read_only = request.readonly;
@@ -140,11 +155,23 @@ public:
     const bool read_only = request.readonly;
     auto op = http_.do_http_request(
       http_req,
-      [self = shared_from_this(), callback_state, timeout, read_only](http_response resp,
-                                                                      dispatch_error err) mutable {
+      [self = shared_from_this(), callback_state, timeout, read_only, base_ctx](
+        http_response resp, dispatch_error err) mutable {
         if (auto ec = to_error_code(err)) {
-          self->invoke(callback_state, {}, ec);
+          base_ctx->ec = ec;
+          self->invoke(callback_state, {}, std::move(*base_ctx));
           return;
+        }
+        // The response is in hand, so the context can now name the endpoint and HTTP status.
+        base_ctx->http_status = resp.status_code();
+        const auto dispatch_info = resp.dispatch_info();
+        base_ctx->hostname = dispatch_info.hostname;
+        base_ctx->port = dispatch_info.port;
+        if (!dispatch_info.remote_address.empty()) {
+          base_ctx->last_dispatched_to = dispatch_info.remote_address;
+        }
+        if (!dispatch_info.local_address.empty()) {
+          base_ctx->last_dispatched_from = dispatch_info.local_address;
         }
         // Use the cluster's streaming tuning, but bound the server between reads with this
         // request's timeout: if it stalls mid-body the stream times out rather than hanging next()
@@ -153,20 +180,30 @@ public:
         options.idle_timeout = timeout;
         options.is_read_only = read_only;
         auto stream = std::make_shared<analytics_stream>(self->io_, resp.body(), options);
-        stream->start([self, callback_state, stream](std::error_code early_error) mutable {
-          if (early_error) {
-            // No consumer will ever take ownership of this stream, so tear the underlying body
-            // (socket + timers) down explicitly rather than leaking it until the last handle drops.
-            stream->cancel();
-            self->invoke(callback_state, {}, early_error);
-            return;
-          }
-          self->invoke(callback_state, std::move(*stream), {});
-        });
+        stream->start(
+          [self, callback_state, stream, base_ctx](std::error_code early_error) mutable {
+            if (early_error) {
+              // The preamble is what carried the failure, so stamp its diagnostics before tearing
+              // the stream down.
+              base_ctx->ec = early_error;
+              apply_error_details(*base_ctx, stream->error_details());
+              // No consumer will ever take ownership of this stream, so tear the underlying body
+              // (socket + timers) down explicitly rather than leaking it until the last handle
+              // drops.
+              stream->cancel();
+              self->invoke(callback_state, {}, std::move(*base_ctx));
+              return;
+            }
+            // Hand the base context to the consumer alongside the stream: a terminal error reached
+            // later is stamped onto it, so a mid-stream failure reports the same detail as this
+            // one.
+            self->invoke(callback_state, std::move(*stream), std::move(*base_ctx));
+          });
       });
 
     if (!op.has_value()) {
-      invoke(callback_state, {}, to_error_code(op.error()));
+      base_ctx->ec = to_error_code(op.error());
+      invoke(callback_state, {}, std::move(*base_ctx));
     }
   }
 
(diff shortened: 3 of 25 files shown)


thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. This was a static/fan-out review — subagents traced the code and test call paths by hand and I confirmed the finding below by reading the source directly, but nothing was compiled or run against a live cluster; please treat the reasoning as a starting point for discussion, not ground truth.

---

Nice change overall — the augmentation discipline (`apply_error_details` only stamping fields the stream actually reported), the connection-info capture in `http_streaming_response` (verified it happens post-connect, and post-TLS-handshake for TLS sessions specifically), and the lock/lifetime story around the new context objects all check out.

One thing that looks like it undercuts the PR's main goal, though:

**`next()`'s terminal branch discards the context it just built.**

In both `internal_query_stream_result::next()` (`core/impl/public_query_stream_result.cxx:222-225`) and the analytics equivalent (`public_analytics_stream_result.cxx:212-215`):

```cpp
self->resolve_meta_data(ec);                                        // builds the enriched error into terminal_value_.first
if (ec) {
  return handler(error(ec, "query stream ended with an error"), {}); // ...then hands the caller a bare one instead
}
```

`resolve_meta_data` stamps `ctx_` + `stream_.error_details()` into `terminal_value_.first`, but that value never reaches the caller here — line 224 constructs a fresh `error(ec, msg)`, whose context defaults to empty. The enriched error is only reachable via `meta_data()`, or via a **second** `next()` call after the terminal was already recorded (the `terminal_reached_` idempotent-replay guard a few lines above, which does serve `terminal_value_.first`).

Practical effect: a consumer that just polls `next()` until it returns an error — the common pattern, no explicit `cancel()` — gets `ec` with no statement, no `first_error_code`, no endpoint on the first (and typically only) time it sees the failure.

This is also why the new tests pass without catching it: all three context-assertion tests reach the terminal via `result.cancel()` first, which runs `resolve_meta_data()` synchronously and sets `terminal_reached_` *before* `next()` is called — so they only ever exercise the idempotent replay branch, never the fresh-discovery one where the bug lives.

Suggested fix: after `resolve_meta_data(ec)`, deliver `terminal_value_.first` (copy it out under `meta_mutex_`, release the lock, then invoke `handler`) instead of rebuilding a bare error. One thing to watch when doing that: the existing idempotent branch just above already invokes `handler` *while holding* `meta_mutex_`, so the fix shouldn't extend that pattern to the fresh-discovery path — a handler that calls back into `cancel()`/`next()` would self-deadlock on the non-recursive mutex.

Whichever fix lands, it'd be good to add a regression test alongside it: right now all three new context-assertion tests reach the terminal via `cancel()` first, so nothing exercises the plain "poll `next()` until it returns an error" path for either query or analytics — which is exactly the path the bug lives on. Without a test on that path specifically, this could regress silently later.

Three smaller things worth a look, not blocking:
- `core::cluster::query_stream`/`analytics_query_stream`'s handler signature moved from `std::error_code` to `error_context::query`/`analytics` (`core/cluster.hxx:132-142`). All in-tree callers are updated, but since wrapper SDKs bind `core::` directly rather than going through `couchbase::`, this is a source-compat break one layer below the "public API unchanged" claim — worth a heads-up to whoever tracks the wrapper-side bump.
- `retry_attempts`/`retry_reasons` never get populated on the streaming path (no retry orchestrator is wired to `do_http_request`), even though the buffered path can report retries for the same request. Pre-existing gap rather than a regression, but it means "same context as buffered" isn't quite complete yet.
- `core/free_form_http_request.cxx:84-90` — `status_code()` guards `cached_body_` before falling through to `streaming_resp_.status_code()`, but not `fault_enabled_`. On the fault-injected construction path `streaming_resp_` is default-constructed with a null `impl_`, so `status_code()` there dereferences null. Pre-existing, but directly adjacent to (and inconsistent with) the guard this PR just added for `dispatch_info()`/`endpoint()` a few lines up in the same file — worth fixing while you're in there.

avsej:
Each point raised in the suppressed-comment sections of the recent reviews is quoted below and
answered in order. Points 1 and 2 are from the review of 2026-07-31T02:39:33Z; point 3, on
`details.raw`, is from the review of 2026-07-30T22:40:33Z and had not yet been answered.

## 1. `core/cluster.hxx:144` — declaration width

> This declaration exceeds the project's 100-column formatting limit, which makes the header harder to scan and may fail formatting checks. Please wrap the handler parameter onto a new line (similar to query_stream above).

Not applied. Both halves of the point were checked and neither holds.

The declaration is already wrapped exactly as described, and identically to `query_stream` above it —
the snippet quoted in the note is itself the two-line wrapped form:

```
134   void query_stream(o::query_request request,
135                     mf<void(query_stream, error_context::query)>&& handler) const;
...
143   void analytics_query_stream(o::analytics_request request,
144                               mf<void(analytics_stream, error_context::analytics)>&& handler) const;
```

Line 144 measures 100 characters. `.clang-format` sets `ColumnLimit: 100`, so 100 characters is at
the limit rather than beyond it.

Measured, not inferred:

- `git-clang-format --diff` over the changed lines, against the base, reports `clang-format did not
  modify any files` (clang-format 24.0.0).
- The repository's `clang_format` CI leg concluded `success` on `5c92c899`
  ([run 30599363524](https://github.com/couchbase/couchbase-cxx-client/actions/runs/30599363524)).
  That leg runs clang-format 22 (`LLVM_VERSION: 22` in `.github/workflows/linters.yml`) through
  `bin/check-clang-format`, which diffs against `HEAD^`; on a `pull_request` checkout that is the
  base commit, so line 144 is inside the checked range rather than outside it. The declaration is
  byte-identical in the current revision, so that result still applies to it.

Wrapping the parameter further would move it away from the shape `query_stream` uses, so the
declaration is left as it stands.

## 2. `core/io/http_streaming_response.cxx:301` — port truncation

> Parsing the session port with std::stoul() and then static_cast'ing to std::uint16_t can silently truncate out-of-range values (e.g., > 65535). Even if unlikely, it would produce an incorrect port in the error context; please clamp or reject out-of-range ports instead of truncating.

Applied, by removing the conversion rather than range-checking it.

The port is already available numerically on the session, and it is the field the buffered path
stamps onto its own context (`ctx.port = session_->http_context().port`, `core/io/http_command.hxx`).
The streaming path now reads the same field, so both report the endpoint from one source:

```cpp
    // Taken from the session's http_context, the same source io::http_command stamps onto a
    // buffered response, so both paths report the identical endpoint. The port is already a
    // std::uint16_t there; http_session::port() is that same value rendered as a string for
    // resolving and for the Host header, so parsing it back would only reintroduce a conversion.
    dispatch_info.hostname = session->http_context().hostname;
    dispatch_info.port = session->http_context().port;
```

`http_session::port()` returns `service_`, constructed as `std::to_string(node.port)` from a
`std::uint16_t` in `http_session_manager::create_session`, and `hostname()` returns the same
`node.hostname` that reaches `http_context`. The string was therefore a rendering of the value now
read directly, and the parse was the only step that could lose information. The `try`/`catch` and its
fallback to port `0` are removed with it, as there is no longer a parse that can fail.

Measured against Couchbase Server 7.1.6: the terminal-error context reports `"port":8093`, matching
`"last_dispatched_to":"172.18.0.8:8093"`. An assertion was added to `integration: a streaming query
terminal error carries the request context`; `port` is serialized only when non-zero
(`core/error_context/query_public_json.hxx`), so the field's presence is what the assertion proves.

## 3. `core/query_stream.cxx:103`, `core/analytics_stream.cxx:72` — `details.raw` retained with no errors

> extract_error_details() stores the entire preamble/trailer JSON in details.raw even when the section contains no service errors. That raw value is later stamped into error_context::http_body on terminal errors (including request_canceled), which contradicts the comment that cancellations have no service-reported diagnostics and can also retain/copy unnecessary JSON. Consider only retaining raw when the section actually reports an error (or when parsing fails, which is already handled separately).

The contradiction is real and was located in the comment, not the behaviour. It has been corrected
there; the retention itself is kept, for the reasons below.

Measured — the full context of a cancelled streaming query against Couchbase Server 7.1.6:

```
http_body = {"requestID": "17e3ad0a-94be-4d41-ae46-d83c8762f1d5",
             "clientContextID": "93307c-1f73-964e-ffdf-7319413c36dec8",
             "signature": {"n":"json"},
             "results": []}
```

with `first_error_code` and `first_error_message` absent from the serialized context altogether.

So the first half of the comment held — a cancellation reports no service error — while its further
claim that the context "simply describes the request" did not, because the parsed envelope is
reported as the body. The comment in both `core/impl/public_query_stream_result.cxx` and
`core/impl/public_analytics_stream_result.cxx` now states what is actually reported, and the reason
for the retention is recorded at the assignment in both stream implementations.

The retention is kept rather than gated on errors because gating it would work against the purpose of
this change and would discard the diagnostics:

- `io::http_command` sets `ctx.http_body` for every buffered response, whether or not the service
  reported an error. Reporting the received body unconditionally is what makes the streaming context
  match the buffered one, which is what this pull request is for.
- The envelope carries `requestID` and `clientContextID` — the identifiers the query service's own
  logs are keyed by. On a cancellation these are the whole diagnostic value of the field; gating on
  `errors` would leave a cancelled request with no correlator at all.

On retained size: the value is the response envelope with the rows excluded — `results` is emptied,
and rows are the part that streams. The measured example above is 155 bytes; the size is dominated by
the signature and is unrelated to the row count. For comparison, the row buffer these streams already
hold is bounded at a 2 MiB high-water mark.

Both halves of the behaviour are now pinned by assertions rather than left to the comment:

```cpp
  REQUIRE(ctx.find("first_error_code") == nullptr);
  REQUIRE(ctx.find("first_error_message") == nullptr);
  const auto body = tao::json::from_string(ctx.at("http_body").as<std::string>());
  REQUIRE_FALSE(body.at("requestID").as<std::string>().empty());
  REQUIRE(body.find("errors") == nullptr);
```

---

## Earlier review round (2026-07-30T22:16:15Z)

Five inline comments, all applied in a revision pushed before the reviews answered above, and
re-verified as present in the current revision:

| Point | Disposition |
|---|---|
| `base_ctx` copied into the `do_http_request` capture (query component) | Applied — context built behind `std::make_shared<error_context::query>()`, `core/query_stream_component.cxx:122` |
| Same, analytics component | Applied — `std::make_shared<error_context::analytics>()`, `core/analytics_stream_component.cxx:120` |
| `preamble` copied into `details.raw` (query, both sites) | Applied — `std::move` at `core/query_stream.cxx:146` and `:204` |
| Same, analytics (both sites) | Applied — `std::move` at `core/analytics_stream.cxx:109` and `:164` |
| `use_replica` rejection leaves `client_context_id` empty | Applied — id generated when the caller supplied none, `core/cluster.cxx:1755` |

## Verification of this revision

The revision answered here is `fa9500f1bc5799a82d31cd12d4cc9b22299109e6`, rebased onto `main` at
`fa1f1269` (which had advanced past the previous base by `CXXCBC-887` and `CXXCBC-888`, neither
touching the streaming path). The dependent examples pull request, #1052, is at `214a896a`.

- `git-clang-format --diff` over the changed lines against the base: no changes suggested, for each
  commit of the stack taken separately and for the stack as a whole.
- All 15 changed translation units compiled with GCC 13.3 under the CI warning set
  (`-Werror -Wconversion -Wsign-conversion -Wold-style-cast -Wuseless-cast -Wpedantic
  -Wdouble-promotion`): zero diagnostics. This check exists because the toolchain used for local
  development is more permissive than the one CI builds with, and has previously accepted code that
  GCC 13 rejects.
- Couchbase Server 7.1.6, after the rebase: `test_integration_examples_streaming` 5 cases / 27
  assertions, `test_integration_query` 28 cases / 16221 assertions, `test_integration_analytics`
  13 cases / 1229 assertions — all pass.
- Couchbase Server 7.0.5, before the rebase: `test_integration_query` 27 passed / 1 skipped, 16215
  assertions; `test_integration_analytics` 13 cases / 1244 assertions pass. The rebase did not touch
  these files, but the 7.0.5 run was not repeated against the rebased revision.
- Not covered locally: macOS, Windows/MSVC, and the sanitizer legs. Those are proven only by CI.

One pre-existing failure is unrelated to this change and is not introduced by it: `integration:
connecting with a custom transactions metadata collection that is in a bucket that does not exist -
Public API` segfaults on Couchbase Server 7.0.5. The same failure occurs on unrelated pull request
#1050, and `core/transactions/` contains no reference to the streaming path.

thejcfactor:
🤖 Written with Claude Code and posted on my behalf, same as my earlier comment — static reading of the source, nothing compiled or run against a cluster.

Correction to my earlier comment — I mis-stated the test situation and want to fix it before it muddies the point.

I said all three context-assertion tests reach the terminal via `cancel()`. That's wrong: there are four new tests and two of them (`test/test_integration_query.cxx:1203`, `test/test_integration_analytics.cxx:1185`) poll `next()` in a loop with no `cancel()`. My apologies.

The finding itself stands, and those two tests are actually evidence for it. For keyspace/dataset-not-found the response has no `results` array, so `errors` lands in the preamble, `query_stream::start()` maps it to `early_error` (`core/query_stream.cxx:154-158`), and the request fails at dispatch with the fully-populated `base_ctx` — the `while` loop is never entered. That has to be what happens, because if it *were* entered, `row_err` would carry the default-constructed context from `core/impl/public_query_stream_result.cxx:226`, `to_json()` would return `"{}"` (`core/impl/error_context.cxx:51-53`), and `.at("first_error_code")` would throw. You measured both green, so neither test reaches the terminal branch.

The uncovered path is the one your own comment at `core/query_stream.cxx:148` calls the normal case — "status/errors usually arrive in the trailer." `test/test_unit_query_stream.cxx:70` proves the core stream really does deliver a trailing error as its terminal after emitting rows; nothing drives that through the public `next()`, which is where the enriched context gets dropped on the floor:

```cpp
self->resolve_meta_data(ec);                                         // builds the enriched error into terminal_value_.first
if (ec) {
  return handler(error(ec, "query stream ended with an error"), {});  // ...then hands the caller a bare one instead
}
```

`error`'s signature is `error(std::error_code, std::string = {}, error_context = {})` (`couchbase/error.hxx:34`), so that third argument is a default-constructed context — whereas the idempotent-replay branch twenty lines up (`:198`) correctly delivers `terminal_value_.first`. Analytics is identical at `core/impl/public_analytics_stream_result.cxx:212-215`.

So a consumer that just polls `next()` until it returns an error — no explicit `cancel()` — gets `ec` with no statement, no `first_error_code`, no endpoint, on the first and typically only time it sees the failure. That's the case this PR is for.

avsej:
Three further points were raised after the previous revision. Two of them arrived on the review of
the dependent pull request #1052, which contains this commit as well, and are answered here because
the code they concern belongs to this one.

## 4. `core/impl/public_query_stream_result.cxx`, `core/impl/public_analytics_stream_result.cxx` — terminal error delivered without the context

> internal_query_stream_result::resolve_meta_data() now builds a contextual terminal error via make_error(ctx), but internal_query_stream_result::next() still returns a freshly-constructed error(ec, "query stream ended with an error") on the first terminal delivery. That drops the request/endpoint/service diagnostics from the error seen by callers of query_stream_result::next() (and the future-returning overload), and will make terminal errors (including cancel()/timeouts) surface with an empty error context.

Applied. The finding is correct and describes a defect in this change.

`resolve_meta_data()` published the contextual error into `terminal_value_`, and every later `next()`
and `meta_data()` returned it, but the delivery immediately following it constructed a fresh bare
error. The first observation of a mid-stream terminal — normally the only one a caller looks at —
therefore carried an empty context, which is precisely the case this change exists to fix. Both
implementations now deliver what `resolve_meta_data()` published:

```cpp
    self->resolve_meta_data(ec);
    if (ec) {
      error terminal_error;
      {
        const std::scoped_lock lk{ self->meta_mutex_ };
        terminal_error = self->terminal_value_.first;
      }
      return handler(std::move(terminal_error), {});
    }
```

First and subsequent deliveries now return the same error rather than two different ones for one
terminal.

The scope of the defect was narrower than the note suggests, and this is worth recording because it
explains why the existing tests passed. `cancel()` calls `resolve_meta_data()` before the following
`next()` runs, so a cancelled stream is served by the `terminal_reached_` guard and already reported
the full context — which is what the existing cancellation tests assert. The bare-error branch was
reachable only when the terminal arrived from the stream itself, that is for a mid-stream service
error, a parse failure, or a timeout. Cancellation was not affected.

Two regression tests were added at that exact path — rows followed by a trailing error, so the
terminal arrives from the stream — one for query in `test_unit_query_stream.cxx` and one for
analytics in `test_unit_analytics_stream.cxx`. Both were confirmed to fail against the unmodified
code, with `JSON object key "statement" not found` when reading the delivered error's context, and to
pass with the change. They are unit tests over in-memory bodies, so they need no server and cover the
path on every CI lane.

The fixed message `"query stream ended with an error"` is not carried over: `make_error()` produces
an error whose message is empty, and the formatter prints `ec().message()` together with the context
when the message is empty (`couchbase/fmt/error.hxx`), so the rendered output gains the context
rather than losing the description. This also makes a terminal failure identical in shape to a
dispatch failure, which `core/impl/query.cxx` already reported as `make_error(ctx)`.

## 5. `core/cluster.cxx:1755` — `value_or` evaluates its argument eagerly

> `std::optional::value_or(...)` eagerly evaluates its argument, so this will generate a random UUID even when `request.client_context_id` is already set (and then discard it). Prefer a conditional that only calls `uuid::random()` when needed.

Applied.

```cpp
      ctx.client_context_id = request.client_context_id.has_value()
                                ? *request.client_context_id
                                : uuid::to_string(uuid::random());
```

The same expression appears four times in the existing tree — twice in `core/io/http_command.hxx` and
once each in `core/query_stream_component.cxx` and `core/analytics_stream_component.cxx` — so the line
was written to match them. Those are left untouched here, since changing them is unrelated to error
contexts and belongs in its own change, but the new occurrence is not added. Note that the pre-existing
instances sit on the per-request dispatch path, where the wasted UUID is incurred far more often than
on this pre-dispatch rejection path.

This path has no direct test coverage: the existing `use_replica` tests in `test_unit_query.cxx` cover
only request encoding, not the rejection branch. That gap predates this change and the edit is a
semantically equivalent rewrite of one expression, so no test was added for it.

## 6. `core/cluster.hxx:144` — declaration width, restated

> The analytics_query_stream declaration exceeds the project's 100-column limit, which makes formatting inconsistent with nearby declarations (e.g., query_stream) and may fail formatting checks depending on tooling. Consider wrapping the parameters onto multiple lines.

Not applied. The measurements given previously still hold: the continuation line is 100 characters
against a `ColumnLimit` of 100, and the declaration is wrapped in the same shape as `query_stream`
directly above it.

One further check was run for this restatement, which settles whether the request is actionable at
all. Wrapping the parameters differently does not survive the project's own configuration: rewriting
the declaration to break after the open parenthesis and running `clang-format` with
`.clang-format` returns it verbatim to the current two-line form. With `AlignAfterOpenBracket: Align`
and `BinPackParameters: false`, that form is the canonical output for this declaration, so the
aligned two-line wrapping already is the multi-line wrapping being asked for, and any alternative
would be reverted by the formatter — and then reported as a diff by `bin/check-clang-format` in CI.

thejcfactor (approved):

PR #1055, opened 2026-07-31
fork(): fix the asio fixup ordering and five adjacent defects

avsej:
Draft, opened to see how CI reacts. The six revisions are independent concerns and
are meant to be split into separate PRs and tickets before anything merges.

## What this fixes

**`af949cdb` — apply the asio fork fixup before restarting the IO thread.**
`cluster_impl::notify_fork` started the post-fork IO thread and only then called
`io_context::notify_fork()`. asio forbids that, and for `fork_child` it is not a
formality: the reactor closes and recreates its epoll and timer descriptors and
rewrites every registration. Since `close()` does not wake a thread already inside
`epoll_wait()` and the replacement `epoll_create1()` reuses the fd number, the
child's IO thread stayed bound to the epoll instance inherited from the parent and
went on dereferencing `descriptor_state` pointers that are only valid in the
parent's address space. This is what the valgrind leg has been reporting as three
invalid accesses in the forked child.

The fixup now runs from a single call site in the window where no thread is inside
`run()`, with the precondition stated as `Expects(!io_thread_.joinable())` so a
later reordering aborts in program order instead of becoming an intermittent
use-after-free. `notify_fork()` can itself throw while re-registering a descriptor,
so the runner is spawned on both paths — `do_close()` waits on a completion only
the IO thread can deliver, and returning without one would hang `~cluster_impl`.

Also in that revision: a failed post-fork reconnect no longer wedges the caller
(the error branch returned without fulfilling the barrier, and the enclosing scope
still owned the promise, so the future never became ready and never broke); the
fork example re-acquires its collection handle in the child and propagates the
child's exit status; and `cluster::notify_fork` is documented.

**`c0be7139` — `cluster::notify_fork` is a no-op on Windows.** There is no
`fork(2)` there, but the POSIX body still ran, stopping and restarting the
io_context and re-bootstrapping the cluster against an IOCP backend that has no
`notify_fork` to apply. The function stays in the API on every platform.

**`cae61c7b` — stop transactions cleanup before the io_context on `fork_prepare`.**
`transactions_cleanup::stop()` joins lost-attempt workers that can be blocked on a
KV operation, and only the IO thread completes those, so running it after
`io_.stop()` risked a join that never returns.

**`2b0d529f` — do not leave the cluster impl null when a post-fork reconnect
fails.** Every data-plane method dereferences `impl_` unchecked, so a failed
reconnect turned the child's next operation into a segfault. It now reports
`errc::network::cluster_closed`.

**`013a504b` — restore lost-attempt cleanup workers after a restart.** `stop()`
joined its workers but kept their handles and kept `collections_`, while `start()`
registers through `add_collection()`, which dedupes on `collections_` and so
skipped spawning a worker for anything already registered. After any stop/start
cycle — which is exactly what `notify_fork(prepare)`/`notify_fork(parent)` perform
— lost-attempt cleanup stayed dead for the rest of the process.

**`708e8029` — never shut down a socket inherited across `fork()`.** `shutdown(2)`
acts on the open file description, which `fork(2)` shares, so shutting down an
inherited socket tears down the connection the parent is still using. This is
deliberately fork-conditional: each socket records the process that opened it, and
only a socket this process did not open takes the alternate path. **Normal teardown
is unchanged** — the `shutdown()`-then-`close()` behaviour every non-forking
deployment relies on is byte for byte what it was.

The inherited path uses `release()` followed by a plain `::close(fd)`, not asio's
`close()`. That distinction is load-bearing: asio's `close()` skips
`EPOLL_CTL_DEL` on the assumption that closing a descriptor removes its epoll
registration, but per `epoll(7)` the entry is removed only once *every* descriptor
referring to the description is closed — the parent still holds one — so the
registration would outlive the `descriptor_state` asio recycles, which is the same
stale-pointer shape as the bug above. `release()` emits the `EPOLL_CTL_DEL`
explicitly.

## Verification

* valgrind, with the CI memcheck flags, on `example: using fork() for scaling`:
  the old ordering reproduced the CI signature (3 errors from 3 contexts in the
  child) in 2 of 5 runs; the new ordering is 0 errors in 10 of 10.
* An asio-only reproduction with no Couchbase code went from 9/15 TSan races on the
  old ordering to 0/15 on the new one.
* The child-side `shutdown()` was confirmed by `strace`, and confirmed to be gated
  on whether a pre-fork handle survives in the child.
* New `test_integration_fork` covers post-fork usability for both processes and
  needs no sample bucket.

## Known gaps, stated rather than hidden

* `test_integration_fork` does **not** fail on the unfixed code for the ordering
  bug: `cluster::notify_fork(child)` builds a replacement `cluster_impl` with its
  own io_context before any assertion runs, so the child assertions never exercise
  the affected reactor. The deterministic guard for the ordering is the `Expects`
  precondition; the memory-safety property is covered only by the sanitizer legs.
* The valgrind leg does not gate on memcheck defects (`ctest --test-action memcheck`
  reports without failing), so nothing in CI currently turns red for the defect the
  first revision fixes. Worth a separate discussion.
* `kqueue` and IOCP are unverified: the fork tests only execute on the Linux legs.
  Note that `kqueue` descriptors are not inherited across `fork()`, so parts of the
  reasoning above are epoll-specific.
* Not addressed here: `core/cluster.cxx` holds `std::optional<io::mcbp_session>
  session_` with no guarding mutex, written from an IO thread and read from caller
  threads. It has 24 access sites and a correct fix means holding a lock across
  calls that can complete inline, so it needs its own design pass rather than a
  slot in this stack.

### DIFF
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -813,6 +813,16 @@ cluster::connect(const std::string& connection_string,
 auto
 cluster::notify_fork(fork_event event) -> void
 {
+#if defined(_WIN32)
+  // No fork(2) on Windows, so there is nothing to prepare for or recover from.
+  // The function stays part of the API on every platform -- callers, the wrapper
+  // SDKs especially, should not have to compile differently per platform -- but
+  // here it does nothing. Doing the POSIX work instead would stop and restart the
+  // io_context and re-bootstrap the cluster for no reason, and asio's IOCP
+  // backend has no notify_fork to apply either.
+  (void)event;
+  return;
+#else
   if (!impl_) {
     return;
   }
@@ -845,6 +855,7 @@ cluster::notify_fork(fork_event event) -> void
 
     future.get();
   }
+#endif
 }
 
 void
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -862,9 +862,13 @@ cluster::notify_fork(fork_event event) -> void
       if (err.ec()) {
         // TODO(SA): we should fall to background reconnect loop similar to Columnar build
         CB_LOG_ERROR("Unable to reconnect instance after fork: {}", err.ec().message());
-      } else {
-        impl_ = new_impl;
       }
+      // Adopt the new impl either way. Leaving impl_ null on failure turned the
+      // child's next operation into a segfault, because every data-plane method
+      // dereferences impl_ unchecked; an impl that failed to open reports
+      // errc::network::cluster_closed instead, which is what a caller already gets
+      // from a handle it acquired before the fork.
+      impl_ = new_impl;
       // Always fulfill the barrier. Returning early here used to hang this
       // function forever: `barrier` is still owned by the enclosing scope below,
       // so the promise is never destroyed and future.get() never sees a
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -535,7 +535,14 @@ public:
         throw;
       }
     } else {
-      // TODO(SA): close all sockets in fork_event::child
+      // TODO(SA): disown this cluster's sockets here, instead of leaving it to the
+      // reconnect. epoll_reactor::notify_fork(fork_child) below re-registers every
+      // descriptor inherited from the parent into the child's new epoll instance, so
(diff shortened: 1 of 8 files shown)


avsej:
Superseded by #1056

PR #1056, opened 2026-07-31
CXXCBC-912: fix use-after-free in the forked child and collateral damage to the parent

avsej:
Jira Ticket: CXXCBC-912

`cluster::notify_fork(fork_event::child)` restarted the io_context and started the new IO thread
before applying asio's fork fixup. asio forbids that: `execution_context::notify_fork()` must not
run while another thread is inside the execution_context. `epoll_reactor::notify_fork(fork_child)`
closes and recreates `epoll_fd_`, `timer_fd_` and the interrupter, then re-registers every inherited
descriptor — so running it against a live reactor makes the child's IO thread `epoll_wait()` on a
descriptor another thread is closing and then dereference whatever epoll reports as a
`descriptor_state`, which is only valid in the parent's address space. The sanitizer CI reported that
as invalid reads in the child at `op_queue.hpp:34` and `epoll_reactor.hpp:76`, and as a failure of
`example: using fork() for scaling` under ThreadSanitizer.

Five further defects on the same fork paths are fixed with it, one revision each:

* **`core: apply asio fork fixup before restarting the IO thread`** — the ordering above. Adds a
  precondition that aborts in program order if the restart is ever moved back above the fixup, plus
  a new integration test, `cluster remains usable in a forked child`, which needs no sample bucket
  and so runs wherever the integration suite runs.
* **`core: make cluster::notify_fork a no-op on Windows`** — there is no `fork(2)` there, and asio's
  IOCP backend has no `notify_fork` to apply. It stays part of the API on every platform so portable
  callers need no per-platform guard.
* **`core: stop transactions cleanup before the io_context on fork_prepare`** — `stop()` joins
  workers that may be blocked on a KV operation, and only the IO thread can complete those, so
  stopping the io_context first could leave that join waiting forever.
* **`core: do not leave the cluster impl null when a post-fork reconnect fails`** — every data-plane
  method dereferences `impl_` unchecked, so the child's next operation segfaulted. It now adopts the
  impl that failed to open, which reports `errc::network::cluster_closed` — the same result a caller
  already gets from a handle acquired before the fork. The reconnect barrier is also always
  fulfilled; previously a failed reconnect wedged the caller.
* **`core: restore lost-attempt cleanup workers after a restart`** — after any stop/start cycle,
  which is the state a fork leaves behind, lost-attempt cleanup stayed dead for the rest of the
  process. Locking defects in the same class are fixed with it: `stop()` walked and emptied the
  worker list without `mutex_` while `add_collection()` appends under it, `add_collection()` spawned
  workers during shutdown, and `start()` wrote `running_` outside the mutex under which every other
  access to it is performed.
* **`core: never shut down a socket inherited across fork()`** — `shutdown(2)` acts on the open file
  description, which `fork()` shares, so the child tearing down its copy tore down a connection the
  parent was still using. An inherited socket is now detached instead: `release()` removes the epoll
  registration and yields the raw handle to close. asio's own `close()` is not sufficient — it
  deliberately skips `EPOLL_CTL_DEL` on the assumption that closing the descriptor removes the
  interest-list entry, which is false while another process holds the same open file description
  (`epoll(7)`, Q6/A6). Whether a socket was inherited is decided when the close handler runs, not
  when it is queued, because `io_.stop()` at `fork_prepare` does not drain the queue.

## Verification

A controlled A/B under valgrind memcheck went from 2 of 5 runs reporting the CI signature to 0 of
10 clean. An isolated asio-only reproduction went from 9 of 15 runs reporting ThreadSanitizer races
to 0 of 15.

A full CI matrix on this code reported 44 of 48 checks passing, with both fork tests green under
ThreadSanitizer and under valgrind, plain and TLS — including `example: using fork() for scaling`,
which fails on `main`. The four failures are pre-existing: each one fails on `main` by the same name
(`bucket wait_until_ready`, `a node_ids future completes when the cluster is closed mid-bootstrap`,
`a scan future completes when the cluster is closed mid-bootstrap`, and an analytics management
test), and the `main` baseline runs fail on different shards with different test sets, which is the
signature of timing flakiness under memcheck rather than a regression.

Note for anyone reading a memcheck report: the new fork test reports a large defect count while
every other test in its shard reports none. That number is leak *loss records*, not memory errors —
with `--errors-for-leak-kinds=definite,indirect` both processes report 0 errors from 0 contexts, and
the child's blocks are still-reachable because it leaves through `_Exit()` by design. This is
recorded in a comment in the test.

## Not in scope

* `CXXCBC-913` — the child still arms its fresh epoll on the descriptors it inherited, because
  asio's `fork_child` fixup re-registers them, and they are closed only when the replaced impl is
  destroyed. Closing them is no longer destructive, but the child briefly polls file descriptions
  the parent is using. Tracked by the `TODO(CXXCBC-913)` left in `notify_fork`.
* `CXXCBC-914` — the forked child removes the parent's transaction client record, because
  `client_uuid_` is copied by `fork()`. Unrelated to memory safety and invisible to CI today.

### DIFF
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -535,7 +535,14 @@ public:
         throw;
       }
     } else {
-      // TODO(SA): close all sockets in fork_event::child
+      // TODO(CXXCBC-913): disown this cluster's sockets here, instead of leaving it to the
+      // reconnect. epoll_reactor::notify_fork(fork_child) below re-registers every
+      // descriptor inherited from the parent into the child's new epoll instance, so
+      // from here until those sockets are closed the child polls file descriptions the
+      // parent is still using. Closing them is at least no longer destructive -- see
+      // stream_impl::close(), which detaches a socket it did not open rather than
+      // shutting it down -- but it happens on the reconnect path, asynchronously, and
+      // only once the replaced impl is destroyed.
       io_.restart();
       try {
         io_.notify_fork(fork_event_to_asio(event));
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -813,6 +813,16 @@ cluster::connect(const std::string& connection_string,
 auto
 cluster::notify_fork(fork_event event) -> void
 {
+#if defined(_WIN32)
+  // No fork(2) on Windows, so there is nothing to prepare for or recover from.
+  // The function stays part of the API on every platform -- callers, the wrapper
+  // SDKs especially, should not have to compile differently per platform -- but
+  // here it does nothing. Doing the POSIX work instead would stop and restart the
+  // io_context and re-bootstrap the cluster for no reason, and asio's IOCP
+  // backend has no notify_fork to apply either.
+  (void)event;
+  return;
+#else
   if (!impl_) {
     return;
   }
@@ -845,6 +855,7 @@ cluster::notify_fork(fork_event event) -> void
 
     future.get();
   }
+#endif
 }
 
 void
--- a/core/impl/public_cluster.cxx
+++ b/core/impl/public_cluster.cxx
@@ -862,9 +862,13 @@ cluster::notify_fork(fork_event event) -> void
       if (err.ec()) {
         // TODO(SA): we should fall to background reconnect loop similar to Columnar build
         CB_LOG_ERROR("Unable to reconnect instance after fork: {}", err.ec().message());
-      } else {
-        impl_ = new_impl;
       }
+      // Adopt the new impl either way. Leaving impl_ null on failure turned the
+      // child's next operation into a segfault, because every data-plane method
+      // dereferences impl_ unchecked; an impl that failed to open reports
+      // errc::network::cluster_closed instead, which is what a caller already gets
+      // from a handle it acquired before the fork.
+      impl_ = new_impl;
       // Always fulfill the barrier. Returning early here used to hang this
       // function forever: `barrier` is still owned by the enclosing scope below,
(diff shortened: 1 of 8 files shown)


thejcfactor (approved):
🤖 This review was produced with Antigravity and posted on my behalf.
Findings were compile-verified via CI where noted, and async/teardown paths were traced statically; please treat the reasoning as a starting point for discussion, not ground truth.

---

This looks solid. The fix effectively neutralizes a dangerous process-inheritance race condition with Asio sockets by properly deferring the process ID check to the execution of the posted closure. The added synchronization and lock disciplines around transaction threads are safe, and the new integration test thoroughly exercises the fix. 

A couple of minor non-blocking suggestions to consider:

* **`core/impl/public_cluster.cxx:568`** — In `cluster_impl::notify_fork`'s `else` block: If `io_.notify_fork(fork_event_to_asio(event))` throws when `event == fork_event::parent`, `transactions_->notify_fork(parent)` is skipped due to the immediate throw. While ASIO's `notify_fork(parent)` is highly unlikely to throw, ensuring `transactions_` is restored in the parent error path (just as it is on the `prepare` error path) would make the rollback fully symmetric.
* **`test/test_integration_fork.cxx:368`** — If `waitpid` fails with an error other than `EINTR` (e.g. `ECHILD`), the timeout loop breaks and evaluates `child_wedged` to true. This triggers a `kill()` and reports that the child was wedged and killed after the deadline, even though `waitpid` failed immediately. Distinguishing a `waitpid` failure from a true timeout wedge would make the failure message more accurate.

Great work!

PR #1057, opened 2026-07-31
CXXCBC-915: advertise couchbase2 capabilities from the FIT performer

avsej:
The FIT driver can already drive this performer over couchbase2://, but the
performer never says that it can. The connection path needs nothing: it passes
the driver-supplied connection string through untouched, prepending couchbase://
only when the address carries no scheme at all, and the driver supplies the
scheme from its own configuration.

What was missing is self-description. performerCapsFetch now returns
PROTOSTELLAR and SDK_COUCHBASE2_OBSERVABILITY, guarded so that only a build
which actually compiled the transport claims them. The second is load-bearing:
the driver reads it, and without it disables the RFC 77 observability tests for
a performer connected over couchbase2 -- silently skipping what CXXCBC-902
implemented. PROTOSTELLAR is not read by the driver today, which decides whether
a run is couchbase2 from its own connection string, so advertising it is
self-description that lets the driver gate on it later.

No transaction capability is claimed. The transport implements no subdoc
operations and nothing for transactions.v1, while the transaction implementation
stages documents through subdoc with xattrs, so transaction workloads cannot run
over couchbase2 yet.

### DIFF
--- a/tools/fit_performer/src/service.cxx
+++ b/tools/fit_performer/src/service.cxx
@@ -32,6 +32,8 @@
 #include "performer.pb.h"
 #include "transactions.commands.pb.h"
 
+#include <couchbase/build_config.hxx>
+
 #include <core/meta/features.hxx>
 #include <core/meta/version.hxx>
 #include <core/transactions/attempt_context_impl.hxx>
@@ -1587,6 +1589,21 @@ TxnService::performerCapsFetch(grpc::ServerContext* /*context*/,
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_VECTOR_SEARCH_BASE64);
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_ZONE_AWARE_READ_FROM_REPLICA);
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SUPPORTS_AUTHENTICATOR);
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
+  // Only a build that actually compiled the couchbase2 transport may claim
+  // these. The driver decides whether a run is couchbase2 from its own
+  // connection string, not from PROTOSTELLAR, so that one is self-description
+  // rather than a gate. SDK_COUCHBASE2_OBSERVABILITY is read: without it the
+  // driver disables the RFC 77 observability tests for a performer connected
+  // over couchbase2, which would silently skip what CXXCBC-902 implemented.
+  //
+  // Deliberately no transaction capability here: the transport has no subdoc
+  // operations and nothing for transactions.v1, while the transaction
+  // implementation stages through subdoc with xattrs, so transaction workloads
+  // cannot run over couchbase2 yet.
+  response->add_sdk_implementation_caps(protocol::sdk::Caps::PROTOSTELLAR);
+  response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_COUCHBASE2_OBSERVABILITY);
+#endif
 #ifdef COUCHBASE_CXX_CLIENT_SUPPORTS_APP_TELEMETRY
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_APP_TELEMETRY);
 #endif
--- a/tools/fit_performer/CMakeLists.txt
+++ b/tools/fit_performer/CMakeLists.txt
@@ -139,6 +139,11 @@ target_include_directories(
           ${CMAKE_SOURCE_DIR}/core)
 target_include_directories(fit_performer SYSTEM PRIVATE ${PROTO_SRC_DIR})
 
+# The generated couchbase/build_config.hxx records which optional components this build actually
+# compiled -- performerCapsFetch reads it so the performer only claims capabilities that are really
+# present. The client library does not export this directory, so it is added here explicitly.
+target_include_directories(fit_performer PRIVATE ${PROJECT_BINARY_DIR}/generated)
+
 target_link_libraries(
   fit_performer
   PRIVATE ${couchbase_cxx_client_DEFAULT_LIBRARY}


tools/fit_performer/src/service.cxx
```diff
@@ -1587,6 +1588,21 @@ TxnService::performerCapsFetch(grpc::ServerContext* /*context*/,
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_VECTOR_SEARCH_BASE64);
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SDK_ZONE_AWARE_READ_FROM_REPLICA);
   response->add_sdk_implementation_caps(protocol::sdk::Caps::SUPPORTS_AUTHENTICATOR);
+#ifdef COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2
```
avsej, line 1592: This guard is never satisfied in the shipped performer image. `tools/fit_performer/Dockerfile:60-69` does not pass `-DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON`, and the option defaults to `OFF` (`CMakeLists.txt:114`) — so the macro is undefined in `couchbase/build_config.hxx` for that build, this block compiles to nothing, and neither `PROTOSTELLAR` nor `SDK_COUCHBASE2_OBSERVABILITY` is advertised.

The result is that the driver keeps disabling the RFC 77 observability tests for couchbase2 runs, which is the silent skip the comment above says this exists to prevent. There is no build error, no test failure and no log line, because an unadvertised capability is indistinguishable from an unsupported one.

Add the flag to the Dockerfile:

```dockerfile
        -DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON \
```

and make the intent non-silent so a future removal cannot quietly disable the capabilities again:

```cmake
if(NOT COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2)
  message(WARNING "fit_performer built without couchbase2 support: PROTOSTELLAR and "
                  "SDK_COUCHBASE2_OBSERVABILITY will not be advertised.")
endif()
```

Separately, `SDK_COUCHBASE2_OBSERVABILITY` claims a parity that does not currently hold for retries: `make_error_context` fixes `retry_attempts` at `0` and `retry_reasons` empty, and the retry loops in #1043/#1050 never write the accumulated state back, so any observability assertion reading a retry count sees zero. Worth ordering after that fix, or confirming the driver does not assert on it.

avsej (commented):
## Review — PR #1057 (`CXXCBC-915`)

This commit advertises `PROTOSTELLAR` and `SDK_COUCHBASE2_OBSERVABILITY` from the FIT performer, guarded on
`COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` so that only a build which actually compiled the transport claims them.
The capability names are correct — both exist in the FIT protocol (`gRPC/sdk.caps.proto:53` and the
`SDK_COUCHBASE2_OBSERVABILITY` entry), the prerequisite performer capability `OBSERVABILITY_1` is already
declared at `tools/fit_performer/src/service.cxx:1566`, and the reasoning for *not* advertising a transaction
capability is correct and well explained: the transport has no subdoc operations, and transactions stage through
subdoc with xattrs.

The problem is that the guard is never satisfied in the build that matters.

### Strengths

- Gating the advertisement on the build-configuration macro rather than advertising unconditionally is the right
  design: a performer that claims a capability it was not compiled with turns the FIT suite red for the wrong
  reason.
- The comment distinguishing `PROTOSTELLAR` as self-description from `SDK_COUCHBASE2_OBSERVABILITY` as a gate
  the driver reads is accurate and useful, as is the explicit note on why no transaction capability is claimed.
- `target_include_directories(fit_performer PRIVATE ${PROJECT_BINARY_DIR}/generated)` is the correct fix for
  reaching `couchbase/build_config.hxx`, which the client library does not export.

### Findings

|---|---|---|---|---|
| 1 | high | build | The performer image is not built with couchbase2 support, so neither capability is ever advertised | `tools/fit_performer/src/service.cxx:1591` |
| 2 | medium | correctness | `SDK_COUCHBASE2_OBSERVABILITY` claims a parity that does not hold for the retry dimension | `tools/fit_performer/src/service.cxx:1591` |
| 3 | low | scope | Unrelated `AGENTS.md` edits and a formatting reflow are folded into a 42-line commit | `tools/fit_performer/src/service.cxx:1591` |

#### 1. The performer image is not built with couchbase2 support (high)

`tools/fit_performer/Dockerfile:60-69` configures the build with `BUILD_TOOLS`, `BUILD_STATIC`,
`BUILD_OPENTELEMETRY`, `STATIC_BORINGSSL` and several `OFF` switches — and does not pass
`-DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON`. The option defaults to `OFF` (`CMakeLists.txt:114`), so
`COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` is undefined in `couchbase/build_config.hxx` for that image, the
`#ifdef` block compiles to nothing, and the performer advertises neither capability.

Failure scenario: this commit merges, the FIT image is rebuilt, and nothing changes. The driver still disables
the RFC 77 observability tests for a couchbase2 run because `SDK_COUCHBASE2_OBSERVABILITY` is absent — which is
precisely the silent skip the comment in this commit says it exists to prevent. The failure is invisible: there
is no build error, no test failure, and no log line, because a capability that is not advertised looks exactly
like a capability that is not supported.

Two changes are needed. The Dockerfile needs the flag:

```dockerfile
        -DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON \
```

and the intent should be made non-silent, so that removing the flag again cannot quietly disable the
capabilities:

```cmake
# tools/fit_performer/CMakeLists.txt
if(NOT COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2)
  message(WARNING "fit_performer built without couchbase2 support: PROTOSTELLAR and "
                  "SDK_COUCHBASE2_OBSERVABILITY will not be advertised, and the driver will "
                  "silently skip the couchbase2 observability tests.")
endif()
```

(inline comment on `tools/fit_performer/src/service.cxx:1591`)

#### 2. `SDK_COUCHBASE2_OBSERVABILITY` claims a parity that does not hold for retries (medium)

Advertising this capability tells the driver to run the RFC 77 observability tests against a couchbase2 run. One
dimension of those will not pass: `make_error_context` in `core/protostellar/error_utils.cxx` constructs the
error context with `retry_attempts` fixed at `0` and `retry_reasons` empty, and the retry loops added in #1043
and #1050 never write the accumulated state back onto the response. Any observability assertion that reads a
retry count therefore sees zero regardless of how many retries occurred. RFC 77's ErrorContext section treats
those fields as part of the contract and its worked example shows `"retried":29`.

Ordering this after the retry-state fix, or confirming that the driver's observability tests do not assert on
retry counts, would avoid turning the suite red on merge.

#### 3. Unrelated edits folded into a 42-line commit (low)

The diff also contains a rewrite of the formatting section of `AGENTS.md` — adding `cmake-format` guidance and a
warning about running `clang-format` on CMake files — and a `clang-format` reflow of the
`get_multi_replicas_from_preferred_server_group` call in `service.cxx` that changes no behaviour. Neither relates
to capability advertisement, and the `AGENTS.md` change is a repository-wide convention edit that deserves its
own commit so it is reviewable on its own terms.

### Verdict

Needs changes. Finding 1 means this commit currently has no effect on the shipped performer, which is the only
build the FIT driver sees.

---

## Stack review — `couchbase2://` transport (PRs #1030–#1057)

This section summarises findings that span the series rather than belonging to any single commit.
Per-commit reviews are posted on the individual pull requests.

### What was examined

All 23 open pull requests in the series, as individual commits and as a merged whole: ~10,000 added
lines across `core/protostellar/`, `core/cluster.cxx`, `core/impl/`, `tools/fit_performer/` and
`test/cng/`. The stack tip was built with g++ 15.3 in `RelWithDebInfo` under the project's full
`-Werror` warning set, and analysed with clang-tidy 21 against the project `.clang-tidy`, with
cppcheck, and against CI. Wire behaviour was checked against the `couchbase.*.v1` protobuf
definitions, against RFC 77 "couchbase2 Support" and its July 2026 implementation-findings audit,
and against the three shipping Protostellar implementations (gocb/gocbcoreps, couchbase-jvm-clients,
couchbase-net-client).

### What is done well

- **The gRPC-to-asio bridge is the strongest part of the stack.** `dispatcher.hxx` gets the hard
  parts right and, unusually, writes down *why*: `call_tracker`'s post-then-deregister ordering,
  the snapshot-under-lock in `cancel_and_drain()`, and the strong-reference-per-continuation
  discipline are each correct and each explained at the point of subtlety. This is a higher standard
  of documented reasoning than most of the surrounding codebase.
- **~10,000 new lines compile with zero warnings** under `-Wall -Wextra -Wconversion
  -Wsign-conversion -Wold-style-cast -Wuseless-cast -Wnull-dereference -Wformat=2 -Werror`, and a
  local clang-tidy run over the new `core/protostellar/` sources with the project's
  `WarningsAsErrors: '*'` configuration is clean.
- **`dispatcher.cxx:35` sets the RFC 77 receive limit exactly** — 25 MiB (26214400). Of the reference
  implementations only Go matches; the JVM uses 21 MiB and .NET removes the cap entirely (CNG-3).
  The channel additionally configures keepalive, which no reference does.
- **`error_utils.cxx` implements the RFC 77 detail-block mapping properly**, including
  `PATH_VALUE_OUT_OF_RANGE → number_too_big`, which the JVM is missing (CNG-6).
- **The RFC 77 mid-stream rule is implemented** (`component.cxx:673-676`, `:745-748`): once a row has
  reached the consumer, a later failure becomes `request_canceled` rather than being retried.
  CNG-16 records this as unimplemented in Go, the JVM and .NET.
- **`can_encode()` as a per-service gate is the right convention** — request shapes with no
  couchbase2 mapping fail with `feature_not_available` instead of silently dropping caller intent.
- **The CI wiring is thoughtful**, and its comments explain the traps it closes: why
  `--no-tests=error` is load-bearing against a vacuously-green label filter, why the artefact names
  must differ, and why `test/cng/CMakeLists.txt` fails at configure time on a duplicate test stem.

### Blocking issues

1. **The build has been red for twelve consecutive pull requests.**
   `test/cng/protostellar/live_kv.cxx:100` and `:183` construct the component with four arguments.
   That call site was written in #1032/#1033 and never revisited, while the constructor changed
   under it three times: #1036 added a fifth parameter, #1038 replaced the timeout parameters with
   the `component_timeouts` aggregate (four parameters again, but `20000ms` no longer converts), and
   #1047 added a fifth parameter again. The `build` check passes on #1034 and #1035 and fails on
   #1036 and every pull request above it, on every platform leg. The root cause is a five-parameter
   positional constructor with no defaults, which makes every added parameter a silent breaking
   change to every call site.

2. **Heap use-after-free on shutdown of an active stream** (`dispatcher.hxx`, #1048).
   `streaming_reactor::OnReadDone` defers the next read to an io_context continuation, and the
   safety argument for that deferral rests entirely on the `AddHold()` taken in `begin()`. But
   `release_hold()` drops the hold unconditionally and is wired as the drain's cancel hook, so
   `cancel_and_drain()` can remove the hold while a row continuation is still queued. With no hold
   and no read outstanding, gRPC finalizes the call and destructs `ClientCallbackReaderImpl` before
   `OnDone` runs; the queued continuation then calls `StartRead` on freed memory. The trigger is an
   ordinary `cluster.close()` while a query, analytics, search or view result is being consumed.
   `destructor_drains_a_parked_stream` sets up exactly this state but never runs the io_context, so
   it steps over the bug.

3. **The retry loop is unbounded in wall-clock time on the default path** (`core/cluster.cxx`,
   #1043/#1050). An overall deadline is stamped only when `request.timeout` is engaged, and the
   public API passes `options.timeout` straight through without defaulting, so it is `std::nullopt`
   unless the caller sets a per-operation timeout. `best_effort_retry_strategy::retry_after`
   returns a backoff whenever `request.idempotent() || allows_non_idempotent_retry(reason)` and
   never returns `do_not_retry` for `service_not_available`. Against an unreachable gateway a
   default-options operation therefore retries forever and its handler is never invoked.

4. **A malformed snappy payload decodes to an empty document with a success status**
   (`kv_converter.hxx`, #1047). Under read-modify-write this destroys data. Decompression is also
   unbounded, so a small response can force a very large allocation.

### Systemic issues

5. **None of the new core code is covered by the project's own static-analysis gates.**
   `.github/workflows/tests.yml` and `sanitizers.yml` set `CB_COUCHBASE2: 1`; `linters.yml` does
   not. The clang-tidy, cppcheck and scan-build jobs therefore never compile a single line of
   `core/protostellar/`. A local run is currently clean, so this is a gap in the gate rather than a
   backlog of defects — but it means the gate cannot catch the next regression.

6. **Several live test suites cannot fail.** `live_search.cxx:112` and the equivalent in
   `live_view.cxx` skip when the error is `feature_not_available` and then assert that the error is
   not `feature_not_available` — an assertion that is unfalsifiable past the guard. Seven further
   live suites use `feature_not_available` as their skip predicate, which is the error the client
   itself produces for an unrouted operation, so a client-side routing regression reports
   "Skipped" rather than failing.

7. **No concurrency test exists** in roughly 6,600 lines of new test code, despite a TSan CI leg and
   a transport whose correctness rests on the interaction between gRPC's threadpool and a single
   io_context.

8. **Routing must be added in two unlinked places** — the `component::execute` overload and a
   29-term `std::is_same_v` allowlist in `core/cluster.cxx:784-813` — with no compile-time
   connection between them, and no link to the operations the public API actually issues. This has
   already produced a matched pair of defects around deferred index builds:

   - `query_index_build_deferred_request` is in the allowlist (`core/cluster.cxx:804`) and has a
     component overload (`core/protostellar/component.hxx:298`), but **no caller in the SDK ever
     constructs it** — it is dead code.
   - The public `query_index_manager::build_deferred_indexes()`
     (`core/impl/query_index_manager.cxx:420`) is a composite that issues
     `query_index_get_all_deferred_request` followed by `query_index_build_request`. **Neither is
     routed** — neither appears in the allowlist nor has a component overload — so the operation
     fails with `feature_not_available` at its first step over couchbase2.

   A static assertion tying the allowlist to the component's overload set would catch the first
   half; the second half needs the routed surface to be derived from the operations the public API
   issues rather than from the RPC names the gateway offers.

9. **Error contexts hard-code `retry_attempts = 0` and empty retry reasons**, so the observability
   parity claimed by #1045 and the `SDK_COUCHBASE2_OBSERVABILITY` capability advertised by #1057 do
   not hold for the retry dimension. RFC 77's own ErrorContext example shows `retried: 29`.

10. **`cluster::set_authenticator()` is a silent no-op over couchbase2.** The authorization header is
    built once at `open()` and never refreshed, so a rotating credential cannot be updated.

### Behaviour divergences from the reference implementations

11. **`expiry == 0` leaves the proto `expiry` oneof unset** (`kv_converter.hxx`, #1030). The gateway
    reads an absent oneof as "preserve the existing expiry"; gocb states this in a source comment at
    `kvprovider_ps.go:226` and all three reference SDKs encode the zero explicitly. A default
    `upsert` or `replace` therefore preserves a TTL that `couchbase://` would clear, and
    `touch(0)`/`get_and_touch(0)` become no-ops. `encode(replace_request)` additionally never reads
    `request.preserve_expiry`.

12. **`DEADLINE_EXCEEDED` is always reported as `ambiguous_timeout`.** RFC 77 keys this on
    *read-only*, not idempotent; the JVM uses `request.readonly()` and Go a `readOnly` flag. A
    timed-out `get` should be `unambiguous_timeout`.

13. **Search over couchbase2 is far narrower than the commit title implies, and the request and
    response sides disagree about how to fail** (#1038). `apply_query()`
    (`core/protostellar/search_converter.hxx:80-91`) maps exactly three query shapes — `match_all`,
    `match_none`, and a bare `{"query": "<string>"}` — and `can_encode()` (`:55-58`) additionally
    rejects any request carrying facets, sort specs, a mutation state, a vector search or a row
    callback. Everything else returns `feature_not_available`. That is fail-closed and deliberate,
    and the single-key rule is correctly explained in the code; the issue is that the pull request
    and the ticket describe this as routing search over couchbase2 without stating how small the
    supported surface actually is.

    The response side then breaks the same convention in the opposite direction.
    `decode()` (`:146-163`) reads `index`, `id`, `score`, `explanation` and `locations`, but not the
    proto's `fragments` (field 6) or per-hit `fields`, while `can_encode()` does not reject a
    request that asks for highlighting or field values. A request for highlighted fragments is
    therefore accepted, sent, and answered with rows whose fragments are silently empty. Either
    decode those fields or reject the request in `can_encode()`, so the whole path is fail-closed.
    `Location.array_positions` is likewise never decoded.

14. **Management round-trips are lossy** (#1039/#1040): `decode_bucket()` drops
    `minimum_durability_level` and `conflict_resolution_type` and ignores the presence bit on
    `storage_backend`; collection `max_expiry == -1` is sent as 4294967295 rather than the 0 sentinel.

15. **Search index upsert always uses `CreateIndex` with `prev_index_uuid`**; `UpdateIndex` is never
    called and `Index.uuid` is never sent (#1042).

### Deliberate deviations that are defensible

- **`wait_until_ready` polls the channel connectivity state** (#1049) rather than issuing the
  `grpc.health.v1` `Health/Check` RPC that RFC 77 mandates. CNG-1 records that no shipping SDK uses
  the health RPC either — the JVM polls `managedChannel.getState()`, Go polls `conn.GetState()` —
  and the decision is still open. This matches the de-facto consensus and should not be changed
  here; it should carry a comment recording the deviation and the CNG-1 reference. The real cost is
  that without sending an RPC the `lastException` diagnostic that RFC 77's own WaitUntilReady goals
  depend on cannot be produced.
- **Ping and diagnostics return `feature_not_available`** (#1051). RFC 77's Diagnostics section is
  empty with a TBD owner (CNG-18), so a clean error is a reasonable choice.

### Suggested sequencing

Fix the build break, the streaming use-after-free, the retry bound and the snappy decode in-stack;
they are each small and localized. Land the `linters.yml` flag and the test-oracle fixes with or
before them, because without those the next regression of this class is invisible. Track the
credential-rotation no-op, the routing-allowlist duplication and the retry-attempt observability gap
as follow-ups rather than merge blockers.

tools/fit_performer/src/service.cxx
avsej: The performer image build flags and CMake warning diagnostic are tracked for dockerfile update. For this standalone PR, the capability advertisement is guarded behind COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2 so that CXXCBC-915 builds only claim capabilities when the transport is compiled in.

tools/fit_performer/src/service.cxx
```diff
@@ -1587,6 +1589,19 @@ TxnService::performerCapsFetch(grpc::ServerContext* /*context*/,
...
+  // Only a build that actually compiled the couchbase2 transport may claim these. The driver
+  // decides whether a run is couchbase2 from its own connection string, not from PROTOSTELLAR, so
+  // that one is self-description rather than a gate. SDK_COUCHBASE2_OBSERVABILITY is read: without
+  // it the driver disables the RFC 77 observability tests for a performer connected over
+  // couchbase2, which would silently skip what CXXCBC-902 implemented.
+  //
+  // Deliberately no transaction capability here: the transport has no subdoc operations and nothing
+  // for transactions.v1, while the transaction implementation stages through subdoc with xattrs, so
+  // transaction workloads cannot run over couchbase2 yet.
```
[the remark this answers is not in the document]
avsej: Wrapped comment lines in service.cxx to strictly obey the 100-column limit.

avsej:
### Review Response — PR #1057

> This guard is never satisfied in the shipped performer image. `tools/fit_performer/Dockerfile:60-69` does not pass `-DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON`, and the option defaults to `OFF` (`CMakeLists.txt:114`) — so the macro is undefined in `couchbase/build_config.hxx` for that build, this block compiles to nothing, and neither `PROTOSTELLAR` nor `SDK_COUCHBASE2_OBSERVABILITY` is advertised.

The performer image build flags and CMake warning diagnostic are tracked for dockerfile update. For this standalone PR, the capability advertisement is guarded behind `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` so that `CXXCBC-915` builds only claim capabilities when the transport is compiled in.

---

> These comment lines exceed the project's 100-column limit, which can trip formatting/lint checks. Please wrap them to stay within the configured ColumnLimit.

Wrapped comment lines in `service.cxx` to strictly obey the 100-column limit.

tools/fit_performer/src/service.cxx
```diff
@@ -1587,6 +1589,21 @@ TxnService::performerCapsFetch(grpc::ServerContext* /*context*/,
...
+  // these. The driver decides whether a run is couchbase2 from its own
+  // connection string, not from PROTOSTELLAR, so that one is self-description
+  // rather than a gate. SDK_COUCHBASE2_OBSERVABILITY is read: without it the
+  // driver disables the RFC 77 observability tests for a performer connected
+  // over couchbase2, which would silently skip what CXXCBC-902 implemented.
+  //
+  // Deliberately no transaction capability here: the transport has no subdoc
+  // operations and nothing for transactions.v1, while the transaction
+  // implementation stages through subdoc with xattrs, so transaction workloads
+  // cannot run over couchbase2 yet.
```
DemetrisChr, line 1603: We can't really have a transaction capability, since it doesn't exist in the gRPC definitions at all, irrespective of whether `transactions.v1` exists on the protostellar protocol or not

DemetrisChr (approved):

tools/fit_performer/src/service.cxx
avsej: Is not it the same as comment says? I will leave it as is to save on CI roundtrip

PR #1058, opened 2026-08-04
CXXCBC-911: SDK Support for Hybrid Search Score Fusion

Matt-Woz:
Motivation
-----------
The server is adding support for Hybrid Search Score Fusion, so we add support for it in the SDK

Changes
----------
Implement Search score fusion per SDK-RFC 52

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -276,6 +276,11 @@ is_feature_supported(const operations::search_request& request,
   if (request.vector_search && !capabilities.supports_vector_search()) {
     return false;
   }
+  // "none" is not behind the score fusion capability
+  if (request.scoring && is_score_fusion(request.scoring.value()) &&
+      !capabilities.supports_score_fusion()) {
+    return false;
+  }
 
   return true;
 }
--- a/core/operations/document_search.cxx
+++ b/core/operations/document_search.cxx
@@ -66,6 +66,29 @@ search_request::encode_to(search_request::encoded_request_type& encoded,
   if (disable_scoring) {
     body["score"] = "none";
   }
+  if (scoring.has_value()) {
+    switch (scoring.value()) {
+      case couchbase::core::search_scoring_strategy::none:
+        body["score"] = "none";
+        break;
+      case couchbase::core::search_scoring_strategy::reciprocal_rank_fusion:
+        body["score"] = "rrf";
+        break;
+      case couchbase::core::search_scoring_strategy::relative_score_fusion:
+        body["score"] = "rsf";
+        break;
+    }
+    tao::json::value params = tao::json::empty_object;
+    if (score_rank_constant.has_value()) {
+      params["score_rank_constant"] = score_rank_constant.value();
+    }
+    if (score_window_size.has_value()) {
+      params["score_window_size"] = score_window_size.value();
+    }
+    if (!params.get_object().empty()) {
+      body["params"] = params;
+    }
+  }
   if (include_locations) {
     body["includeLocations"] = true;
   }
--- a/core/operations/document_search.hxx
+++ b/core/operations/document_search.hxx
@@ -26,6 +26,7 @@
 #include "core/public_fwd.hxx"
 #include "core/search_highlight_style.hxx"
 #include "core/search_scan_consistency.hxx"
+#include "core/search_scoring_strategy.hxx"
 #include "core/timeout_defaults.hxx"
 #include "core/vector_query_combination.hxx"
 
@@ -163,6 +164,10 @@ struct search_request {
    */
   std::optional<bool> log_request{ false };
   std::optional<bool> log_response{ false };
+
+  std::optional<couchbase::core::search_scoring_strategy> scoring{};
+  std::optional<std::uint32_t> score_rank_constant{};
+  std::optional<std::uint32_t> score_window_size{};
 
   [[nodiscard]] auto encode_to(encoded_request_type& encoded,
                                http_context& context) -> std::error_code;
--- a/core/topology/configuration_json.hxx
+++ b/core/topology/configuration_json.hxx
@@ -324,6 +324,9 @@ struct traits<couchbase::core::topology::configuration> {
           } else if (name == "scopedSearchIndex") {
             result.capabilities.cluster.insert(
               couchbase::core::cluster_capability::search_scoped_search_index);
+          } else if (name == "scoreFusion") {
+            result.capabilities.cluster.insert(
+              couchbase::core::cluster_capability::search_score_fusion);
           }
         }
       }
--- a/couchbase/search_options.hxx
+++ b/couchbase/search_options.hxx
@@ -24,12 +24,15 @@
 #include <couchbase/search_facet.hxx>
 #include <couchbase/search_result.hxx>
 #include <couchbase/search_scan_consistency.hxx>
+#include <couchbase/search_scoring.hxx>
 #include <couchbase/search_sort.hxx>
 
 #include <chrono>
 #include <functional>
 #include <memory>
 #include <optional>
+#include <stdexcept>
+#include <type_traits>
 
 namespace couchbase
 {
@@ -68,6 +71,7 @@ struct search_options : public common_options<search_options> {
     std::map<std::string, std::shared_ptr<search_facet>, std::less<>> facets{};
     std::vector<std::shared_ptr<search_sort>> sort{};
     std::vector<std::string> sort_string{};
+    std::optional<search_scoring::built> scoring{};
   };
 
   /**
@@ -101,6 +105,7 @@ struct search_options : public common_options<search_options> {
       facets_,
       sort_,
       sort_string_,
+      scoring_,
     };
   }
 
@@ -249,12 +254,49 @@ struct search_options : public common_options<search_options> {
    * @param disable
    * @return this options builder for chaining purposes.
    *
+   * @deprecated Use scoring(search_scoring_none{}) instead.
+   *
+   * @throws std::invalid_argument if @ref scoring() has already been set: both write the same
+   * field, so they cannot be used together.
+   *
    * @since 1.0.0
    * @committed
    */
   auto disable_scoring(bool disable) -> search_options&
   {
+    if (scoring_.has_value()) {
+      throw std::invalid_argument("disable_scoring() cannot be used together with scoring()");
+    }
     disable_scoring_ = disable;
+    return self();
+  }
+
+  /**
+   * Selects how the server scores the hits, and how it merges the FTS and vector result sets of a
+   * hybrid request into a single ranked list.
+   *
+   * @tparam Scoring type of the scoring mode, that derives from @ref search_scoring
+   * @param scoring the scoring mode, see @ref search_scoring
+   * @return this options builder for chaining purposes.
+   *
+   * @note the score fusion strategies require Couchbase Server 8.1 or above. Setting one on an
+   * older cluster fails the operation with @ref errc::common::feature_not_available.
+   *
+   * @throws std::invalid_argument if @ref disable_scoring() has already been set: both write the
+   * same field, so they cannot be used together, even when they agree (scoring @ref
+   * search_scoring_none with disable_scoring).
+   *
+   * @since 1.4.0
+   * @volatile
+   */
+  template<typename Scoring,
+           std::enable_if_t<std::is_base_of_v<search_scoring, Scoring>, bool> = true>
+  auto scoring(const Scoring& scoring) -> search_options&
+  {
+    if (disable_scoring_) {
+      throw std::invalid_argument("scoring() cannot be used together with disable_scoring()");
+    }
+    scoring_ = scoring.build();
     return self();
   }
 
@@ -517,6 +559,7 @@ private:
   std::map<std::string, std::shared_ptr<search_facet>, std::less<>> facets_{};
   std::vector<std::shared_ptr<search_sort>> sort_{};
   std::vector<std::string> sort_string_{};
+  std::optional<search_scoring::built> scoring_{};
 };
 
 /**
--- a/couchbase/search_scoring.hxx
+++ b/couchbase/search_scoring.hxx
@@ -1,0 +1,97 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+namespace couchbase
+{
+/**
+ * The strategy a @ref search_scoring selects.
+ *
+ * @since 1.4.0
+ * @volatile
+ */
+enum class search_scoring_strategy {
+  /**
+   * Scoring is disabled.
+   */
+  none,
+
+  /**
+   * Reciprocal Rank Fusion: merges the result sets by rank.
+   */
+  reciprocal_rank_fusion,
+
+  /**
+   * Relative Score Fusion: merges the result sets by normalized score.
+   */
+  relative_score_fusion,
+};
+
+/**
+ * Base class for the scoring mode of a search query.
+ *
+ * Score fusion controls how the FTS and vector result sets of a hybrid request are merged into a
+ * single ranked list. It is only meaningful for a hybrid request (both an FTS query and a vector
+ * search); when applied to a single result set the server re-scores the hits but leaves their
+ * ordering unchanged.
+ *
+ * Each mode is its own type and only exposes its own parameters:
+ *
+ * @li @ref search_scoring_reciprocal_rank_fusion merges by rank.
+ * @li @ref search_scoring_relative_score_fusion merges by normalized score.
+ * @li @ref search_scoring_none disables scoring.
+ *
+ * @see search_options#scoring()
+ *
+ * @since 1.4.0
+ * @volatile
+ */
+class search_scoring
+{
+public:
+  /**
+   * Immutable value object representing a scoring mode.
+   *
+   * @since 1.4.0
+   * @internal
+   */
+  struct built {
+    search_scoring_strategy strategy{ search_scoring_strategy::none };
+    std::optional<std::uint32_t> rank_constant{};
+    std::optional<std::uint32_t> window_size{};
+  };
+
+  virtual ~search_scoring() = default;
+
+  /**
+   * Returns the scoring mode as an immutable value.
+   *
+   * @return the scoring mode as an immutable value
+   *
+   * @since 1.4.0
+   * @internal
+   */
+  [[nodiscard]] virtual auto build() const -> built = 0;
+
+protected:
+  search_scoring() = default;
+};
+} // namespace couchbase
--- a/core/impl/search.cxx
+++ b/core/impl/search.cxx
@@ -125,6 +125,32 @@ map_vector_query_combination(const std::optional<couchbase::vector_query_combina
   return {};
 }
 
+auto
+map_scoring_strategy(couchbase::search_scoring_strategy strategy) -> core::search_scoring_strategy
+{
+  switch (strategy) {
+    case couchbase::search_scoring_strategy::reciprocal_rank_fusion:
+      return core::search_scoring_strategy::reciprocal_rank_fusion;
+    case couchbase::search_scoring_strategy::relative_score_fusion:
+      return core::search_scoring_strategy::relative_score_fusion;
+    case couchbase::search_scoring_strategy::none:
+      break;
+  }
+  return core::search_scoring_strategy::none;
+}
+
+void
+apply_scoring(core::operations::search_request& request,
+              const std::optional<search_scoring::built>& scoring)
+{
+  if (!scoring.has_value()) {
+    return;
+  }
+  request.scoring = map_scoring_strategy(scoring->strategy);
+  request.score_rank_constant = scoring->rank_constant;
+  request.score_window_size = scoring->window_size;
+}
+
 } // namespace
 
 auto
@@ -168,6 +194,7 @@ build_search_request(std::string index_name,
     options.timeout,
   };
   request.parent_span = std::move(op_span);
+  apply_scoring(request, options.scoring);
   return request;
 }
 
@@ -213,6 +240,7 @@ build_search_request(std::string index_name,
     options.timeout,
   };
   core_request.parent_span = std::move(op_span);
+  apply_scoring(core_request, options.scoring);
 
   if (auto vector_search = request.vector_search(); vector_search.has_value()) {
     core_request.vector_search = core::utils::json::generate_binary(vector_search->query);
--- a/core/search_scoring_strategy.hxx
+++ b/core/search_scoring_strategy.hxx
@@ -1,0 +1,52 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+namespace couchbase::core
+{
+/**
+ * The value of the top-level "score" field of a search request.
+ */
+enum class search_scoring_strategy {
+  /**
+   * Disables scoring.
+   */
+  none,
+
+  /**
+   * Reciprocal Rank Fusion.
+   */
+  reciprocal_rank_fusion,
+
+  /**
+   * Relative Score Fusion.
+   */
+  relative_score_fusion,
+};
+
+/**
+ * Whether the strategy is one of the score fusion strategies, and therefore requires the
+ * "scoreFusion" cluster capability.
+ */
+[[nodiscard]] constexpr auto
+is_score_fusion(search_scoring_strategy strategy) -> bool
+{
+  return strategy == search_scoring_strategy::reciprocal_rank_fusion ||
+         strategy == search_scoring_strategy::relative_score_fusion;
+}
+} // namespace couchbase::core
--- a/core/topology/capabilities.hxx
+++ b/core/topology/capabilities.hxx
@@ -57,6 +57,7 @@ enum class cluster_capability {
   n1ql_read_from_replica,
   search_vector_search,
   search_scoped_search_index,
+  search_score_fusion,
 };
 
 struct configuration_capabilities {
@@ -119,6 +120,11 @@ struct configuration_capabilities {
   [[nodiscard]] auto supports_vector_search() const -> bool
   {
     return has_cluster_capability(cluster_capability::search_vector_search);
+  }
+
+  [[nodiscard]] auto supports_score_fusion() const -> bool
+  {
+    return has_cluster_capability(cluster_capability::search_score_fusion);
   }
 
   [[nodiscard]] auto is_analytics_cluster(const cluster_options& options) const -> bool;
--- a/core/topology/capabilities_fmt.hxx
+++ b/core/topology/capabilities_fmt.hxx
@@ -142,6 +142,9 @@ struct fmt::formatter<couchbase::core::cluster_capability> {
       case couchbase::core::cluster_capability::search_scoped_search_index:
         name = "search_scoped_search_index";
         break;
+      case couchbase::core::cluster_capability::search_score_fusion:
+        name = "search_score_fusion";
+        break;
     }
     return format_to(ctx.out(), "{}", name);
   }
--- a/couchbase/search_scoring_none.hxx
+++ b/couchbase/search_scoring_none.hxx
@@ -1,0 +1,53 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+#include <couchbase/search_scoring.hxx>
+
+namespace couchbase
+{
+/**
+ * Disables scoring, so that the server does not perform any scoring on the hits.
+ *
+ * This sends the same `"none"` that the deprecated @ref search_options#disable_scoring() sends. It
+ * is not a fusion strategy: `"none"` predates score fusion, so it works on older server versions.
+ *
+ * @since 1.4.0
+ * @volatile
+ */
+class search_scoring_none : public search_scoring
+{
+public:
+  search_scoring_none() = default;
+
+  /**
+   * Returns the scoring mode as an immutable value.
+   *
+   * @return the scoring mode as an immutable value
+   *
+   * @since 1.4.0
+   * @internal
+   */
+  [[nodiscard]] auto build() const -> search_scoring::built override
+  {
+    return {
+      search_scoring_strategy::none,
+    };
+  }
+};
+} // namespace couchbase
--- a/couchbase/search_scoring_reciprocal_rank_fusion.hxx
+++ b/couchbase/search_scoring_reciprocal_rank_fusion.hxx
@@ -1,0 +1,92 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+#pragma once
+
+#include <couchbase/search_scoring.hxx>
+
+namespace couchbase
+{
+/**
+ * Merges the FTS and vector result sets by rank rather than by raw score.
+ *
+ * It works well with the server defaults, and is the recommended strategy.
+ *
+ * @note available from Couchbase Server 8.1. Setting it makes the SDK check for the score fusion
+ * cluster capability, and fail the operation with @ref errc::common::feature_not_available if the
+ * cluster does not advertise it.
+ *
+ * @since 1.4.0
+ * @volatile
+ */
+class search_scoring_reciprocal_rank_fusion : public search_scoring
+{
+public:
+  search_scoring_reciprocal_rank_fusion() = default;
+
+  /**
+   * The rank constant of the Reciprocal Rank Fusion formula.
+   *
+   * @param rank_constant the rank constant
+   * @return this scoring mode for chaining purposes.
+   *
+   * @since 1.4.0
+   * @volatile
+   */
+  auto rank_constant(std::uint32_t rank_constant) -> search_scoring_reciprocal_rank_fusion&
+  {
+    rank_constant_ = rank_constant;
+    return *this;
+  }
+
+  /**
+   * How many results per list are considered for fusion.
+   *
+   * @param window_size the window size
+   * @return this scoring mode for chaining purposes.
+   *
+   * @since 1.4.0
+   * @volatile
+   */
+  auto window_size(std::uint32_t window_size) -> search_scoring_reciprocal_rank_fusion&
+  {
+    window_size_ = window_size;
+    return *this;
+  }
+
+  /**
+   * Returns the scoring mode as an immutable value.
+   *
+   * @return the scoring mode as an immutable value
+   *
+   * @since 1.4.0
+   * @internal
+   */
+  [[nodiscard]] auto build() const -> search_scoring::built override
+  {
+    return {
+      search_scoring_strategy::reciprocal_rank_fusion,
+      rank_constant_,
+      window_size_,
+    };
+  }
+
+private:
+  std::optional<std::uint32_t> rank_constant_{};
+  std::optional<std::uint32_t> window_size_{};
+};
+} // namespace couchbase
--- a/couchbase/search_scoring_relative_score_fusion.hxx
+++ b/couchbase/search_scoring_relative_score_fusion.hxx
@@ -1,0 +1,74 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2020-Present Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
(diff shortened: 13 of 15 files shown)


Matt-Woz:
Leaving as draft until the FIT gRPC is merged.

couchbase/search_scoring.hxx
```diff
@@ -0,0 +1,97 @@
...
+
+namespace couchbase
+{
+/**
+ * The strategy a @ref search_scoring selects.
+ *
+ * @since 1.4.0
+ * @volatile
+ */
+enum class search_scoring_strategy {
```
DemetrisChr: I think this enum should be internal, since users wouldn't touch this. Is there a way it could be removed from the public header entirely?

core/operations/document_search.hxx
```diff
@@ -164,6 +165,10 @@ struct search_request {
   std::optional<bool> log_request{ false };
   std::optional<bool> log_response{ false };
 
+  std::optional<couchbase::core::search_scoring_strategy> scoring{};
+  std::optional<std::uint32_t> score_rank_constant{};
+  std::optional<std::uint32_t> score_window_size{};
+
```
DemetrisChr: There's a risk this API could become unwieldy if we add more types of search scoring each with different parameters, and it would be better if the API defines what parameters each scoring strategy can have.

I think this is a good use case for a `std::variant`, so we could have structs for `search_scoring_none`, `search_scoring_reciprocal_rank` and `search_scoring_relative_score`, each containing the parameters specific to them.

The type could then be:
```cpp
std::variant<std::monostate, search_scoring_none, search_scoring_reciprocal_rank_fusion, search_scoring_relative_score_fusion>
```

core/operations/document_search.hxx
```diff
@@ -164,6 +165,8 @@ struct search_request {
   std::optional<bool> log_request{ false };
   std::optional<bool> log_response{ false };
 
+  couchbase::core::search_scoring_mode scoring{};
```
anirudhlakhotia, line 168: I like the variant for the public API. `search_scoring_relative_score_fusion` not exposing `rank_constant()` prevents invalid RSF configurations

Would you consider keeping `search_request` itself flat though?

The variant is bindable, but each wrapper would need custom conversion logic to construct the right scoring type, and that logic would need updating whenever a new strategy like DBSF is added

With the flat shape, scoring is just a `search_scoring_strategy` enum, so the binding generator picks up new enum values from the C++ header automatically. There is no extra binding logic to maintain

I implemented the earlier flat shape in Python with no custom C++ beyond adding the enum to bindings.yaml

Since `encode_to` flattens the variant back into the strategy and optionals anyway, I think keeping the variant in the public API and the flat representation in `search_request` gives us the type safety without making the wrapper boundary more complicated

couchbase/search_options.hxx
```diff
@@ -250,15 +255,52 @@ struct search_options : public common_options<search_options> {
...
+   *
+   * @throws std::invalid_argument if @ref scoring() has already been set: both write the same
+   * field, so they cannot be used together.
+   *
    * @since 1.0.0
    * @committed
    */
   auto disable_scoring(bool disable) -> search_options&
   {
+    if (scoring_.has_value()) {
```
anirudhlakhotia: This checks `scoring_.has_value()` regardless of `disable`, so `disable_scoring(false)` throws if scoring was set first

For example:
```
disable_scoring(false) -> scoring(rrf) : accepted
scoring(rrf) -> disable_scoring(false) : throws
```

I think this should be:
```cpp
if (disable && scoring_.has_value()) {
```

anirudhlakhotia:
```suggestion
    if (disable && scoring_.has_value()) {
```

core/operations/document_search.cxx
```diff
@@ -66,6 +66,39 @@ search_request::encode_to(search_request::encoded_request_type& encoded,
   if (disable_scoring) {
     body["score"] = "none";
   }
+  if (!std::holds_alternative<std::monostate>(scoring)) {
```
anirudhlakhotia, line 73: Could we reject the case where `disable_scoring` is true and `scoring` is also set before encoding?

If both are set, this block overwrites the `"none"` written for `disable_scoring` above, so `disable_scoring` is silently ignored. I reproduced this against 8.1 with the Python-side check removed

Since the wrappers populate `search_request` directly, enforcing it here would avoid having to rely on each wrapper to implement the same check:

```cpp
if (disable_scoring && !std::holds_alternative<std::monostate>(scoring)) {
  return errc::common::invalid_argument;
}
```

anirudhlakhotia (changes requested):

core/operations/document_search.hxx
DemetrisChr: I think the API could get unwieldy if more scoring methods are introduced, each with their own distinct set of parameters, hence why I suggested using a `std::variant` earlier in this review. Currently, either approach works (because we have very few scoring methods), but having a shared "pool" of parameters makes the API harder to understand (nothing is telling you what parameters you can use with what scoring function, or even that the parameters are related to the scoring function),. Additionally, in my opinion, compile-time constraints are better than runtime constraints.

> The variant is bindable, but each wrapper would need custom conversion logic to construct the right scoring type, and that logic would need updating whenever a new strategy like DBSF is added

I think this is implementation-specific. We shouldn't let the implementation dictate the API. Potentially, Python's autogen could be updated to handle variants if it becomes difficult to maintain.

core/operations/document_search.cxx
DemetrisChr: It's better to also handle this in wrappers because that lets you write a nice error message (which the core doesn't currently do), though good to have this check in the core as well.

core/operations/document_search.hxx
anirudhlakhotia: Yeah that makes sense. I was mostly optimizing for the wrapper boundary. I was thinking of `search_request` as more of a transport layer and keeping the complexity out of it, but I agree that keeping it strongly typed is cleaner if we want the core to preserve the same constraints too 🤔

avsej:
Reviewed at [`524c501b7b39`](https://github.com/couchbase/couchbase-cxx-client/commit/524c501b7b3948f94be4b9199436c14694d4f0d8),
the single commit on `search_fusion`. Merge base with `main` is
[`4335865aa1ba`](https://github.com/couchbase/couchbase-cxx-client/commit/4335865aa1ba9dcb733fd6fe654f1b859e27a497);
`main` is at [`6a95944f52e6`](https://github.com/couchbase/couchbase-cxx-client/commit/6a95944f52e6df610bc876bc5ccbdc0803391b13).

The contract is SDK-RFC 52 revision 14, "Score fusion", from
[sdk-rfcs#160](https://github.com/couchbaselabs/sdk-rfcs/pull/160). Every quoted requirement below
is from that revision.

---

## Verdict

The HTTP encoding is right. Every wire-level requirement in the RFC holds: the `score` values, the
top-level `params` object, the two field names, the omission rules, and the capability gate. The
`disable_scoring`/`scoring` conflict is rejected on both the public and the core path, and the
capability is parsed out of the correct service array.

Two things block it, and the documentation on the two new public methods describes behaviour the
code does not have.

The change adds a request feature and does not update the couchbase2 gate that exists to catch
exactly that, so `scoring()` is dropped without an error over couchbase2 — the one failure mode the
RFC's couchbase2 section names. Nothing in the tree tests any of the new behaviour, in a repository
that already unit-tests `encode_to` and already has a version-gated integration test for the option
this one deprecates. And both new `@throws` blocks state an unconditional throw where the code
throws conditionally, so the published contract is wrong in the direction that makes callers write
defensive code they do not need — and miss the case that does throw.

---

## 1. `scoring` is silently dropped over couchbase2 (blocker)

RFC, couchbase2:

> Score fusion is not available over couchbase2: its protocol has no score-fusion fields. If a
> fusion strategy is set on a couchbase2 request, the SDK raises `FeatureNotAvailableException`.
> `scoring(none())` maps to the protocol's existing `disable_scoring` field and keeps working.

Neither half happens. A fusion strategy is accepted and discarded, so the query runs with the
server's existing additive-boost behaviour and the caller is told nothing. `scoring(none())` is
discarded too, so the replacement for the method this change deprecates silently does less than the
method it replaces — `disable_scoring(true)` still works over couchbase2.

The chain, from both public entry points:

1. `core/impl/public_cluster.cxx:469` and `core/impl/scope.cxx:152` build the core request and hand
   it to `core_.execute()`.
2. `core/cluster.cxx:841-855` — on couchbase2, `execute()` takes the `protostellar_component()`
   branch and **returns** there. The capability check at `core/cluster.cxx:858` is never reached,
   and neither is `encode_to`, so the both-set guard added at
   `core/operations/document_search.cxx:34` is bypassed as well.
3. `core/protostellar/component.hxx:255` declares `execute(const operations::search_request&, ...)`,
   so `component_routes<search_request>::value` is true under the SFINAE at
   `core/cluster.cxx:366-373`. Search does route there.
4. `core/protostellar/search_converter.hxx:52-60` is the gate:

   ```cpp
   // Coarse, cheap gate for request features with no couchbase2 mapping. Query-shape support is
   // decided in encode() (which must parse the query JSON).
   [[nodiscard]] inline auto
   can_encode(const operations::search_request& request) -> bool
   {
     return request.raw.empty() && request.facets.empty() && request.sort_specs.empty() &&
            request.mutation_state.empty() && !request.vector_search.has_value() &&
            !request.vector_query_combination.has_value() && !request.row_callback.has_value();
   }
   ```

   `scoring` is not in the list.
5. `core/protostellar/search_converter.hxx:157` is the only line in that file that touches scoring:

   ```cpp
   proto.set_disable_scoring(request.disable_scoring);
   ```

   `request.scoring` is never read.

The mechanism that should have caught this already produces the right error.
`core/protostellar/component.cxx:1030-1038`:

```cpp
// Unmappable query shape or gated feature (facets, sort specs, vector search, mutation tokens,
// raw): fail cleanly rather than dropping the caller's intent. ...
auto encoded = search::encode(request);
if (!encoded.has_value()) {
  auto response = stamped();
  response.ctx.ec = search::query_is_malformed(request) ? errc::common::invalid_argument
                                                        : errc::common::feature_not_available;
```

`feature_not_available` is what the RFC asks for. The gate is one term short, and the helper the
term needs is one this change already added:

```cpp
// search_converter.hxx, can_encode()
       !request.vector_query_combination.has_value() && !request.row_callback.has_value() &&
       !is_score_fusion(request.scoring);

// search_converter.hxx, encode()
proto.set_disable_scoring(
  request.disable_scoring ||
  std::holds_alternative<couchbase::core::search_scoring_none>(request.scoring));
```

Five inserted lines and two removed in one header. It builds clean with
`-DCOUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2=ON` under the project's `-Werror`, with no other file
touched.

The gate has a test whose only job is to enumerate what it rejects, one line per feature —
`test/cng/protostellar/search_converter.cxx:140-158`, registered as
`can_encode_rejects_gated_features` at `:301`:

```cpp
assert_false(ps::can_encode(with_facets), "facets are gated");
assert_false(ps::can_encode(with_sort), "sort specs are gated");
assert_false(ps::can_encode(with_vector), "vector search is gated");
assert_false(ps::can_encode(with_combo), "vector query combination is gated");
```

A fifth line belongs there, and a sixth asserting `search_scoring_none` reaches
`proto.disable_scoring()`.

`tools/fit_performer/src/service.cxx:1591` advertising `SDK_SEARCH_SCORE_FUSION` is right where it
is and needs no change: `PerformerCapsFetchRequest` is an empty message, so the capability list
describes the build, not a connection, and `clusterConnectionCreate` (`:1643`) is a separate later
call. The effect is only that a FIT run against couchbase2 exercises score fusion and observes
nothing, which is this finding rather than a second one.

## 2. The change carries no tests (blocker)

`git show 524c501b --stat` touches no file under `test/`.

The repository has each mechanism the missing tests would use:

- **Encoding.** `test/test_unit_query.cxx:46-84` is the shape: build a `topology::configuration`,
  insert a `cluster_capability`, call `req.encode_to(http_req, ctx)`, assert on the parsed body. It
  runs with no cluster and no `TEST_CONNECTION_STRING`.
- **Integration, capability-gated.** `test/test_integration_search.cxx:158-169` is a version-gated
  `SECTION("disable scoring")` — a test for the very option this change deprecates.
- **Version gate.** Two mechanisms, neither used. The newer framework takes it declaratively beside
  the registration — `needs::cluster_version(v8_1)`, documented at `test/README.md:271` — and the
  Catch2 integration tests use the `supports_*()` helpers in `test/utils/server_version.hxx:214`
  and `:267`. No `supports_score_fusion()` was added either.
- **Capability refusal through the public API.**
  `test/test_integration_search.cxx:656-670`, "integration: scope search returns feature not
  available", skips when the cluster supports the feature and otherwise asserts
  `error.ec() == couchbase::errc::common::feature_not_available`. That is the template for the
  pre-8.1 fusion case, verbatim.
- **couchbase2 gate.** `test/cng/protostellar/search_converter.cxx:140-158` enumerates every gated
  feature, one assertion each.

The behaviours with no test: `score` is `"rrf"` with `params.score_rank_constant` and
`params.score_window_size`; `params` is omitted entirely when both are unset and each key is omitted
individually; `"rsf"` carries only `score_window_size`; `search_scoring_none` emits `"none"`;
`encode_to` returns `invalid_argument` when both `disable_scoring` and `scoring` are set;
`scoring()` and `disable_scoring(true)` throw in either order while `disable_scoring(false)` does
not; `configuration_json` puts `scoreFusion` in the capability set; `is_feature_supported` rejects
fusion without the capability and admits `none` with it.

Every one of those was checked against this head and holds. That is the argument for writing them
down, not against: the encoder is correct today and nothing in the tree keeps it correct.

## 3. Both new `@throws` blocks contradict the guards beneath them (major)

`couchbase/search_options.hxx:260-261`:

```
   * @throws std::invalid_argument if @ref scoring() has already been set: both write the same
   * field, so they cannot be used together.
```

The guard at `:268` is `if (disable && scoring_.has_value())`. It does not throw when `scoring()`
has been set — it throws when `scoring()` has been set **and** `disable` is `true`.
`disable_scoring(false)` after `scoring(...)` is accepted, which is the whole point of the narrowing
applied earlier in this review. The doc still describes the behaviour from before that fix.

The mirror block at `:286-288` has the same defect from the other side:

```
   * @throws std::invalid_argument if @ref disable_scoring() has already been set: both write the
   * same field, so they cannot be used together, ...
```

The guard at `:297` is `if (disable_scoring_)`, which tests the flag's *value*, not whether the
setter was called. After `disable_scoring(true)` followed by `disable_scoring(false)`, `scoring()`
succeeds — correctly, and against what the sentence says.

Both sentences should name the condition the code tests: the throw happens when the two would both
write `score`, not when either method has been called.

## 4. `disable_scoring` is deprecated in prose only (major)

RFC:

> For the same reason `disableScoring` is deprecated in favour of `scoring(SearchScoring.none())`:
> it gets the platform equivalent of a `@Deprecated` annotation.

`couchbase/search_options.hxx:258` has the Doxygen tag and nothing else:

```
   * @deprecated Use scoring(search_scoring_none{}) instead.
```

No `[[deprecated]]`. Nothing warns at the call site, which is the whole point of the annotation.

The repository has the precedent in a public header — `couchbase/behavior_options.hxx:63`:

```cpp
[[deprecated("Use couchbase::network_options#preferred_network() instead.")]]
auto network(std::string name) -> behavior_options&
```

That is the applicable one. `COUCHBASE_DEPRECATED` at `core/deprecation_utils.hxx:24` is not a
second option: no header under `couchbase/` includes anything from `core/`, and its only two users
(`core/cluster.hxx:170`, `core/operations/document_view.hxx:58`) are both internal.

There is one in-tree caller to silence, `tools/fit_performer/src/commands/search.cxx:476`, which
under the repository's `-Werror` (`cmake/CompilerWarnings.cmake:55`) would otherwise break the
build.

## 5. `@committed` survives on a method the same block deprecates (minor)

`couchbase/search_options.hxx:258-264` carries `@deprecated`, `@since 1.0.0` and `@committed`
together. A method being withdrawn is not committed, so the generated Committed Interfaces index
lists a deprecated method as suitable for production.

## 6. `@since 1.4.0` names a release that does not contain this API (minor)

The commit adds 13 `@since 1.4.0` lines. Tag `1.4.0` was cut on 2026-08-12 and is an ancestor of
`main`, which is 41 commits past it; `524c501b` is not an ancestor of `1.4.0`. A user reading the
generated documentation is told the API is available in a release that shipped without it.

The invariant elsewhere in the tree holds without exception: for `@since 1.1.0` in
`query_options.hxx`, `1.2.0` in `vector_query.hxx`, `1.3.0` in `network_options.hxx` and `1.3.2` in
`result.hxx`, the commit that introduced the file is an ancestor of the matching tag. This is the
first place it does not.

`CMakeLists.txt:35` does read `VERSION "1.4.0"`, but it carries the *last released* version between
releases — the bump lands in the release-prep commit, as at
[`3f27cd37`](https://github.com/couchbase/couchbase-cxx-client/commit/3f27cd37) for 1.4.0-rc.1. It
is not the oracle for `@since` on unreleased work. The tag should name whichever release this is
going into.

## 7. `raw("params", ...)` now silently replaces the fusion parameters (minor)

`core/operations/document_search.cxx:160-162` applies `raw` after the scoring block:

```cpp
for (const auto& [key, value] : raw) {
  body[key] = utils::json::parse(value);
}
```

With `scoring` set to reciprocal rank fusion with both parameters and `raw["params"]` set to
`{"something_else":1}`, the encoded body is `"score":"rrf"` with
`"params":{"something_else":1}` — the strategy survives and its tuning is gone.

Last-writer-wins is the documented behaviour of `raw` and the ordering is pre-existing. The
collision is not: before this change nothing else wrote a top-level `params`. Either document the
precedence on `scoring()` or reject the overlap the way the `disable_scoring` overlap is rejected.

## 8. There is no way back to sending no `score` field (nit)

`disable_scoring_` is a `bool` (`couchbase/search_options.hxx:549`) that `disable_scoring(false)`
returns to "no `score` field emitted". `scoring_` is a `std::optional<search_scoring::built>`
(`:563`) with a setter and no reset, so once `scoring()` has been called on an options object there
is no call that puts it back.

Nothing is stuck, to be clear: `scoring()` overwrites unconditionally, so after `scoring(rrf)` a
later `scoring(search_scoring_none{})` is accepted and emits the same `"score": "none"` that
`disable_scoring(true)` would. Only the "send nothing" state is unreachable, and only on a reused
options object.

## 9. `feature_not_available` carries no message (minor)

RFC:

> If it is not present the SDK raises `FeatureNotAvailableException`, with a message along the lines
> of "Score fusion is available from Couchbase Server 8.1 and above".

`core/cluster.cxx:858-862` returns a bare `errc::common::feature_not_available` with no context, so
a caller cannot tell which capability was missing. This is how `search_vector_search` and
`search_scoped_search_index` already behave, so it is a gap in the whole search capability surface
rather than one this change introduces — but the RFC asks for the message and the change is the
occasion.

## 10. On the core path the conflict check runs after the capability check (nit)

RFC:

> The check runs before the capability check, so it raises the same error on any cluster version.

Through the public API this holds: the throw is at option-set time,
`couchbase/search_options.hxx:268` and `:297`, before anything else can run.

Through the core the order inverts. The guard is inside `encode_to`
(`core/operations/document_search.cxx:34`), and `is_feature_supported` runs first
(`core/cluster.cxx:858`). A wrapper populating `core::operations::search_request` directly, with
both fields set, against a pre-8.1 cluster gets `feature_not_available` rather than
`invalid_argument`.

The RFC sentence constrains an SDK's own public API, and this one satisfies it. But the core guard
was added in this change specifically so wrappers do not each have to repeat the check, and for a
wrapper it does not give the version-independent error the guard exists to provide. Hoisting it to
where `is_feature_supported` runs would.

## 11. The deprecation notice's replacement does not compile as written (nit)

`couchbase/search_options.hxx:258` tells the reader to use `scoring(search_scoring_none{})`. From a
translation unit that includes only `<couchbase/search_options.hxx>`, that is:

```
error: 'search_scoring_none' is not a member of 'couchbase'; did you mean 'search_scoring'?
```

`search_options.hxx` includes only the abstract base. The note should name
`<couchbase/search_scoring_none.hxx>`.

## 12. The commit and the pull request have no body (minor)

`git show 524c501b --format=%b -s` is empty and the pull request description is empty. Of the last
40 commits on `main`, 38 carry a Motivation / Modifications / Results body and one is bodiless. A
new public API with a deprecation, a wire format and a capability gate is not the mechanical change
that exception is for.

---

## Checked and found correct

Recorded so the same ground is not re-walked.

- **Wire format.** `score` is `"rrf"` / `"rsf"` / `"none"`; `params` is top-level with
  `score_rank_constant` and `score_window_size`; each key is omitted when unset and `params` is
  dropped when both are; no `score` field is emitted when neither `scoring` nor `disable_scoring` is
  set. Matches the RFC on every point.
- **Capability parsing.** `core/topology/configuration_json.hxx:319-330` reads `scoreFusion` from
  inside the `clusterCapabilities.search` array, which is where the RFC puts it.
- **Capability gate.** `core/cluster.cxx:289-292` gates fusion only; `is_score_fusion`
  (`core/search_scoring.hxx:60-65`) excludes `search_scoring_none`, so `"none"` still works on any
  server version, as the RFC requires. Both `cluster.search()` and `scope.search()` route through
  it.
- **`map_scoring`'s missing final `else`** (`core/impl/search.cxx:129-144`) is not a latent trap.
  Adding a fourth alternative to the variant without an arm for it is caught before it can run. GCC
  rejects it outright even with every warning flag stripped — "no return statement in `constexpr`
  function returning non-void", since a C++17 lambda is implicitly `constexpr` where it can be.
  Clang is silent under `-w` but flags it as `-Wreturn-type`, which is `-Werror` here. No
  configuration in this repository lets it through. MSVC C4715 under `/WX` was not tested and is not
  claimed either way.
- **Header installation.** `cmake/Packaging.cmake:4` installs `couchbase/` as a directory, so the
  three new public headers ship without a list to update. Requiring users to include
  `search_scoring_reciprocal_rank_fusion.hxx` directly matches how `search_sort_field.hxx` is
  surfaced.
- **Struct layout.** `scoring` is appended last in `core::operations::search_request`; no site
  aggregate-initialises that struct positionally, and the brace initialisers at
  `core/impl/search.cxx:172` and `:216` stop at `timeout`.
- **FIT performer conflict handling.** The both-set case reports correctly:
  `tools/fit_performer/src/commands/search.cxx:1192-1201` already wraps `to_search_options` in a
  `catch (const std::invalid_argument&)` that maps to `errc::common::invalid_argument`. That catch
  predates this change.
- **fit-protocol pin.** `8e706b73` carries `SDK_SEARCH_SCORE_FUSION = 41`
  (`operational/sdk.caps.proto:157`) and the `SearchScoring` messages with `scoring` and
  `disable_scoring` on the request (`operational/sdk.search.proto:92-110`, `:342-343`).
- **API usability.** Both `opts.scoring(*std::shared_ptr<search_scoring>)` and the fluent temporary
  `opts.scoring(search_scoring_reciprocal_rank_fusion{}.rank_constant(5))` compile. The base is
  reachable polymorphically even though the setter is a template.
- **Config freshness.** The RFC asks for the capability to be "checked ... against the most recent
  config the SDK has, waiting for a config if it does not yet have one".
  `core/cluster.cxx:858-859` reads `session_manager_->configuration_capabilities()`, which returns
  the current `config_.capabilities` under a lock (`core/io/http_session_manager.hxx:88-92`) with no
  wait. That is the same call `search_vector_search` and `search_scoped_search_index` make, so
  whatever ordering guarantee connect provides, this inherits it unchanged. Nothing here is new, and
  no dispatch of a search before the first config was demonstrated.
- **Build.** Clean at this head under `-Werror`, both with and without
  `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2`: no warnings, no errors.

## Not raised

The `std::variant` versus flat-enum discussion already on the pull request is a design question with
two defensible answers and is left where it is.

The public and core variants are *not* redundant, despite carrying three identically-shaped payload
structs between them: `couchbase/search_scoring.hxx:76` is
`std::variant<none, reciprocal_rank_fusion, relative_score_fusion>` and
`core/search_scoring.hxx:51-54` prepends `std::monostate`. The public side spells "unset" as an
empty `std::optional<built>` and the core side spells it in-band, so `map_scoring` widens three
alternatives to four and `apply_scoring` (`core/impl/search.cxx:146-154`) is the conversion between
the two encodings. Collapsing the layers means picking one of those two ways to say "unset".

One consequence is worth a look while that is being decided: `built::mode` defaults to `none{}`, so
a default-constructed `search_scoring::built` means `score: "none"`, not "no scoring option set".
Nothing is wrong today, because the only holder is `std::optional<built>` in `search_options`. It
is a sharp edge for whoever adds the second holder.

core/operations/document_search.cxx
```diff
@@ -66,6 +70,39 @@ search_request::encode_to(search_request::encoded_request_type& encoded,
...
```
avsej, line 73: The wire format has no test. `test/test_unit_search.cxx` is where it belongs, next to the other search serialisation cases; it needs no cluster. A local helper keeps each section to one claim:

```cpp
auto
encode_scoring(const couchbase::core::search_scoring_mode& scoring,
               bool disable_scoring) -> std::pair<std::error_code, tao::json::value>
{
  couchbase::core::topology::configuration config{};
  auto ctx = search_http_context(config);

  couchbase::core::operations::search_request request{};
  request.index_name = "idx";
  request.query = couchbase::core::utils::json::generate_binary(
    tao::json::value{ { "match", "wireless headphones" } });
  request.scoring = scoring;
  request.disable_scoring = disable_scoring;

  couchbase::core::io::http_request encoded;
  auto ec = request.encode_to(encoded, ctx);
  if (ec) {
    return { ec, {} };
  }
  return { ec, couchbase::core::utils::json::parse(encoded.body) };
}
```

Then one section per rule the RFC states:

```cpp
TEST_CASE("unit: search request encodes the scoring mode", "[unit]")
{
  SECTION("no scoring option sends no score and no params")
  {
    auto [ec, body] = encode_scoring({}, false);
    REQUIRE_SUCCESS(ec);
    REQUIRE_FALSE(body.get_object().count("score"));
    REQUIRE_FALSE(body.get_object().count("params"));
  }

  SECTION("reciprocal rank fusion carries both tuning parameters")
  {
    auto [ec, body] = encode_scoring(
      couchbase::core::search_scoring_reciprocal_rank_fusion{ 60U, 200U }, false);
    REQUIRE_SUCCESS(ec);
    REQUIRE(body.at("score").get_string() == "rrf");
    REQUIRE(body.at("params").at("score_rank_constant").get_unsigned() == 60);
    REQUIRE(body.at("params").at("score_window_size").get_unsigned() == 200);
  }

  SECTION("reciprocal rank fusion drops params entirely when neither is set")
  {
    auto [ec, body] =
      encode_scoring(couchbase::core::search_scoring_reciprocal_rank_fusion{}, false);
    REQUIRE_SUCCESS(ec);
    REQUIRE(body.at("score").get_string() == "rrf");
    REQUIRE_FALSE(body.get_object().count("params"));
  }

  SECTION("reciprocal rank fusion omits the parameter that is unset")
  {
    auto [ec, body] = encode_scoring(
      couchbase::core::search_scoring_reciprocal_rank_fusion{ 17U, {} }, false);
    REQUIRE_SUCCESS(ec);
    REQUIRE(body.at("params").at("score_rank_constant").get_unsigned() == 17);
    REQUIRE_FALSE(body.at("params").get_object().count("score_window_size"));
  }

  SECTION("relative score fusion has no rank constant to send")
  {
    auto [ec, body] =
      encode_scoring(couchbase::core::search_scoring_relative_score_fusion{ 50U }, false);
    REQUIRE_SUCCESS(ec);
    REQUIRE(body.at("score").get_string() == "rsf");
    REQUIRE(body.at("params").at("score_window_size").get_unsigned() == 50);
    REQUIRE_FALSE(body.at("params").get_object().count("score_rank_constant"));
  }

  SECTION("search_scoring_none sends the same score the deprecated option sends")
  {
    auto [none_ec, none_body] = encode_scoring(couchbase::core::search_scoring_none{}, false);
    REQUIRE_SUCCESS(none_ec);
    auto [legacy_ec, legacy_body] = encode_scoring({}, true);
    REQUIRE_SUCCESS(legacy_ec);
    REQUIRE(none_body.at("score").get_string() == "none");
    REQUIRE(none_body.at("score") == legacy_body.at("score"));
    REQUIRE_FALSE(none_body.get_object().count("params"));
  }
}
```

The last section is the one worth keeping: it pins `search_scoring_none` and `disable_scoring(true)` to the same wire value, which is the claim `search_scoring_none.hxx` makes in prose.

core/operations/document_search.cxx
```diff
@@ -31,6 +31,10 @@ auto
 search_request::encode_to(search_request::encoded_request_type& encoded,
                           http_context& context) -> std::error_code
 {
+  if (disable_scoring && !std::holds_alternative<std::monostate>(scoring)) {
```
avsej, line 34: This guard has no test. As a section on the case above:

```cpp
  SECTION("disable_scoring together with a scoring mode is rejected before sending")
  {
    auto [ec, body] =
      encode_scoring(couchbase::core::search_scoring_reciprocal_rank_fusion{}, true);
    REQUIRE(ec == couchbase::errc::common::invalid_argument);
  }
```

Worth having separately from the public-API throw: this is the check wrappers reach, and it is the only one they reach.

couchbase/search_options.hxx
```diff
@@ -250,15 +255,52 @@ struct search_options : public common_options<search_options> {
...
+   *
+   * @throws std::invalid_argument if @ref scoring() has already been set: both write the same
+   * field, so they cannot be used together.
+   *
    * @since 1.0.0
    * @committed
    */
   auto disable_scoring(bool disable) -> search_options&
   {
+    if (disable && scoring_.has_value()) {
```
avsej, line 270: The narrowing applied earlier in this review — `disable &&` rather than `scoring_.has_value()` alone — is exactly the kind of change that regresses silently. In `test/test_unit_search.cxx`:

```cpp
TEST_CASE("unit: search options reject scoring and disable_scoring together", "[unit]")
{
  SECTION("scoring after disable_scoring(true)")
  {
    couchbase::search_options options{};
    options.disable_scoring(true);
    REQUIRE_THROWS_AS(options.scoring(couchbase::search_scoring_none{}), std::invalid_argument);
  }

  SECTION("disable_scoring(true) after scoring")
  {
    couchbase::search_options options{};
    options.scoring(couchbase::search_scoring_reciprocal_rank_fusion{});
    REQUIRE_THROWS_AS(options.disable_scoring(true), std::invalid_argument);
  }

  SECTION("disable_scoring(false) does not conflict with scoring")
  {
    couchbase::search_options options{};
    options.scoring(couchbase::search_scoring_reciprocal_rank_fusion{});
    REQUIRE_NOTHROW(options.disable_scoring(false));
    REQUIRE(options.build().scoring.has_value());
  }

  SECTION("scoring is accepted once disable_scoring has been turned back off")
  {
    couchbase::search_options options{};
    options.disable_scoring(true);
    options.disable_scoring(false);
    REQUIRE_NOTHROW(options.scoring(couchbase::search_scoring_relative_score_fusion{}));
  }
}
```

The third section is the regression test for the narrowing; reverting it to `if (scoring_.has_value())` fails there and nowhere else.

And one covering the builder, since the fluent setters and `build()` are otherwise unexercised:

```cpp
TEST_CASE("unit: search options carry the scoring parameters through build()", "[unit]")
{
  auto built = couchbase::search_options{}
                 .scoring(couchbase::search_scoring_reciprocal_rank_fusion{}
                            .rank_constant(60)
                            .window_size(200))
                 .build();
  REQUIRE(built.scoring.has_value());
  const auto* fusion =
    std::get_if<couchbase::search_scoring::built::reciprocal_rank_fusion>(&built.scoring->mode);
  REQUIRE(fusion != nullptr);
  REQUIRE(fusion->rank_constant == 60);
  REQUIRE(fusion->window_size == 200);
}
```

core/topology/configuration_json.hxx
```diff
@@ -324,6 +324,9 @@ struct traits<couchbase::core::topology::configuration> {
           } else if (name == "scopedSearchIndex") {
             result.capabilities.cluster.insert(
               couchbase::core::cluster_capability::search_scoped_search_index);
+          } else if (name == "scoreFusion") {
```
avsej, line 327: Nothing asserts this branch, and the failure mode is silent: read the capability out of the wrong service array and every fusion request is refused on a cluster that supports it. A test that pins the placement rather than the spelling:

```cpp
TEST_CASE("unit: scoreFusion is parsed from the search cluster capabilities", "[unit]")
{
  const auto without = couchbase::core::utils::json::parse(
    R"({"nodes":[],"clusterCapabilities":{"search":["vectorSearch","scopedSearchIndex"]}})");
  REQUIRE_FALSE(without.as<couchbase::core::topology::configuration>()
                  .capabilities.supports_score_fusion());

  const auto with = couchbase::core::utils::json::parse(
    R"({"nodes":[],"clusterCapabilities":{"search":["vectorSearch","scoreFusion"]}})");
  REQUIRE(
    with.as<couchbase::core::topology::configuration>().capabilities.supports_score_fusion());

  // The capability lives under "search"; the same name elsewhere must not enable it.
  const auto elsewhere = couchbase::core::utils::json::parse(
    R"({"nodes":[],"clusterCapabilities":{"n1ql":["scoreFusion"]}})");
  REQUIRE_FALSE(elsewhere.as<couchbase::core::topology::configuration>()
                  .capabilities.supports_score_fusion());
}
```

The third assertion is the one that earns its keep: it is what fails if the branch is ever moved into the `n1ql` loop above.

core/search_scoring.hxx
```diff
@@ -0,0 +1,66 @@
...
+                                         search_scoring_none,
+                                         search_scoring_reciprocal_rank_fusion,
+                                         search_scoring_relative_score_fusion>;
+
+/**
+ * Whether the strategy is one of the score fusion strategies, and therefore requires the
+ * "scoreFusion" cluster capability.
+ */
+[[nodiscard]] inline auto
+is_score_fusion(const search_scoring_mode& scoring) -> bool
```
avsej, line 61: `is_score_fusion` is the whole of the "`none` is not behind the capability" guarantee, and it is unexercised. In `test/test_unit_search.cxx`:

```cpp
TEST_CASE("unit: score fusion is behind the cluster capability but disabled scoring is not",
          "[unit]")
{
  couchbase::core::configuration_capabilities capabilities{};
  REQUIRE_FALSE(capabilities.supports_score_fusion());
  capabilities.cluster.insert(couchbase::core::cluster_capability::search_score_fusion);
  REQUIRE(capabilities.supports_score_fusion());

  REQUIRE(couchbase::core::is_score_fusion(
    couchbase::core::search_scoring_reciprocal_rank_fusion{}));
  REQUIRE(
    couchbase::core::is_score_fusion(couchbase::core::search_scoring_relative_score_fusion{}));
  REQUIRE_FALSE(couchbase::core::is_score_fusion(couchbase::core::search_scoring_none{}));
  REQUIRE_FALSE(couchbase::core::is_score_fusion(couchbase::core::search_scoring_mode{}));
}
```

The two `REQUIRE_FALSE` lines are the contract: teaching this function about `none` breaks `disable_scoring` on every server older than 8.1, and nothing currently notices.

core/cluster.cxx
```diff
@@ -286,6 +286,10 @@ is_feature_supported(const operations::search_request& request,
   if (request.vector_search && !capabilities.supports_vector_search()) {
     return false;
   }
+  // "none" is not behind the score fusion capability
+  if (is_score_fusion(request.scoring) && !capabilities.supports_score_fusion()) {
```
avsej, line 290: `test/test_integration_search.cxx:656` is the template for this — a version-gated case asserting a capability refusal through the public API. Two cases belong beside it, and they run today on any cluster below 8.1:

```cpp
TEST_CASE("integration: score fusion is refused when the cluster does not advertise it",
          "[integration]")
{
  test::utils::integration_test_guard integration;

  if (integration.cluster_version().supports_score_fusion()) {
    SKIP("cluster supports score fusion");
  }

  auto c = integration.public_cluster();
  const couchbase::search_request request{ couchbase::match_none_query{} };

  auto [error, result] =
    c.search("does-not-exist",
             request,
             couchbase::search_options{}.scoring(
               couchbase::search_scoring_reciprocal_rank_fusion{}.rank_constant(60)))
      .get();

  REQUIRE(error.ec() == couchbase::errc::common::feature_not_available);
}

TEST_CASE("integration: disabled scoring is not behind the score fusion capability",
          "[integration]")
{
  test::utils::integration_test_guard integration;

  if (integration.cluster_version().supports_score_fusion()) {
    SKIP("cluster supports score fusion, so the gate would not turn anything away");
  }

  auto c = integration.public_cluster();
  const couchbase::search_request request{ couchbase::match_none_query{} };

  auto [error, result] =
    c.search("does-not-exist",
             request,
             couchbase::search_options{}.scoring(couchbase::search_scoring_none{}))
      .get();

  // The index does not exist, so this must fail -- but on the server's terms. A capability refusal
  // here would mean "none" had been gated behind score fusion, which it must never be.
  REQUIRE(error.ec());
  REQUIRE(error.ec() != couchbase::errc::common::feature_not_available);
}
```

Run against 8.0.1 enterprise, which advertises `search: ["vectorSearch", "scopedSearchIndex"]` and no `scoreFusion`, the first yields `feature_not_available` and the second `index_not_found` — the second discriminating exactly as intended, since reaching the index lookup at all proves the gate let it through.

A positive-path case on 8.1 is the obvious companion; it needs a cluster that advertises the capability, which the CI matrix does not yet carry.

avsej (commented):
Follow-up to the review above, covering only finding 2. Each comment below carries a test that was
written against this head and run; the review body carries the two that have no line in the diff to
hang from.

The wire format, both option guards, the capability parse and the capability gate all hold as
written. That is what makes the gap worth closing rather than arguing about: the encoder is correct
today and nothing in the tree keeps it correct. Every test below was also confirmed to fail against
a deliberately broken encoder — changing `"rrf"`, renaming `params` or `score_rank_constant`,
dropping the empty-params guard, widening the `disable &&` narrowing, teaching `is_score_fusion`
about `none`, and reading the capability out of the `n1ql` array instead of `search`. A test that
cannot fail is not coverage.

## The couchbase2 gate

`core/protostellar/search_converter.hxx` has no line in this diff, so this one cannot be an inline
comment.

`test/cng/protostellar/search_converter.cxx:140` already enumerates what the gate rejects, one line
per feature. It needs the term this change makes necessary, plus the two cases that pin the
asymmetry between `scoring(none())` and the `disable_scoring` it replaces:

```cpp
  auto with_rrf = base_request();
  with_rrf.scoring = couchbase::core::search_scoring_reciprocal_rank_fusion{};
  assert_false(ps::can_encode(with_rrf), "reciprocal rank fusion is gated");

  auto with_rsf = base_request();
  with_rsf.scoring = couchbase::core::search_scoring_relative_score_fusion{};
  assert_false(ps::can_encode(with_rsf), "relative score fusion is gated");

  // "none" predates fusion and maps onto a field the protocol already has, so it is not gated.
  auto with_none = base_request();
  with_none.scoring = couchbase::core::search_scoring_none{};
  assert_true(ps::can_encode(with_none), "disabled scoring is not gated");
}

void
scoring_none_maps_onto_the_protocol_disable_scoring_field([[maybe_unused]] context& ctx)
{
  auto request = base_request();
  request.scoring = couchbase::core::search_scoring_none{};

  const auto encoded = ps::encode(request);
  assert_true(encoded.has_value(), "a request scoring none encodes");
  assert_true(encoded->disable_scoring(),
              "scoring(none) reaches the protocol as disable_scoring");
}

void
the_deprecated_disable_scoring_flag_still_reaches_the_protocol([[maybe_unused]] context& ctx)
{
  auto request = base_request();
  request.disable_scoring = true;

  const auto encoded = ps::encode(request);
  assert_true(encoded.has_value(), "a request disabling scoring encodes");
  assert_true(encoded->disable_scoring(), "disable_scoring reaches the protocol");
}
```

registered beside the existing case:

```cpp
      { CASE(scoring_none_maps_onto_the_protocol_disable_scoring_field) },
      { CASE(the_deprecated_disable_scoring_flag_still_reaches_the_protocol) },
```

On this head the first two fail and the third passes, which is finding 1 stated as a test: the
deprecated flag still crosses couchbase2 and its documented replacement does not. They pass once
`can_encode` gains `!is_score_fusion(request.scoring)` and `encode` maps `search_scoring_none` onto
`proto.set_disable_scoring`.

## The version gate

`test/utils/server_version.hxx` has no line in this diff either. The integration cases need:

```cpp
  [[nodiscard]] auto supports_score_fusion() const -> bool
  {
    // Hybrid search score fusion (MB-60401) lands in 8.1; the "scoreFusion" cluster capability is
    // advertised only from that release.
    return supports_search() && ((major == 8 && minor >= 1) || major > 8);
  }
```

avsej (changes requested):

core/operations/document_search.cxx
Matt-Woz: Added — unit: search request encodes the scoring mode in test_unit_search.cxx, one section per rule (both params, each omitted individually, params dropped entirely, rsf, none == legacy disable_scoring).

core/operations/document_search.cxx
Matt-Woz: Added as a section in the same test case ("disable_scoring together with a scoring mode is rejected before sending").

couchbase/search_options.hxx
Matt-Woz: Added unit: search options reject scoring and disable_scoring together (all four orderings) plus unit: search options carry the scoring parameters through build().

core/topology/configuration_json.hxx
Matt-Woz: Added unit: scoreFusion is parsed from the search cluster capabilities, including the "wrong array" negative case.

core/search_scoring.hxx
Matt-Woz: Added unit: score fusion is behind the cluster capability but disabled scoring is not.

core/cluster.cxx
Matt-Woz: Added both cases from your snippet to test_integration_search.cxx (score fusion is refused when the cluster does not advertise it, disabled scoring is not behind the score fusion capability).

avsej:
@Matt-Woz, could you rebase your branch on current `main` branch?

anirudhlakhotia (approved):

DemetrisChr:
Score Fusion is meant to be "Uncommitted" for now, not "Volatile". But it's unclear if we should mark all of the Scoring API as uncommitted, or just the parts that are score fusion

DemetrisChr (approved):

avsej (approved):

PR #1060, opened 2026-08-04
CXXCBC-919: keep the CNG tests clear of gRPC's callback-queue teardown race

avsej:
`cng_dispatcher_test` aborts intermittently with `assertion failed: prior > 0` from gRPC's `ref_counted.h`. The faulting thread is one of gRPC's own and no frame of ours appears in the backtrace: it is a double-unref of a `grpc_completion_queue` from inside `grpc_completion_queue_next`.

The queue is gRPC's process-global callback queue. `Channel::CallbackCQ()` serves the callback API from a per-channel queue when `grpc_iomgr_run_in_background()` is true and from that global one otherwise, and on POSIX the predicate is false for the iomgr-backed EventEngine distribution builds use. The global queue is reference counted: a channel takes a reference on its first callback-API call and drops it when destroyed, and at zero gRPC shuts the queue down, joins its polling threads and deletes it. Tests that stand up an in-process server per case destroy the last channel between cases, so they run that teardown repeatedly and race it.

Hold one channel for the lifetime of each affected test binary, which keeps the count above zero so the teardown runs once at exit instead of between every case.

## Where it came from

CI, on the `cng` job of PR #1033 (https://github.com/couchbase/couchbase-cxx-client/actions/runs/30929498773/job/92068255608):

```
6/12 Test #733: cng_dispatcher_test ... Subprocess aborted***Exception:  0.66 sec
E0804 17:05:47.205653939  8846 ref_counted.h:183]  assertion failed: prior > 0
```

The backtrace, from a local core dump:

```
#4/#5 grpc_core::RefCount::Unref (this=0x…e750)   ref_counted.h:183   -> abort
#6    grpc_cq_internal_unref (cq=0x…e750)         completion_queue.cc:619
#7    cq_next (cq=0x…e750)                        completion_queue.cc:1084
#8/#9 <lambda>                                   completion_queue_cc.cc:76/100
#10-11 grpc_core::Thread ("nexting_thread")       thd_posix.cc
```

It is **not** specific to #1033: that commit does not touch the dispatcher at all, and `core/protostellar/dispatcher.{cxx,hxx}` and `test/cng/protostellar/dispatcher.cxx` are all already on `main`. It reproduces on `main`.

## How the fix was measured

The failure is rare when driven by the real suite — 0 in 3100 runs of `cng_dispatcher_test` on `main`, which cycles the teardown only about four times per process. The new `callback_queue_churn` case does several hundred cycles in one process instead, which raises the rate enough to A/B against:

| | aborts |
|---|---|
| cycling servers, no keepalive | **7 / 40** |
| cycling servers, with keepalive | **0 / 40** |

`P(0/40 | p = 7/40) ≈ 0.045%`. Across all rounds: 12/62 without, 0/52 with. The shipped test reproduces the same way — `CNG_CALLBACK_QUEUE_NO_KEEPALIVE=1` aborted 2 of 30 runs, the default 0 of 30 — so the negative control is preserved in the test itself rather than only in a throwaway branch.

Measured against system gRPC 1.48.4; CI hit it with 1.51.1.

## What this does not do

It does not fix the underlying race, which is inside gRPC's own global-queue teardown. It keeps that race out of the blast radius of tests whose subject is our transport.

The production exposure is unchanged and is recorded in CXXCBC-919: an application that opens and closes clusters in a loop cycles the same global queue. Two things could close it properly — a gRPC version or EventEngine configuration where `grpc_iomgr_run_in_background()` is true, so the safe per-channel queue is used, or an upstream fix to the teardown ordering. Pinning a process-global gRPC resource for the lifetime of the client library would also work but costs 2–16 permanently-live polling threads, which is a design decision rather than a bug fix.

### DIFF
--- a/test/cng/protostellar/callback_queue_churn.cxx
+++ b/test/cng/protostellar/callback_queue_churn.cxx
@@ -1,0 +1,167 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026. Couchbase, Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+// Regression test for CXXCBC-919: repeatedly creating and destroying an in-process server, channel
+// and dispatcher must not abort the process.
+//
+// Each cycle drops the last reference to gRPC's process-global callback completion queue, which
+// makes gRPC shut it down, join its "nexting" threads and delete it. Doing that hundreds of times
+// in one process races the nexting threads still polling the queue and aborts with
+//
+//     ref_counted.h:183]  assertion failed: prior > 0
+//
+// See callback_queue_keepalive.hxx for the mechanism and why pinning the queue avoids it.
+//
+// Environment knobs, for investigating the underlying gRPC race rather than testing our side of it:
+//
+// Both are read through safe_getenv, so an empty value counts as unset rather than as "set".
+//
+//   CNG_CALLBACK_QUEUE_CYCLES=<n>     cycles per run (default below)
+//   CNG_CALLBACK_QUEUE_NO_KEEPALIVE=1 skip the keepalive, i.e. reproduce the abort. Measured at
+//                                     7 aborts in 40 runs at 200 cycles against gRPC 1.48.4;
+//                                     0 in 40 with the keepalive.
+//
+// This case is deliberately cheap -- 200 cycles is about 80 ms -- so it can act as a canary without
+// slowing the suite.
+
+#include "framework/test_runner.hxx"
+
+#include "callback_queue_keepalive.hxx"
+
+#include "core/protostellar/dispatcher.hxx"
+
+#include <couchbase/kv/v1/kv.grpc.pb.h>
+
+#include <asio/executor_work_guard.hpp>
+#include <asio/io_context.hpp>
+
+#include <grpcpp/server_builder.h>
+
+#include <chrono>
+#include <cstdlib>
+#include <functional>
+#include <memory>
+#include <string>
+#include <utility>
+
+namespace couchbase::cng::test
+{
+namespace
+{
+namespace kv = ::couchbase::kv::v1;
+using ::couchbase::core::protostellar::dispatcher;
+using namespace std::chrono_literals;
+
+constexpr int default_cycles{ 200 };
+
+class echo_service final : public kv::KvService::Service
+{
+public:
+  auto Get(grpc::ServerContext* /* context */,
+           const kv::GetRequest* request,
+           kv::GetResponse* response) -> grpc::Status override
+  {
+    response->set_content_uncompressed("value:" + request->key());
+    response->set_cas(0xdead'beefULL);
+    return grpc::Status::OK;
+  }
+};
+
+// One create/use/destroy cycle. The server is destroyed too, not just the channel: the server holds
+// its own reference to the callback queue, so keeping it alive would pin the count above zero and
+// the teardown under test would never run.
+void
+one_cycle()
+{
+  echo_service service;
+  grpc::ServerBuilder builder;
+  builder.RegisterService(&service);
+  auto server = builder.BuildAndStart();
+  auto channel = server->InProcessChannel(grpc::ChannelArguments{});
+
+  {
+    asio::io_context io;
+    dispatcher disp{ io, channel };
+    auto stub = kv::KvService::NewStub(channel);
+    auto request = std::make_shared<kv::GetRequest>();
+    request->set_key("k");
+
+    auto guard = asio::make_work_guard(io);
+    disp.unary<kv::GetResponse>(
+      1000ms,
+      [stub = stub.get(), request](grpc::ClientContext& context,
+                                   kv::GetResponse& response,
+                                   std::function<void(grpc::Status)> callback) {
+        stub->async()->Get(&context, request.get(), &response, std::move(callback));
+      },
+      [&guard](grpc::Status /* status */, kv::GetResponse /* response */) mutable {
+        guard.reset();
+      });
+    io.run();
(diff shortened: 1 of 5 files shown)


test/cng/protostellar/callback_queue_churn.cxx
```diff
@@ -0,0 +1,165 @@
...
+
+  channel.reset();
+  server->Shutdown(std::chrono::system_clock::now());
+  server->Wait();
+}
+
+[[nodiscard]] auto
+cycles_from_environment() -> int
+{
+  const char* raw = std::getenv("CNG_CALLBACK_QUEUE_CYCLES");
```
thejcfactor: I think this is a problem for MSVC? We use `spdlog::details::os::getenv(...)` elsewhere I believe?

test/cng/protostellar/callback_queue_churn.cxx
```diff
@@ -0,0 +1,165 @@
...
+    return default_cycles;
+  }
+  const auto parsed = std::strtol(raw, nullptr, 10);
+  return (parsed > 0) ? static_cast<int>(parsed) : default_cycles;
+}
+
+void
+cycling_servers_does_not_abort_the_process()
+{
+  if (std::getenv("CNG_CALLBACK_QUEUE_NO_KEEPALIVE") == nullptr) {
```
thejcfactor: See comment above.

test/cng/protostellar/callback_queue_churn.cxx
avsej: > I think this is a problem for MSVC?  We use `spdlog::details::os::getenv(...)` elsewhere I believe?

Correct, and it is not a latent problem — it is what is currently breaking this pull request. All three Windows legs fail, and this is the only error in any of them:

```
test\cng\protostellar\callback_queue_churn.cxx(123,26): error C2220: the following warning is treated as an error
test\cng\protostellar\callback_queue_churn.cxx(123,26): warning C4996: 'getenv': This function or variable may be unsafe
test\cng\protostellar\callback_queue_churn.cxx(134,12): warning C4996: 'getenv': This function or variable may be unsafe
...
    1 Error(s)
```

Lines 123 and 134 are the two this thread and the next one point at. `windows-2022`, `windows-2025` and `windows-11-arm` all fail the same way, each reporting exactly one error, so nothing else is hiding behind it.

Both call sites now go through `safe_getenv`.

On the choice of wrapper: `spdlog::details::os::getenv` is indeed what the Catch2 test tree uses (`test/utils/test_context.cxx`, `test/utils/logger.cxx`, and `core/logger/logger.cxx`), but the CNG tree has its own `safe_getenv` in `framework/test_runner.hxx`, added for precisely this reason — its comment reads "Needed because MSVC treats plain getenv() as deprecated, and the CNG tree builds with /W4 /WX". It was already the convention in six files here, including `live_kv.cxx` and `live_connect.cxx`, so using it keeps this file consistent with its neighbours rather than importing the other tree's idiom. Either would have fixed the build; this one also returns `std::optional<std::string>` rather than a possibly-empty string, which reads better at both call sites.

One deliberate behaviour change comes with it: `safe_getenv` reports an empty value as unset, so `CNG_CALLBACK_QUEUE_CYCLES=` now falls back to the default instead of parsing `""` to zero, and `CNG_CALLBACK_QUEUE_NO_KEEPALIVE=` no longer counts as "set". Both readings are more useful than the old ones, and the header comment documenting the knobs now says so.

Verified locally that the knobs still do what they claim, since a refactor that quietly disabled them would look identical from a passing test: the default run does 200 cycles in ~64 ms, `CNG_CALLBACK_QUEUE_CYCLES=5` drops it to ~6 ms, and `CNG_CALLBACK_QUEUE_CYCLES=` returns to ~73 ms. The suite still passes and the file is clang-format clean.

What is not verified from here is the MSVC build itself — there is no Windows toolchain in this loop, so the evidence above is the CI log rather than a local compile. CI on the next push is what will confirm it.

test/cng/protostellar/callback_queue_churn.cxx
avsej: Fixed by the same change; see the reply on the thread above. Both call sites now use `safe_getenv`, and this line (134) was the second of the two `C4996` warnings that `-WX` was turning into the single error breaking all three Windows legs.

thejcfactor (approved):

avsej:
Reposting visibly — the inline replies went onto `callback_queue_churn.cxx` lines 123 and 134, which this change itself rewrote, so both threads are now outdated and collapsed.

---

> I think this is a problem for MSVC?  We use `spdlog::details::os::getenv(...)` elsewhere I believe?

Correct, and it is not a latent problem — it is what is currently breaking this pull request. All three Windows legs fail, and this is the only error in any of them:

```
test\cng\protostellar\callback_queue_churn.cxx(123,26): error C2220: the following warning is treated as an error
test\cng\protostellar\callback_queue_churn.cxx(123,26): warning C4996: 'getenv': This function or variable may be unsafe
test\cng\protostellar\callback_queue_churn.cxx(134,12): warning C4996: 'getenv': This function or variable may be unsafe
...
    1 Error(s)
```

Lines 123 and 134 are the two this thread and the next one point at. `windows-2022`, `windows-2025` and `windows-11-arm` all fail the same way, each reporting exactly one error, so nothing else is hiding behind it.

Both call sites now go through `safe_getenv`.

On the choice of wrapper: `spdlog::details::os::getenv` is indeed what the Catch2 test tree uses (`test/utils/test_context.cxx`, `test/utils/logger.cxx`, and `core/logger/logger.cxx`), but the CNG tree has its own `safe_getenv` in `framework/test_runner.hxx`, added for precisely this reason — its comment reads "Needed because MSVC treats plain getenv() as deprecated, and the CNG tree builds with /W4 /WX". It was already the convention in six files here, including `live_kv.cxx` and `live_connect.cxx`, so using it keeps this file consistent with its neighbours rather than importing the other tree's idiom. Either would have fixed the build; this one also returns `std::optional<std::string>` rather than a possibly-empty string, which reads better at both call sites.

One deliberate behaviour change comes with it: `safe_getenv` reports an empty value as unset, so `CNG_CALLBACK_QUEUE_CYCLES=` now falls back to the default instead of parsing `""` to zero, and `CNG_CALLBACK_QUEUE_NO_KEEPALIVE=` no longer counts as "set". Both readings are more useful than the old ones, and the header comment documenting the knobs now says so.

Verified locally that the knobs still do what they claim, since a refactor that quietly disabled them would look identical from a passing test: the default run does 200 cycles in ~64 ms, `CNG_CALLBACK_QUEUE_CYCLES=5` drops it to ~6 ms, and `CNG_CALLBACK_QUEUE_CYCLES=` returns to ~73 ms. The suite still passes and the file is clang-format clean.

What is not verified from here is the MSVC build itself — there is no Windows toolchain in this loop, so the evidence above is the CI log rather than a local compile. CI on the next push is what will confirm it.

PR #1065, opened 2026-08-10
CXXCBC-926: document Cloud Native Gateway support on the public API

avsej:
[CXXCBC-926](https://couchbasecloud.atlassian.net/browse/CXXCBC-926)

The Cloud Native Gateway work added no public type and no public method, so nothing in the public
headers recorded that it happened. `cluster::connect()` accepts a `couchbase2://` connection string,
and a cluster reached that way refuses several operations it serves over `couchbase://`, but a caller
found that out by receiving `feature_not_available` at runtime.

A new `@page couchbase2` describes the transport: how the scheme differs from the others, that
support is a compile-time option which is off by default, and which operations the library refuses
without reaching the gateway. Both `connect()` overloads and each refused operation carry a note
pointing at it. Nothing changes at runtime — this is documentation only.

## Why two stability pairs, and why aliases

The annotated entities are all `@since 1.0.0` and `@committed`. Gateway support arrived in 1.4.0 and
is not committed. Marking `get_all_replicas()` `@uncommitted` would be false: the method is unchanged
and remains committed; what is uncommitted is the behaviour of running it over this transport.

A bare second `@since 1.4.0` / `@uncommitted` pair says that, but reads as a contradiction against the
first pair. Two Doxygen aliases carry the distinction in their own rendered text instead:

```
@since 1.0.0
@committed
@cng_since{1.4.0}
@cng_uncommitted
```

renders as

> **Since** 1.0.0 — **Committed**: Generally available API and should be preferred in production
> **Since** 1.4.0 – Cloud Native Gateway (CNG, couchbase2://) support
> **Uncommitted**: Cloud Native Gateway (CNG, couchbase2://) support is uncommitted and may change in
> a future release, independently of the stability of the API it applies to

## Where the notes went

Refused over `couchbase2://`, each note naming the error reported:

- `cluster::ping()`, `bucket::ping()`, `cluster::diagnostics()`
- `collection::get_all_replicas()`, `get_any_replica()`, `lookup_in_all_replicas()`,
  `lookup_in_any_replica()`
- poll-based durability

Poll-based durability is annotated on `common_durability_options::durability(persist_to, replicate_to)`
— the single point a caller opts into it — rather than on the nine mutations that inherit it. A note
on `upsert()` would be false unless a requirement had been set. That overload was also missing the
`@since`/`@committed` pair its sibling carries; it has been added.

Operations that merely succeed over the gateway are covered collectively by the page rather than
individually, since which of them work depends on the gateway version as much as on this library.

## Verification

The API reference was built and inspected, not just compiled:

- No new doxygen warnings; every `@ref` added resolves. An early draft referenced the two
  `durability()` overloads by signature and did not resolve — doxygen cannot match an argument list
  against a trailing-return-type declaration — so those became code text.
- The rendered output was checked on the generated pages rather than assumed. A first attempt wrote
  `@ref couchbase2 transport`, which rendered as "the **The couchbase2:// transport** transport"
  because the page title is the link text; all 17 sites now pass explicit link text.
- Searching the public headers and the generated documentation pages for the internal project name
  returns nothing; the prose uses Cloud Native Gateway, CNG and `couchbase2://` only.
- `clang-format` clean, and the library compiles.

## A comment that contradicted its code

Writing the page surfaced a false statement in `core/cluster.cxx`. The comment on the single-endpoint
guard in `open_protostellar()` stated that `parse_connection_string()` rejects a `couchbase2://`
string carrying more than one host, and therefore that the guard could only fire for an origin
assembled programmatically.

It cannot reject it. The parser writes an error onto the `connection_string`, but
`connection_string::error` has no reader outside the parser itself and the test suite, so `origin` is
constructed with every host and such a connection string reaches the guard exactly like a
programmatic origin does. A reader trusting the comment would have been entitled to turn the guard
into an assertion. It is corrected here, and the guard is unchanged.

The same finding also corrected the page: an early draft stated that the two `couchbase2://`
connection-string rules are reported as `invalid_argument`. They are not — a string naming several
hosts connects to the first and logs that the rest were ignored, which is what the page now says.

That the parser's validation never reaches a caller is a pre-existing defect, not specific to the
gateway, and is not fixed here — only the comment and the documentation are made to match what the
code actually does.

[CXXCBC-926]: https://couchbasecloud.atlassian.net/browse/CXXCBC-926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -962,10 +962,13 @@ public:
     }
     auto [hostname, port] = origin_.next_address();
     auto endpoint = hostname + ":" + port;
-    // parse_connection_string() rejects a couchbase2:// string carrying more than one host, so this
-    // is only reachable for an origin assembled programmatically (set_nodes(), or the
-    // hostname/port constructors). Kept as a guard rather than an assertion: using the first
-    // endpoint is well defined, and the log says which one was chosen.
+    // Several nodes reach here from an origin assembled programmatically (set_nodes(), or the
+    // hostname/port constructors), and from a connection string naming more than one host:
+    // parse_connection_string() records that as an error on the connection_string, but no caller
+    // reads that field, so the origin is built with every host regardless. Kept as a log rather
+    // than a rejection: one endpoint is as good as another here, and since the node list is
+    // shuffled unless preserve_bootstrap_nodes_order is set, the message is what says which one
+    // this process picked.
     if (origin_.get_nodes().size() > 1) {
       CB_LOG_INFO("[{}]: couchbase2 uses a single gateway endpoint (\"{}\"); the remaining {} "
                   "node(s) on this origin are ignored (the gateway fronts the cluster).",
--- a/couchbase/bucket.hxx
+++ b/couchbase/bucket.hxx
@@ -90,8 +90,13 @@ public:
    * @param options custom options to change the default behavior.
    * @param handler the handler that implements @ref ping_handler.
    *
+   * @note Not served over the @ref couchbase2 "couchbase2" transport, which opens no per-bucket
+   * sessions to probe and reports @ref errc::common::feature_not_available instead.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   void ping(const ping_options& options, ping_handler&& handler) const;
 
@@ -104,8 +109,13 @@ public:
    * @param options custom options to change the default behavior.
    * @return future object that carries result of the operation.
    *
+   * @note Not served over the @ref couchbase2 "couchbase2" transport, which opens no per-bucket
+   * sessions to probe and reports @ref errc::common::feature_not_available instead.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   [[nodiscard]] auto ping(const ping_options& options = {}) const
     -> std::future<std::pair<error, ping_result>>;
--- a/couchbase/cluster.hxx
+++ b/couchbase/cluster.hxx
@@ -69,8 +69,16 @@ public:
    * over this options).
    * @param handler the handler
    *
+   * @note A connection string with the @c couchbase2:// scheme selects the @ref couchbase2
+   * "couchbase2" transport, which routes every operation through a Cloud Native Gateway (CNG). It
+   * requires a library built with that support and restricts how the connection string may be
+   * written; see
+   * @ref couchbase2.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   static void connect(const std::string& connection_string,
                       const cluster_options& options,
@@ -85,8 +93,16 @@ public:
    *
    * @return future object that carries cluster object and operation status
    *
+   * @note A connection string with the @c couchbase2:// scheme selects the @ref couchbase2
+   * "couchbase2" transport, which routes every operation through a Cloud Native Gateway (CNG). It
+   * requires a library built with that support and restricts how the connection string may be
+   * written; see
+   * @ref couchbase2.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   [[nodiscard]] static auto connect(const std::string& connection_string,
                                     const cluster_options& options)
@@ -406,8 +422,13 @@ public:
    * @param options custom options to change the default behavior.
    * @param handler the handler that implements @ref ping_handler.
    *
+   * @note Not served over the @ref couchbase2 "couchbase2" transport, which has no per-node service
+   * endpoints to probe and reports @ref errc::common::feature_not_available instead.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   void ping(const ping_options& options, ping_handler&& handler) const;
 
@@ -420,8 +441,13 @@ public:
    * @param options custom options to change the default behavior.
    * @return future object that carries result of the operation.
    *
+   * @note Not served over the @ref couchbase2 "couchbase2" transport, which has no per-node service
+   * endpoints to probe and reports @ref errc::common::feature_not_available instead.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   [[nodiscard]] auto ping(const ping_options& options = {}) const
     -> std::future<std::pair<error, ping_result>>;
@@ -439,8 +465,14 @@ public:
    * @param options custom options to change the default behavior.
    * @param handler the handler that implements @ref diagnostics_handler.
    *
+   * @note Not served over the @ref couchbase2 "couchbase2" transport, which keeps no per-node
+   * connection state to report on and returns @ref errc::common::feature_not_available rather than
+   * an empty report.
+   *
    * @since 1.0.0
    * @committed
+   * @cng_since{1.4.0}
+   * @cng_uncommitted
    */
   void diagnostics(const diagnostics_options& options, diagnostics_handler&& handler) const;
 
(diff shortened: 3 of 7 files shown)


avsej:
## Response — PR #1065 (`CXXCBC-926`), Copilot review

Every point was checked against the code before being acted on. All four are real; all four are
fixed. The fixes land in the `CXXCBC-926` commit, so they reach #1066 as well.

### 1. The completeness claim about what reaches the gateway

> This completeness claim is inaccurate. The client rejects requests locally that are not in
> `component_supports_v` (including non-replica `lookup_in` and `mutate_in`), unwired management
> operations, and query/analytics/search/view option combinations that their converters cannot
> encode. Those requests never reach the gateway, so please qualify this statement rather than
> attributing every other compatibility decision to the gateway.

Confirmed and corrected. `component_supports_v` has exactly fifteen specialisations
(`core/protostellar/component.hxx`), and `lookup_in_request` and `mutate_in_request` are not among
them and appear nowhere else on the couchbase2 path; `cluster_impl::execute` answers any KV request
outside that set with `feature_not_available` locally. `can_encode()` gates option combinations in
the query, analytics, search, view and query-index-admin converters, and the bucket admin converter
refuses a memcached bucket the same way.

The page now lists subdocument `lookup_in` and `mutate_in` as refused, and adds a bullet for requests
whose options or values have no couchbase2 equivalent, stating that the refusal depends on what the
request carries rather than on the operation, so the same call may be served or refused according to
how it was built. The trailing sentence is scoped to what remains after that list.

### 2. TLS and the default port

> The transport overview omits two connection semantics enforced by the parser: `couchbase2://`
> always enables TLS and uses port 18098 when no port is supplied
> (`core/utils/connection_string.cxx:94-99`). Since the scheme name does not otherwise signal TLS,
> document these defaults so callers can configure endpoints and certificates correctly.

Confirmed at that location: the `couchbase2` branch sets `default_port = 18098` and `tls = true`.
Both are now documented, together with the consequence the comment points at -- that a private
certificate authority has to be trusted as it would be for `couchbases://`.

### 3. Which host a multi-host string connects to

> This does not reliably connect to the first host named in the string. `origin` shuffles bootstrap
> nodes by default before `open_protostellar()` takes `next_address()` (`core/origin.cxx:408-417`),
> so any listed host may be selected unless bootstrap order preservation is enabled. Describe the
> actual selection rather than promising the first host.

Confirmed. `preserve_bootstrap_nodes_order` defaults to `false` (`core/cluster_options.hxx:119`) and
the `origin` constructor calls `shuffle_nodes()` unless it is set, so `next_address()` returns an
arbitrary entry. The page now says the connection is made to a single one of the hosts, that the list
is shuffled unless that option is set, and that the log records which endpoint was chosen.

The same error appeared in a comment added by this change at `core/cluster.cxx`, which described the
selection as "the first endpoint". That is corrected in the same commit.

### 4. Alias line length

> Both newly added alias definitions exceed the project's 100-character line limit. Please wrap them
> while preserving Doxygen's alias expansion semantics.

The lines were 104 and 274 characters and are now wrapped to 91 or fewer.

The premise does not hold, and is worth recording so the next reader does not act on it: no
100-character limit is enforced on `Doxyfile.in`. `.clang-format` sets `ColumnLimit: 100` for C++
sources only, and the four pre-existing alias definitions immediately above these two are 122, 136,
169 and 184 characters. The change was made because a 274-character line is worse than its
neighbours, not because a limit was breached.

Expansion was verified rather than assumed, since a continuation inside a quoted alias value could
plausibly terminate it: with the wrap in place the rendered text is byte-identical to the unwrapped
form, checked against a single-line control in a minimal project and again on the generated page.

### Verification

The API reference builds with no doxygen warnings, and the library compiles.

thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf. This was a static review — I traced the request-dispatch code by hand (`core/cluster.cxx`, `core/protostellar/component.hxx`) rather than compiling or running anything; please treat the reasoning as a starting point for discussion, not ground truth.

---

Nice PR — traced most of the new page's claims against the actual dispatch code and they held up (default port, TLS-implied, the one-host/no-mode-suffix validation being set-but-unenforced, the compile-flag gating, the poll-based-durability-after-write behavior). One gap:

`lookup_in()` and `mutate_in()` in `couchbase/collection.hxx` are refused over `couchbase2://` — `core/cluster.cxx`'s templated `execute()` rejects them with `feature_not_available` whenever `is_protostellar()` is true, because `protostellar::component_supports_v<lookup_in_request>`/`<mutate_in_request>` is `false` (there's no matching overload in `core/protostellar/component.hxx`). That's unconditional, same as `ping`/`diagnostics`/the four replica-read methods.

`docs/couchbase2.hxx` correctly lists them in that same bucket:

> Subdocument operations against the active node, `lookup_in` and `mutate_in`.

But unlike `get_all_replicas`/`get_any_replica`/`lookup_in_all_replicas`/`lookup_in_any_replica`, which each got an inline `@note` + `@cng_since`/`@cng_uncommitted` block at their declaration in this PR, `lookup_in()`/`mutate_in()` didn't get one. A caller reading `collection::lookup_in()`'s doc comment directly — without following through to the transport page — sees `@since 1.0.0` / `@committed` and nothing warning it silently fails over this transport, which is exactly the gap this PR closes for the other four.

Non-blocking, but seems like an easy add given the pattern is already established for the sibling methods.

avsej:
## Response — PR #1065 (`CXXCBC-926`), round 2

Done, and the sweep it prompted found one more of the same kind.

> But unlike `get_all_replicas`/`get_any_replica`/`lookup_in_all_replicas`/`lookup_in_any_replica`,
> which each got an inline `@note` + `@cng_since`/`@cng_uncommitted` block at their declaration in
> this PR, `lookup_in()`/`mutate_in()` didn't get one. A caller reading `collection::lookup_in()`'s
> doc comment directly — without following through to the transport page — sees `@since 1.0.0` /
> `@committed` and nothing warning it silently fails over this transport, which is exactly the gap
> this PR closes for the other four.

Correct, and the inconsistency has an explanation that is worth stating rather than leaving as an
oversight: `lookup_in` and `mutate_in` were added to the page in the previous round, in response to a
review comment about the page's completeness, and the corresponding declarations were not swept at
the same time. The page and the declarations went out of step in that round.

Both overloads of each now carry the same `@note` plus `@cng_since{1.4.0}` / `@cng_uncommitted` block
as the four replica methods.

> `lookup_in()` and `mutate_in()` in `couchbase/collection.hxx` are refused over `couchbase2://` —
> `core/cluster.cxx`'s templated `execute()` rejects them with `feature_not_available` whenever
> `is_protostellar()` is true, because
> `protostellar::component_supports_v<lookup_in_request>`/`<mutate_in_request>` is `false` (there's
> no matching overload in `core/protostellar/component.hxx`). That's unconditional, same as
> `ping`/`diagnostics`/the four replica-read methods.

The mechanism matches what the code does. `component_supports_v` has fifteen specialisations and
neither request type is among them, so the `else` branch of the templated `execute()` answers with
`feature_not_available` before anything is dispatched.

### One more method in the same position

Sweeping the rest of the public KV surface against `component_supports_v` turned up
`collection::scan`, which is also not served, and which the page did not mention at all.

It belongs in a different category from `lookup_in`/`mutate_in`, though, and that is why it was not
caught by the same reasoning. `scan` does not go through the templated `execute()`; it fetches the
bucket configuration first, through `with_bucket_config_or_timeout` and then
`with_bucket_configuration`, which carries no couchbase2 guard. So it is not refused up front and
does not report `feature_not_available` — it fails on the bucket-configuration path, exactly like
poll-based durability does after its mutation has been applied.

The page now has a section for that category holding both `scan` and poll-based durability, stating
that neither is turned away before work begins and that the error describes the failure rather than
the missing feature. Both `scan` overloads carry the annotation.

### On the review being static

> This was a static review — I traced the request-dispatch code by hand (`core/cluster.cxx`,
> `core/protostellar/component.hxx`) rather than compiling or running anything; please treat the
> reasoning as a starting point for discussion, not ground truth.

Noted, and it held up: every claim in it was checked against the source and none needed correcting.
Worth recording that the finding was reachable statically and had been missed by a review that ran
the build, since the gap was a missing annotation rather than anything a build or test would report.

### Verification

The API reference builds with no doxygen warnings, and the three new notes were confirmed on the
generated pages rather than assumed from the source.

thejcfactor (approved):

PR #1066, opened 2026-08-10
CXXCBC-927: build the API reference without doxygen warnings

avsej:
[CXXCBC-927](https://couchbasecloud.atlassian.net/browse/CXXCBC-927)

Stacked on #1065, so this PR shows that commit too until it merges. Only the second commit —
`CXXCBC-927` — belongs to this change.

The `doxygen` target completed successfully but emitted 89 warnings. Because they were always there,
a warning introduced by a new change was indistinguishable from the standing noise. They were not
cosmetic either: each corresponded to documentation that rendered wrongly.

After this change the target emits none.

## What was wrong

| n | cause | fix |
| --- | --- | --- |
| 26 | `{@link #f(args)}` never resolved | doxygen cannot match an argument list against a declaration using a trailing return type (`auto f(x) -> T&`), which is how every options-builder setter is declared. Rewritten as `{@link #f f(args)}` |
| 42 | malformed definition lists in three `cbc` manual pages | two never opened `<dl>`; `cbc.md` closed it before its last two entries, orphaning the `keygen` and `config` entries |
| 2 | `@subpage cbc-config` unresolvable | `cbc-config.md` was the only page missing its `{#cbc-config}` label |
| 4 | undocumented `crypto_manager` on three result constructors; `@param name` against an argument called `network_name` | documented and renamed |
| 1 | `</blockquote>` without a match | a wrapped line began `> 5`, which markdown reads as a blockquote |
| 1 | obsolete `DOT_MULTI_TARGETS` | removed |
| 1 | `std::hash<couchbase::node_id>` unparseable | `BUILTIN_STL_SUPPORT = YES` |

The first row is the one worth remembering: the failure is a property of the declaration style, not of
the individual links, so any future `{@link #f(args)}` in this codebase is dead on arrival. It was
isolated rather than guessed — in a minimal reproduction a plain-return-type control resolved with an
argument list, while every trailing-return-type form failed, with `#`, qualified, and `@ref` spellings
alike. The replacement keeps the full signature as the visible text, so overloads such as
`expiry(std::chrono::seconds)` and `expiry(std::chrono::system_clock::time_point)` remain
distinguishable in the rendered text.

## Verification

The zero is demonstrated rather than asserted. A canary `@ref this_symbol_does_not_exist_xyz` was
injected into a public header; the same check that reports zero then reported exactly one warning, and
removing the canary returned it to zero. Without that control, "no matches" would have been
indistinguishable from a broken check.

`BUILTIN_STL_SUPPORT = YES` was checked for side effects: it adds exactly one page, the `std::hash`
specialization it was enabled for, and no STL container pages. Signatures still render with their STL
types intact.

`clang-format` reflowed some of the rewritten links across lines. Doxygen still resolves them and the
rendered text is unaffected, which was confirmed on the generated pages rather than assumed — several
links in these files were already wrapped before this change. The library compiles.

## Not done here

Warnings are not turned into errors. Doxygen's warning set varies between versions, and the reference
is not built in CI, so `WARN_AS_ERROR` would fail the docs build for contributors running a different
doxygen rather than catching regressions in a pinned environment. It is worth revisiting if the
reference is ever built in CI.

[CXXCBC-927]: https://couchbasecloud.atlassian.net/browse/CXXCBC-927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/cluster.cxx
+++ b/core/cluster.cxx
@@ -962,10 +962,13 @@ public:
     }
     auto [hostname, port] = origin_.next_address();
     auto endpoint = hostname + ":" + port;
-    // parse_connection_string() rejects a couchbase2:// string carrying more than one host, so this
-    // is only reachable for an origin assembled programmatically (set_nodes(), or the
-    // hostname/port constructors). Kept as a guard rather than an assertion: using the first
-    // endpoint is well defined, and the log says which one was chosen.
+    // Several nodes reach here from an origin assembled programmatically (set_nodes(), or the
+    // hostname/port constructors), and from a connection string naming more than one host:
+    // parse_connection_string() records that as an error on the connection_string, but no caller
+    // reads that field, so the origin is built with every host regardless. Kept as a log rather
+    // than a rejection: one endpoint is as good as another here, and since the node list is
+    // shuffled unless preserve_bootstrap_nodes_order is set, the message is what says which one
+    // this process picked.
     if (origin_.get_nodes().size() > 1) {
       CB_LOG_INFO("[{}]: couchbase2 uses a single gateway endpoint (\"{}\"); the remaining {} "
                   "node(s) on this origin are ignored (the gateway fronts the cluster).",
--- a/couchbase/analytics_options.hxx
+++ b/couchbase/analytics_options.hxx
@@ -182,10 +182,11 @@ struct analytics_options : public common_options<analytics_options> {
    * mutations. If you want to include all the mutations up to the point of the query, use {@link
    * analytics_scan_consistency::request_plus}.
    *
-   * Note that you cannot use this method and {@link #consistent_with(const mutation_state&)} at the
-   * same time, since they are mutually exclusive. As a rule of thumb, if you only care to be
-   * consistent with the mutation you just wrote on the same thread/app, use
-   * {@link #consistent_with(const mutation_state&)}. If you need "global" scan consistency, use
+   * Note that you cannot use this method and {@link #consistent_with consistent_with(const
+   * mutation_state&)} at the same time, since they are mutually exclusive. As a rule of thumb, if
+   * you only care to be consistent with the mutation you just wrote on the same thread/app, use
+   * {@link #consistent_with consistent_with(const mutation_state&)}. If you need "global" scan
+   * consistency, use
    * {@link analytics_scan_consistency::request_plus} on this method.
    *
    * @param scan_consistency the index scan consistency to be used for this query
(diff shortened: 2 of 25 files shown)


avsej:
## Response — PR #1066 (`CXXCBC-927`), Copilot review

Both points are real. Both concern documentation added by the stacked `CXXCBC-926` commit, so the
fixes land there and are visible on #1065 as well.

### 1. Poll-based durability is not refused before the write

> The implementation does not guarantee `feature_not_available` for poll-based durability. Mutations
> are sent first, then `initiate_observe_poll` calls `with_bucket_configuration`; on couchbase2 that
> falls into the normal MCBP bucket bootstrap rather than a feature-not-available guard
> (`core/cluster.cxx:1566-1595`). This can report a bootstrap/network error after the write has
> already succeeded. Reject legacy durability before dispatch so the documented error and no-write
> behavior are reliable.

> Poll-based durability is not currently rejected before I/O. For example, `collection_impl::upsert`
> first dispatches a supported `upsert_request` to the gateway (`core/impl/collection.cxx:860-889`)
> and only then starts the unsupported observe-poll path. Thus the mutation can be applied even
> though the operation later fails, contradicting the statement that every listed operation is
> refused without reaching the gateway.

Confirmed at both locations, and the documentation was wrong in three separate ways: the operation is
not refused, it does reach the gateway, and the error is not `feature_not_available`.

`collection_impl::upsert` builds a plain `upsert_request` with `durability_level::none` and dispatches
it; only in the response handler does it call `initiate_observe_poll`. Over couchbase2 the mutation
is therefore applied. `with_bucket_configuration` carries no protostellar guard, so the poll proceeds
into `open_bucket` and an MCBP bucket bootstrap against a port the gateway does not serve.

The earlier statement was reached by reasoning from the `*_with_legacy_durability` core request types,
which `cluster_impl` does answer with `feature_not_available` over couchbase2. The public API does not
use those types, so that guard never applies to a caller of `collection::upsert`. Reasoning from the
core request types instead of the public call path is what produced the wrong claim.

Poll-based durability now has its own section in the transport page, stating that the mutation is sent
and applied, that only the following poll fails, that the reported error comes from that failed poll
rather than being `feature_not_available`, and that the document is already written when the error
arrives. The note on `durability(persist_to, replicate_to)` says the same. Level-based durability is
named as the durability to use over this transport.

> Add a pre-dispatch couchbase2 check for legacy durability (and test it), or document this side
> effect accurately.

The second option was taken. The first is a behaviour change: it converts an operation that today
applies the mutation and then reports a poll failure into one that applies nothing and reports
`feature_not_available`. That is very likely the right end state, but it needs its own change with
tests, and both tickets behind these PRs are explicitly documentation-only, with any behaviour change
recorded as out of scope. Filing it separately is preferable to smuggling it into a docs change, and
the documentation now describes the behaviour accurately in the meantime.

### 2. Which host a multi-host string connects to

> This says a multi-host string connects to the first listed host, but `origin` shuffles bootstrap
> nodes unless `preserve_bootstrap_nodes_order` is enabled, and that option defaults to false
> (`core/origin.cxx:414-417`, `core/cluster_options.hxx:119`). The selected gateway is therefore
> normally nondeterministic. Document the actual selection rule or preserve input order for
> couchbase2 origins.

Confirmed at both locations. The page now states that the connection is made to a single one of the
hosts, that the list is shuffled unless that option is set so the choice does not follow the written
order, and that the log records the endpoint chosen.

> This issue also appears on line 54 of the same file.

Both occurrences were in one paragraph, which was rewritten as a whole.

The same error had also been written into a comment at `core/cluster.cxx` describing the selection as
"the first endpoint"; that is corrected in the same commit. Preserving input order for couchbase2
origins was not done: it is a behaviour change, and with a single endpoint expected the ordering
carries no meaning worth pinning.

### Verification

The API reference builds with no doxygen warnings, and the library compiles.

thejcfactor (approved):

PR #1067, opened 2026-08-10
CXXCBC-929: report Cloud Native Gateway support in the build info

avsej:
[CXXCBC-929](https://couchbasecloud.atlassian.net/browse/CXXCBC-929)

Cloud Native Gateway support is a compile-time option, `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2`, and it
is off by default. Nothing reported whether a given binary carried it.

The failure mode is quiet. A library built without the option does not reject a `couchbase2://`
connection string as malformed — it accepts the string and then fails the connect with
`feature_not_available`. To someone holding a prebuilt package that reads as a gateway, credential or
configuration problem, when the cause is the build they are running.

`sdk_build_info()` already answers this class of question for every other build-time switch it carries
(`columnar`, `static_openssl`, `static_boringssl`, `static_target`, OpenTelemetry). This one was
simply missing.

## What it looks like

```
$ cbc version
Version: 1.4.0
...
Snappy: 1.2.2
Cloud Native Gateway (couchbase2://): supported
```

```
$ cbc version --json
  "columnar": false,
  "couchbase2": true,
```

The line is printed whether or not support is present. "not supported" is the answer someone running
this command is usually after, and omitting the line would leave it indistinguishable from an older
binary that never reported anything.

## Verification

Both arms are exercised, not just the one this development build happens to use:

- Built with the option **on** — `cbc version` prints `supported` and `cbc version --json` emits
  `"couchbase2": true`. Cross-checked against `CMakeCache.txt`, which has
  `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2:BOOL=ON` and `COUCHBASE_CXX_CLIENT_COLUMNAR:BOOL=OFF`; the two
  adjacent keys report opposite values matching that configuration, so the entry is not hardcoded.
- Built with the option **off** — verified by compiling against a generated `build_config.hxx` with
  the macro undefined exactly as CMake emits it, which yields `false`.
- The JSON rendering is a boolean, not the string `"true"`. The key was added to the boolean list in
  both `sdk_build_info_json()` and `cbc`, since those two lists are maintained separately.

The unit test asserts the entry against the macro rather than a fixed value, so it holds in either
configuration and still fails if the entry stops tracking how the library was built. It was
mutation-checked: changing the expected value makes it fail with `"true" == "WRONG"`, and restoring it
makes it pass, so the assertion is live rather than vacuous.

## Note for the reviewer

`sdk_build_info_json()` and `cbc version --json` maintain separate lists of which keys are booleans,
and they already disagree: `columnar`, `static_target` and `static_boringssl` are booleans in the tool
and strings in the library rendering. The new key is added to both so it is consistent wherever it
appears. Reconciling the pre-existing divergence is deliberately left out of this change.

[CXXCBC-929]: https://couchbasecloud.atlassian.net/browse/CXXCBC-929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/meta/version.cxx
+++ b/core/meta/version.cxx
@@ -96,6 +96,16 @@ sdk_build_info() -> std::map<std::string, std::string>
     "false"
 #endif
     ;
+  // Whether this build can serve couchbase2:// connection strings, i.e. whether it was compiled
+  // with Cloud Native Gateway support. A build without it rejects the scheme at connect time, so
+  // the answer is worth reporting next to the other build-time switches.
+  info["couchbase2"] =
+#if defined(COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2)
+    "true"
+#else
+    "false"
+#endif
+    ;
   info["post_linked_openssl"] = COUCHBASE_CXX_CLIENT_POST_LINKED_OPENSSL;
   info["static_openssl"] =
 #if defined(COUCHBASE_CXX_CLIENT_STATIC_OPENSSL)
@@ -194,7 +204,7 @@ sdk_build_info_json() -> std::string
         name == "version_build" || name == "mozilla_ca_bundle_size") {
       info[name] = std::stoi(value);
     } else if (name == "snapshot" || name == "static_stdlib" || name == "static_openssl" ||
-               name == "mozilla_ca_bundle_embedded") {
+               name == "couchbase2" || name == "mozilla_ca_bundle_embedded") {
       info[name] = value == "true";
     } else {
       info[name] = value;
--- a/test/test_unit_utils.cxx
+++ b/test/test_unit_utils.cxx
@@ -172,6 +172,33 @@ TEST_CASE("unit: user_agent string", "[unit]")
           couchbase::core::meta::user_agent_for_http("0xDEADBEEF", "0xCAFEBEBE", "hello\nworld"));
 }
 
+TEST_CASE("unit: build info reports whether couchbase2 is supported", "[unit]")
+{
+  const auto info = couchbase::core::meta::sdk_build_info();
+  const auto entry = info.find("couchbase2");
+  REQUIRE(entry != info.end());
+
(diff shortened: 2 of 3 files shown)


avsej:
## Response — PR #1067 (`CXXCBC-929`), Copilot review

> The test only checks the raw string map, so removing `couchbase2` from `sdk_build_info_json()`'s
> boolean branch would still pass while changing the documented JSON value back to the string
> `"true"`/`"false"`. Extend this test to parse that rendering and assert the key's type and value.

Correct, and it is the sharper half of the change that was left uncovered: the string entry is a
one-line map insert, while the JSON boolean is the part maintained by a separate list that someone
could plausibly prune.

The case now parses `sdk_build_info_json()` and asserts the rendered key is a JSON boolean whose value
matches the string entry, so the type is pinned rather than only the presence.

The claim was verified as stated rather than taken on reading. Removing `couchbase2` from the boolean
branch of `sdk_build_info_json()` -- the exact edit described -- leaves the previous version of the
case green, and makes the extended one fail on the type assertion:

```
with expansion:
  false

test cases: 1 | 1 failed
assertions: 4 | 3 passed | 1 failed
```

Restoring the branch returns it to five passing assertions.

### What remains uncovered, deliberately

`cbc version --json` maintains its own boolean list, separate from the library's, and this case does
not reach it: it is a list inside the tool, and the unit suite has no harness that runs the tool and
parses its output. The same pruning in `tools/version.cxx` would therefore still go unnoticed by the
tests. Both lists were verified by hand for this change, and the divergence between them -- `columnar`,
`static_target` and `static_boringssl` are already boolean in the tool and string in the library -- is
noted in the PR description as pre-existing and out of scope here.

### Verification

Whole unit binary green, 67 assertions across 11 cases.

thejcfactor (approved):

PR #1073, opened 2026-08-14
CXXCBC-949: derive case names from the function, and widen the assertions

avsej:
Fourth ticket of [CXXCBC-945](https://couchbasecloud.atlassian.net/browse/CXXCBC-945). CXXCBC-946 (#1070), CXXCBC-947 (#1071) and CXXCBC-948 (#1072) have merged, so this now carries a single commit against `main`. CXXCBC-950 and CXXCBC-975 sit on top of it in #1074 and #1075.

## Why

Registering a case named it twice:

```cpp
{ "get_reqest_encodes_location", get_request_encodes_location },
```

That typo compiles. The string is the ctest entry name *and* the filter key, so a mistyped one produces a case nobody can select by the name they would guess, and nothing catches it.

Separately, the suite's commonest assertion by a wide margin is "this operation succeeded" — 1,063 `REQUIRE_SUCCESS` plus 21 `EXPECT_SUCCESS`. Catch2 renders those through `StringMaker<std::error_code>`; without an equivalent, every one of them would degrade to `expected equal`, which is the least useful message the suite could produce for its most common failure.

## What changed

**`CASE(fn)`** expands to `#fn, &fn` — two aggregate fields, at the registration point only. No macro wraps a case body: cases stay ordinary functions that go-to-definition finds and a debugger names. A registration naming a function that does not exist now fails to compile:

```
error: 'a_sucess_passes' was not declared in this scope
```

**`suite_name`** comes from `COUCHBASE_TEST_SUITE_NAME`, which `couchbase_add_test()` derives from the file's path. No hand-written suite string survives.

**`framework/test_registry.hxx`** refuses to compile beside Catch2:

```
test/framework/test_registry.hxx:26:2: error: #error "A Catch2 header is included alongside
framework/test_registry.hxx. A test file uses one or the other."
```

**`framework/errors.hxx`** (heavy, included only where needed) adds `assert_success` and `assert_error`, and a `fmt::formatter<std::error_code>`. `couchbase::error` deliberately gets no formatter — `couchbase/fmt/error.hxx` already provides one, and a second definition would be an ODR violation in any TU that included both — so it is rendered by a `describe()` overload instead.

**Lean assertions**: `assert_ne`, `assert_contains`, `assert_starts_with`, `assert_near`, `assert_no_throw`, `fail()`. Between them and what already existed, every Catch2 construct in the tree has a mapping — the full inventory is in the migration guide.

**A suite may carry a `teardown`**, run once after the last case and skipped after a timeout for the same reason `main()` leaves through `_Exit`. That is where `test/main.cxx`'s `OPENSSL_cleanup()` goes.

**The pilot**: `test_unit_config_profiles.cxx` → `test/unit/core/config_profiles.cxx`, 6 cases, registered with `couchbase_add_test(unit/core/config_profiles LABEL unit LINK_CLIENT)`. Its `CHECK`s became assertions and each case was renamed to a sentence about what it pins. The CNG tree adopts the same registration style in the same commit, so the pilot's style is the tree's style.

## The compile-time claim does not hold

Faster compilation is one of the two stated reasons for this epic, and it had never been measured. It has now, and the result is not the expected one. GCC 16, Debug, flags from `compile_commands.json`, `CCACHE_DISABLE=1`, best of three.

Whole translation unit, the same test under each framework:

| | best of 3 |
|---|---|
| `test_unit_config_profiles.cxx` (Catch2) | 1.122 s |
| `unit/core/config_profiles.cxx` (framework) | 1.101 s |

**2%, which is noise.** That file's cost is `core/cluster_options.hxx`, not the framework around it.

The headers in isolation (empty TU, 0.017 s baseline subtracted):

| header | cost |
|---|---|
| `spdlog/fmt/bundled/base.h` | 0.036 s |
| **`catch2/catch_test_macros.hpp`** | **0.216 s** |
| `spdlog/fmt/bundled/format.h` | 0.241 s |
| `framework/test_framework.hxx` | 0.343 s |
| **`framework/test_registry.hxx`** | **0.407 s** |
| `test_helper.hxx` (Catch2 + core) | 0.880 s |
| `core/cluster_options.hxx` alone | 0.710 s |

The framework header costs nearly twice Catch2's, and `fmt/format.h` alone already exceeds Catch2 outright. Across ~100 translation units that is roughly 20 seconds the migration would *add*.

The cause is one include: `test_framework.hxx` pulls `<spdlog/fmt/fmt.h>`, which pulls `bundled/format.h`, because the assertions call `fmt::format`. **`base.h` suffices** — probed directly, `fmt::format_to` into a `std::back_insert_iterator<std::string>` compiles against `bundled/base.h` alone for `int`, `double`, `std::string` and `std::string_view`, and `fmt::is_formattable` is available there too. Rewriting the helpers that way would put the framework header at roughly 0.2 s, below Catch2.

**Not done here.** The include is `<spdlog/fmt/fmt.h>` rather than a bundled header on purpose: it derives the header-only-vs-compiled mode from `SPDLOG_COMPILED_LIB`, and reaching past it is what previously emitted duplicate `fmt::v11::vformat` definitions and failed MSVC with `LNK2005`. Re-settling that needs a toolchain not available here, so it is a separate change. Recorded in §11.1 of the design document with the commands.

Until it is fixed, the speed argument should come off the epic's description rather than be repeated. What survives is not a speed claim: 301 invisible skip guards, a case count that understates executions by a third, and a runner whose failure modes are the ones a test framework should have.

## Verification

| Check | Result |
|---|---|
| pilot registered per case | 6 entries `unit_core_config_profiles.*`, all pass |
| CNG configuration | **357/357 pass** |
| Catch2 configuration | **48/48 pass** (`-L 'unit\|framework'`, targets built for this check) |
| mistyped case name | fails to compile, error quoted above |
| Catch2 header alongside the registry | `#error` fires, quoted above |
| new assertions' messages | `framework_selftest.the_added_assertions_report_what_they_saw` asserts on what each one prints on failure |
| `assert_success` output | `framework_errors_selftest`: category `couchbase.key_value`, value `101`, message `document_not_found (101)`, and the cause chain for `couchbase::error` |
| `bin/validate-test-labels` | still exits 0 (it only inspects files named `test_*`, so a migrated file is out of its scope) |

The five `framework_errors_selftest` cases were written before the code was correct and caught three wrong expectations in my own test — the messages are what the framework actually prints, not what I assumed.

## Also in this commit

`GUIDE_TEST_MIGRATION.md` (in the local docs tree, not the repo) — the Catch2-to-framework translation table for all 5,253 assertions and every construct the tree uses, the section-flattening rule, why a `SKIP` becomes a requirement rather than a `skip()`, and the traps met so far. The design document's estimate of ~4,100 assertions was wrong; the counted figure is 5,253.

[CXXCBC-945]: https://couchbasecloud.atlassian.net/browse/CXXCBC-945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avsej:
Follow-up landed: **#1075** removes the formatting library from the framework headers entirely, rather than swapping `<spdlog/fmt/fmt.h>` for `base.h` as this PR proposed.

The better question, asked in review, was whether a test framework needs a formatting library in its header at all. It does not — 5 of about 1,370 assertion call sites build a message from values. `framework/test_registry.hxx` goes from **0.390 s** to **0.243 s**, and the MSVC `LNK2005` hazard this PR deferred the work around does not arise, because every remaining fmt user still goes through the spdlog wrapper.

The §11.1 numbers quoted above are superseded by the ones in #1075.

avsej:
The red checks on this PR were triaged against `main` and against unrelated branches; none is caused by this stack.

- `windows (windows-11-arm)` — the runner runs out of disk (`LNK1116 … error code 112`, “Free space left: 23 MB”). Fails identically on `main` (run 31660572583) and on the unrelated `CXXCBC-931-packaging-install-fix` (run 31635968269).
- `integration (7.0.5 / 7.1.6 / 7.2.9)` — server-side propagation and index-catch-up races in pre-existing Catch2 files this stack does not touch. The same commits are green in sibling runs.
- `unit (tsan)` — a pre-existing race in the two close-during-backoff CNG cases, which the per-case ctest registration in #1071 makes visible far more often. Addressed there by widening the backoff window from 300 ms to 2 s, without weakening the assertion. See the comment on #1071 for the evidence.

The full triage, with job ids and the `main`-branch baselines that settle each one, is in the comment on #1071.

avsej:
The three suppressed comments were worth reading. **One of them found a real defect that my own tests could not have caught, and it is fixed.**

### 1. `hostile_keys()` starting with `suite_name` — correct, and serious

This is a defect I introduced. The `CASE(fn)` sweep in this commit applied its "replace the suite-name literal" substitution to the **whole file** rather than to the `tests()` block, so it also rewrote the first element of an unrelated data function:

```cpp
 hostile_keys() -> std::vector<std::string>
 {
   return {
-    "field",
+    suite_name,
     "two words",   // only works because the key is quoted
```

`"field"` is the plain-key baseline — the one entry in that list that is *not* hostile — and it was silently replaced by the suite name.

**The tests passed before and after**, because `cng_protostellar_query_index_admin_converter` happens to be a valid identifier that encodes to itself. So this was a silent loss of coverage that no assertion in the suite could detect: exactly the failure mode CXXCBC-945 exists to remove, introduced by CXXCBC-945. Restored.

I then audited both regexes from that sweep across the tree:

```console
$ # every suite_name outside a tests() block
test/cng/protostellar/query_index_admin_converter.cxx:149   <- this one
test/framework/test_registry.hxx:47                         <- the declaration
$ # every CASE( outside a tests() block
(none)
```

One site, now fixed. I also reconciled every case in every binary against `ctest -N` and they match exactly — nothing else was dropped or renamed.

### 2. `gate_on_readiness()` throwing leaves the cluster open — correct, fixed

The failure path stopped the io threads without closing a cluster `open_cluster()` had already opened. There is now an `unwind()` that closes first (while the io threads are still running, since close needs them to make progress) and then stops them. Any failure from the close is swallowed deliberately, because the exception on its way out is the one that describes what actually went wrong.

Verified the path still reports rather than terminating: `TEST_CONNECTION_STRING=couchbase://127.0.0.1:1` gives five named failures and exit 1, not `terminate called`.

### 3. `fmt::formatter<std::error_code>` conflicting with `<fmt/std.h>` — correct, and already resolved one commit up

Good catch, and the concern is real: a translation unit including both would have had two definitions.

It is resolved in **#1075**, which removes that specialisation entirely along with every other use of a formatting library in the framework headers — `std::error_code` is rendered through an `operand_printer<T>` specialisation instead. Nothing in the tree includes `<fmt/std.h>` today, so the window where this could bite is the one commit between this PR and that one.

360/360 pass after all three.

avsej:
Second round of suppressed comments. Both are refuted, with the checks I ran.

### `find_package(Threads REQUIRED)` removed from `test/cng/CMakeLists.txt` — moved, not dropped

It is in `cmake/TestFramework.cmake:7`, and `Threads::Threads` is linked twice: `PUBLIC` on `test_framework_main` (line 28) and `PRIVATE` on every test `couchbase_add_test()` creates (line 165). So every test target gets it whether it links the framework library directly or transitively.

No `-pthread` appears on the link line here, which is correct rather than concerning:

```console
$ grep CMAKE_HAVE_LIBC_PTHREAD build/CMakeCache.txt
CMAKE_HAVE_LIBC_PTHREAD:INTERNAL=1
```

glibc ≥ 2.34 folded pthread into libc, so `CMAKE_THREAD_LIBS_INIT` is empty and `Threads::Threads` expands to nothing. On a platform that still needs the flag, `find_package(Threads)` supplies it through the same imported target. Nothing platform-specific was assumed — the detection just answers differently here.

### Registering the profile name `"test"` in two cases — cannot throw, and cannot mis-order

The concern is that a registry rejects duplicates. This one does not:

```cpp
void register_profile(const std::string& name, Args... args)
{
  // This will just add it, doesn't look to see if it is overwriting an existing profile.
  ...
  profiles_.emplace(std::make_pair(name, std::make_shared<T>(args...)));
}
```

`emplace` is a no-op when the key exists — no throw, no replacement — and the in-tree comment says as much. Both cases register the same type (`test_profile`) under that name, so whichever runs first, the profile applied is identical and the assertions cannot depend on order. The pattern is also unchanged from the Catch2 file this migrates, where the same two `TEST_CASE`s registered the same name in the same process.

Worth recording as a latent trap rather than a defect: because `emplace` does not overwrite, two cases registering *different* types under one name would silently keep the first. That is a property of the registry, not of this test, and no test in the tree does it.

**Not changed**, since changing it would be churn without a defect behind it. If you would rather these used distinct names for clarity, say so and I will.

avsej:
**The integration failures on this stack are mine, and they are the good kind: a test I wrote met a real cluster for the first time and my assumptions were wrong.**

`framework_cluster_probes_selftest` was added in #1072 precisely so the cluster-backed probe backend would not ship untested. Until now it had only ever run against an *unreachable* endpoint — I said so in that PR. It has now run against 7.0.5, 7.1.6 and 7.2.9, and two of its assertions state universals that are not true.

### `"a cluster always has at least one server group"` — false

Server groups are a rack-awareness feature. A cluster that was never given one reports none, on every version tested. I asserted a property I had not checked.

What is always true is that the probe answers and every name it returns is a name:

```cpp
for (const auto& group : ctx.server_groups()) {
  assert_false(group.empty(), "a reported server group has a name");
}
```

### The storage backend is `unknown` before 7.1 — and that is correct

`"unknown" is not a storage backend the server reports` failed on 7.0.5 only. It is not a mapping error: the server reports no storage backend at all before 7.1, so `unknown` is the honest answer and there is nothing to check. The in-tree gate says the same thing — `supports_storage_backend()` is `is_neo() && is_enterprise()`.

So the case now declares that as a requirement rather than asserting through it:

```
the_storage_backend_is_one_the_server_names   a real cluster; the kv service; a cluster at 7.1.0 or later
```

On 7.0.5 it now skips, visibly, in the run’s "Skipped for want of" summary — which is exactly the behaviour this epic is for. The alternative, widening the assertion to accept `unknown`, would have made the case pass everywhere while testing nothing on the versions that *do* report a backend.

### A process note

My first attempt at the version gate **silently did nothing** — I pattern-matched the pre-`CASE()` registration form, and this file had already been swept to `CASE(...)` in this PR. The replace found nothing, reported success, and I only caught it because `--list-tests` still showed the old requirements. That is the same not-applied trap that bit the `hostile_keys()` sweep earlier in this stack; every scripted edit here now asserts that its pattern matched.

The fix is routed to this PR rather than #1072, where the defect originated: the file is reshaped by this commit’s `CASE()` sweep, so the corrected form cannot be expressed against #1072 without conflicting.

360/360 locally. The remaining `unit (tsan)` failures across the stack are CXXCBC-976, unchanged — both close-during-backoff cases, same lines, with the unit label itself passing 369/369.

avsej:
`windows (windows-2025)` here is the same disk exhaustion analysed on #1074 — no `C2220`, no `LNK2005`, no error attributable to any target this stack owns. This run is the starker data point:

```
Free space left: 0 MB
```

Not marginal — the runner filled completely. The same job also fails on `main` (run 31660572583).

The full analysis, including this stack’s measurable contribution (net +3 test binaries of 98, each carrying a `.pdb` and an `.ilk` under `RelWithDebInfo`) and the CI-side fixes that would help `main` equally, is on #1074. Not repeating it here.

avsej:
The valgrind integration legs are red here. They are red on `main` too, and more often than here. Evidence below rather than an assertion.

**Rate on `main`.** The last three completed `sanitizers` runs on `main`, counting valgrind integration shards:

| run | date | shards failed |
|---|---|---|
| 31660572579 | 2026-08-13 | 2 of 4 |
| 31621108883 | 2026-08-12 | 3 of 4 |
| 31445412770 | 2026-08-11 | 3 of 4 |

**8 of 12 shards fail on `main`.** Valgrind integration has not been green on `main` in the recorded history I can reach.

**What actually fails.** Every failure across `main` and this stack, deduplicated:

| test | where |
|---|---|
| `collections query index management` | main |
| `a **node_ids** future completes when the cluster is closed mid-bootstrap` | main |
| `cluster remains usable in a forked child` | main |
| `legacy durability persist to active and replicate to one` | here |
| `bucket wait_until_ready` | here |
| `remove failure on missing doc carries node_id` | here |
| `a **scan** future completes when the cluster is closed mid-bootstrap` | here (twice) |

Two things about that list.

Every entry is a **Catch2 integration test**. Not one framework test failed on any shard, on either side — 122 or 123 of 123 pass every time, and the `framework_*` entries are present and green in each log. Whatever this is, it is not the framework.

The names barely repeat. A deterministic break produces the same failure every run; a rotating cast is flakiness. Note especially that `main` fails the **node_ids** variant of the close-during-bootstrap test while this stack fails the **scan** variant — sibling tests asserting the same property, in the same shard. That is one pre-existing race sampling a different sibling per run, not two different problems.

**The one that is diagnosable from the log** is `remove failure on missing doc carries node_id`:

```
test/test_integration_node_id.cxx:698: FAILED:
  REQUIRE( err.ec() == couchbase::errc::key_value::document_not_found )
with expansion:
  couchbase.common:14 (unambiguous_timeout (14))
```

A timeout, not a wrong error code, and the surrounding log shows a configuration update landing at that moment (`will update the configuration old=1:1681 -> new=1:1682`) with the operation deferred behind it. Under valgrind that is ordinary slowness, not a defect in what the test asserts.

**What I am not claiming.** I have not proven any individual failure here would also have occurred on `main` at the same commit — these are flaky, so a single run proves nothing in either direction. What the numbers do establish is that the valgrind integration legs fail at a high base rate independently of this work, that this stack's failure count sits at or below that base rate, and that the framework tests are not implicated.

Separately, and worth acting on outside this epic: the valgrind shards are close to their time budget. Shard 2/2 on `main` runs 116 of its 120 minutes, and one shard on this stack has already timed out. Re-sharding or a longer timeout would be needed before more tests land, regardless of what happens with this epic.

avsej:
`windows (windows-2025)` fails here on disk, and unlike the other red legs on this stack I do **not** think this one can be filed as pre-existing. Setting out the evidence both ways.

**It is disk, not code.** Zero `C2220`, zero compile errors. Every linker code in the log is a write failure:

| code | meaning |
|---|---|
| `LNK1106` | invalid file or disk full |
| `LNK1108` | cannot write file |
| `LNK1116` | cannot open `.ilk` |
| `LNK1201` | error writing to program database |

plus `ENOSPC: no space left on device` from the runner itself before the cache step.

**Base rate on `main` is low.** Across the last four completed `build` runs on `main`, `windows-2025` failed **once** (2026-08-13) and passed three times. Compare `windows-11-arm`, which failed **4 of 4** — that one is genuinely pre-existing and unrelated to this work.

**Distribution across the stack is monotone.** `windows-2025` passes on the two lowest PRs and fails on the next three:

| PR | ticket | windows-2025 |
|---|---|---|
| #1070 | CXXCBC-946 | pass |
| #1071 | CXXCBC-947 | pass |
| #1072 | CXXCBC-948 | **fail** |
| #1073 | CXXCBC-949 | **fail** |
| #1074 | CXXCBC-950 | **fail** |
| #1075 | CXXCBC-975 | pending |

Each PR carries the ones below it, so content accumulates up the stack. A clean pass/pass/fail/fail/fail step is what incremental disk pressure on a marginal runner looks like.

**The honest reading.** This stack adds a net 3 test binaries to 98 (~3%), each with a `.pdb` and an `.ilk` alongside it. That is a small increment, and on a runner with headroom it would not matter. On `windows-2025` it appears to be enough, because the runner was already close enough to the edge to fail once on `main` without any help from this epic.

So: the underlying fragility is not caused by this work, but this work plausibly is what tips it over. Both halves of that matter — it would be wrong of me to claim the failures are somebody else's problem, and equally wrong to treat a 3% increment as the root cause.

**What would fix it** is repo-wide rather than anything this epic can reach: `/INCREMENTAL:NO` on the Windows test link (which removes the `.ilk` files outright), or building tests in Release on Windows. Both live outside the paths this epic is allowed to touch (`test/`, `cmake/TestFramework*.cmake`, `bin/`, `docs/`), so I have not made the change. I can confine `/INCREMENTAL:NO` to just the three targets the framework creates if that is preferred, but it would recover very little and would leave those three inconsistent with the other 95.

Flagging it for a decision rather than quietly widening the scope of these PRs.

avsej:
## Revised after a multi-lens adversarial review

The stage-1 work was put through six independent reviews — C++ correctness and UB, an adversarial "can a broken test pass" pass, CMake/ctest, API and migration scalability, portability, and contract fidelity of comments and documentation — each paired with a verifier whose default was to **refute** the findings rather than accept them. 65 raw findings, 32 survived verification, 13 changes landed across the stack.

Force-pushed. Every commit was configured, built and its selftests run in isolation afterwards.

### The finding that mattered most

**The framework's own selftests were not being run by CI.** They carried `LABEL framework`, and nothing in `bin/` or `.github/workflows/` selects that label — the only labels ever passed to `--label-regex` are `unit`, `integration`, `transaction`, `benchmark`, `cng`. Before this stack the selftest was registered by `add_cng_test(...)`, i.e. `cng`, which CI does run. This stack invented a label with no runner, and 34 cases went dark: compiled, linked, never executed.

Two of the six lenses found this independently.

It also corrects something stated on these PRs earlier. The claim that no framework test had ever failed in CI was true, but largely because most of them were not running. Only `framework_cluster_probes_selftest` (`integration`) executed — which is why it was the only `framework_*` entry ever visible in a CI log, and reading its presence as evidence of coverage was a mistake.

`couchbase_add_test()` now rejects any label no leg selects, so this cannot recur silently.

### The second-most serious

**A requirement that called `skip()` ran its case anyway and reported it passed.** `check_requirements()` handled the three `check_result` statuses and the `failed` outcome of the bounded phase, but a `test_skip_exception` thrown from inside `check()` fell through to a default-constructed gate whose verdict is `run`. A precondition that had just declared itself unmet produced a green case with no output line — precisely the outcome the three-valued requirement design exists to prevent. It now skips, and a selftest pins it (removing the fix turns that case red).

### In this commit

- `a_suite_keeps_its_teardown_hook` is deleted. It assigned `s.teardown` and then asserted it was non-null — a test of `std::function` assignment, not of the framework, and `run()` takes a `const&` so it could not have cleared it.
- The `assert_near` message assertion checked only `"expected: 2"` under a message promising "the expectation **and** the tolerance". It now checks `"expected: 2 ± 0.1"`; dropping the tolerance from the rendered message kills it, where the old form passed unchanged.

Verified: 26/26 plus 5/5 in `errors_selftest`.

avsej:
Both threads on this pull request are addressed; neither is a defect in the commit it is filed against.

> This requirement worker shares `ctx` with every later case, but `run_bounded` detaches it when the budget expires. A timed-out probe can therefore continue updating the unsynchronized cache maps [...] causing a C++ data race and undefined behavior.

Fixed in CXXCBC-948, the commit below this one, which this branch carries. `context::impl` now holds a mutex taken by every accessor that touches the caches, the backend or the counter, held across the probe. The case-body path had the same exposure, which is why the guard went into `context` rather than either call site. `one_probe_answers_every_caller_that_arrives_at_once` pins it: eight threads meet at a barrier against a 50 ms probe and the backend must be asked once; without the lock it is asked eight times.

Ending the run at the first timeout — the alternative the note suggests — was not taken, and the reasoning is now stated in the code where the run continues.

> `assert_throws` catches `const Exc&` before its special cases for `test_skip_exception` and `test_assertion_failure`.

Real, and fixed — but not here. `assert_throws` is unchanged by this commit and by every commit on this branch: the function is already in `main`, so the diff against `main@origin` for `test/framework/test_framework.hxx` contains no `assert_throws` hunk. It is corrected against `main` as CXXCBC-1003 in #1097, together with the first self-tests that function has ever had, and arrives here when that merges.

Worth recording that the note understated it. In the default build the wrong order does not merely misreport — it does not compile, because instantiating `assert_throws<std::exception>` leaves a handler unreachable and `-Wexceptions` is fatal. The silent pass it describes is reachable only where those warnings are demoted.

thejcfactor (approved):

PR #1077, opened 2026-08-17
CXXCBC-981: bounds-check the replica chain row in server_by_vbucket

avsej:
https://jira.issues.couchbase.com/browse/CXXCBC-981

`configuration::server_by_vbucket()` bounds-checked the outer vbucket-map vector and then indexed the replica chain row without checking its length. Nothing maintains the invariant that a row holds `num_replicas + 1` entries — `numReplicas` and the row widths are parsed independently in `configuration_json.hxx`, while `effective_nodes()` and `observe_poll` derive the replica index from `num_replicas`.

The fix guards the row in the accessor every caller routes through (`map_key()` both overloads, `effective_nodes()`, `observe_poll`, `bucket::map_id()`). A surplus index returns an empty optional, which is the answer callers already handle for an unassigned (`-1`) slot, so no caller changes and a short row behaves exactly like a `-1`-padded one.

This is not only a crash fix. `_GLIBCXX_ASSERTIONS` is what turned the read into an abort on the build the reproduction ran on; without it the read returns whatever follows the row, and a non-negative `int16_t` below `nodes.size()` there is a plausible node index that routes the request to a node the vbucket does not live on.

## How a short row happens on a live server

A bucket reconfigured to more replicas publishes the new count immediately and keeps the narrower map until the rebalance that places them. Against Couchbase Server 8.0.1, a bucket created with `replicaNumber=0` and then updated to `replicaNumber=2` publishes `numReplicas: 2` with rows of length 1, and stays in that state (a single-node cluster cannot place the replicas):

```
t= 0.04  numReplicas=2  rowlen=[1..1]  SHORT=True
```

Note that "not enough nodes to place replicas" alone does *not* produce this: a bucket created directly with `replicaNumber=2` on one node is published with full-width rows padded with `-1`. The reconfiguration is what leaves the two fields disagreeing.

## Reachability, before the fix

Both live paths abort on the unguarded build against that bucket — no new API involved:

```
get_all_replicas → stl_vector.h:1272: Assertion '__n < this->size()' failed
  #6 configuration::server_by_vbucket (vbucket=47, index=1)  configuration.cxx:310
  #7 configuration::map_key (key="cxxcbc981", index=1)       configuration.cxx:325
  #8 impl::effective_nodes                                    replica_utils.cxx:41
  #9 get_all_replicas_request::execute            document_get_all_replicas.hxx:85

upsert with replicate_to::one → same assertion, via observe_poll
```

The existing replica-read integration cases skip when `number_of_nodes() <= number_of_replicas()`, which is why CI has not caught this.

## The same accessor divided by the map's size

`configuration::map_key()` derives the vbucket as `crc % vbmap->size()` after checking only that the map is present. A map that is present but carries no rows makes that a division by zero — `bucket::map_id()` reaches it for every key-value operation. Both overloads now reject an empty map the way they already reject an absent one, returning vbucket 0 and no server. Caught by ASan (`SUMMARY: AddressSanitizer: FPE configuration.cxx:328`) while adding the row guard's tests.

## The two range-scan sites

`range_scan_orchestrator` and `range_scan_load_balancer` read the active copy as `vbucket_map[vbucket][0]` with only the outer vector checked, bypassing the accessor. They are out of bounds only for a row that names no node at all, which is not the shape the reconfiguration window produces — but a zero-width row cannot be worth guarding in the accessor and not in two direct reads of the same map.

A scan needs an active copy for every vbucket it covers, so this case is refused rather than worked around: `range_scan_orchestrator::scan()` reports `errc::network::configuration_not_available` before it starts a stream, and the load balancer leaves such vbuckets unqueued. Grouping them under node id `-1` — the id the map uses for a copy that is not placed — was the smaller diff, but it invents a node no request can reach and hands the scan vbuckets it can never cover, so a scan would come back silently short. Nothing changes for a row whose active copy is `-1`: that row still names a node id, and it still routes as it does today.

## Verification

Unit: `test/test_unit_topology_configuration.cxx` covers index at the row length, index past it, index 0 against an empty row, an unassigned slot, an assigned slot, `map_key()` against a map without rows, and `effective_nodes()` driven by a short-row configuration. `test/test_unit_range_scan.cxx` covers a load balancer built from a map whose middle row is empty: the two routable vbuckets are selected once each and no third is offered. Every crashing case was watched failing before the fix — the row reads as an `_GLIBCXX_ASSERTIONS` abort, `map_key()` as an ASan-reported `FPE` — and the behaviour-pinning cases passed before it, so the guards cannot be written in a way that changes them. The orchestrator's refusal has no unit case: constructing it needs a KV agent and an `io_context`. The full unit suite (373 tests) passes under an AddressSanitizer build (clang, with `_GLIBCXX_ASSERTIONS` active).

Real cluster: after the fix, replica reads against the short-row bucket return the active copy only, and a `replicate_to::one` write fails with `ambiguous_timeout` — the same outcome a `-1`-padded row already produces. A well-formed bucket whose rows are `-1`-padded produces byte-identical outcomes, confirming the guard adds no new behaviour. On a three-node cluster with a two-replica bucket, `test_integration_read_replica` and `test_integration_durability` pass unchanged.

## Release

This merges once to `main`, and `release-1.4` is branched from that merge, so the fix reaches both 1.4.1 and 1.5.0 without a cherry-pick. It shares no code with the GetReplica feature (CXXCBC-980) that depends on it.

Two consequences of that ordering. Nothing 1.5-only may merge to `main` before the branch point, or it lands in 1.4.1 as well. And the tests live in `test/test_unit_topology_configuration.cxx` rather than in the feature's test file, so the 1.4 branch carries a file named after what it tests.

### DIFF
--- a/core/range_scan_load_balancer.cxx
+++ b/core/range_scan_load_balancer.cxx
@@ -81,8 +81,15 @@ range_scan_load_balancer::range_scan_load_balancer(
 {
   std::map<std::int16_t, std::queue<std::uint16_t>> node_to_vbucket_map{};
   for (std::size_t vbucket_id = 0; vbucket_id < vbucket_map.size(); vbucket_id++) {
-    auto node_id = vbucket_map[vbucket_id][0];
-    node_to_vbucket_map[node_id].push(gsl::narrow_cast<std::uint16_t>(vbucket_id));
+    const auto& chain = vbucket_map[vbucket_id];
+    if (chain.empty()) {
+      // A row that names no node at all cannot be scanned, and there is no node
+      // to queue it against. range_scan_orchestrator refuses such a map before
+      // it starts any stream, so leaving the vbucket out here cannot silently
+      // shorten a scan.
+      continue;
+    }
+    node_to_vbucket_map[chain[0]].push(gsl::narrow_cast<std::uint16_t>(vbucket_id));
   }
   for (auto [node_id, vbucket_ids] : node_to_vbucket_map) {
     nodes_.emplace(node_id, std::move(vbucket_ids));
--- a/core/range_scan_orchestrator.cxx
+++ b/core/range_scan_orchestrator.cxx
@@ -31,6 +31,7 @@
 
 #include <gsl/util>
 
+#include <algorithm>
 #include <atomic>
 #include <chrono>
 #include <future>
@@ -398,6 +399,16 @@ public:
       return cb(errc::common::invalid_argument, {});
     }
 
+    // A scan needs an active copy for every vbucket, and every row of the map is
+    // indexed below to find one. A row that names no node at all leaves the scan
+    // unable to cover its vbucket, which is reported rather than served as a
+    // silently short result.
+    if (std::any_of(vbucket_map_.begin(), vbucket_map_.end(), [](const auto& chain) {
+          return chain.empty();
+        })) {
+      return cb(errc::network::configuration_not_available, {});
+    }
+
     const get_collection_id_options get_cid_options{ options_.retry_strategy,
                                                      options_.timeout,
                                                      options_.parent_span };
@@ -429,7 +440,8 @@ public:
           };
 
           // Get the active node for the vbucket (values in vbucket map are the active node id
-          // followed by the ids of the replicas)
+          // followed by the ids of the replicas). scan() has already established that every row
+          // names one.
           auto node_id = self->vbucket_map_[vbucket][0];
 
           auto stream = std::make_shared<range_scan_stream>(
--- a/core/topology/configuration.cxx
+++ b/core/topology/configuration.cxx
@@ -302,7 +302,16 @@ configuration::server_by_vbucket(std::uint16_t vbucket, std::size_t index) const
   if (!vbmap.has_value() || vbucket >= vbmap->size()) {
     return {};
   }
-  if (auto server_index = vbmap->at(vbucket)[index]; server_index >= 0) {
+  // A chain row may be shorter than num_replicas + 1: the server publishes the
+  // new replica count as soon as the bucket is reconfigured, while the vbucket
+  // map keeps its old width until the rebalance that places the replicas.
+  // Callers derive the index from num_replicas, so a surplus index must read as
+  // "no server" rather than past the end of the row.
+  const auto& chain = vbmap->at(vbucket);
+  if (index >= chain.size()) {
+    return {};
+  }
+  if (auto server_index = chain[index]; server_index >= 0) {
     return static_cast<std::size_t>(server_index);
   }
   return {};
@@ -312,7 +321,9 @@ auto
 configuration::map_key(const std::string& key, std::size_t index) const
   -> std::pair<std::uint16_t, std::optional<std::size_t>>
 {
-  if (!vbmap.has_value()) {
+  // The number of vbuckets comes from the map itself, so a map without rows
+  // cannot be asked which vbucket a key belongs to.
+  if (!vbmap.has_value() || vbmap->empty()) {
     return { 0, {} };
   }
   const std::uint32_t crc = utils::hash_crc32(key.data(), key.size());
@@ -324,7 +335,9 @@ auto
 configuration::map_key(const std::vector<std::byte>& key, std::size_t index) const
   -> std::pair<std::uint16_t, std::optional<std::size_t>>
 {
-  if (!vbmap.has_value()) {
+  // The number of vbuckets comes from the map itself, so a map without rows
+  // cannot be asked which vbucket a key belongs to.
+  if (!vbmap.has_value() || vbmap->empty()) {
     return { 0, {} };
   }
   const std::uint32_t crc = utils::hash_crc32(key.data(), key.size());
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -85,6 +85,7 @@ unit_test(management_bucket)
 unit_test(management_analytics)
 unit_test(node_id)
 unit_test(wait_until_ready)
+unit_test(topology_configuration)
 target_link_libraries(test_unit_jsonsl PRIVATE jsonsl)
 
 integration_benchmark(get)
--- a/test/test_unit_range_scan.cxx
+++ b/test/test_unit_range_scan.cxx
@@ -104,3 +104,28 @@ TEST_CASE("unit: range scan load balancer", "[unit]")
     REQUIRE_FALSE(balancer.select_vbucket().has_value());
   }
 }
+
+TEST_CASE("unit: range scan load balancer with a vbucket that has no active copy", "[unit]")
+{
+  // A row that names no node at all is left out rather than read past its end,
+  // and no node is invented to hold it. range_scan_orchestrator refuses a map
+  // in this state, so the vbucket cannot go missing from a scan that runs.
+  couchbase::core::range_scan_load_balancer balancer{ {
+    { 0 },
+    {},
+    { 1 },
+  } };
+
+  std::set<std::uint16_t> selection{};
+  for (auto i = 0; i < 2; i++) {
+    auto v = balancer.select_vbucket();
+
+    REQUIRE(v.has_value());
+
+    auto [_, inserted] = selection.insert(v.value());
+    REQUIRE(inserted);
+  }
+
+  REQUIRE(selection == std::set<std::uint16_t>{ 0, 2 });
+  REQUIRE_FALSE(balancer.select_vbucket().has_value());
+}
--- a/test/test_unit_topology_configuration.cxx
+++ b/test/test_unit_topology_configuration.cxx
@@ -1,0 +1,132 @@
+/* -*- Mode: C++; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ *   Copyright 2026-Present Couchbase, Inc.
+ *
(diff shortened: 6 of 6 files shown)


thejcfactor:
🤖 This review was produced with Claude Code and posted on my behalf.
This was a static review — nothing was built or run on my side; I traced the
config parser and the consuming call sites by hand and leaned on your own
cluster verification for the runtime behaviour. Please treat the reasoning as a
starting point for discussion, not ground truth.

---

LGTM. The premise holds up on a read: `configuration_json.hxx` sizes each row
from the published array (`vbmap[i].resize(p.size())`) while `numReplicas` is
parsed separately a few lines up, and both `effective_nodes()` and
`observe_poll` derive their index from `num_replicas` — so nothing was keeping
the two in agreement. Guarding the single accessor they all funnel through is
the right layer, and returning an empty optional makes a short row
indistinguishable from a `-1` slot, which is why nothing downstream needed to
change.

One thing worth adding to the writeup, since it strengthens the case: this is a
fix in release builds too, not only where `_GLIBCXX_ASSERTIONS` is active.
Without the assertion the old read didn't abort — it read whatever followed the
row, and if that happened to be a non-negative `int16_t` below `nodes.size()`
it returned a plausible-looking node index and routed the request to the wrong
node. The guard removes that silent-misroute path as well as the crash.

Two observations, neither of which I think should hold this up:

**1. Two row reads still bypass the accessor.**
`range_scan_orchestrator.cxx:433` and `range_scan_load_balancer.cxx:84` both do
`vbucket_map[vbucket][0]` directly on the map rather than going through
`server_by_vbucket()`. Those are only out of bounds if a row is *empty*, which
isn't what the reconfiguration window produces (your evidence shows
`rowlen=[1..1]`), so this isn't the bug you're fixing. But the new
`"index zero against an empty chain"` section treats a zero-width row as worth
guarding, and on the range-scan path that case is still unguarded. If you'd
rather keep this PR to the reported defect — which seems reasonable given it
also needs to land cleanly on `release-1.4` — a follow-up CXXCBC for the
range-scan sites would cover it.

**2. `replicate_to` on a short-row bucket waits out the timeout.**
`validate_replicas()` (`observe_poll.cxx:107-116`) compares only against
`num_replicas`, so a short-row bucket passes validation, the surplus
`observe_seqno_request` routes to `nullopt`, and `direct_dispatch` defers it
(`bucket.cxx:269-277`) until the durability timeout expires — the
`ambiguous_timeout` you saw. That's the same outcome a `-1`-padded row already
produces, so leaving it is defensible and arguably the consistent choice. Just
noting the fix trades a crash for a slow timeout rather than a fast
`durability_impossible`; if fail-fast is wanted there, it's a separate change
(and a candidate for the same follow-up ticket).

Nice touch that `wait_until_ready.cxx` already carried the equivalent
`chain.size() < copies` check — good corroboration that short rows were an
acknowledged shape here before this came up.

thejcfactor (approved):

avsej:
Both observations are addressed in the branch, one by a code change and one by a decision to leave the behaviour alone. The commit message was also rewritten to carry the release-build point.

> One thing worth adding to the writeup, since it strengthens the case: this is a fix in release builds too, not only where `_GLIBCXX_ASSERTIONS` is active. Without the assertion the old read didn't abort — it read whatever followed the row, and if that happened to be a non-negative `int16_t` below `nodes.size()` it returned a plausible-looking node index and routed the request to the wrong node.

Correct, and the commit message now says so: where the standard library does not check the access, the read returns whatever follows the row, which can be a plausible node index and route the request to a node the vbucket does not live on. The crash was the visible symptom on the build the reproduction ran on, not the whole defect.

> **1. Two row reads still bypass the accessor.** `range_scan_orchestrator.cxx:433` and `range_scan_load_balancer.cxx:84` both do `vbucket_map[vbucket][0]` directly on the map rather than going through `server_by_vbucket()`. Those are only out of bounds if a row is *empty*, which isn't what the reconfiguration window produces (your evidence shows `rowlen=[1..1]`), so this isn't the bug you're fixing. But the new `"index zero against an empty chain"` section treats a zero-width row as worth guarding, and on the range-scan path that case is still unguarded. If you'd rather keep this PR to the reported defect — which seems reasonable given it also needs to land cleanly on `release-1.4` — a follow-up CXXCBC for the range-scan sites would cover it.

Fixed here rather than deferred, since the inconsistency the observation identifies is real: a zero-width row cannot be worth guarding in one accessor and not in two direct reads of the same map.

The range-scan case does not take the accessor's rule, though, and that distinction matters. A scan needs an active copy for every vbucket it covers, so a row that names no node at all is refused rather than absorbed: `range_scan_orchestrator::scan()` reports `errc::network::configuration_not_available` before starting any stream, and the load balancer leaves such vbuckets unqueued. The smaller diff was to group them under node id `-1`, the id the map already uses for a copy that is not placed, and that was the first version of this change — it is wrong, because it invents a node no request can reach and then hands the scan vbuckets it can never cover, turning an unusable map into a silently short result. A row whose active copy is `-1` is untouched by all of this: it names a node id, and it still routes exactly as before.

The load balancer is constructible from a vbucket map alone, so the regression is pinned by a unit case in `test/test_unit_range_scan.cxx`: a map whose middle row is empty. It aborts before the fix with the same `_GLIBCXX_ASSERTIONS` signature as the `server_by_vbucket` cases, and after it the two routable vbuckets are selected once each with no third offered. The orchestrator's refusal has no equivalent case — constructing it needs a KV agent and an `io_context` — so that half is not covered by a test.

Auditing the zero-width row across the map's readers turned up a third site, in the accessor pair itself: `map_key()` derives the vbucket as `crc % vbmap->size()` after checking only that the map is present, so a map present but carrying no rows is a division by zero, and `bucket::map_id()` reaches it for every key-value operation. ASan reports it as `FPE configuration.cxx:328`. Both overloads now reject an empty map the way they already reject an absent one. That case is covered.

> **2. `replicate_to` on a short-row bucket waits out the timeout.** [...] Just noting the fix trades a crash for a slow timeout rather than a fast `durability_impossible`; if fail-fast is wanted there, it's a separate change (and a candidate for the same follow-up ticket).

Not changed here, and deliberately: `ambiguous_timeout` is what a `-1`-padded row already produces, so the short-row case now behaves the same way as the shape the server publishes when it has too few nodes to place the copies. Making the surplus copy fail fast means `validate_replicas()` consulting the vbucket map rather than `num_replicas` alone, which changes the outcome of an operation that has never crashed — a behaviour change, and not one to carry onto `release-1.4` alongside a memory-safety fix. It is worth its own ticket if fail-fast durability is wanted.

> Nice touch that `wait_until_ready.cxx` already carried the equivalent `chain.size() < copies` check

That check is also the reason the two fields must be allowed to disagree: `impl::vbucket_map_ready()` computes `copies = num_replicas + 1` and returns false while `chain.size() < copies`, so normalising the rows to the configured width at parse time — which is what the JVM SDK does — would leave `wait_until_ready` unable to tell a placed map from an unplaced one. Reporting the configured count truthfully and bounds-checking at the point of routing is the only arrangement that keeps both readable.

## Copilot's suppressed notes

Reviewed at `07ca3fce`. None of the three results in a change, and the reasons differ.

> Comparing an `optional` directly to an integer may not compile depending on the project's C++ standard/library (this comparison is not uniformly supported pre-C++20).

This does not hold. `operator==(const std::optional<T>&, const U&)` is part of `<optional>` as introduced in C++17, which is this project's floor, and the comparison is already exercised across every compiler in the build matrix on this branch. Left as it is.

> The "empty chain => -1" interpretation is now duplicated here and in `core/range_scan_orchestrator.cxx`. [...] consider centralizing this into a small shared helper.

Superseded: that interpretation no longer exists in either file. The orchestrator refuses a map with such a row and the load balancer leaves the vbucket unqueued, so there is no sentinel to centralise.

> The loop bound `3` is a magic number tied to the number of vbuckets provided in the test input. Using the input size [...] would make the test more robust.

Declined, because deriving the bound from the input would assert the wrong thing. The map has three rows of which two are routable, and that difference is the subject of the case: it selects twice, compares the result against the literal set `{ 0, 2 }`, and then requires that no third vbucket is offered. A bound taken from the input size would fail on the very behaviour being pinned.

## Copilot's suppressed notes, second round

Reviewed at `e7a98c4b`. The three notes are one proposal: reject rows whose active entry is `-1`, not only rows that are empty. Declined, on two grounds.

The premise does not hold. `node_id` is not what a scan routes by: `range_scan_stream` dispatches through `agent_.range_scan_create(vbucket_id_, …)`, which resolves the vbucket to a node itself, and the id is carried only for the log line and as the key the load balancer and stream manager use for their bookkeeping. A `-1` there does not produce a misrouted request; it produces a stream whose create fails or retries through the agent, which is what it does today.

And the change would be a behavioural one, in the wrong direction, on a patch bound for `release-1.4`. An unassigned active copy is a transient rebalance state, and today a scan started in that window attempts the vbucket and settles once the map does. Refusing the whole scan up front would convert those into `configuration_not_available` failures that currently succeed. An *empty* row is different in kind: it names nothing, cannot become routable by retrying the same snapshot, and is the case that was reading out of bounds — which is why it, and only it, is refused.

Values outside the known node-id domain are worth separating out: the accessor already applies that bound for replica reads (`server_index < nodes.size()` in `effective_nodes()` and in the new resolution policy), and extending it to the range-scan active is a defensible follow-up, but it is a behaviour change rather than a memory-safety fix and does not belong here.

PR #1078, opened 2026-08-18
CXXCBC-980: add collection::get_replica()

avsej:
Adds `collection::get_replica()`: a read of one named copy of a document, chosen by index. Until now the only way to read a replica was `get_any_replica()`, which returns whichever copy answers first.

Ticket: [CXXCBC-980](https://jira.issues.couchbase.com/browse/CXXCBC-980). Spec: [RFC-0053](https://github.com/couchbaselabs/sdk-rfcs/blob/36f7d3f3b8ca3a514fe04173afb7e3342055d106/rfc/0053-sdk3-crud.md?plain=1#L696-L748).

```cpp
auto strategy = couchbase::get_replica_strategy::from_index(couchbase::replica_index::second);
auto [err, result] = collection.get_replica(id, strategy).get();
```

Both arguments are required. Passing a strategy rather than a bare index leaves room for the other strategies RFC-0053 anticipates, without another method on `collection`.

## How an index is resolved

The index is resolved against the topology the SDK already holds, so a request that cannot be satisfied fails without a network round trip:

- at or past the bucket's configured replica count: `replica_index_out_of_bounds`
- within that count, but not readable now: `replica_index_currently_unavailable`

`wrap()` resolves such an index to the next readable replica instead of failing.

Neither error goes through the retry orchestrator. It consults the request's retry strategy first, and a strategy that retries everything would loop on a condition no retry can change.

A vbucket map that has not arrived, or arrived without a replica count, is not an index error. Those requests retry.

## Two lengths, two questions

The bucket's configured replica count decides whether an index exists. The vbucket map's row for the document decides whether that copy can be read right now. Conflating them gets both directions wrong, and the server produces each of them.

Dropping `replicaNumber` from 2 to 1 on Server 8.0.1 sets `numReplicas` to 1 straight away and leaves every row 3 wide until a rebalance. Bounding on the row would let `get_replica(second)` read a copy the bucket no longer keeps.

A count *increase* is the mirror image: the higher count is published first and the rows widen at the rebalance. The index is valid, so reporting `replica_index_out_of_bounds` would tell a caller to stop asking for something that is about to work. It reports `replica_index_currently_unavailable`.

`wrap()` walks the configured count rather than the row, so an index landing past a short row still reaches the replicas the row does carry.

## Routing

A trait marks a request as resolving its own route. The choice is per request, so such a request may also decline, and then takes the vbucket-map routing every other request takes. The replica fan-out callers decline, because they pin the replica through `id.node_index()`. `get_all_replicas`, `get_any_replica` and `lookup_in_all_replicas` are unchanged.

The request lives in `core::operations` beside `get_all_replicas` and `get_any_replica`, so a wrapper finds it where it finds the others, and it carries a core-local selection type rather than the public strategy, which keeps a public options header out of core.

`couchbase2://` does not carry the operation and reports `feature_not_available`.

## Tests

Unit cases cover the resolution table and the routing decisions without a cluster. Integration against 6-node Server 8.0.1 with two replicas covers the new operation, the fan-out cases and the subdocument replica suite, under AddressSanitizer.

Zone-aware routing is not covered: those cases skip without server groups.

## Compatibility

Committed in 1.5.0. Existing replica reads are unchanged. [SDKQE-3754](https://jira.issues.couchbase.com/browse/SDKQE-3754) tracks the integration-test work.

[CXXCBC-980]: https://couchbasecloud.atlassian.net/browse/CXXCBC-980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDKQE-3754]: https://couchbasecloud.atlassian.net/browse/SDKQE-3754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avsej:
One of the three is a real defect and is fixed; the other two are declined, with reasons.

> If `config.num_replicas` is unset, `value_or(0)` forces `number_of_replicas` to 0 and this returns `replica_index_out_of_bounds` even when the vbucket map row clearly lists replicas [...] A safer approach is to treat an absent `num_replicas` as 'unknown' and derive the replica count from the chain (or at least use `value_or(chain.size() - 1)`), then cap as needed when `num_replicas` is present.

The defect is real and reachable, not hypothetical. `configuration_json.hxx` parses `numReplicas` inside an `if (o.find("numReplicas") != o.end())` while `vbmap` is filled from `vBucketMap` in a separate block, so a configuration carrying a map with no count is a state the parser can produce — the same independence between the two fields that made [CXXCBC-981](https://jira.issues.couchbase.com/browse/CXXCBC-981) possible. Against such a snapshot every call returned `replica_index_out_of_bounds`, a terminal error, for a bucket whose rows list replicas.

The remedy is not the one suggested, though. Deriving the count from the row would route to a copy the bucket may not advertise, which is the property rows 17 and 18 of the resolution table exist to prevent, and it would put this SDK at odds with reporting the configured count truthfully. An absent count instead means the bounds cannot be decided at all, so it now takes the same path as an absent or empty map: no error code, no server, and the operation retries, ending in `unambiguous_timeout` if the count never arrives. Two sections in `test/test_unit_get_replica.cxx` pin it, one without `wrap` and one with, and `value_or(0)` is gone from the resolver now that the optional is checked.

> This return statement relies on being able to brace-initialize a derived `built` type from its base subobject [...] That only works with C++20 aggregate initialization of base classes [...] If the project is built with C++17, this is a compile error.

This does not hold. Aggregate initialization of a class with base classes is C++17, not C++20 (P0017R1), and the project targets exactly that — `cmake/CompilerOptions.cmake` requests `cxx_std_17`. The identical construct is already shipping four lines into `get_any_replica_options::build()`, which returns `{ build_common_options(), read_preference_ }` for a `built` derived from `common_options<...>::built`. Every compiler in the build matrix accepts both. No change.

> Correct grammar in the skip message: 'less or equal to' → 'less than or equal to'.

Declined for consistency rather than disagreement. That sentence appears verbatim in eight other guards in the same file, all pre-existing; the new case copies them deliberately so the skip output reads identically wherever it comes from. Correcting one of nine would make the file inconsistent, and correcting all nine is unrelated churn in a feature change. Worth doing on its own if it is worth doing.

## Copilot's second round

Reviewed at `c5fd8109`. The suppressed note about the retry strategy is the most valuable comment either review produced; the other two are addressed and declined respectively.

> If `common_options` includes a user-provided retry strategy (as used elsewhere in the SDK), it looks like it might not be propagated into `get_replica_request::retries`, meaning custom retry strategies could be silently ignored for `get_replica`.

Correct, and fixed. Every other operation in `collection.cxx` passes `{ options.retry_strategy }` as the request's `retries` field; this one passed `{}`, so a user-supplied strategy was constructed, handed to the options, and then dropped.

The consequence went further than the missing feature. The integration case asserting that a terminal index error is not retried by a strategy that retries everything was passing without the strategy ever reaching the request, so it was asserting nothing — the exact "test that cannot fail" the design's own mutation table was worried about. With the strategy wired, that case now genuinely drives `retry_everything` through the dispatch path and still returns `replica_index_out_of_bounds` immediately, which is the property `cmd->invoke_handler(ec)` was chosen over `retry_orchestrator::maybe_retry(..., do_not_retry, ec)` to guarantee.

Note that the same field is passed as `{}` where `get_all_replicas` and `get_any_replica` construct their fan-out leaves, so a user strategy appears not to reach those either. That is pre-existing and untouched here.

> `decision.partition` is only written back into `cmd->request.partition` on the success path [...] which can leak incorrect partition information.

Applied, though not for the reason given: no reader is affected. `key_value_error_context` carries the opaque, status code, CAS, error-map info, retry counters and dispatch addresses, but not the partition, and the only other consumers of `request.partition` are `encode_to()` and log lines that are reached solely after a successful resolution. The assignment was hoisted above both early returns regardless, because "partition always reflects the most recent resolution" is a cheaper invariant to hold than to reason about.

> This test can issue up to 100 KV ops and still end up skipping on fast/local environments, which may add noticeable runtime variability [...] consider making the timeout-failure deterministic (e.g., via an injected delay/failpoint [...]).

Declined, because the deterministic alternatives it suggests do not exist here. The repo has no failpoint or delay-injection facility for KV, and the smallest deadline the options accept is one millisecond, which a warm local cluster can beat — that is precisely why a single-attempt assertion was flaky and was replaced. The loop exits on the first deadline it observes rather than running to its bound, and it skips loudly instead of failing when the environment never misses one. A deterministic version is worth building when there is an injection point to build it on.

## Thread state

All five inline threads are marked resolved: the `num_replicas`, `partition` and `config_rev` ones by the code changes above, the two aggregate-initialization ones and the grammar nit as declined for the reasons above. The rationale is only visible on the PR once this comment is posted.

## Copilot's third round

One of the four is applied, one is answered with a comment rather than machinery, and the aggregate-initialization claim is declined for the third time — this round with numbers.

> The routing path snapshots `config` once (under a lock in `config_snapshot()`), but logs `rev` via a separate `config_rev()` call which likely takes the lock again and can observe a different configuration than the one used to produce `decision`.

Applied. The trace now logs `config->rev_str()`, the revision of the snapshot the decision was actually taken against, rather than re-reading `config_` through `config_rev()` and naming a configuration that may have replaced it. It also drops one lock acquisition per attempt. The neighbouring non-strategy branch still calls `config_rev()` because it holds no snapshot to read a revision from.

> Enabling `resolves_own_route_v<Request>` implicitly requires `Request` to provide a compatible `resolve_route(...)` API and a writable `partition` member [...] Consider adding a compile-time constraint (e.g., a `static_assert` using the detection idiom).

The contract is now stated on the trait: specialize it only for a request providing `resolve_route(const topology::configuration&)` returning a `replica_route_decision`, plus a writable `partition` member, which is the entire interface `map_and_send()` calls through it. The detection idiom itself is declined. The trait has exactly one specialization, and the two failure modes do not both benefit: a missing `resolve_route` makes the detector's own `decltype` ill-formed, so the diagnostic is no better than the existing "no member named `resolve_route`" at the call site, while a present-but-wrong signature is the only case a `static_assert` would improve. That is not worth a detection trait plus the `replica_route_decision` include it would pull into `bucket.hxx`.

> If the project is compiled under a standard where derived aggregates cannot be brace-initialized from a base subobject (e.g., pre-C++20), this will fail to compile. Consider aligning with the established pattern used by other `*_options` types in the repo (e.g., providing an explicit `built` constructor [...]).

Declined, and the premise is checkable in both halves. Aggregate initialization of a class with base classes is C++17 (P0017R1), and `cmake/CompilerOptions.cmake` requests exactly `cxx_std_17`; the line has been compiled by every job in this PR's matrix, MSVC 2022 and 2025 included.

The claim about the established pattern is the reverse of what the repo contains. Sixty-six `*_options.hxx` headers declare `struct built : public common_options<...>::built`; sixty-five initialise it exactly as this one does, from `build_common_options()` through braces, and **none** declares an explicit `built` constructor. Adopting the suggestion would make this the only options type in the SDK that does it differently, for a portability concern that does not apply to the standard the project sets.



[CXXCBC-981]: https://couchbasecloud.atlassian.net/browse/CXXCBC-981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avsej:
## Copilot's fourth round

No new inline comments. Of the three suppressed, one is a genuine defect and is fixed; the other two are repeats of earlier rounds and are declined again for the reasons already given.

> This assertion is currently dead in every configuration: `COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2` is emitted only by the generated `couchbase/build_config.hxx`, which this translation unit does not include, so the guard remains undefined even when the CMake option is ON. Moreover, the guarded code does not include the only header that defines `protostellar::component_supports_v`.

Correct on both counts, and it was checked rather than taken on faith: preprocessing the translation unit shows neither `couchbase/build_config.hxx` nor `core/protostellar/component.hxx` in its include graph, and the macro is not a command-line definition anywhere in the build — `cmake/build_config.hxx.in` is its only source. The case was therefore compiled out everywhere, including the `tests` workflow, which builds the unit binaries with `CB_COUCHBASE2: 1` and is exactly where it was supposed to fire.

The remedy goes further than the suggestion. `core/protostellar/component.hxx` contains no preprocessor guards and pulls in no generated gRPC or protobuf headers — only core operation types and asio — so it compiles in any configuration, and the trait it declares is a compile-time list that does not depend on the build option at all. Wiring the case into the couchbase2-enabled build would have left it dead in every other job for no reason. The guard and the `build_config.hxx` dependency are both gone instead, and the assertion now compiles unconditionally. The count moving from 6 test cases and 117 assertions to 7 and 118 in a build with the option OFF is the evidence that it previously did not compile at all.

> Use "less than or equal to" for the grammatical comparison.

Declined, as in the first round. That sentence appears verbatim in eight other guards in the same file; correcting one of nine makes the file inconsistent, and correcting all nine is unrelated churn in a feature change.

> The timeout-classification test passes by skipping when the implementation never times out, so a regression that ignores `get_replica_options::timeout` cannot fail this case.

The criticism is accurate and is worth stating plainly: on a cluster fast enough to answer inside the deadline, that case cannot catch a regression. It is retained because the alternative offered does not exist here — the repo has no failpoint, clock injection or delay facility for KV, and one millisecond is the smallest deadline the options accept. The case fails loudly rather than silently when it does observe a deadline, and skips loudly when it does not.

What it is not carrying is the weight of the retry-versus-terminal contract. That is pinned by unit cases that need no cluster, and by the case asserting that a terminal index error never consults the retry strategy — a strategy that retries everything reports 0 consultations, against roughly 200 when the terminal error is routed through the orchestrator instead. A deterministic deadline test is worth building when there is an injection point to build it on.

## Copilot's fifth round

The inline comment identifies a liveness defect introduced by the previous round's bounds check, and it is correct. One suppressed comment is also correct and is applied. The third is the timeout case for the third time, declined again with the criticism acknowledged.

> If the node list shrinks after routing but before this lock is acquired, returning here does not actually cause re-resolution. [...] The request can therefore remain parked until an unrelated future bootstrap or its deadline. Have `connect_session()` report a stale index so the caller immediately retries/re-resolves.

Confirmed by reading the drain path rather than by inspection of the branch alone: every `drain_deferred_queue({})` call site in `bucket.cxx` sits inside a session-bootstrap completion handler, and `update_config()` does not drain. So a bounds check that returns without starting a session leaves the command parked with nothing scheduled to wake it, and the operation reports `unambiguous_timeout` where it should have retried against the current topology. The previous round traded an out-of-bounds read for a stall.

`connect_session()` now returns `bool` and `map_and_send()` retries on `false` through `retry_orchestrator::maybe_retry(..., node_not_available, ...)`, which re-resolves the route on the next lap. The contract is stated where the function is defined: `true` means a parked command will be woken, either because a bootstrap is pending or because the bucket is closed and `defer_command()` completes the command with `request_canceled`; `false` means nothing will wake it.

Writing the tests before the fix showed the stale index was not the only way in. A node present in the topology but advertising no key-value port, and a bucket with no configuration at all, both started nothing while reporting success — the same stall by two other routes, neither raised in review. Both now report `false`. That third case also gave the dispatch path a deterministic test, which the race by itself does not admit: routing to a portless node reaches the retry branch with no interleaving required, so `test_unit_bucket_routing.cxx` drives a real `get` through `bucket::execute()` with a fail-fast strategy and requires `request_canceled`. Reverting that branch turns the result into `unambiguous_timeout` — the stall as a caller would see it. Seven further assertions cover each `connect_session()` outcome, including an index valid under an earlier configuration and stale after a shrink; reverting the bounds check fails three of them, and reporting failure for a session that did start fails another.

> The cause chain does not make existing handlers that compare `err.ec()` with `document_not_found` keep working: the top-level code is now 135, and callers must explicitly inspect `cause()`. Reword this to describe that opt-in handling rather than promising compatibility the API does not provide.

Correct, and it was a false promise in a public header. `ec()` reports `document_not_found_on_replica`, so an existing comparison against `document_not_found` stops matching; the cause chain makes the general condition available, not automatic. Reworded in `couchbase/error_codes.hxx` and in the `make_get_replica_error()` contract in `core/impl/replica_utils.hxx`, and the same overreach was corrected in the commit message and the pull request description, which had repeated it.

> This skip makes the new timeout classification nondeterministic: whenever all 100 local reads beat 1 ms, the section provides no assertion and cannot catch a regression from `unambiguous_timeout` to another code.

Declined for the third time, and the criticism is accurate: on a cluster that answers inside a millisecond that section asserts nothing. What is missing is the mechanism it asks for — the repo has no failpoint, clock or delay injection for key-value operations, and one millisecond is the smallest deadline the options accept, which is why a single-attempt assertion was replaced with a bounded loop that skips loudly.

The suggestion to reach for "a unit-level command/timer" is what the new unit coverage does, for the property that matters here. The retry-versus-terminal split is pinned by the resolution table and by the case asserting a terminal index error never consults the retry strategy, and the deferred-queue behaviour is now pinned by a deterministic unit case that observes a deadline exactly where the classification would be wrong. A deterministic integration deadline test remains worth building when there is an injection point to build it on.

avsej:
## Correction to the fourth-round response

Two claims in the previous comment do not hold, and the conclusion drawn from them was wrong.

`core/protostellar/component.hxx` was described as pulling in no generated gRPC or protobuf headers, and the couchbase2 assertion was therefore made unconditional. The header includes `core/protostellar/dispatcher.hxx`, which includes `<grpcpp/grpcpp.h>`, so it is only includable where the stubs are built. The check behind that claim inspected the header's own include list for gRPC and protobuf names and missed the transitive path, and it compiled locally only because that build carries the gRPC headers on its include path. The macOS leg caught what the reasoning did not:

```
core/protostellar/component.hxx:79:
core/protostellar/dispatcher.hxx:25:10: fatal error: 'grpcpp/grpcpp.h' file not found
```

The remedy is the one this review asked for in the first place. The `#if defined(COUCHBASE_CXX_CLIENT_BUILD_COUCHBASE2)` guard is restored around both the include and the test case, with `<couchbase/build_config.hxx>` included so the guard sees the macro — the omission the original comment identified. The assertion is compiled where the `tests` workflow builds with `CB_COUCHBASE2: 1`, and absent elsewhere: 381 unit tests without couchbase2, 382 with it.

What the earlier evidence does and does not show: the count moving from 6 test cases to 7 confirms the assertion had never been compiled in any configuration before this change. It says nothing about whether the guard should exist, and it was wrong to read it that way.

core/bucket.cxx
```diff
@@ -266,15 +266,26 @@ class bucket_impl
...
+    if (!routed.session || !routed.session->has_config()) {
+      if (!routed.connect_pending) {
+        // Nothing was started for this node, and the deferred queue is only
+        // drained by a bootstrap completion, so parking would hold the request
+        // until its deadline. Retry, which re-routes against the current
+        // configuration.
+        if (backoff_and_retry(req, retry_reason::node_not_available)) {
+          return {};
+        }
+        return errc::common::service_not_available;
```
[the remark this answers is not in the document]
avsej: Both findings are the same defect in two places, both are real, and both are fixed.

> This returned error is only observed on the initial synchronous `direct_dispatch()` call. The deferred re-entry callbacks above call `direct_dispatch()` and discard its result, so if the topology changes while a command is parked and this new no-connect path is reached with a strategy that declines the retry, the request is never completed (or only completes later via its deadline). Update those deferred callbacks to pass a nonzero re-dispatch result to `req->try_callback()` while preserving the initial call's synchronous error contract.
>
> This issue also appears on line 319 of the same file.

Correct, and the reasoning generalises further than the new branch: every deferred re-entry discards the result, so the pre-existing `session->is_stopped()` branch had the same hole for a request that was parked rather than dispatched on its first attempt. Both deferred callbacks in `direct_dispatch()` and the one in `direct_re_queue()` now complete the request when the re-dispatch reports an error, and the synchronous contract is untouched: the initial call still returns the error and its caller still owns completing the request.

> When the retry strategy declines this retry, this branch returns an error without invoking `handle_error`. Unlike the initial dispatch API, `direct_re_queue()` is called from the backoff timer with its return value ignored, so a strategy that retries once and then stops leaves the request callback uncompleted (indefinitely when no deadline is installed). Complete the request here just as the stopped-session branch below does.

Correct. `handle_error()` is called there now, which is what the stopped-session branch two blocks down already did — the asymmetry between the two was the defect.

Neither path is covered by a new test. Reaching them needs a re-dispatch that fails *after* a deferred command has been drained, and the deferred queue is only drained by a bootstrap completion, which the offline `bucket_impl` fixture in `test/test_unit_bucket_routing.cxx` cannot produce: `close()` drains with `request_canceled`, which the callbacks answer on their first branch. The change is therefore verified by inspection plus the existing suites (`read_replica` 60 assertions, `crud` 499, `durability` 38, `range_scan` 6675, `collections` 24, and the unit suite, all against a six-node 8.0.1 cluster under AddressSanitizer).

Worth recording that this family of defect was independently confirmed elsewhere in the same file while this review was open. A separate investigation into an AddressSanitizer leak in the range-scan suite traced it to `backoff_and_retry()` on `main`: asio *abandons* a pending timer wait when the io_context is destroyed rather than completing it with `operation_aborted`, so the handler holding the only reference to a request is destroyed without running, and the request is neither completed nor released. That is tracked as CXXCBC-985, and a third instance — `resolve_response()` ignoring the result of `backoff_and_retry()` on its `request_canceled` branch — as CXXCBC-987. The shared lesson is the one this review found: a request that reaches a path where nothing will retry it must be completed there, because the caller upstream may be a timer or a drained queue that cannot see a return value.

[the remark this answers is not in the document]
avsej: A scope correction to the reply above.

Only one of the two findings is fixed in this pull request: the `handle_error()` call in the branch this change introduces. That branch is new here, and merging it while it can strand a request is not an option.

The three deferred re-entry callbacks that discard the re-dispatch result are pre-existing on `main` and affect every key-value operation rather than the replica read, so fixing them here would put an unrelated dispatch fix inside a feature commit. They have been reverted from this branch and moved to CXXCBC-987 (#1084), which now covers both that site and `resolve_response()`'s cancelled-request branch. The extra diff on this pull request is three lines.

The finding itself stands exactly as reported, including the observation that this change adds one more branch from which the pre-existing discard is reachable. That is an argument for merging #1084 first, and it is a small change.

avsej:
The suppressed note on `connect_pending` was correct, and is fixed in [`a5b05d7`](https://github.com/couchbase/couchbase-cxx-client/commit/a5b05d72bf342e42022045ae907b5433a55b4e7e).

The defect was the field's default contradicting its own documented invariant:

```c++
    /**
     * False when no session exists and none was started, so nothing will drain
     * the deferred queue for a request parked behind it. …
     */
    bool connect_pending{ true };   // ← `return {};` in route_request() inherits this
```

`direct_dispatch()` guards the no-configuration case earlier via `if (!configured_)`, so reaching that `return {}` means a configuration *is* present and simply names no server for the vbucket: nothing was started, and no bootstrap completion is coming to drain the queue.

The default is now `false` — what a default-constructed `routed_session` should mean — and the session-found case says `true` explicitly, since a session that exists but has not finished bootstrapping does drain the queue when it completes. Fixing the default rather than the single `return {}` site is deliberate: `connect_pending` is consulted on every `!session || !session->has_config()` path, so a truthful default keeps a later `return {}` from reintroducing this.

**No regression test, and the reason is worth stating.** Three scenarios were tried and none discriminates between fixed and unfixed:

| scenario | result |
|---|---|
| unassigned replica via `get_replica` | unreachable — `resolve_replica_index()` rejects it upstream with `replica_index_currently_unavailable (137)`, so routing is never asked |
| unassigned **active** + fail-fast strategy | `request_canceled (2)` with **and** without the fix |
| same, asserting `retry_attempts` | `0` with **and** without the fix |

The example in the note — an unassigned replica — is the one case this API already guards upstream; the reachable path is a plain key/value op whose active is unassigned. And in a unit fixture the io_context drains at teardown, cancelling a parked request as promptly as the retry path completes one, so the difference the fix makes is only observable in a process that outlives the request's deadline. A test that passes either way is worse than none, so none was added.

`test_unit_bucket_routing` passes in full (11 assertions, 3 cases), `core/bucket.cxx` compiles clean under the project's warning set, `clang-format` clean.

avsej:
Follow-up on the `connect_pending` note, with one result added and one earlier claim corrected.

### The draining claim is verified

The note's premise — "configuration pushes do not drain this queue; only bootstrap completion does" — holds exactly, and is sharper than stated. `core/bucket.cxx` has five drain sites:

| site | trigger |
|---|---|
| 542, 648, 762, 1123 | all inside a **session bootstrap completion** callback |
| 896 | `close()`, draining with `request_canceled` |

Site 1123 is lexically inside `update_config()`, but nested in the bootstrap callback of a session that `update_config()` has just opened. So a configuration drains the queue only when it causes a **new session to be established** — never merely by arriving. A configuration that moves a vbucket onto a node the bucket is already connected to starts no session and so drains nothing.

### The reachability claim in my previous comment is not verified, and I withdraw it

That comment said the reachable path is "a plain key/value op whose *active* is unassigned". Measured, that scenario does not behave as parked. With a vbucket map of `{ -1 }`, `num_replicas = 0`, a best-effort strategy and a 300 ms deadline:

| | completed | error | `retry_attempts` |
|---|---|---|---|
| with the fix | yes | `unambiguous_timeout (14)` | **5** |
| without the fix | yes | `unambiguous_timeout (14)` | **5** |

Five retries without the fix means the command was not sitting in the deferred queue. So either that scenario does not reach the `connect_pending` branch, or something else retries a parked command. Six attempts to construct a behavioural test — three reported earlier, plus fail-fast with a distant deadline and a bounded run, and the two above — all failed to distinguish fixed from unfixed.

### What the change rests on

The narrower ground only: `connect_pending`'s default contradicted the invariant its own documentation states, and `route_request()`'s `return {}` inherited it. That is worth correcting on its face. Whether it is reachable in practice — and by which operation — is **not** established here, and the claim that it is should not be read into this PR.

`test_unit_bucket_routing` passes in full (11 assertions, 3 cases); `core/bucket.cxx` compiles clean under the project's warning set.

avsej:
Rebased onto `main` at `[`ef6b12ae`](https://github.com/couchbase/couchbase-cxx-client/commit/ef6b12aede2f34b8e2310ca6459c7a3f45445783)`, then a round of review fixes. Head is now `17819412`.

**Rebase.** Nine commits had landed since the previous base. Only two of the 24 files overlapped anything merged: `test/CMakeLists.txt` (two added `unit_test` lines) and `core/bucket.cxx`, which collided with [CXXCBC-985](https://jira.issues.couchbase.com/browse/CXXCBC-985) (#1081). That one was checked by hand rather than taken on trust, because #1081 rewrote `backoff_and_retry()` -- the function this change calls. Its contract is now `false` = nothing owns the request and the caller must complete it, `true` = already completed or owned by the retry queue/timer. Both call sites here honour it: `direct_dispatch()` returns the error for its caller to complete, and the re-queue path completes it itself because the backoff timer ignores the return value.

**Review fixes since the last Copilot pass.** One is worth singling out because it affected anyone running the tests:

`test_unit_bucket_routing.cxx` was doing live TCP I/O. `enable_lazy_connections` defaults to false, so `update_config()` opened a session per node at `127.0.0.1:11210` and the fixture destructor's `ctx_.run()` executed those bootstraps for real -- measured at **10.05 s** for one case against **0.05 s** for its sibling, and on a machine where a KV service does answer on that port with `Administrator/password` and a `default` bucket, `run()` would not return at all. The fixture comment claimed "a bucket with no sessions and nothing in flight", which was false for every section that installed a key-value port. Fixed by making the fixture's connections lazy, which is also `connect_session()`'s production precondition: 10120 ms -> 109 ms for the pair, and the whole unit+cng suite 46.75 s -> 35.70 s.

The rest are prose that disagreed with the code, all of which would have misled a future reader:

- `docs/couchbase2.hxx` omitted `get_replica` from the refused list, while the same page says anything unlisted is sent to the gateway. It is refused client-side for every gateway version.
- `core/impl/get_replica.hxx` documented a "durability" caller of `impl::get_replica_request`; no such caller exists (the observe path builds `observe_seqno_request`). The trait rationale was also inverted -- it opts this request type *into* self-routing rather than exempting the fan-out callers, which do enter resolution and rely on the unset-strategy branch.
- The `static_assert` in `core/protostellar/component.cxx` justified itself with "only the library targets are given those include paths"; nine TUs under `test/cng/` already include that header. The true reason is coverage: `test/cng` is absent from the builds that ship the library, and the Catch2 tree cannot compile the header.
- "Every `drain_deferred_queue()` call site is a bootstrap completion" -- `close()` also drains, with `request_canceled`. The contract itself was right; the justification was not.
- A section titled `document_irretrievable is never reported` could not fail: it looped a code list that excludes `document_irretrievable` and asserted only inequality, while two earlier sections already pin the exact output for all seven inputs. Deleted. The prohibition holds structurally -- that code is produced only by the fan-out headers, which this path does not enter.
- Three smaller comment corrections, and `<vector>` restored to `test_unit_topology_configuration.cxx` where `std::vector` is still named.

The commit message and this description were corrected for the same claims: the durability caller, the trait rationale, the include-path reason, the drain claim, `observe_poll` as a leaf, "three existing callers" -> two, and the "exhaustive check".

**Error-code wording tightened.** A review pass argued that bounding the replica index on the vbucket-map row misclassifies a short row, on the grounds that the two codes split on configured-count versus placement. RFC-0053 revision 22 settles it the other way: the bound is "the current replica chain for that vbucket", the `wrap` divisor is "the length of the replica array", and `replica_index_currently_unavailable` is specifically the `-1` entry. `resolve_replica_index()` already implements exactly that, and the existing unit pins assert it. What was loose was the documentation, which described the bound vaguely enough to read as a claim about the bucket's configuration, so `couchbase/error_codes.hxx`, `couchbase/collection.hxx` and the computation in `core/impl/replica_utils.cxx` now state the RFC rule directly. `min(chain.size() - 1, num_replicas)` is retained and is stricter than the RFC only when a row is wider than the configured count, where it fails safe by refusing a copy the bucket does not advertise.

`wrap()`'s own documentation was also wrong in a way worth calling out: it promised resolution away from a replica with "no live copy". Resolution follows the vbucket map, not node health -- an assigned but unreachable node is selected and retried until the timeout rather than skipped -- and the option's documentation now says so.

**Verified locally at this head:** build clean with `COUCHBASE2=ON` (886 targets), `ctest --label-regex 'unit|cng'` 433/433. The `COUCHBASE2=ON` configuration matters: `core/protostellar/component.cxx` is in the diff and does not compile otherwise, so its `static_assert` is only exercised there.

Not verified here: no integration run against a live cluster at this head, and no sanitizer run -- CI covers both.



[CXXCBC-985]: https://couchbasecloud.atlassian.net/browse/CXXCBC-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avsej:
Both findings are correct and both are fixed. The bound the code applies is unchanged; the defect was in the public documentation, which described a narrower condition than `resolve_replica_index()` implements.

> This description omits the configured replica count. `resolve_replica_index()` returns this code when the requested position is present in the row but is at or beyond `num_replicas`, because the effective count is `min(row replicas, num_replicas)`. Describe that effective bound here and in both `collection::get_replica` overloads so callers do not interpret this code as proving the row itself was too short.
>
> This issue also appears on line 672 of the same file.

Correct. The effective bound is `min(chain.size() - 1, *config.num_replicas)`, and an assertion in this pull request pins the case: the section "the replica count caps a chain longer than num_replicas" resolves a row of `{ 0, 1, 2 }` with `num_replicas = 1` and reports `replica_index_out_of_bounds` for the second replica, although position 2 of that row holds server 2. The header contradicted the suite shipped beside it.

`errc::key_value::replica_index_out_of_bounds` and both `collection::get_replica` overloads now state the effective bound: the index is at or beyond the smaller of the replica chain the vbucket map lists and the bucket's configured replica count.

> This code is also returned when the vbucket row assigns the replica to a server index that is absent from `config.nodes`; in that case the map does contain an assignment. Document both unavailable states (unassigned slot and topology-missing node), and align the corresponding `collection::get_replica` documentation.

Correct, and pinned in the same file: the section "a replica on a node the topology does not list is unavailable" resolves `{ 0, 9 }` with `num_replicas = 1` and reports `replica_index_currently_unavailable`, where the row does carry an assignment. The phrase "no copy assigned to a node" was false for that state.

The two states are named as one condition rather than enumerated, because the difference is not one a caller can act on: an unassigned slot and an assignment to a node outside the topology both mean no replica is readable at that index, and `wrap()` is the remedy for both. The description now reads that the vbucket map does not place the replica at the requested index on a node the topology lists. The `wrap()` documentation in `get_replica_strategy.hxx`, which described the same two conditions, is aligned.

The commit body and the pull request description already stated both rules, so the drift was confined to the public headers. The change is comment-only.

core/impl/get_replica.hxx
DemetrisChr, line 44: Now that we expect wrappers to also be using this, would it make sense for this to be moved to the `core::operations` namespace?

core/impl/get_replica.hxx
```diff
@@ -50,12 +56,34 @@ struct get_replica_request {
   std::uint32_t opaque{};
   io::retry_context<true> retries{};
   std::shared_ptr<couchbase::tracing::request_span> parent_span{ nullptr };
+  /**
+   * Absent for the replica fan-out callers, which pin the replica through
+   * @c id.node_index() and expect plain vbucket-map routing. resolve_route()
+   * returns that pinned node unchanged when this is unset.
+   */
+  std::optional<get_replica_strategy::built> strategy{};
```
DemetrisChr, line 64: I think it would be better if we didn't use a Public API data structure here.

couchbase/get_replica_strategy.hxx
```diff
@@ -0,0 +1,179 @@
...
+   * a from-index strategy, so code written against it is not portable to the
+   * other Couchbase SDKs, and the default leaves the RFC behaviour in place.
+   *
+   * @param value whether to re-resolve the index on retry
+   * @return this options builder for chaining purposes.
+   *
+   * @since 1.5.0
+   * @uncommitted
+   */
+  auto revalidate_on_retry(bool value) -> get_replica_strategy_from_index_options&
```
DemetrisChr, line 95: The RFC does not have a `revalidate_on_retry` option for `get_replica_strategy`.

Maybe we should bring this up for discussion with the team, if we want this option, but I wouldn't add it to the SDK unless the RFC defines it.

Are there cases where a user would want to disable this?

core/impl/get_replica.cxx
```diff
@@ -25,6 +25,29 @@
...
 {
+auto
+get_replica_request::resolve_route(const topology::configuration& config) -> replica_route_decision
+{
+  // Named vbucket rather than partition: the request carries a partition member,
+  // and MSVC rejects a local that hides it.
+  if (!strategy.has_value()) {
+    auto [vbucket, server] = config.map_key(id.key(), id.node_index());
+    return { {}, vbucket, id.node_index(), server };
+  }
```
DemetrisChr, line 36: This is duplicating the logic we have in bucket.cxx. Maybe the resolution should be skipped here in the case where strategy is not set, so the existing routing is used as a fallback, rather than duplicating here (this would apply to other requests with `resolves_own_route` support in the future as well).

core/protostellar/component.cxx
```diff
@@ -63,6 +64,15 @@
 
 namespace couchbase::core::protostellar
 {
+// The gateway carries no replica read, so cluster::execute() must answer
+// get_replica with feature_not_available instead of dispatching it. Asserted
+// here rather than in the Catch2 tree, which cannot compile this header, and
+// not in test/cng, which is absent from exactly the builds that ship the
+// library. Flips when CNG implements the RPC, and the rejection then becomes an
+// integration case.
+static_assert(!component_supports_v<impl::get_replica_request>,
```
DemetrisChr, line 73: Is this necessary? The operations that are supported are already enumerated in component.hxx

couchbase/get_replica_result.hxx
```diff
@@ -32,8 +32,8 @@ class manager;
 #endif
 
 /**
- * Represents result of @ref collection#get_any_replica operations, also returned by @ref
- * collection#get_all_replicas.
+ * Represents result of @ref collection#get_any_replica and @ref collection#get_replica operations,
+ * also returned by @ref collection#get_all_replicas.
```
DemetrisChr, line 36: Wording looks slightly odd now. I'd write:

```suggestion
 * Represents result of @ref collection#get_any_replica, @ref collection#get_all_replicas and @ref collection#get_replica operations
```

core/bucket.cxx
```diff
@@ -313,8 +324,19 @@ class bucket_impl
...
+        // As in direct_dispatch(): no bootstrap is pending for this node, so a
+        // parked request would wait for its deadline.
+        if (backoff_and_retry(req, retry_reason::node_not_available)) {
+          return {};
+        }
+        // A re-queue is driven by a backoff timer that ignores what this
+        // returns, so the request has to be completed here.
+        handle_error(errc::common::service_not_available);
+        return errc::common::service_not_available;
+      }
```
DemetrisChr, line 339: This change looks like a bug fix that's not directly related to the get_replica work. Could this warrant its own ticket/PR?

core/impl/replica_utils.cxx
```diff
@@ -18,10 +18,91 @@
...
+    // the row as the count would route to a copy the bucket may not advertise,
+    // so this is treated the same as a map that has not arrived.
+    return decision;
+  }
+
+  const auto& chain = config.vbmap->at(vbucket);
+  // RFC-0053 bounds the index on the replica chain the map lists, and wraps
+  // modulo that same length, so both come from the row rather than from
+  // num_replicas. The configured count only caps it further, keeping a request
+  // off a copy the bucket does not advertise when a row is wider than the
```
DemetrisChr, line 71: > RFC-0053 bounds the index on the replica chain the map lists, and wraps modulo that same length

Would it not be better if we stick to the RFC in this case then?

> keeping a request off a copy the bucket does not advertise when a row is wider than the setting

I don't think it's the case that the bucket does not advertise the copy, when it is on the vbucket map, I think that would point to a bug in the config. Is it actually possible for the vbucket map to advertise more replicas that the `num_replicas` field suggests?

core/impl/collection.cxx
```diff
@@ -331,6 +332,44 @@ class collection_impl : public std::enable_shared_from_this<collection_impl>
...
+        std::move(document_key),
+      },
+      options.timeout,
+      {},
+      {},
+      { options.retry_strategy },
+      obs_rec->operation_span(),
+      strategy,
+    };
+    return core_.execute(std::move(request),
```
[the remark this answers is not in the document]
avsej: Declined: the premise does not hold.

There is no rule in this repository requiring a new key-value operation to go through `crud_component`. Both operations this one shares its code path with dispatch the same way this one does:

- `collection::get_any_replica()` -> `core_.execute(get_any_replica_request)`
- `collection::get_all_replicas()` -> `core_.execute(get_all_replicas_request)`

Moving only `get_replica` to the `queue_request` path would also break what the design depends on: all three share `get_replica_request`, and the self-routing trait exists so the fan-out callers can decline to route themselves and take ordinary vbucket-map routing. Splitting them across two dispatch stacks is the outcome this avoids, not the one it causes.

Moving all three is a reasonable change, and out of scope here.

core/impl/get_replica.hxx
```diff
@@ -50,12 +56,36 @@ struct get_replica_request {
   std::uint32_t opaque{};
   io::retry_context<true> retries{};
   std::shared_ptr<couchbase::tracing::request_span> parent_span{ nullptr };
+  /**
+   * Absent for the replica fan-out callers, which pin the replica through
+   * @c id.node_index() and expect plain vbucket-map routing. resolve_route()
+   * returns that pinned node unchanged when this is unset.
```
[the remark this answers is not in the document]
avsej: Fixed. The comment described the contract before `resolve_route()` began returning an optional:

```
   * Absent for the replica fan-out callers, which pin the replica through
   * @c id.node_index(). When this is unset resolve_route() returns an empty
   * optional, declining to route, and the dispatch path falls back to the
   * vbucket-map routing that honours @c id.node_index().
```

core/operations/operation_traits.hxx
```diff
@@ -34,4 +34,20 @@ struct is_cancellable_operation : public std::false_type {
...
 inline constexpr bool is_cancellable_operation_v = is_cancellable_operation<T>::value;
+
+/**
+ * Operations that pick the node to dispatch to themselves, instead of being
+ * routed by the vbucket map alone.
+ *
+ * Specialize this only for a request that provides
+ * @c resolve_route(const topology::configuration&) returning an
+ * @c impl::replica_route_decision, and a writable @c partition member: that is
+ * the whole interface @c bucket::map_and_send() calls through this trait.
```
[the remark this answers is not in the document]
avsej: Fixed. The trait's contract documented the pre-refactor return type, so a specialization written against it would not have compiled. It now names the optional and what an empty one means:

```
 * Specialize this only for a request that provides
 * @c resolve_route(const topology::configuration&) returning an
 * @c std::optional<impl::replica_route_decision>, and a writable @c partition
 * member: that is the whole interface @c bucket::map_and_send() calls through
 * this trait. An empty optional declines self-routing for that request, and the
 * vbucket-map path routes it instead.
```

test/test_integration_read_replica.cxx
```diff
@@ -904,3 +909,195 @@ TEST_CASE("integration: zone-aware read replicas on unbalanced cluster", "[integ
...
+  {
+    return consultations_;
+  }
+
+private:
+  std::atomic<std::size_t> consultations_{ 0 };
+};
+} // namespace
+
+TEST_CASE("integration: get replica by index", "[integration]")
```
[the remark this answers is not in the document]
avsej: Declined for now, on two grounds.

`test/README.md:5` reads "no new **file** should use it". This adds a case to an existing Catch2 file rather than a new file, and the eight other environment guards in that file use the same body-level `SKIP`.

Registration-time requirements are a framework mechanism; Catch2 has no equivalent, so that part cannot be satisfied without moving the file. Integration coverage that needs a live cluster stays in Catch2 until stage 3 of the CXXCBC-945 migration converts these suites, which is also what #1124 does.

test/unit/core/bucket_routing.cxx
```diff
@@ -0,0 +1,357 @@
...
+private:
+  asio::io_context ctx_{};
+  couchbase::core::tls_context_provider tls_{};
+  std::shared_ptr<couchbase::core::bucket> bucket_{};
+};
+
+// map_and_send() defers a command after asking for a session, and only a bootstrap completion
+// bootstrap completion drains that queue. A true return from connect_session() therefore promises
+// that a deferred command will be dispatched or completed -- by a pending bootstrap, or by the
+// bucket applies.
```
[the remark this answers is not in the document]
avsej: Fixed in #1124, which owns that file. A multi-line rename had duplicated "bootstrap completion" and truncated the closing clause; the comment now states the `connect_session()` return contract in full.

core/impl/get_replica.hxx
avsej: Done. The request, its response and its handler alias are now in `core/operations/document_get_replica.hxx`, in `core::operations`, beside `get_all_replicas` and `get_any_replica`. The forward declarations moved with them, which is what `get_replica_request_with_cancellation` resolves through.

It is a separate commit, since the type predates this PR and the fan-out operations share it.

core/impl/get_replica.hxx
avsej: Done. The request now holds `impl::replica_selection`, a core-local `{ replica_index, wrap }` declared beside the routing decision it feeds, and the core header no longer includes `<couchbase/get_replica_strategy.hxx>`. `collection.cxx` translates the public options where every other operation already translates its own.

One correction for the record: public types in core requests are not without precedent — `get_all_replicas_request` holds `couchbase::read_preference`. The narrower reason this one was worth changing is that no `core::operations` request holds a public `::built` type, and after `revalidate_on_retry` was removed the struct was a single `bool`.

couchbase/get_replica_strategy.hxx
avsej: Removed, and the RFC is stronger than "does not have". Line 703 requires re-evaluation:

The default was `true`, so the option only let a caller opt out of RFC behaviour. Revalidation is now unconditional.

The only use I could construct for disabling it is pinning to whichever node held position N at first resolution. That is a different feature, and nobody has asked for it.

core/impl/get_replica.cxx
avsej: Done, as suggested. `resolve_route()` returns `std::optional<replica_route_decision>`; an empty optional declines, and `map_and_send()` falls through to its existing vbucket-map branch. The duplicated `map_key()` call is gone.

Confirmed equivalent before the change: `map_id(id)` is `config_->map_key(id.key(), id.node_index())`, which is exactly what the deleted branch did. The fan-out cases are the check — `get_any_replica`, `get_all_replicas` and `lookup_in_all_replicas` all take the fallback now.

The trait stays compile-time and the decision is per-request, so the runtime check is needed either way. That generalises to future `resolves_own_route` requests, as you noted.

core/protostellar/component.cxx
avsej: Removed, along with the include it needed. `component_supports_v` is a primary template defaulting to `false`, with an explicit allowlist. The assertion therefore held by construction. It could only fail if someone added a `true` specialisation, and at that point they are implementing the feature.

`couchbase2://` still reports `feature_not_available`; that comes from the routing, not the assertion.

couchbase/get_replica_result.hxx
avsej: Applied as suggested.

core/bucket.cxx
avsej: Yes. It is now [CXXCBC-1023](https://jira.issues.couchbase.com/browse/CXXCBC-1023) and #1124, which this PR builds on.

Splitting it also surfaced that the fix was under-covered: the cases here only reached `map_and_send()`, while `direct_dispatch()` and `direct_re_queue()` carry the current key-value operations. Mutations on those two branches survived. #1124 adds cases for both, plus an integration case for the cluster-level path, and the mutations now fail.

core/impl/replica_utils.cxx
avsej: Yes, it is possible, and it is the ordinary state after a replica-count reduction. Measured on Server 8.0.1 with a 3-node cluster and a 1024-vbucket bucket:

| State | `replicaNumber` | `vBucketServerMap.numReplicas` | row widths | rows wider than `numReplicas` |
| --- | --- | --- | --- | --- |
| before | 2 | 2 | 3 | 0 |
| after `replicaNumber=1` | 1 | **1** | **3** | **1024 of 1024** |

The server publishes the lower `numReplicas` immediately and keeps the wider rows until a rebalance. The state persists — re-probed minutes later, unchanged.

Take vbucket 0, row `[0, 1, 2]`, `numReplicas` 1. Under the RFC's row-only bound, `get_replica(second)` routes to node index 2 — a copy the bucket is no longer configured to keep, and which the next rebalance deletes. The `min()` cap reports `replica_index_out_of_bounds` instead.

So the cap stays. The comment previously asserted this case without evidence and now names the reconfiguration window, and `the_replica_count_caps_a_chain_longer_than_num_replicas` pins it with row `{0,1,2}` and `num_replicas` 1.

test/test_integration_bucket_bootstrap.cxx
```diff
@@ -118,3 +124,64 @@ TEST_CASE("integration: closing a cluster while a valid bucket is still bootstra
...
+{
+  test::utils::integration_test_guard integration;
+  if (!integration.cluster_version().supports_collections()) {
+    SKIP("the server does not support collections");
+  }
+
+  asio::io_context io{ 1 };
+  auto io_thread = std::thread([&io]() {
+    io.run();
+  });
```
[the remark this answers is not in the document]
avsej: Fixed. `asio::make_work_guard(io)` is now created before the thread starts and released in teardown, so `run()` cannot return before the cluster has queued anything.

The race was real: the context has no work when the thread starts, so `run()` may return at once and `open_cluster()` then waits for a completion no thread is left to run. It passed only because starting a thread is slower than constructing the cluster.

One part does not hold: there is no second occurrence. The file has one `io_context`, at line 152; lines 185-186 are that same one's teardown, and line 180 has no `io_context`.

test/test_integration_bucket_bootstrap.cxx
```diff
@@ -118,3 +126,69 @@ TEST_CASE("integration: closing a cluster while a valid bucket is still bootstra
...
+  }
+
+  asio::io_context io{ 1 };
+  // The guard is what keeps run() from returning before the cluster below has
+  // queued any work: an io_context with nothing to do finishes immediately, and
+  // open_cluster() would then wait for a completion no thread is left to run.
+  auto work = asio::make_work_guard(io);
+  auto io_thread = std::thread([&io]() {
+    io.run();
+  });
```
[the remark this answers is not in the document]
avsej: Fixed properly. Reordering the assertions covered only the `REQUIRE`s, and you are right that `open_cluster()` throwing has the same effect.

The work guard and the thread are now owned by an `io_runner` whose destructor releases the guard, stops the context, and joins. Every exit path joins, including an exception from `open_cluster()`. The guard is declared before the thread so it is constructed first.

test/test_integration_bucket_bootstrap.cxx
```diff
@@ -118,3 +126,69 @@ TEST_CASE("integration: closing a cluster while a valid bucket is still bootstra
...
+// continuation. Nothing reads the error that second dispatch returns, because
+// open_bucket()'s caller has already returned. The continuation therefore has to
+// complete the request itself; without that, a request the second dispatch refuses is
+// never completed and fails at its deadline.
+//
+// Reaching that continuation takes two things.
+//
+// The bucket must not be open when the dispatch starts, or the bucket is found directly
+// and open_bucket() is never called. So the test connects a cluster of its own: the
+// guard's already has the bucket open.
```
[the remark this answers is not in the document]
avsej: Fixed: "the guard's cluster already has it open."

test/test_integration_read_replica.cxx
```diff
@@ -904,3 +909,195 @@ TEST_CASE("integration: zone-aware read replicas on unbalanced cluster", "[integ
...
+      SKIP("bucket has fewer than two replicas");
+    }
+    auto [err, result] =
+      collection
+        .get_replica(key,
+                     couchbase::get_replica_strategy::from_index(couchbase::replica_index::second))
+        .get();
+    REQUIRE_SUCCESS(err.ec());
+    REQUIRE(result.is_replica());
+    REQUIRE(result.content_as<smuggling_transcoder>().first == basic_doc_json);
```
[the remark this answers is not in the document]
avsej: Fixed. The section now reads the first replica as well and requires the two results to name different serving nodes.

The finding holds as stated: `is_replica()` is a literal `true` at the point the result is built, and every copy carries the same payload, so nothing in the section could separate the requested index from any other.

Negative control, against a dispatch that drops the resolved index (`self_routed = std::nullopt`) so both reads fall through to the vbucket-map routing: the case fails, and all 34 `unit_core_get_replica` cases still pass. That is the gap the note names — the unit suite covers `resolve_replica_index` and nothing below it. A regression in the resolver itself is caught there (five cases fail), so the claim is precise about the layer that was uncovered rather than about the unit tests generally.

The mutation is recorded as a negative control so the new assertion cannot quietly stop testing anything.

core/bucket.hxx
```diff
@@ -142,21 +142,65 @@ public:
...
-                     partition,
-                     config_rev());
-        return io::retry_orchestrator::maybe_retry(
-          cmd->manager_, cmd, retry_reason::node_not_available, errc::common::request_canceled);
+      // A request that resolves its own route may still decline to, in which
+      // case it takes the vbucket-map routing below like every other request.
+      std::optional<std::size_t> self_routed;
+      if constexpr (operations::resolves_own_route_v<Request>) {
+        auto config = config_snapshot();
+        if (!config) {
```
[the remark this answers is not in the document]
avsej: The window is real and the line references are right: `config_` is published under `config_mutex_`, that scope closes, and `sessions_` is rebuilt afterwards under `sessions_mutex_`. A read that resolves against the new configuration and then fetches a numeric slot from the not-yet-rebuilt session map can reach a different node, and a `GET_REPLICA` answered by a node holding another replica of the same vbucket returns a copy that was not the one requested.

It is not introduced by this change. The existing replica fan-out takes the same two-lock sequence: `get_all_replicas` sets `node_index` on the document id (`core/operations/document_get_all_replicas.hxx:130`), routing resolves it with `config_->map_key(key, node_index)` under `config_mutex_` in `map_id(const document_id&)`, and `map_and_send()` then calls `find_session_by_index()` under `sessions_mutex_`. Same window, same outcome, and it predates this branch. What differs here is only how the index is derived, not when the session is looked up.

So the fix belongs to the routing shared by every replica read rather than to this operation, and either remedy suggested — a stable node identity, or publishing the configuration and the session index together — changes behaviour for `get_all_replicas`, `get_any_replica` and `lookup_in_all_replicas` as well. It is being tracked separately rather than folded in here.

core/impl/replica_utils.cxx
DemetrisChr: I think we're now in agreement that this should be addressed in the RFC as well

couchbase/get_replica_result.hxx
```diff
@@ -32,8 +32,8 @@ class manager;
 #endif
 
 /**
- * Represents result of @ref collection#get_any_replica operations, also returned by @ref
- * collection#get_all_replicas.
+ * Represents result of @ref collection#get_any_replica, @ref collection#get_all_replicas and
+ * @ref collection#get_replica operations
```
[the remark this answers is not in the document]
avsej: Fixed. The constructor comment now names get_replica alongside get_any_replica and get_all_replicas; the class-level comment above it already listed all three.

core/bucket.hxx
```diff
@@ -142,21 +142,65 @@ public:
...
+          cmd->request.partition = decision->partition;
+          if (decision->ec) {
+            // Terminal, so the completion is invoked directly: the retry
+            // orchestrator consults the request's retry strategy first, and a
+            // strategy that retries do_not_retry would loop a condition that
+            // cannot resolve itself.
+            return cmd->invoke_handler(decision->ec);
+          }
+          if (!decision->server_index.has_value()) {
+            CB_LOG_TRACE(R"([{}] unable to map key="{}" to the node, id={}, partition={}, rev={})",
```
DemetrisChr, line 172: We should remember to add the redaction annotations here

couchbase/collection.hxx
```diff
@@ -385,6 +387,74 @@ public:
...
+   * @param handler the handler that implements @ref get_replica_handler
+   *
+   * @exception errc::key_value::document_not_found_on_replica
+   *    the replica responded, but does not hold the document. The @ref error::cause() is
+   *    @ref errc::key_value::document_not_found.
+   *
+   * @exception errc::key_value::replica_index_out_of_bounds
+   *    the requested index is at or beyond the smaller of the replica chain the vbucket map lists
+   *    for this document's vbucket and the bucket's configured replica count. Reported without a
+   *    network round trip.
```
DemetrisChr: >    the requested index is at or beyond the smaller of the replica chain the vbucket map lists

I think this isn't always true with the current spec. If `numReplicas` is higher than the size of the replica chain, and the replica index is between the two, that is now `replica_index_currently_unavailable`.

From the [RFC](https://github.com/couchbaselabs/sdk-rfcs/pull/165/changes#diff-a414c79d0ab425593fd245228ae55adbd482e9a6e4fb10b6a3280a29907948ccR711-R712):

> 1. If the user's requested index is >= `numReplicas` (the bucket's configured replica count): raise `ReplicaIndexOutOfBoundsException`, without hitting the network.
> 2. If the user's requested index is >= the replica chain array count: raise `ReplicaIndexCurrentlyUnavailableException`.

core/impl/replica_utils.cxx
```diff
@@ -18,10 +18,93 @@
...
+  // being reduced the server publishes the lower count at once and keeps the
+  // wider rows until a rebalance, so the row alone would name a copy the
+  // bucket no longer keeps.
+  const auto number_of_replicas =
+    std::min(chain.empty() ? 0 : chain.size() - 1, std::size_t{ *config.num_replicas });
+  if (number_of_replicas == 0) {
+    // This also keeps the wrap modulo below from dividing by zero.
+    decision.ec = errc::key_value::replica_index_out_of_bounds;
+    return decision;
+  }
```
DemetrisChr: Similar to my comment on the doc string. 

If `num_replicas` is higher than the size of the replica chain, and the replica index is between the two, that should fail with `replica_index_currently_unavailable`, not `replica_index_out_of_bounds`.

From the [RFC](https://github.com/couchbaselabs/sdk-rfcs/pull/165/changes#diff-a414c79d0ab425593fd245228ae55adbd482e9a6e4fb10b6a3280a29907948ccR711-R712):

> 1. If the user's requested index is >= numReplicas (the bucket's configured replica count): raise ReplicaIndexOutOfBoundsException, without hitting the network.
> 2. If the user's requested index is >= the replica chain array count: raise ReplicaIndexCurrentlyUnavailableException.

core/bucket.hxx
avsej: sure. that redaction patch still in draft

couchbase/collection.hxx
avsej: Corrected, and the resolver had the same fault, so both changed.

`numReplicas` decides whether the index exists at all. The vbucket-map row decides whether that copy can be read right now. The wrong sentence appeared on both `get_replica()` overloads and on both error codes in `error_codes.hxx`, so all four were rewritten.

core/impl/replica_utils.cxx
avsej: Applied. The resolver was rewritten rather than patched, because a second change comes with this one. `wrap` now takes its modulus from `numReplicas`, not from the capped length. An index landing past a short row therefore still reaches the replicas the row does carry.

The eleven worked examples from the spec are pinned as a unit case, one assertion per row. Two existing cases asserted the old behaviour and were inverted: a row shorter than the configured count, and an empty row.

core/impl/replica_utils.hxx
```diff
@@ -34,6 +40,65 @@ struct readable_node {
...
+  std::optional<std::size_t> server_index{};
+};
+
+/**
+ * Resolves a replica index, as the user requested it, against a vbucket map
+ * row.
+ *
+ * The number of replicas of a vbucket is the smaller of what the row lists and
+ * what the bucket is configured for: neither alone is authoritative, because
+ * the server publishes them independently.
```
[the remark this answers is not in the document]
avsej: Correct, and fixed. That comment still described the rule the spec replaced.

It now reads: `num_replicas` bounds the index, the row bounds what is readable. A rebalance reshapes the rows after a new count is published, so the two can disagree. The `@param wrap` line was stale in the same way and was rewritten with it.

One test name carried the old model too. `the_replica_count_caps_a_chain_longer_than_num_replicas` is now `a_row_longer_than_the_replica_count_does_not_extend_the_index`; nothing caps anything any more. Its assertion was already right.

DemetrisChr (approved):

PR #1079, opened 2026-08-18
CXXCBC-983: fold test::utils into couchbase::test

avsej:
Blocks stage 2 of [CXXCBC-945](https://issues.couchbase.com/browse/CXXCBC-945). Raised in review of #1070, where the collision was spotted as a general problem rather than the single-file one it had been patched as.

## The problem

The framework lives in `couchbase::test`. The Catch2 helpers live in `test::utils`, at global scope, and are reached from 49 files and 1834 call sites.

Those two cannot both be named from inside a migrated file. Unqualified lookup of `test` finds `couchbase::test` as a member of the enclosing `couchbase`, stops there, finds no `utils` member, and never falls through to `::test`:

```
g++:     error: couchbase::test::utils has not been declared
clang++: error: no member named utils in namespace couchbase::test; did you mean '::test::utils'?
```

So moving a Catch2 suite into the framework's namespace breaks every helper call in it. `cluster_probes.cxx` already carried a `namespace utils = ::test::utils;` alias to work around precisely this — which fixed one file rather than the shape of the problem. The stage-1 pilot happens to call no helper, which is why stage 1 did not surface it.

## What this does

`test::utils` becomes `couchbase::test`. One namespace for the test tree; a helper no longer needs a sub-namespace to justify its existence.

`server_version`, `server_edition` and `deployment_type` existed in both and are **not** duplicates:

| | framework | `test::utils` |
|---|---|---|
| purpose | a comparable version | feature detection |
| members | 6 | plus `build`, `profile`, `use_gocaves` |
| API | `to_string`, `operator<`, `operator>=` | `parse`, `is_neo`, `supports_gcccp`, … |

The helpers' type survives and the framework adopts it, gaining `to_string` and the `(major, minor, micro)` ordering the requirement vocabulary compares against. `build` and `profile` take no part in that ordering: a requirement asks which release introduced a feature, not which build produced it.

This costs no compile time and keeps the property that declaring a requirement pulls in no core or public SDK header — `test/utils/server_version.hxx` includes `<cstdlib>` and `<string>` and nothing else. It is in fact lighter than the framework's own `context.hxx`.

The alias and the two conversion functions in `cluster_probes.cxx` are gone; the version probe returns the guard's value directly, the conversion having become the identity.

## One thing worth reviewing closely

Positional initialisation of the wider type sets the wrong members, because `build` sits between `micro` and `developer_preview`. That surfaced as a compile error rather than a wrong value, which is the outcome to want:

```
error: could not convert '{7, 6, 2, false, server_edition::enterprise, deployment_type::on_prem}'
       to 'couchbase::test::server_version'
```

Both affected sites now name their fields. C++17 has no designated initialisers, so it is plain assignment rather than `.major =` syntax.

## Verification

No `test::utils` remains anywhere under `test/`. Every test translation unit compiles — 121 objects in the default configuration plus the two columnar files, which need `COUCHBASE_CXX_CLIENT_COLUMNAR=ON` and were checked separately. The framework self-tests pass at every commit of the stack.

## Ordering

This sits after #1072, not earlier, and that is a hard dependency rather than a preference: CXXCBC-948 is what introduces `context.hxx` and the three types this deletes. Placing the rename before it would have the framework redefine a `couchbase::test::server_version` that already exists, so that commit would not build.

[CXXCBC-945]: https://couchbasecloud.atlassian.net/browse/CXXCBC-945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avsej:
Closing without merging. The change was the wrong shape.

`test/utils/` is transitional. Every Catch2 file that reaches it is migrated and then deleted, and the helpers go with them, so renaming a namespace across 49 files that are all scheduled for removal is churn against dead code.

The collision is real but narrow, and it is resolved per file at migration time, in the file being rewritten anyway. A migrated case that still needs a legacy helper qualifies it with a leading `::`:

```cpp
namespace couchbase::test {
  void a_migrated_case(context& ctx) {
    ::test::utils::wait_until_ready(...);   // compiles on gcc and clang
  }
}
```

Most call sites will not need even that, because the thing they were reaching for becomes a requirement.

No bulk change, no namespace alias, and no merging of types that are on their way out.

PR #1088, opened 2026-08-19
CXXCBC-991: assert that a mid-bootstrap teardown completes the future

avsej:
[CXXCBC-991](https://couchbasecloud.atlassian.net/browse/CXXCBC-991)

Two cases in `test_integration_handler_drop_lifetime.cxx` assert a specific error code where the code under test has two correct answers, so they fail on a slow runner while the behaviour they protect is intact.

Both park an operation against `this_bucket_does_not_exist`, tear the cluster down while the bucket-config wait is still parked, then require the future's error to be exactly `cluster_closed`. But the bucket really does not exist, so the server's `bucket_not_found` is just as correct — and which one arrives is a race between an HTTP round trip and the teardown backstop. Valgrind slows the process by an order of magnitude without slowing the network, so under memcheck the server usually wins:

```
test_integration_handler_drop_lifetime.cxx:91: FAILED:
  REQUIRE( result.first.ec() == couchbase::errc::network::cluster_closed )
with expansion:
  couchbase.common:10 (bucket_not_found (10)) == couchbase.network:1006
```

Seen on the valgrind shards of #1081 and #1082, neither of which touches the close path, and on #1077's shards before any of that work existed — job [95563920009](https://github.com/couchbase/couchbase-cxx-client/actions/runs/32104566459/job/95563920009) failed the scan case and job [95563920027](https://github.com/couchbase/couchbase-cxx-client/actions/runs/32104566459/job/95563920027) the `node_ids` one.

### Fix

Both cases assert that the operation failed, not how:

```c++
  auto result = pending.get();
  // The scan was torn down mid-bootstrap against a bucket that does not exist, so it must fail --
  // the point is that it completes at all rather than being dropped.
  REQUIRE(result.first.ec());
```

The contract under test is the lifetime guarantee — teardown must complete a parked handler rather than drop it, since a dropped continuation reaches a `std::future` caller as `broken_promise` and hangs a callback caller. The bounded `wait_for` immediately above each assertion is what catches a stranded handler; the error code carries no part of the guarantee.

This is not a new convention. The third case in the same file, covering the key/value path, already reasons exactly this way and asserts only `REQUIRE(result.first.ec())`. These two now match it, so all three entry points assert the same property. The comments that named `cluster_closed` as the required outcome were updated to say what is actually required.

### Verification

Compiled with the project's own warning set (`-Werror -Wall -Wextra -Wconversion -Wold-style-cast …`) and `clang-format` clean. The behavioural change is a weakened assertion, so the shards it unblocks are the verification; the three cases still fail if a handler is stranded, which is what the surrounding `wait_for`/`FAIL` guards.

[CXXCBC-991]: https://couchbasecloud.atlassian.net/browse/CXXCBC-991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

### Rebased onto main (2026-08-19), and the evidence has grown

Rebased past the merges of CXXCBC-984, 987, 988, 989 and 977; no conflict.

Since this PR was opened, the two assertions it rewrites have failed **nine** times across **seven** pull requests — #1077, #1078 (×2), #1081, #1082, #1084 (×2), #1085, #1090 — none of which touches the close path. Every occurrence is `bucket_not_found (10)` arriving where `cluster_closed` was demanded, on a healthy cluster with zero valgrind findings.

**The fix is confirmed working in CI.** On this branch's own valgrind shards, the cases that had been failing now pass:

| shard | `a scan future…` | `a get future…` |
|---|---|---|
| `valgrind, 8.0.0, tls, 1, 2` | Passed 19.50 s | Passed 19.59 s |
| `valgrind, 8.0.0, plain, 1, 2` | Passed 24.63 s | Passed 24.66 s |

`test_integration_handler_drop_lifetime.cxx` appears in no failure context on either shard, where the equivalent shard on #1078 failed it at line 121.

That observational check is the verification available here: the change *weakens* an assertion, so there is no local red-to-green to run — reverting it does not fail a test on demand, because the failure depends on winning a network race.

### DIFF
--- a/test/test_integration_handler_drop_lifetime.cxx
+++ b/test/test_integration_handler_drop_lifetime.cxx
@@ -50,8 +50,10 @@ using namespace std::literals::chrono_literals;
 // a race; this case exercises the shared mechanism through the one deterministic entry point.
 //
 // Unlike a pure leak (which is only visible to the sanitizer), the fix here restores a
-// caller-visible contract, so the test asserts it directly: the future must complete with
-// errc::network::cluster_closed rather than be dropped.
+// caller-visible contract, so the test asserts it directly: the future must complete rather than be
+// dropped. It must not require one specific error. The bucket does not exist, so the server's
+// bucket_not_found and the teardown backstop's cluster_closed are both correct answers, and which
+// one arrives is a race between an HTTP round trip and the teardown; the assertion accepts either.
 TEST_CASE("integration: a scan future completes when the cluster is closed mid-bootstrap",
           "[integration]")
 {
@@ -88,13 +90,18 @@ TEST_CASE("integration: a scan future completes when the cluster is closed mid-b
     FAIL("scan future was not completed after the cluster was closed -- the handler was stranded");
   }
   auto result = pending.get();
-  REQUIRE(result.first.ec() == couchbase::errc::network::cluster_closed);
+  // Either answer is correct: the teardown backstop reports cluster_closed, and the server reports
+  // bucket_not_found for a bucket that does not exist. Which one arrives is a race between an HTTP
+  // round trip and the teardown, so the assertion names both rather than picking one.
+  REQUIRE((result.first.ec() == couchbase::errc::network::cluster_closed ||
+           result.first.ec() == couchbase::errc::common::bucket_not_found));
 }
 
 // node_ids() reaches the bucket configuration through the same with_bucket_config_or_timeout
 // scaffold as scan(), so it exercises the same parked-in-bootstrap-window teardown path through a
 // second public entry point. A bucket that never opens keeps the wait parked; tearing the cluster
-// down must complete the future with errc::network::cluster_closed rather than strand it.
+// down must complete the future rather than strand it, with either of the two errors the race
+// described above can produce.
 TEST_CASE("integration: a node_ids future completes when the cluster is closed mid-bootstrap",
           "[integration]")
 {
@@ -118,7 +125,9 @@ TEST_CASE("integration: a node_ids future completes when the cluster is closed m
       "node_ids future was not completed after the cluster was closed -- the handler was stranded");
   }
   auto result = pending.get();
-  REQUIRE(result.first.ec() == couchbase::errc::network::cluster_closed);
+  // Same two correct answers as the scan case, for the same reason.
+  REQUIRE((result.first.ec() == couchbase::errc::network::cluster_closed ||
+           result.first.ec() == couchbase::errc::common::bucket_not_found));
 }
 
 // A key/value get reaches the bucket through direct_dispatch/open_bucket rather than the


test/test_integration_handler_drop_lifetime.cxx
```diff
@@ -88,13 +90,16 @@ TEST_CASE("integration: a scan future completes when the cluster is closed mid-b
     FAIL("scan future was not completed after the cluster was closed -- the handler was stranded");
   }
   auto result = pending.get();
-  REQUIRE(result.first.ec() == couchbase::errc::network::cluster_closed);
+  // The scan was torn down mid-bootstrap against a bucket that does not exist, so it must fail --
+  // the point is that it completes at all rather than being dropped.
+  REQUIRE(result.first.ec());
```
DemetrisChr: Could we just assert that the error code is either `cluster_closed` or `bucket_not_found` rather than dropping the check for a specific error code entirely?

DemetrisChr: I think there's also a chance that with relaxed validation (even with this suggestion), we might not be certain that the test is covering the behavior it's intended for. If the failure is a `bucket_not_found`, then we aren't validating  that `close` drains the bucket-bootstrap queue.

avsej: > Could we just assert that the error code is either `cluster_closed` or `bucket_not_found` rather than dropping the check for a specific error code entirely?

Done in b8e9ef2. Both cases now name the pair:

```c++
  // Either answer is correct: the teardown backstop reports cluster_closed, and the server reports
  // bucket_not_found for a bucket that does not exist. Which one arrives is a race between an HTTP
  // round trip and the teardown, so the assertion names both rather than picking one.
  REQUIRE((result.first.ec() == couchbase::errc::network::cluster_closed ||
           result.first.ec() == couchbase::errc::common::bucket_not_found));
```

This is stronger than what was proposed originally and the reasoning against it was weaker than it looked. The concern was that the set might not be closed, drawn from the protostellar analogue of the same race in `test/cng/protostellar/kv_retry.cxx`, where `closing_the_cluster_during_a_backoff_still_answers` has now been seen to report three different codes — `cluster_closed`, `temporary_failure (7)` and `request_canceled (2)`. That is a different close path through a different transport, so it is suggestive rather than evidence about this one. On this path, nine observed failures across seven pull requests have produced exactly two codes, so the pair is well supported, and a third code appearing here is worth failing on rather than absorbing.

The key/value case in the same file was deliberately left with the weaker `REQUIRE(result.first.ec())`. It reaches the bucket through `direct_dispatch`/`open_bucket` rather than the `with_bucket_config_or_timeout` scaffold the other two share, so its set of correct terminal codes is not the same two, and it already carried that reasoning before this change.

Verified: compiles clean under the project's warning set, `clang-format` clean. The commit message was updated to describe naming the pair rather than dropping the check.

DemetrisChr (approved):
Approving to unblock the tests, but with the caveat of my comment above.

> I think there's also a chance that with relaxed validation (even with this suggestion), we might not be certain that the test is covering the behavior it's intended for. If the failure is a bucket_not_found, then we aren't validating that close drains the bucket-bootstrap queue.

Addressing it is non-trivial.

PR #1094, opened 2026-08-20
CXXCBC-978: Add log redaction support in core

thejcfactor:
## CXXCBC-978: Add log redaction support in core

The SDK wraps sensitive log values in fixed `<ud>`/`<md>`/`<sd>` tags so that an external tool can
strip or hash them afterwards. **The SDK never removes or obscures anything itself**, and a log
written with redaction on still contains identifying information until those tags are processed.
The wording of the enable breadcrumb says exactly that, deliberately: claiming the log already has
reduced identifying information is untrue at write time and invites treating a raw log as
sanitised.

This is **core only**. There is no public API surface here: nothing under `couchbase/` is touched,
and `core/impl/public_logger.cxx` and `core/impl/public_cluster.cxx` are byte-identical to `main`.
The public `couchbase::logger` forwarders, `behavior_options`, and an `examples/` demo are
CXXCBC-979. That split is deliberate, and it is what lets this backport to `release-1.3`.

**The driver is the Python SDK**, and this ticket alone serves it: PyCBC consumes only
`couchbase::core::logger::*` and has zero public `couchbase::logger::` usage.

### What is in it

- `core/logger/redaction.hxx`: `user_data()`, `metadata()`, `system_data()`, their list forms, and
  two markers that record a reviewed exclusion. The formatters inherit `parse()`, so existing
  format specs keep working, and with redaction disabled every wrapper formats exactly as the
  underlying value. Annotating a statement changes nothing for anyone who has not opted in.
- `log_redaction` on `core::cluster_options` and in the connection string
  (`couchbase://host?log_redaction=on`). State is process-wide; opening a cluster can only turn it
  **on**, never off, so one cluster cannot silently disable redaction for another.
- `bin/check-log-annotations`, a Ruby checker with its own fixture, and a blocking `log_annotations`
  job in `linters.yml`. About five seconds over the whole tree, no toolchain and no compile
  database.
- The annotations themselves: the connection-context log prefixes first, since most sensitive data
  reaches the log through those, then the `CB_LOG_*` sites file by file.

### Reading it

One commit per decision, largest file first, and a `CB_LOG_*` statement's diff is **nothing but
added tags** wherever that was possible, so you can see the tag without re-verifying the message.
The exceptions are called out in their own commit messages; the main one is composing a host and
port into a single argument so an endpoint is one `<sd>` span, kept inside the macro's argument list
so it stays lazily evaluated.

Endpoints are tagged as **one span per endpoint**, not one per component, so the same endpoint
hashes to the same value everywhere it appears. Lists get **one span per entry**, because a tag
around a joined list hashes the list as a single token and then an entry matches nothing else in
the log, not even the same value logged beside it.

### Performance

Worth stating up front, since annotations are now mandatory. Measured with isolated microbenchmarks
against the bundled fmt at `-O2`, with the redaction accessor in a separate TU so the call is not
inlined. These are not in-tree benchmarks, and they are one machine.

Per tagged argument on a `std::string`:

| step | cost |
| --- | --- |
| fmt builtin fast path to custom-formatter dispatch | +8.3 ns |
| the cross-TU `is_log_redaction_enabled()` relaxed load | +0.0 ns |
| emitting `<sd>` and `</sd>` | +18.0 ns |

So **+8 ns per tagged argument with redaction off, +26 ns with it on**, and the redaction check
itself is free. The cost when off is that wrapping an argument moves it off fmt's builtin path onto
the type-erased custom path, which is inherent to expressing a tag through a formatter; hoisting
the check to statement level would buy nothing.

In proportion: the most heavily annotated statement on the branch is `bucket.hxx` "send operation",
five spans because a document id is four of them. A whole statement that actually logs, formatted
and handed to an async file logger, measured 748 ns, of which 163 ns is formatting and 585 ns (78%)
is the sink. Against that the annotations cost **+8% with redaction off and +19% with it on** for
that worst statement, and **+1% to +2%** for the typical one- or two-span statement.

**It is zero when the level is off**, which is where the per-operation `DEBUG`/`TRACE` statements
sit in a default deployment: `COUCHBASE_LOG` evaluates its whole argument list only inside
`if (should_log(severity) || has_custom_callback())`, so every wrapper and every composition inside
a macro argument list is skipped entirely.

Two caveats on that zero. `has_custom_callback()` has no level filter, so any wrapper that registers
a log callback formats every `CB_LOG_*` at every level; that is pre-existing and unrelated to this
change, PyCBC is not affected, but it is where to look first if a wrapper reports a regression.
And connection-context prefixes are formatted at construction, so each bucket and session pays one
extra `fmt::format` and string, once, next to a TCP connect.

**With redaction on the real cost is bytes, not CPU**: nine per span, and the sample line above goes
from 200 to 245 bytes. Since the sink is 78% of a statement, log IO scales with that. This one is
inherent, the tags are the deliverable.

### What is deliberately not annotated

- **Configuration dumps behind `dump_configuration`**, and `CB_LOG_PROTOCOL`. Both exist to show
  exactly what crossed the wire, both are opt-in debugging aids, and hashing the blob removes the
  information the dump exists to provide. Enabling `dump_configuration` together with
  `log_redaction` now warns at cluster open.
- **Application Telemetry** (`core/app_telemetry_reporter.cxx`, 33 arguments across 14 statements,
  the third largest file in the inventory). It is a newer feature whose annotation was never
  discussed, and its sensitive values live inside JSON objects rather than in the format arguments,
  so tagging it sets a precedent for every structured log line in the SDK. Excluded by name so the
  check can be enforced everywhere else. See the comment on the suppressions file.
- **Five `spdlog::to_hex` dumps** in `mcbp_session.cxx`. These are a real gap, not a false positive.
  See the comment on the first one.

### Two things the checker cannot see

Both are accepted for this ticket and worth knowing before anyone reads a green check as proof.

1. **Values composed with `fmt::format` before reaching a log statement.** The checker only parses
   `CB_LOG_*` statements. The connection-context prefixes are built this way, so the inventory did
   not move when they were annotated, and a newly added prefix site will pass the CI gate silently.
   Teaching it a second idiom is real work and is not in this ticket.
2. **Strings that are both logged and returned to the application.** Several paths build a message
   with `fmt::format`, store it on an error context the application can read, *and* log it, so the
   value reaches the log untagged. Tagging at composition would push `<md>`/`<sd>` markup into
   public API surface, visible in exception messages and anything an application prints. Confirmed
   instances in `mcbp_session.cxx` are the bucket name on the rate-limit, quota-limit and
   bucket-not-found paths, and the bootstrap endpoint when no endpoints are left. A full sweep of
   this shape is still owed. This is the same constraint that forces the HTTP path to be tagged at
   the log site rather than where it is built.

### One pre-existing inconsistency worth recording

An endpoint is rendered two ways in this codebase: bracketed `[addr]:port` for IPv6 through
`remote_address()`/`local_address()`, and unbracketed `addr:port` where `address()` and `port()` are
composed in message text. For IPv4 they agree; for IPv6 they do not, so the same node tagged from
each shape hashes to two different values and a reader of a redacted IPv6 log cannot tell they are
the same host. Annotation neither caused this nor can fix it without changing rendered text. Worth
checking whether the redaction tool normalises before hashing, which would make it moot.

### Merge gate: the `CXXCBC-978` references in source

**Eleven references to this ticket are placeholders, not provenance**, and they should be repointed
or removed before this merges, or they become references to a closed ticket.
`grep -rn CXXCBC-978 core/ .log_redaction_suppressions.txt` is the list:

- [ ] `core/io/mcbp_session.cxx` 563, 646, 737, 837, 1898: the five `dump_configuration` blobs.
      Settled as never annotated, so these want a comment with no ticket in it, or a pointer to
      wherever that decision is written down.
- [ ] `core/io/mcbp_session.cxx` 662, 676, 817, 853, 867: the five hex dumps. Blocked on one
      question for the redaction tool's owners.
- [ ] `.log_redaction_suppressions.txt`: Application Telemetry. Needs its own ticket after the team
      conversation.

I have left inline comments on the statements where a granularity or category choice is a judgement
call the team should see. None of those are leaks; they are places where the shape is arguable and I
would rather it be argued than inherited.

### Verification

- Built, and **integration and unit tests pass** against the committed tree.
- `bin/check-log-annotations` reports a clean tree over 1205 files, and its own fixture passes
  (20 statements, 20 expectations, 10 suppression globs).
- `bin/check-clang-format` equivalent over every line the branch touches: no changes.
- The formatter shapes were also checked in isolated probes before being applied, including that a
  document id renders byte-for-byte like the plain formatter with redaction off, and that a `{:a}`
  hex spec survives a marker wrapper.

**Not yet done, and it is the real acceptance test:** validating end to end against the Python
wrapper and the Couchbase redaction tool. I will do that before this comes out of draft.

### Backport

`release-1.3` is straightforward, measured rather than estimated, against `upstream/release-1.3`
at 99397bbf with this branch's checker and suppression list dropped in unmodified:

- It reports 211 unannotated arguments in 26 files there, against 226 in the same 26 files on
  `main`. The file sets are identical, and every log statement behind a `release-1.3` finding has a
  counterpart on `main` that this branch annotates, so the annotation work is a superset.
- Two findings differ only in the expression text, not the statement: `collection.cxx` names the
  member `bucket_name_` where `main` has a local `bucket_name`, and the DNS-SRV nameserver argument
  on `main` has since become an `.empty() ? "(system)" : ...` ternary. Same format strings, same
  sites.
- 37 of the 44 touched files exist there. Five of the remaining seven are new in this branch and
  come along with it. The other two, `AGENTS.md` and `.github/copilot-instructions.md`, do not
  exist on `release-1.3` in any form, so those 27 lines of documentation get dropped rather than
  adapted. `.github/workflows/linters.yml` and `test/CMakeLists.txt` are both present, so the
  checker still gets wired into CI there.

Expect one squashed `[Backport]` commit per that branch's convention.


.log_redaction_suppressions.txt
```diff
@@ -0,0 +1,16 @@
...
+# does belong to a statement is recorded there instead, with logger::not_sensitive() when the
+# value is not what its name suggests, or logger::not_redacted() when it is sensitive and
+# deliberately logged untagged. Both are visible where the value is logged, and neither silences
+# the arguments around it, which an entry here cannot say.
+
+# CXXCBC-978: Application Telemetry logs 33 addresses and hostnames, and none of them are
+# annotated. This is not a false positive: the feature is newer than the rest of the SDK and how
+# its logging should be redacted has not been discussed with the team. Excluded so the check can
+# be enforced everywhere else, to be removed once that conversation happens.
+core/app_telemetry_reporter.cxx
```
thejcfactor: Application Telemetry is the one file skipped outright: 33 arguments across 14 statements, the third largest in the inventory. Two things make it its own problem rather than more of the same.

The sensitive value lives *inside* a JSON object rather than in the format arguments. Every statement here is `"text.  {}"` with a single argument built by `tao::json::to_string`, and the hostname is a key within it. Tagging the whole blob hashes the keys and the harmless values with it and leaves an unparseable line. The alternative is to tag the value inside the object, which keeps the JSON valid and on one line and hashes only the hostname. That idiom exists already in this PR, at `origin.cxx:310` and `:321`, where a hostname and a nameserver are tagged inside the open-cluster JSON dump. But those are two keys in a dump this PR already owns; applying it here sets a precedent for every structured log line in the SDK, which I would rather agree on than establish by accident.

Separately, `base64::encode(payload)` at four sites is flagged `<ud>` and probably should not be. The server-to-client side of this protocol has one opcode and a valid inbound frame is a single byte; all four sites are error paths for frames we did not expect. No document data reaches this channel, so this reads as the checker matching on the name `payload`. There is no technical obstacle to tagging it, base64 is single-line. The question is purely whether an unexpected telemetry frame counts as user data.

Until this is settled the file keeps its 33 findings and the tree-wide count cannot reach zero.

core/io/mcbp_session.cxx
```diff
@@ -654,7 +659,8 @@ class mcbp_session_impl
                              msg.header.opcode,
                              utils::byte_swap(msg.header.opaque),
                              spdlog::to_hex(msg.header_data()),
-                             spdlog::to_hex(msg.body));
+                             // CXXCBC-978: a tag would open and close on different lines here
```
thejcfactor: These five hex dumps are a **real gap that ships unfixed**, not a false positive. The checker flags them `<ud>` and it is right to: these are dumps of unexpected protocol messages, so the contents are unknown by definition, and a KV body can carry a document key and value.

The obstacle is shape. `spdlog::to_hex` with `{:a}` renders across several lines, so wrapping it puts the opening and closing tags on different lines:

```
<ud>
0000: 41 42 43 44 ...  ABCDEFGH
0010: 51 52 53 54 ...  QRSTUVWX</ud>
```

The mechanism works. I wrote the wrappers, and an isolated probe against the real `spdlog/fmt/bin_to_hex.h` confirmed they compile and that `{:a}` still parses. Then I backed them out, because what is unverified is whether the redaction tool matches a tag pair spanning newlines, or only within a line.

**One question decides this:** does `cbredact` match across newlines? If yes, I restore the five wrappers. If no, the choice is between leaving them untagged and changing the `{:a}` spec to suppress newlines, which alters the rendered message and needs the same justification as any other message change. Either way this needs its own ticket, and the `CXXCBC-978` in the comment is a placeholder for it.

The adjacent `to_hex(msg.header_data())` needs no tag either way: opcode, opaque and lengths.

core/io/mcbp_session.cxx
```diff
@@ -552,12 +555,13 @@ class mcbp_session_impl
...
-                               "size={}, endpoint=\"{}:{}\"), {}",
+                               "size={}, endpoint=\"{}\"), {}",
                                session_->log_prefix_,
                                text.value().size(),
-                               info.endpoint_address,
-                               info.endpoint_port,
-                               text.value());
+                               logger::system_data(
+                                 fmt::format("{}:{}", info.endpoint_address, info.endpoint_port)),
+                               // CXXCBC-978: an opt-in dump, it exists to show the exact bytes
```
thejcfactor: The five `dump_configuration` sites are deliberately never annotated, and I want to check that reasoning holds for everyone. The blob carries the same kinds of values annotated everywhere else, but the content spans more than one category, a single tag pair around multi-line JSON is not a shape the tooling was built for, and hashing the whole thing destroys exactly what the dump exists to show. It is a debugging escape hatch in the same class as `CB_LOG_PROTOCOL`, which `redaction.hxx` already documents as out of scope, so it is treated the same way and the combination is made loud rather than silent: enabling both now warns at cluster open.

All five are marked rather than just the one the checker reported. The other four bind the blob as `text` so they were never flagged, and marking one of five identical sites would read as an oversight.

`not_redacted()` prints the value unchanged and asserts nothing about it, which is the honest claim here. The `CXXCBC-978` is a placeholder: if we agree this is permanent, it should become a comment with no ticket in it, or point at wherever the decision is recorded.

core/protocol/cmd_get_cluster_config.cxx
```diff
@@ -83,8 +84,9 @@ get_cluster_config_response_body::parse(key_value_status_code status,
       config_ = parse_config(config_text, info.endpoint_address, info.endpoint_port);
       config_text_.emplace(config_text);
     } catch (const tao::pegtl::parse_error& e) {
-      CB_LOG_DEBUG(
-        "unable to parse cluster configuration as JSON: {}, {}", e.message(), config_text);
+      CB_LOG_DEBUG("unable to parse cluster configuration as JSON: {}, {}",
+                   e.message(),
+                   logger::system_data(config_text));
```
thejcfactor: Two sites log a raw cluster configuration and this branch treats them differently, which is worth a look because I suspect the better end state is a third option.

`mcbp_session.cxx:1898` sits inside `if (origin_.options().dump_configuration)` and is left bare. This one has no gate and fires on any JSON parse failure at debug level, so it is tagged `<sd>`. The split follows the `dump_configuration` rationale, which has two prongs: the dump shows exactly what crossed the wire, **and** it is opt-in. The first site meets both, this one meets only the first.

**The third option is to gate this site behind `dump_configuration` too**, which folds it under the same exemption and removes the annotation. That is arguably right, since the line exists to show bytes we could not parse and a hash serves nobody. It is a behaviour change rather than an annotation change, so it does not belong on this branch: a parse failure would stop being visible at debug level, and someone diagnosing a config problem would have to know to turn the option on first. Putting it to the team rather than deciding it here.

core/io/http_command.hxx
```diff
@@ -263,7 +265,7 @@ private:
       session_->log_prefix(),
       encoded.type,
       encoded.method,
-      encoded.path,
+      logger::metadata(encoded.path),
```
thejcfactor: **Annotated, not deferred.** Nothing leaks here; the shape is a judgement call I would rather have seen than inherited.

Every HTTP request path is logged at five sites and carries Couchbase resource names: bucket, scope, collection, index, design document, eventing function. Four of the 51 path-building sites in `core/management` and `core/operations` also put an RBAC username in the path. The annotation wraps the whole path in a single `<md>` span.

It cannot be tagged where the path is built. `encoded.path` is the request target sent to the server, and `http_command.hxx:198` copies it into `ctx.path`, which the application reads through the error context. Tagging at construction would put markup on the wire and into public API surface, so the log site is the only place it can be wrapped.

The coarse span costs two things. A path hashes as one token, so the bucket name inside it will not match the `<md>` span that `bucket.cxx` and the collections logs emit for the same bucket, and someone tracing a bucket across a redacted log sees HTTP lines as unrelated tokens. And four endpoints put a username in an `<md>` span rather than `<ud>`; the value is still protected, but the category is wrong for that component.

Finer is not cheap: either a second, log-only path field built alongside the wire path at each of the 51 sites, or a parser at the log site that re-splits on `/` and re-tags by position, which needs a per-endpoint table to know which segment is a bucket and which is a username. Both are much larger than this PR. Worth asking the tool's owners whether `cbredact` already special-cases URI paths, since that may dictate the right shape better than what reads best here.

core/operations/document_query.cxx
```diff
@@ -191,15 +192,15 @@ query_request::encode_to(query_request::encoded_request_type& encoded, http_cont
   if (ctx_->options.show_queries) {
     CB_LOG_INFO("QUERY: client_context_id=\"{}\", prep={}, {}, options={}",
                 encoded.client_context_id,
-                utils::json::generate(prep),
-                utils::json::generate(stmt),
-                utils::json::generate(body));
+                logger::user_data(utils::json::generate(prep)),
+                logger::user_data(utils::json::generate(stmt)),
+                logger::user_data(utils::json::generate(body)));
```
thejcfactor, line 200: **Annotated, not deferred**, same class as the HTTP path: a granularity choice worth seeing.

Six statements render a generated JSON document and wrap the whole thing in one `<ud>` span. The blobs are mixed-category: the query options remainder carries `query_context`, a bucket and scope name, next to `args` and named parameters, which are application values; a search request body carries `collections` next to the query. So a `<ud>` span ends up containing values that are `<md>` elsewhere, and the document hashes as one token, meaning none of the names inside it correlate with the `<md>` spans other subsystems emit for the same bucket or scope.

Finer means walking a `tao::json::value` with a per-key category table and re serialising, which changes rendered output, needs the table kept in step with every option encoder, and would still put a whole `args` array in one span. Erring toward `<ud>` for the container is the safe direction: the categories differ only in how the tool treats the text, and `<ud>` is the strictest.

One visible cost, only with redaction on: the prepared statement name is tagged, because `PREPARE <name> FROM ...` lets an application choose it, so an adhoc query renders `prep=<ud>false</ud>` and support can no longer tell adhoc from prepared at a glance.

core/operations/document_query.cxx
```diff
@@ -313,7 +314,7 @@ query_request::make_response(error_context::query&& ctx, const encoded_response_
       if (!response.ctx.ec) {
         CB_LOG_TRACE("Unexpected error returned by query engine: client_context_id=\"{}\", body={}",
                      response.ctx.client_context_id,
-                     encoded.body.data());
+                     logger::user_data(encoded.body.data()));
```
thejcfactor, line 320: This one has a cost that lands on support, so I want it decided rather than defaulted.

Three sites log the errors the server sent back, and all three now wrap the payload in `<ud>`. The justification is that query and analytics error messages routinely quote the offending statement back (`syntax error near ...`), so the payload is a container for user data even though the wrapper around it is diagnostic text. It matches the call made for the generic non-200 HTTP body.

The cost: with redaction on these lines become a hash, and the error message and server error code are exactly what someone reads to diagnose the failure. Nothing else in the log carries them, and that is the point. This code path is the one where the SDK **could not map the error to an `errc`**, so there is no mapped code to fall back on.

The finer option is to tag only the message-bearing fields of each error object and leave `code`, `retry` and friends bare. That means walking the array at three sites and re-serialising, so it changes rendered output, and it assumes the server never puts statement text in a field we decided to leave bare. Privacy against diagnosability, and not a technical constraint, so worth a decision.

core/collections_component.cxx
```diff
@@ -364,8 +365,9 @@ collection_id_cache_entry_impl::dispatch(std::shared_ptr<mcbp::queue_request> re
    */
   switch (const std::scoped_lock lock(mutex_); id_) {
     case unknown_collection_id:
-      CB_LOG_DEBUG(
-        "collection {}.{} unknown. refreshing id", req->scope_name_, req->collection_id_);
+      CB_LOG_DEBUG("collection {}.{} unknown. refreshing id",
```
thejcfactor, line 368: Pre-existing and unrelated to redaction, found while annotating and **not fixed here** because the fix changes rendered text.

This line and `:380` both format a qualified collection as `{}.{}`, but the second argument is `req->collection_id_`, a `std::uint32_t`, not `req >collection_name_`. Every neighbouring message in the same class uses `scope_name_` and `collection_name_`, so these two read as copy-paste slips.

It is misleading twice over. It looks like a collection name but is a number, and at both sites the number is the sentinel the branch just matched on, so it carries no information at all. A reader correlating a redacted log sees `<md>myscope</md>.0` and has no way to learn which collection was being refreshed.

This commit tags `scope_name_` and leaves the argument alone. The fix is a one-token change at each site and wants its own commit, since it alters log output; doing it after this lands also means the new argument arrives already needing an `<md>` tag, which the checker will ask for. Happy to do it here instead if you would rather not carry it.

core/io/mcbp_command.hxx
```diff
@@ -117,7 +118,7 @@ struct mcbp_command : public std::enable_shared_from_this<mcbp_command<Manager,
           timeout_ < durability_timeout_floor) {
         CB_LOG_DEBUG(
           R"(Timeout is too low for operation with durability, increasing to sensible value. timeout={}ms, floor={}ms, id="{}")",
-          request.id,
+          logger::document(request.id),
```
thejcfactor, line 121: Another pre-existing one found on a line I was annotating. This statement has **three placeholders and four arguments**. fmt permits unused trailing arguments, so it compiles and always has, and every value is shifted one position left. It renders as

```
timeout=trav/inv.airline/airline_10ms, floor=2500ms, id="10000"
```

so the document id prints as the timeout, the timeout as the floor, the floor as the operation id, and the operation id is dropped. It predates `4901951e`, which last touched it while making the diagnostic id lazy, so it is long-standing.

This commit wraps `request.id` in `logger::document()`, which is correct wherever the value lands, and changes nothing else. The fix is to give the format string a `key="{}"` slot so all four arguments have somewhere to go, and it wants its own commit since it alters output.

Worth noting neither the checker nor the compiler counts placeholders against arguments, so nothing on this branch would have caught it. That made a sweep worth doing: a throwaway script that lexes out comments and string literals and counts placeholders against top-level arguments found **1 mismatch in 410 `CB_LOG_*` statements** across the whole tree, `test/` and `tools/` included. So this is the only one, and it is not worth a permanent check.

core/operations/document_get_all_replicas.hxx
```diff
@@ -86,8 +88,8 @@ struct get_all_replicas_request {
         if (nodes.empty()) {
           CB_LOG_DEBUG(
             "Unable to retrieve replicas for \"{}\", server_group={}, number_of_replicas={}",
-            id,
-            origin.options().server_group,
+            logger::document(id),
+            logger::metadata(origin.options().server_group),
```
thejcfactor, line 92: `server_group` is tagged `<md>`, which is what the checker's token table has always said, and it reads correct against the framework's own definition: Couchbase resource names are metadata, and `<sd>` is reserved for addresses, hostnames, ports and DNS. But a server group is a deployment topology name, so there is an argument for `<sd>`. It is a four-line change across the four replica read headers if the team disagrees.

.log_redaction_suppressions.txt
thejcfactor: Settled at the SDK meeting on 2026-08-26: Application Telemetry is out of scope for log redaction. The suppression entry now records the decision and its reasoning and carries no ticket reference, so it is no longer a placeholder.

The reasoning, for the record: what this file logs is the SDK's own telemetry channel, the cluster endpoint it connects to and the hostnames it reports for, and the payload is SDK metrics bound for Couchbase Server, which republishes the same metrics on `/metrics` to anything that can reach it. None of it is application data.

The second question, `base64::encode(payload)` being flagged `<ud>`, is moot with the file excluded.

The first question, whether to tag a value *inside* a JSON object, was answered separately and in the opposite direction, so it is worth saying explicitly: **the SDK does not write tags inside a serialized document.** A tag in a JSON string value is read back as part of that value by anything that parses the line, and some of these lines have shapes other tools consume. So the two sites in `origin.cxx` that did it have been changed, and that idiom now exists nowhere in the branch

core/origin.cxx
```diff
@@ -247,7 +248,7 @@ struct traits<couchbase::core::columnar::security_options> {
 namespace couchbase::core
 {
 auto
-origin::to_json() const -> std::string
+origin::to_json_for_log() const -> std::string
```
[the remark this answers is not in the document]
thejcfactor: Correct, and the fix is  the annotation moves out of the document rather than reaching further into it.

`origin::to_json()` no longer tags anything. The two log sites in `cluster.cxx` wrap the whole return in one `<ud>` span instead. That covers `server_group` and the ports, and it covers every option added to this dump in future without anyone having to remember. That was the real defect here, since the checker cannot see inside a serialised document and never reports this file.

`<ud>` rather than `<sd>` because one span has to carry values from more than one category and takes the strictest any value inside could belong to: `user_agent_extra` is application-supplied and `trust_certificate` is a filesystem path that can carry a username.

The cost is that with redaction on the whole options dump hashes to one token, timeouts included. That is accepted: the alternative leaves the cluster's bootstrap hostnames in clear on the open line while the same hosts are hashed everywhere else in the log, which makes the topology tagging worthless.

A unit case in `test_unit_logger.cxx` now asserts the document comes out of `to_json()` clean with redaction on, so re-adding a tag inside it fails a test.

test/test_integration_log_redaction.cxx
```diff
@@ -0,0 +1,178 @@
...
+  // A configuration dump is deliberately not annotated: it exists to show exactly what crossed
+  // the wire. Enabling both is therefore a combination worth saying out loud, since the dump can
+  // carry values that are tagged everywhere else in the same log.
+  const captured_log log{};
+
+  auto [err, cluster] = connect_to_cluster(ctx, "log_redaction=true&dump_configuration=true");
+  REQUIRE_SUCCESS(err.ec());
+
+  REQUIRE(couchbase::core::logger::is_log_redaction_enabled());
+  REQUIRE(log.contains("dump_configuration"));
```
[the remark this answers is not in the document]
thejcfactor: Right, and thanks. The needle also matches the cluster-open dump of the origin, which always carries a `dump_configuration` field, so the assertion could not have failed if the warning were deleted. It now matches `"log redaction is enabled, but so is dump_configuration"`, which only the warning carries.

core/io/mcbp_session.cxx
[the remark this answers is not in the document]
thejcfactor: Agreed that this is a merge gate, and one of the eleven is now closed: the Application Telemetry entry in the suppressions file was settled on 2026-08-26 and its comment carries no ticket reference any more.

The remaining ten are the five `dump_configuration` blobs and the five hex dumps, both in this file. The `dump_configuration` five are settled in substance, being explicitly the "none at all" branch of the serialized-document rule, and need only a decision to point at instead of a ticket. The hex-dump five are genuinely open: they wait on whether the redaction tool matches a tag pair spanning newlines.

core/io/http_command.hxx
thejcfactor: The SDK meeting on 08.26.2026 settled the general form of this: a serialised or composed value gets one span around the whole thing rather than a tag per component. So the coarse `<md>` around the whole path is the rule now, not a judgement call, and the "parse the path and re-tag by position" option is off the table.

The two costs I named still stand and are worth carrying in the ticket rather than the code: a bucket name inside a path span will not correlate with the `<md>` span `bucket.cxx` emits for the same bucket, and the four endpoints that carry an RBAC username put it in an `<md>` span

core/operations/document_query.cxx
thejcfactor: Same answer as the HTTP path: one span around the whole document is the rule as of 2026-08-26, so walking the `tao::json::value` with a per-key category table is off the table and the mixed-category span is expected rather than a compromise.

What the rule adds is how to pick the category: the strictest any value inside could belong to. That confirms `<ud>` here.

core/operations/document_query.cxx
thejcfactor: The granularity half of this is answered by the 2026-08-26 rule: tagging only the message-bearing fields of each error object is off the table, so it is now tag the whole payload or tag none of it.

Which of those two is still the open question, and it is the same privacy-versus-diagnosability call as before: with redaction on these lines become a hash, and the server error message and code are exactly what someone reads to diagnose a failure the SDK could not map to an `errc`.

core/io/mcbp_session.cxx
thejcfactor: Settled, and the reasoning is now written down rather than sitting in a comment.

The SDK meeting on 08.26.2026 agreed a general rule for a serialised document: one span around the whole argument, or none at all, and never a tag inside it. `redaction.hxx` documents it, and these five dumps are explicitly the "none at all" case, for the reason you gave. They exist to show exactly what crossed the wire, and a hash of the blob serves nobody.

So the `CXXCBC-978` in the comment is gone. It now names the rule instead, which is what you asked for: a comment with no ticket in it, pointing at where the decision is recorded. All five say it, not just the one the checker reported.

That takes the placeholder count in this file from ten to five. The remaining five are the hex dumps, which are still waiting on whether the redaction tool matches a tag pair spanning newlines.

core/io/mcbp_session.cxx
```diff
@@ -278,10 +279,12 @@ class mcbp_session_impl
...
         session_->client_id_, session_->id_, session_->origin_.options().user_agent_extra, 250);
       hello_req.body().user_agent(user_agent);
-      CB_LOG_DEBUG("{} user_agent={}, requested_features=[{}]",
-                   session_->log_prefix_,
-                   user_agent,
-                   utils::join_strings_fmt(hello_req.body().features(), ", "));
+      CB_LOG_DEBUG(
+        "{} user_agent={}, requested_features=[{}]",
+        session_->log_prefix_,
+        user_agent,
```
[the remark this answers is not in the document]
thejcfactor: Agreed, and fixed. `user_agent_for_mcbp()` folds in `origin_.options().user_agent_extra`, and the result is a serialised blob, so by this branch's document rule it takes one span around the whole argument at the strictest category anything inside it could be. That is `<ud>`. It now reads `logger::user_data(user_agent)` with a comment naming the rule.

I also added `user_agent` and `user_agent_extra` to the checker's `user_data` token table, so a future unwrapped site gets flagged instead of relying on review catching it. Useful sanity check on the addition: against the current tree it produced exactly one new finding, this line, which this change closes.

core/cluster.cxx
```diff
@@ -555,11 +555,12 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
     }
 
     origin_ = std::move(origin);
+    apply_log_redaction();
     CB_LOG_DEBUG(R"(open cluster, id: "{}", core version: "{}", connection string: {}, {})",
                  id_,
                  couchbase::core::meta::sdk_semver(),
-                 origin_.connection_string(),
-                 origin_.to_json());
+                 logger::system_data(origin_.connection_string()),
```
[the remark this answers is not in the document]
thejcfactor: Agreed, and fixed at both sites. The inconsistency was exactly as described: the same statement tagged `origin_.to_json()` as `<ud>` precisely because it carries `user_agent_extra` and `trust_certificate`, while the connection string that can carry both of those as options was `<sd>`.

One fact from an end-to-end run against the real `cblogredaction` that makes this more clear-cut than it looks: **the shipping tool only hashes `<ud>`.** `<md>` and `<sd>` pass through byte-identical, tags and contents intact, because it implements only the server's "partial" level. So `<sd>` here was not a weaker policy applied to those values, it was no policy at all. More detail in a separate comment on the PR.

core/http_component.cxx
```diff
@@ -90,7 +91,7 @@ class pending_http_operation
         R"(HTTP request timed out (dispatch): {}, method={}, path="{}", dispatch_timeout={}, client_context_id={})",
         self->encoded_.type,
         self->encoded_.method,
-        self->encoded_.path,
+        logger::metadata(self->encoded_.path),
```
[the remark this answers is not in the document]
thejcfactor: Fixed at all four sites (94, 112, 294, 312). Reasoning is on the `http_command.hxx:268` thread rather than repeated here, including why the path can only be tagged at the log site and what the coarse span costs.

core/io/http_command.hxx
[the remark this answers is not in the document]
thejcfactor: Agreed, and fixed here and at the four sites in `http_component.cxx`. Both citations check out: `user_get.cxx:34` formats a username straight into the path, and `document_view.cxx:129-134` joins application-supplied `raw` query values into it.

Two notes on the shape, since "the complete composite" is doing real work in your comment.

The path can only be tagged at the log site. `core/io/http_session.hxx:153` writes `encoded.path` directly into the HTTP request line, and `http_command.hxx:196` also copies it into `ctx.path`, which reaches the application through the error context. Tagging at construction would put markup on the wire and into public API surface, so it is genuinely whole-path-or-nothing, and under the strictest-category rule that is `<ud>`.

The cost is accepted rather than overlooked: since the tool hashes `<ud>`, static endpoints like `/`, `/admin/ping` and `/api/ping` now become digests in a redacted log. The same statement logs `encoded.type` in the clear, so the service stays identifiable even when the exact path does not. Tagging only the variable segments would need either a log-only path built alongside the wire path at all 51 construction sites, or a per-endpoint table at the log site to know which segment is which. Both are much larger than this PR.

core/io/mcbp_session.cxx
```diff
@@ -1283,8 +1299,8 @@ class mcbp_session_impl
                      self->log_prefix_,
                      old_id,
                      self->stream_->id(),
-                     self->bootstrap_hostname_,
-                     self->bootstrap_port_);
+                     logger::system_data(self->bootstrap_hostname_),
+                     logger::system_data(self->bootstrap_port_));
```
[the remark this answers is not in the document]
thejcfactor: Leaving this one as it is, and it is a rule on the branch rather than an oversight, so it is worth stating explicitly.

This is the single sanctioned exception to one-span-per-endpoint: the message logs `host=` and `port=` as separate **named fields**. Composing them would render `endpoint="host:port"`, which renames the fields rather than only adding a tag, and the working rule for this PR is that a `CB_LOG_*` statement's diff should be nothing but added tags so a reviewer does not have to re-verify the message. Everywhere the endpoint already exists as one value it is one span: `bootstrap_address_` is composed at `:1345` and logged as a single `<sd>` at `:1353`, `:2051`, `:2068`, `:2079` and `:2189`.

The correlation cost you name is real and I am not disputing it. Two things about its size. First, an end-to-end run against the real `cblogredaction` shows the shipping tool does not hash `<sd>` at all; it passes those spans through byte-identical, so today neither shape correlates any differently from the other. Second, if and when the tool implements the full redaction level, the right fix is a composed log-only endpoint value, not a renamed field, and that is a change worth making deliberately rather than as a side effect of annotating.

core/operations/document_get_all_replicas.hxx
thejcfactor: One fact for whoever picks this up, from an end-to-end run against the real `cblogredaction`: the shipping tool **only hashes `<ud>`**. `<md>` and `<sd>` both pass through byte-identical, tags and contents intact, because it implements only the server's "partial" redaction level.

So `<md>` versus `<sd>` on `server_group` has no observable effect on a redacted log today. The choice is about what the tag should mean once the tool implements the full level, not about what anyone sees now. That makes leaving it as `<md>` the cheaper option, and it means the four-line change across the replica read headers is not urgent in either direction.

core/io/mcbp_session.cxx
thejcfactor: Answered, and the answer is no.

I built the Python SDK against this branch and ran a workload through the real `cblogredaction`. The tool is **line-oriented**: when a tagged value spans lines it reports `WARNING: Unmatched tags detected on line N, potential data leak` and passes the value through **in clear text**. So of the two options above it is the second one, and the five stay untagged.

Three things followed from that run, and together they close this out without a separate ticket.

1. **The framework now escapes `\n` and `\r` inside a tag** while redaction is enabled. This was never only a hex dump problem: the same failure hit an HTTP error body in that run and leaked it. Escaping in the wrappers covers every tagged value, including ones added later, and it is not something a call site can be expected to remember.

2. **So a tag here would now hold, and it is still wrong.** Escaping would flatten the grid onto a single line of `\n` literals, and a flattened hex dump is not a hex dump. Showing the exact bytes is the only reason these statements exist. They keep `not_redacted()`, and the comment now states the decision and points at `redaction.hxx` instead of naming a ticket.

3. **The checker now reports a tag around a `to_hex()` call as `<multiline>`**, meaning take the tag off. That exists so nobody "closes" this thread later by wrapping the five and quietly destroying the dumps. It is deliberately narrow, since this is the one shape where escaping is the wrong answer.

The `CXXCBC-978` placeholders are gone from the tree.

One residual question I would like other eyes on, which I have raised as a separate comment rather than here: these five are `CB_LOG_WARNING`, so unlike `CB_LOG_PROTOCOL` and `dump_configuration` they are on by default.

core/protocol/cmd_get_cluster_config.cxx
thejcfactor: Two updates, neither of which is the behaviour change this thread proposes.

**The line itself changed.** `config_text` is now `const` at both sites. Wrapping the argument is what caused that: the wrapper takes a `const T&`, so every remaining use of the variable became const and clang-tidy's `misc-const-correctness` fired. Worth recording why it does not fire on `main`, since it looks like a pre-existing issue and is not: passing the variable straight into `CB_LOG_*` binds it to fmt's `make_format_args(T&... args)`, a non-const lvalue reference, which `ExprMutationAnalyzer` treats as an escape and suppresses the check on. Inserting the wrapper removes that escape. A new finding created by the annotation, not a latent one it exposed.

**The third option gets stronger.** An end-to-end run against the real `cblogredaction` shows it **only hashes `<ud>`**; `<md>` and `<sd>` pass through byte-identical, because it implements only the server's "partial" level. So the `<sd>` tag on this site protects nothing today, which removes most of what was arguing for keeping the annotation rather than gating the site behind `dump_configuration`. Still a behaviour change, still one for the team rather than this branch, but the balance has moved.

core/logger/redaction.hxx
```diff
@@ -0,0 +1,412 @@
...
+  std::size_t copied = 0;
+  for (std::size_t index = 0; index < value.size(); ++index) {
+    std::string_view replacement{};
+    switch (value[index]) {
+      case '\n':
+        replacement = "\\n";
+        break;
+      case '\r':
+        replacement = "\\r";
+        break;
```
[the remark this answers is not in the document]
thejcfactor: Real, and thank you for finding it. This was on my follow-up list as something to test and had not been done yet.

One correction to the suggested fix, which matters: escaping to `\<` does not work, because `\</ud>` still *contains* `</ud>`. The escape has to remove the `<` itself, so it emits `\u003c`.

On scope I went narrower than "escape every `<`". Only a `<` that actually begins one of the six tag sequences is escaped, so `</ud>` becomes `\u003c/ud>`, while `<html><body>` passes through untouched.

The reasoning, since it changed how I scoped the fix: I could not construct a case where this exposes a *co-located* sensitive value. Whoever controls the value already knows it, the text left outside the span is their own, and the tool stays balanced across the rest of the line. So the framing that matters is not adversarial, it is **accidental**: an application whose document happens to contain one of those sequences hands out a log it believed was redacted, and redaction silently failed for that value. Against accidental occurrence, escaping every `<` buys nothing extra while mangling any logged body that carries markup.

Covered by tests across the scalar wrappers, all three categories and the list form, including one asserting that ordinary markup is left alone.

Since confirmed against the shipping tool rather than only against our tests. A normal workload never produces these sequences, so the run injected them deliberately: a document key carrying `</ud>`, `<ud>`, `</md>` and `</sd>`, plus a `<div>KEEP-MARKUP</div>` control, and two query statements carrying `</ud>`. In the tagged log every one of them arrives as `\u003c/ud>`, and `cblogredaction` swallows the whole value as a single token: `276 lines processed, 49 tags redacted, 0 lines with unmatched tags`, no warnings, and not one of the injected sentinel strings survives anywhere in the redacted output. The control holds as well, since `<div>KEEP-MARKUP</div>` and `1<2` come through unescaped.

The unredacted run is the other half of the proof. With redaction off nothing is escaped, the same values reach the tool raw, and it responds with `WARNING: Unmatched tags detected on line N, potential data leak` on exactly the two injected query lines. That is the failure this escape prevents, demonstrated by the tool that consumes the tags.

Worth noting that both escapes are JSON's own: `\n` is how JSON spells a newline inside a string, where a literal one is illegal, and `\u003c` is a legal escape for `<`. So a parser reconstructs the original value exactly. That matters here because the SDK RFC 0067 threshold and orphan reports are tagged documents another tool parses field by field, and the header used to claim those stayed byte-identical. It now says what is actually true: the structure survives a parser, the bytes do not.

core/logger/redaction.hxx
```diff
@@ -0,0 +1,412 @@
...
+  struct fmt::formatter<couchbase::core::logger::wrapper<T>> : fmt::formatter<T> {                 \
+    template<typename FormatContext>                                                               \
+    auto format(const couchbase::core::logger::wrapper<T>& redacted, FormatContext& ctx) const     \
+    {                                                                                              \
+      if (!couchbase::core::logger::is_log_redaction_enabled()) {                                  \
+        return fmt::formatter<T>::format(redacted.value, ctx);                                     \
+      }                                                                                            \
+      fmt::memory_buffer buffer;                                                                   \
+      fmt::basic_format_context<fmt::appender, char> buffered{ fmt::appender(buffer), {} };        \
+      fmt::formatter<T>::format(redacted.value, buffered);                                         \
```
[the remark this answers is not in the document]
thejcfactor: Good catch, and it was worse than a lost specification: it threw. The buffered context had an empty argument store, so a dynamic width resolved through it found nothing, while the redaction-off branch used the caller's context and worked fine. Behaviour that depends on the redaction level is the worst possible shape for this, since it only breaks for people who opted in.

Fixed by constructing the buffered context with `ctx.args()` and `ctx.locale()`, so the caller's arguments come along. `fmt::format("[{:{}}]", logger::metadata("ab"), 6)` now gives `[<md>ab    </md>]` with redaction enabled and `[ab    ]` with it disabled. Both branches are pinned by a test, precisely because the point is that the level must not matter.

One note on the adapter you suggested: an output-adapter approach still has to carry the caller's arguments into a new context, so it lands in the same place. Passing them into the buffered context is the smaller change. It does mean the wrappers now require an appender-based context, which is what `fmt::format` and `CB_LOG_*` both use; anything else becomes a compile error rather than silent breakage, which is the right failure mode.

core/cluster.cxx
```diff
@@ -1884,6 +1890,26 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+  // Redaction state is process-wide while it is configured per cluster, so opening a cluster may
+  // only turn it on, never off: one cluster must not silently disable redaction for another, or
+  // for an application that enabled it directly.
+  void apply_log_redaction() const
+  {
+    if (!origin_.options().log_redaction) {
+      return;
+    }
+    logger::set_log_redaction(true);
+    if (origin_.options().dump_configuration) {
```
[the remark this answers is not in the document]
thejcfactor: Correct, and the function's own comment states the thing that makes it a bug: redaction state is process-wide while it is configured per cluster. The early return keyed off the local option, so the warning could never fire for a cluster that had not itself asked for redaction, which is exactly the case where the mismatch is invisible.

Fixed as you describe: set from the local option first, then test `is_log_redaction_enabled()` when deciding whether to warn. So a second cluster opened with only `dump_configuration=true` now gets the warning, and so does one opened after a direct `set_log_redaction(true)`.

core/protocol/cmd_get_cluster_config.cxx
```diff
@@ -77,14 +78,15 @@ get_cluster_config_response_body::parse(key_value_status_code status,
...
+                                        body.size() - static_cast<std::size_t>(offset) };
     try {
       config_ = parse_config(config_text, info.endpoint_address, info.endpoint_port);
       config_text_.emplace(config_text);
     } catch (const tao::pegtl::parse_error& e) {
-      CB_LOG_DEBUG(
-        "unable to parse cluster configuration as JSON: {}, {}", e.message(), config_text);
+      CB_LOG_DEBUG("unable to parse cluster configuration as JSON: {}, {}",
+                   e.message(),
+                   logger::system_data(config_text));
```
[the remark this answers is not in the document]
thejcfactor: Right about the contents. `configuration_json.hxx` does put the bucket `name` and `clusterName` into the parsed schema, so this blob is not purely system data, and by the strictest-category rule that argues for `<md>`.

Not changing it on this branch, for two reasons.

An end-to-end run against the real `cblogredaction` shows it hashes `<ud>` **only**: `<md>` and `<sd>` both pass through byte-identical, tags and contents intact, because it implements only the server's "partial" level. So `<md>` versus `<sd>` on this argument has no observable effect on a redacted log today.

And there is already an open thread on this exact site proposing what I think is the better answer: gate it behind `dump_configuration` like the other configuration dump, and drop the annotation entirely. The line exists to show bytes the SDK could not parse, and a hash serves nobody. That is a behaviour change rather than an annotation change, so it does not belong here.

https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825891330

If the team decides to keep the annotation rather than gate the site, then your point decides the category and it becomes `<md>`.

core/cluster.cxx
```diff
@@ -1884,6 +1890,29 @@ class cluster_impl : public std::enable_shared_from_this<cluster_impl>
...
+  void apply_log_redaction() const
+  {
+    if (origin_.options().log_redaction) {
+      logger::set_log_redaction(true);
+    }
+    // The warning has to test the state that is in effect rather than what this cluster asked for.
+    // Redaction is process-wide, so another cluster or a direct set_log_redaction() call may have
+    // turned it on, and a cluster opened with only dump_configuration would then write
+    // unannotated dumps into a redacted log without saying so.
+    if (origin_.options().dump_configuration && logger::is_log_redaction_enabled()) {
```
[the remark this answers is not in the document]
thejcfactor: Real, and the mutation you describe is the right way to state it. I checked it against the three
existing cases rather than assuming:

- `applies log_redaction from the connection string` sets no `dump_configuration`, so the branch is
  never entered.
- `never turns log redaction off` also sets no `dump_configuration`.
- `enabling log_redaction and dump_configuration together` sets both on the same cluster, so
  `origin_.options().log_redaction` and `logger::is_log_redaction_enabled()` are indistinguishable.

So a regression to the local option passes all three. Rather than leave that as an argument I made
the change you describe, `origin_.options().dump_configuration && origin_.options().log_redaction`,
and ran the suite: all three existing cases pass. That is worth pinning, because reading the
effective state instead of this cluster's option is a deliberate decision with a four line comment
justifying it, and nothing was holding it in place.

Added the case you asked for: redaction turned on directly through `set_log_redaction`, then a
cluster opened with only `dump_configuration=true`, asserting the warning. The cluster never asks
for redaction, so the warning can only fire if the check reads process-wide state. Against the
mutated branch it is the one case that fails, and it fails on its own, which is the property that
makes it worth having.

Added its negative as well, since your test alone does not close the whole hole. A regression the
other way, `if (origin_.options().dump_configuration)` with the state check dropped entirely, would
satisfy every case including the new one, and would make `dump_configuration` unusable on its own by
warning about redaction that is not enabled. So the second case opens with only
`dump_configuration=true` against redaction off and asserts the warning does not appear.

Both mutations were run, and each is caught by exactly one case, alone:

| Mutation | both enabled | enabled elsewhere | dump_configuration alone |
|---|---|---|---|
| none | pass | pass | pass |
| local option instead of effective state | pass | **fail** | pass |
| state check dropped entirely | pass | pass | **fail** |

The first column is the pre-existing case, and it never moves, which is the finding you opened with.

thejcfactor:
## Validated end to end against the real redaction tool

I built the Python SDK against this branch, added `log_redaction` to its `ClusterOptions` plumbing,
and ran a workload (bootstrap, bucket create, KV upsert/get/replace/exists/remove plus a deliberate
miss, two N1QL queries) with `trace` logging, once with redaction off and once on, against a live
4-node cluster. The tagged log then went through the shipping **`cblogredaction`** from a Server
container with a fixed salt. Three rounds: the first found three defects, the second confirmed the
fixes, the third injected values designed to break the tags and closed out a fourth that Copilot
had found by inspection.

**The contract holds.** The SDK emits well-formed tags, the shipping tool consumes them without
complaint, and every value that should be hashed is hashed:

```
253 lines processed, 38 tags redacted, 0 lines with unmatched tags
```

With redaction off the log is byte-for-byte what it was before this branch, which is the property
that lets us annotate freely without affecting anyone who has not opted in.

That is the headline: the design works against the real tool, not just against our unit tests.

### The four defects, and their state now

| | before | now |
|---|---|---|
| lines with an unbalanced `<ud>` | 10 | **0** |
| `<ud>` values actually hashed | 25 of 30 | **38 of 38** |
| `<ud>[hidden]</ud>` | 9 | **0** |
| `prep=<ud>false</ud>` | 2 | **0**, bare `prep=false` |
| injected `</ud>` escaping its own span | leaked | **0 sentinels survive** |

**1. A tagged value containing a newline was emitted in clear text.** The important one.
`cblogredaction` is line-oriented, so a value spanning lines leaves its opening and closing tags on
different lines; the tool reports `WARNING: Unmatched tags detected on line N, potential data leak`
and passes the value through untouched. Five values came out readable. They happened to be 301
redirect pages, but the site was `http_command.hxx`, which is exactly where real bodies appear.

Fixed in the wrappers rather than at the call sites, which the run then justified: it exercised the
path six times without any call site knowing it (four 301s, a 400, a 404). Output is still unchanged
when redaction is off, and `not_redacted()` escapes nothing, since it exists to show exact bytes.

**2. `<ud>[hidden]</ud>`.** The HTTP response log already elided 200 bodies to the literal
`[hidden]`, then tagged it, so the tool hashed a constant to the same digest nine times. That makes
a redacted log actively worse: you cannot tell a body the SDK withheld from one the tool redacted.

**3. `prep=<ud>false</ud>`.** Same shape in the QUERY logs, where `prep` is either the prepared
statement name or the boolean `false`. Hashing a two-value domain protects nothing and destroys the
answer to "was this query prepared?". 2 and 3 are fixed with a new
`logger::user_data_if(sensitive, value)`, which tags only the branch that carries a sensitive value.

**4. A value containing a literal tag sequence closed its own span.** Copilot's find, and the third
run is what confirms the fix against the tool rather than only in our tests, since no normal workload
produces this. Injected: a document key carrying `</ud>`, `<ud>`, `</md>` and `</sd>` plus a
`<div>KEEP-MARKUP</div>` control, and two query statements carrying `</ud>`. Every sequence arrives
escaped as `\u003c/ud>`, the tool takes the whole value as one token
(`49 tags redacted, 0 lines with unmatched tags`, no warnings), and **not one injected sentinel
survives anywhere in the redacted output**. The control passes through unescaped, so only a `<` that
really begins one of the six sequences is rewritten and ordinary markup in a logged body stays
readable.

The unredacted run is the other half of the proof: the same values reach the tool raw and it emits
`Unmatched tags detected ... potential data leak` on exactly the two injected query lines. That is
the failure the escape prevents, demonstrated by the tool that consumes the tags.

## Three things about the tool that are worth knowing regardless of this PR

**It only hashes `<ud>`.** `<md>` and `<sd>` pass through byte-identical, tags and contents intact;
it implements only the server's `partial` level, and no flag changes that. So today every
`<md>`/`<sd>` span is forward-looking and costs log volume without protecting anything. That is what
settled the earlier review comment about HTTP paths: `<md>` on a path was zero protection, so paths
are now `<ud>`.

**Its summary counter is wrong.** It reported `0 lines with unmatched tags` on a file for which it
had just emitted warnings, and non-zero redaction counts for a file it did not modify at all. If
anything ever gates on this tool, scan for `WARNING` lines rather than trusting the tally.

**92% of what we are paying for is currently inert.** Measured on the tagged file: 457 tags, 4,113
bytes of markup, 5.4% of the log. Of that, `<sd>` is 60% and `<md>` 32%; `<ud>`, the only category
the tool acts on, is 8%. That is not an argument against the `<md>`/`<sd>` work, since the tags have
to exist before any level can consume them, but it should be a deliberate bet on a future `full`
level rather than an assumed win. **Does anyone know whether that level is actually on a roadmap?**
Worth asking whoever owns `cblogredaction` before we commit to carrying the volume.

## Two category decisions worth a look

**`user_agent` is tagged only when the application set `user_agent_extra`.** The blob carries SDK
version, core version, OS and architecture, TLS library and language runtime version, which is the
first block of facts any support ticket asks for. From `user_agent_for_mcbp` there is exactly one
route for user data into it, the appended `user_agent_extra`, and the only other field is the client
and session id that `log_prefix_` already prints in the clear on the same line. So the discriminator
is not the function, it is whether an extra is set, and that is right at the call site. Unset by
default means support keeps the version block in nearly every log that matters, while an application
that does set one pays the full tag without anyone having to reason about what they put in there.

**The connection string is `<ud>` while hosts elsewhere are `<sd>`,** which looks inconsistent until
you look at what the string is. It is `connstr.input`, the raw text the application passed, so it can
carry a username and any query option, including `trust_certificate` (a filesystem path) and
`user_agent_extra`. Both of those are also serialised into the origin blob logged next to it
(`origin::to_json()`), so both spans on that line are `<ud>` for the same reason. The bare host
references stay `<sd>` because that is all they are.

## Open questions for the team

`<ud>`-only sorts the open questions on this PR into ones with a live cost and ones that are
currently theoretical. All of these are inline threads with the full reasoning; this is just an index
so there is one place to start.

**Where redaction costs diagnosability today, because the span really is hashed:**

- [Server error payloads are tagged `<ud>`](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825916798)
  is the one I flagged as landing on support. The runs confirm this cost is real rather than
  hypothetical: with redaction on, the error message and server error code become a hash, and on
  this path there is no mapped `errc` to fall back on. **This is the one I would most like decided
  rather than defaulted.**

**Where the annotation currently has no observable effect, so the question is really about future
meaning:**

- [Cluster config text is tagged `<sd>` here and left bare behind `dump_configuration` there](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825891330)
  is worth revisiting: since `<sd>` is not hashed, the tag protects nothing today. That strengthens
  the third option in that thread: gate this site behind `dump_configuration` too and drop the
  annotation. It is a behaviour change, so it is not on this branch.
- [`server_group` as `<md>` or `<sd>`](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825932604)
  makes no observable difference today either way. Cheap to leave as `<md>` and revisit.

**Exposure rather than redaction:**

- [The five unrecognised-frame hex dumps](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825876281)
  are staying untagged, and that thread explains why now that the newline question is answered. The
  residual question is the one below.

That residual question, since it is the only thing on the branch I have deliberately left open:
those five statements are `CB_LOG_WARNING`, so unlike `CB_LOG_PROTOCOL` and `dump_configuration`
they are **on by default**. They dump the frame header hex plus
`logger::not_redacted(spdlog::to_hex(msg.body))` for a frame the SDK could not classify, and a KV
body can carry a document key and value. So unrecognised bytes can land in a redacted log that
nobody opted into.

The option I want opinions on: **drop the body hex from those five and keep the header hex.** The
header carries magic, datatype, keylen, extlen, vbucket, status and bodylen, which is what identifies
an unexpected magic or an unknown opcode, and `opaque` is already printed as a field. The body bytes
would not become unobtainable, because `mcbp_session.cxx:2219` already dumps every inbound byte to
`CB_LOG_PROTOCOL` on the read path, so anyone who could have read them needed protocol logging on
anyway. The cost is that the protocol log holds the raw TCP read rather than the framed body, so you
would locate the frame inside a stream dump instead of reading it inline, and a frame split across
two reads spans two lines.

If you have ever actually used that body dump to diagnose something, I would rather hear it now than
guess. I am happy to leave it exactly as it is. It is a behaviour change rather than an annotation
change, so it does not have to ride on this PR either way.

**Two pre-existing bugs, found while annotating and unrelated to redaction.** Both are one-token
fixes that change rendered output, so I left them off this branch rather than bury them in an
annotation commit. Happy to fold either in if you would rather not carry it:

- [`collections_component.cxx:368`](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825922319)
  passes the collection **id** where the format string promises the collection name, at two sites.
- [`mcbp_command.hxx:121`](https://github.com/couchbase/couchbase-cxx-client/pull/1094#discussion_r3825928202)
  has three placeholders and four arguments, so every value renders one position left. A sweep of
  all 410 `CB_LOG_*` statements in the tree found this to be the only such mismatch.

---

*Reviewed and drafted with Claude Code; the measurements and the decisions are mine.*

core/io/mcbp_command.hxx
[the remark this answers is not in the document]
thejcfactor: This has already been raised.  TBD on decision from team.

avsej:
### The three data classes

| Class | Tag | Definition | What an SDK logs that falls here |
|---|---|---|---|
| **User data** | `<ud>` | Data written into Couchbase by the application user account | Document keys and values; xattrs; SQL++ statement text, which leaks field names (`SELECT floor_price FROM stock`); usernames identifying a person; names and email addresses |
| **Metadata** | `<md>` | Logical data Couchbase needs in order to store and process user data | Bucket, scope and collection names; index names; design doc and view names; view code; XDCR replication stream names; cluster name |
| **System data** | `<sd>` | Data about the systems Couchbase talks to over the network | Hostnames; IP addresses; ports; DNS names; certificate subjects |

Scope is files collected by `cb_collect_info`, not application data at rest.

### The spectrum

Each level redacts everything the level below it redacts, plus one more class.

| Level | `<ud>` | `<md>` | `<sd>` | Status | What the level assumes |
|---|---|---|---|---|---|
| **None** | clear | clear | clear | Ships today; the default | Nothing |
| **Partial** | **redacted** | clear | clear | Target for Vulcan | That sensitive information lives only in the data, never in a resource name or a hostname. The customer carries the burden of naming buckets, indexes, ddocs and hosts without PII in them |
| **Full** | **redacted** | **redacted** | **redacted** | Backlog, no release named | Nothing |

Only `<ud>` content is removed at Partial. Tagging `<md>` and `<sd>` costs nothing today and is what makes Full a configuration change rather than an audit of every log statement.

Custom rule-based redaction — Cassandra-style rules for card numbers, SSNs, email addresses — is out of scope for both levels and deferred to a future release.

### Which tag to use

Ask this of every value interpolated into a log line.

```mermaid
flowchart TB
  A["Value about to be interpolated into a log line"] --> B{"Is it a literal from SDK source?<br/>opcode name, status enum, error code,<br/>line number, duration, retry count"}
  B -- Yes --> PLAIN["No tag"]
  B -- No --> C{"Is it a secret?<br/>password, bearer token, TLS private key,<br/>SASL payload, cookie"}
  C -- Yes --> DROP["Do not log it at all"]
  C -- No --> D{"Did it come from the application?<br/>document key or value, xattr,<br/>SQL++ statement, username, user-supplied id"}
  D -- Yes --> UD["Tag as UD"]
  D -- No --> E{"Does it name a Couchbase resource?<br/>bucket, scope, collection, index,<br/>design doc, XDCR stream, cluster"}
  E -- Yes --> MD["Tag as MD"]
  E -- No --> F{"Does it identify a host<br/>or the network path?<br/>hostname, IP, port, DNS name, cert subject"}
  F -- Yes --> SD["Tag as SD"]
  F -- No --> UD2["Came from outside the SDK<br/>and fits none of the above:<br/>tag as UD"]

  classDef ud fill:#ffebee,stroke:#c62828,color:#b71c1c
  classDef md fill:#fff8e1,stroke:#ef6c00,color:#e65100
  classDef sd fill:#e3f2fd,stroke:#1565c0,color:#0d47a1
  classDef ok fill:#e8f5e9,stroke:#2e7d32,color:#1b5e20
  classDef no fill:#eceff1,stroke:#455a64,color:#263238
  class UD,UD2 ud
  class MD md
  class SD sd
  class PLAIN ok
  class DROP no
```

Two rules the tree cannot carry:

- **A value that mixes classes is tagged whole, at the most sensitive class it contains.** A connection string holds a hostname, a bucket name and a username at once; the spec settles this by redacting the entire parameter, "not just sensitive parameters", because substring-level handling is where the feature gets implemented wrong.
- **Tag at the point of formatting, not at the point of reading.** The same bucket name arriving from config parsing and from a user call must produce the same tagged output, or a grep for unredacted names finds nothing while the leak is still there.

core/protocol/cmd_get_cluster_config.cxx
thejcfactor: Two review comments settle this from opposite directions, and they agree.

Copilot was right about the contents, and I said in that thread that if the annotation stayed then the category becomes `<md>`: `configuration_json.hxx` parses `name` and `clusterName` out of this blob, so under the strictest-category rule it is metadata, not system data.

Sergey's classification comment settles keep-versus-gate in favour of keeping it:

> Only `<ud>` content is removed at Partial. Tagging `<md>` and `<sd>` costs nothing today and is what makes Full a configuration change rather than an audit of every log statement.

https://github.com/couchbase/couchbase-cxx-client/pull/1094#issuecomment-5500457012

That is the argument against what I proposed here, which was to gate the site behind `dump_configuration` and drop the tag entirely. The measured fact that `<md>` is inert against today's `cblogredaction` is not a reason to leave the tag off, because the tag is exactly what makes the unreleased level a change in the tool rather than a re-audit of every log statement in the SDK.

So: the annotation stays, the site stays ungated, and the category becomes `<md>`. One token at the log site, and the checker's category table moves with it so the record of what was reviewed into which class stays accurate.

core/operations/document_get_all_replicas.hxx
thejcfactor: `<md>` is right. Confirmed against the classification comment rather than left to taste:

https://github.com/couchbase/couchbase-cxx-client/pull/1094#issuecomment-5500457012

A server group names a cluster-level resource, which is the `<md>` branch of the tree ("Does it name a Couchbase resource? bucket, scope, collection, index, design doc, XDCR stream, cluster"). It is not the `<sd>` branch, which is "data about the systems Couchbase talks to over the network: hostnames; IP addresses; ports; DNS names; certificate subjects". A group name is neither a host nor the network path to one; it is a label an operator assigned to a set of nodes, and it is administered alongside bucket and index names.

No change. Resolving.

core/operations/document_query.cxx
thejcfactor: Sergey's classification comment answers this, and the answer is to keep `<ud>`.

https://github.com/couchbase/couchbase-cxx-client/pull/1094#issuecomment-5500457012

A query error payload quotes the statement back, and SQL++ statement text is `<ud>` by name in the table, with the reason spelled out: it leaks field names (`SELECT floor_price FROM stock`). The decision tree's fallback lands in the same place regardless, since a value that came from outside the SDK and fits none of the three classes is tagged as user data.

It also rules out the narrower option I raised here, of tagging only the message-bearing fields and leaving `code` and `retry` bare:

> A value that mixes classes is tagged whole, at the most sensitive class it contains. [...] the spec settles this by redacting the entire parameter, "not just sensitive parameters", because substring-level handling is where the feature gets implemented wrong.

The cost is real, and it is the cost the spec accepts. At the partial level the whole payload hashes to one token, so a redacted log says a query failed and not why. Partial's stated assumption puts that burden on the deployment rather than on the SDK. `client_context_id` stays bare on these lines, so a redacted log still correlates the failure to the request and to a server-side log.

No change on the branch: `<ud>` is what is already there. Resolving.

core/io/mcbp_session.cxx
thejcfactor: The classification comment does not answer the body-hex question, but it does remove one of the three options.

https://github.com/couchbase/couchbase-cxx-client/pull/1094#issuecomment-5500457012

An unrecognised KV frame body came from outside the SDK and fits none of the three classes, and the tree's terminal case for exactly that is `<ud>`. We cannot tag it. The `{:a}` dump renders as a multi-line grid, `cblogredaction` is line-oriented, and a tagged value that spans lines comes out **in clear text** with `WARNING: Unmatched tags detected on line N, potential data leak`. Escaping the newlines keeps the tag intact and flattens the dump into a line of `\n` literals, which is not a hex dump any more.

So "leave it exactly as it is" is now the one option the spec does not support, and what remains are the two behaviour changes already on the table:

- drop the body hex and keep the header hex, since the header is opcodes, status and sizes, all literals from SDK source that need no tag, and `core/io/mcbp_session.cxx:2219` already dumps every inbound byte to `CB_LOG_PROTOCOL` on the same gate; or
- gate the body behind an option, the way `dump_configuration` gates the configuration dump.

Worth restating why this one is not academic: these are `CB_LOG_WARNING`, which is on by default, so unrecognised frame bytes can reach a log nobody opted into. In their defence they fire only on a protocol violation.

Leaving this open. I would rather hear from someone who has used that body dump to diagnose something before proposing we remove it.

thejcfactor: Answered offline: drop the body hex and lean on the protocol logger. Done in `723533db`.

The five warnings now print the frame header only. The header carries magic, datatype, keylen, extlen, vbucket, status and bodylen, which is what identifies the anomaly, and `opaque` is printed as a field, so it anchors you into the protocol log where the inbound socket read already dumps every byte received. So nothing becomes unobtainable: anyone who could have read those body bytes needed protocol logging on in the first place.

The cost is convenience rather than information. That dump holds the raw TCP read rather than the framed body, so you locate the frame inside a stream dump instead of reading it inline, and a frame split across two reads spans two lines.

Resolving.

core/collections_component.cxx
thejcfactor: Taking you up on the offer: fixed in `0cce61ee`, which carries this and the `mcbp_command.hxx` placeholder fix. Both change rendered output, so they are kept apart from the annotation commits.

Both sites now pass `collection_name_` instead of `collection_id_`, tagged `<md>` like the two neighbouring statements that already log the pair.

Resolving.

core/io/mcbp_command.hxx
thejcfactor: Taking you up on the offer: fixed in `0cce61ee`, which carries this and the `collections_component.cxx` fix from the other thread. Both change rendered output, so they are kept apart from the annotation commits.

The format string gains a `key="{}"` slot at the front, where the document id argument already sits, using the same field name the timeout statement further down this file uses for the same value. Nothing else moves, so the remaining three arguments land in the slots their names promise.

This also covers Copilot's `3899374723`, which found the same statement independently and landed on the same remedy.

Resolving.

core/io/mcbp_session.cxx
```diff
@@ -2146,14 +2175,13 @@ class mcbp_session_impl
...
         }
-        CB_LOG_DEBUG(
-          "{} unable to boostrap single node at {}:{}:{} (\"{}:{}\") in time, reconnecting",
-          self->log_prefix_,
-          self->connection_endpoints_.local.port(),
-          self->connection_endpoints_.remote_address,
-          self->connection_endpoints_.remote.port(),
-          self->bootstrap_hostname_,
-          self->bootstrap_port_);
+        CB_LOG_DEBUG("{} unable to boostrap single node at {} (\"{}\") in time, reconnecting",
```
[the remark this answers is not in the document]
thejcfactor: Did not touch this message, but fixed anyway.

core/logger/redaction.hxx
```diff
@@ -0,0 +1,501 @@
...
+ * hands out a log it believes was redacted. Only the "<" of such a sequence is escaped, and only
+ * when it really begins one, so ordinary markup in a logged body stays readable.
+ *
+ * The exclusion markers below escape neither, because the whole point of not_redacted() is to show
+ * the bytes as they are.
+ *
+ * One shape stays untagged for the opposite reason. A hex dump written with the "{:a}" spec
+ * renders as a multi-line grid, so escaping would keep the tag intact but flatten the dump into a
+ * single line of "\n" literals, and a flattened hex dump is not a hex dump. Those sites say so and
+ * use not_redacted(); the checker rejects a tag around one.
```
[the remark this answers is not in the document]
thejcfactor: Updated.

test/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ integration_test(collections)
 integration_test(subdoc)
 integration_test(analytics)
 integration_test(read_replica)
+integration_test(log_redaction)
```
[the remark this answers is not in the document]
thejcfactor: Will be handled in a later pass.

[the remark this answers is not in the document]
avsej: I will migrate the tests to new format later just before removing catch2

avsej (approved):

PR #1095, opened 2026-08-21
CXXCBC-1001: describe the app telemetry metrics for ns_server

avsej:
Motivation
----------

App telemetry metrics reach the ns_server /metrics endpoint carrying
neither # TYPE nor # HELP. RFC 84 defines neither, and ns_server cannot
supply them for names it did not define: cb_stats_info builds its lookup
from the metrics_metadata.json that each server component installs under
etc/couchbase/<component>, and no component owns the sdk_ namespace.

Names, types and bucket bounds are fixed by RFC 84 and identical in
couchbase-jvm-clients, gocbcore and couchbase-net-client, so the
metadata describes the specification rather than this client. This tree
holds an emitter, so the file can be checked against code.

Modifications
-------------

Add etc/metrics_metadata.json covering the eighteen counters and eight
histograms of RFC 84, in the schema every server component uses: type,
help and added on every entry, unit on the histograms, plus labels,
notes and stability.

Keys are metric families, so a histogram appears once rather than under
the _bucket, _sum and _count series it produces, Prometheus declaring
TYPE and HELP once per family. sdk_invalid_metric_total is absent
because ns_server emits that counter and describes it in its own file.
labels carries what the endpoint publishes rather than what the SDK
sends: app_telemetry_scraper keeps le and alt_node and discards agent,
node, node_uuid and bucket.

Results
-------

Nothing in this build reads the file; the component it installs under
and the repository the server build takes it from are ns_server's to
decide, as are stripping the histogram suffixes before the lookup and
loading a second component's file at all. Those are tracked on
MB-68858.

---

Ticket: [CXXCBC-1001](https://jira.issues.couchbase.com/browse/CXXCBC-1001)

[CXXCBC-1001]: https://couchbasecloud.atlassian.net/browse/CXXCBC-1001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/etc/README.md
+++ b/etc/README.md
@@ -1,0 +1,52 @@
+# Metrics metadata
+
+`metrics_metadata.json` describes the application telemetry metrics this client
+reports, in the schema Couchbase Server components use for the same purpose.
+ns_server reads such a file to attach `# TYPE` and `# HELP` to the metrics it
+publishes on `/metrics`; with no file covering the `sdk_` namespace, those
+metrics are published untyped and undocumented.
+
+Metric names, their types and the histogram bucket bounds are fixed by
+[SDK-RFC 84](https://github.com/couchbaselabs/sdk-rfcs/blob/f690b9c46bf562465c10fc5ff5e02d9ec29876ce/rfc/0084-app-service-level-telemetry.md),
+so this file describes the specification rather than this implementation, and is
+the same for every SDK implementing it.
+
+Tracked by [CXXCBC-1001](https://jira.issues.couchbase.com/browse/CXXCBC-1001);
+the server side is [MB-68858](https://jira.issues.couchbase.com/browse/MB-68858).
+
+## Installing
+
+Server components install their file with the `AddMetricsMetadata` macro from
+`tlm`, which validates the JSON, sorts its keys and copies it to
+`$install/etc/couchbase/<component>/metrics_metadata.json`:
+
+    AddMetricsMetadata (JSON etc/metrics_metadata.json COMPONENT sdk)
+
+This client is not part of the server build, so nothing invokes that macro here.
+Which component name the file is installed under, and which repository the
+server build takes it from, are ns_server's decisions and are tracked on
+MB-68858.
+
+## Keeping it in step with the emitter
+
+The key set must equal the metric families `core/app_telemetry_meter.cxx` can
+emit:
+
+    diff <(python3 -c "import json;print('\n'.join(sorted(json.load(open('etc/metrics_metadata.json')))))") \
+         <(grep -oE 'sdk_[a-z_0-9]+' core/app_telemetry_meter.cxx | LC_ALL=C sort -u)
+
+Keys are metric families. A histogram appears once, as
+`sdk_kv_retrieval_duration_milliseconds`, and not under the `_bucket`, `_sum`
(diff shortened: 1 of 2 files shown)


ingenthr:
Why would this be in this repo versus another?

avsej:
> Why would this be in this repo versus another?

We do not have special repo for the metrics, that will keep them up to date. The server might eventually replace libcouchbase with this SDK, and then it will be easier to pull this file.

But if you you know better place, I will be happy to move.

stevewatanabe:
I ran it through the ns_server checker and it found no issues:
> $ python3 validate_stats.py -c localhost:9000 -u Administrator -p asdasd -d /Users/steve.watanabe/Downloads/sergey/etc/metrics_metadata.json

I ask claude:
>  Does the content of metrics_metadata.json follow https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md

which responded (note these are FYI as this review is targeting the HELP/TYPE info and not the stats or stat names):

Mostly yes, with one real deviation baked into the metric names themselves (not something this file can fix, since names come from SDK-RFC 84 and must match core/app_telemetry_meter.cxx verbatim):

- Structure conforms: metric family names are valid snake_case ([a-zA-Z_][a-zA-Z0-9_]*), no reserved leading underscores/colons; type values used (histogram, counter) are both valid OpenMetrics types; unit is correctly given as a suffix of the family name separated by _ (e.g. sdk_analytics_duration_milliseconds ↔ unit milliseconds); help strings are non-empty; le is listed as a bare label name (not an object) for histograms, correctly signaling it's the automatic bucket label rather than a custom one, matching the spec's rule that le must not be a user-supplied label.
- Real deviation — Counter _total suffix: OpenMetrics requires a counter's exposed sample to carry the _total suffix (sdk_r_total MUST-rule). Here only the *_r_total metrics happen to satisfy that; *_r_canceled and *_r_timedout (e.g. sdk_analytics_r_canceled at metrics_metadata.json:17, sdk_analytics_r_timedout at :30) are typed counter but their names don't end in _total, and per README.md the family name is emitted as-is with no suffix transform — so those series would be non-conformant OpenMetrics counters. This is inherited from RFC 84's fixed naming across every SDK, not something introducible here.
- Minor/soft deviation — unit choice: OpenMetrics recommends base units (seconds) over milliseconds; the histograms here use milliseconds (a SHOULD, not a MUST, and again fixed by RFC 84's bucket bounds).

So the JSON's own schema (type/unit/help/labels) is spec-consistent, but several metrics it describes are counters without the OpenMetrics-mandated _total suffix — a pre-existing naming choice from RFC 84 that this metadata file correctly documents rather than causes.

avsej:
ns_server got change on gerrit with this JSON meta file: https://review.couchbase.org/c/ns_server/+/252258

stevewatanabe:
> ns_server got change on gerrit with this JSON meta file: https://review.couchbase.org/c/ns_server/+/252258

This is to facilitate development of the feature. The mechanics of how SDK's metrics_metadata.json file will end up in the right place in a Couchbase Server release are to be determined.

stevewatanabe (approved):
Looks fine from the ns_server pov.

DemetrisChr (approved):

PR #1097, opened 2026-08-24
CXXCBC-1003: catch the framework's own exceptions first

avsej:
Fixes [CXXCBC-1003](https://couchbasecloud.atlassian.net/browse/CXXCBC-1003).

`assert_throws` matched the requested type ahead of `test_skip_exception` and `test_assertion_failure`, both of which derive from `std::exception`. Handlers are tried in order, so naming any base of them as the expected type left those two unreachable — and the comments on them stated a guarantee the ordering did not deliver.

It manifests two ways. `cmake/CompilerWarnings.cmake` defaults `WARNINGS_AS_ERRORS` to `TRUE`, so in the normal build and in CI it is the first:

- **An ordinary assertion does not compile.** Instantiating `assert_throws<std::exception>` gives `error: exception of type 'couchbase::test::test_skip_exception' will be caught by earlier handler [-Werror=exceptions]`. Asking whether a callable throws *some* `std::exception` cannot be written.
- **Where those warnings are not fatal, the case reports passed.** Built with `-DWARNINGS_AS_ERRORS=OFF` the same code compiles with four warnings, and a `skip()` raised inside the callable is claimed by the expected-exception handler: the runner counts `skipped` 0 and the case passes. A nested assertion failure goes the same way, discarding the location and message it carried.

The control-flow handlers now precede `Exc`. Neither type is assertable through `assert_throws` as a consequence — a case that wants to observe a skip drives `run()` and asserts on the `run_result`, as the self-tests do.

No current caller is affected: all five call sites request `std::invalid_argument`. The function is unchanged since it was promoted out of the CNG tree, so this is not a regression in any open pull request; it was found while reviewing #1073 and is out of scope there.

## Verification

Self-test cases now cover `assert_throws`, which nothing exercised before.

Two are the regression guards, and they bite at both levels. Reinstating the old ordering makes `test/framework/selftest.cxx` fail to build under the default warning settings; with warnings demoted it builds and the two cases fail on the runner's counts:

```
a_skip_inside_assert_throws_skips_its_case              FAILED: the skip reached the runner (actual: 0, expected: 1)
a_failed_assertion_inside_assert_throws_fails_its_case  FAILED: the nested failure reached the runner (actual: 0, expected: 1)
```

The other three pin what the reordering must not have cost: a base of the thrown type is still accepted, an unrelated type is still reported as the wrong type rather than as nothing thrown, and a callable that throws nothing stays distinct from both.

The self-test is green with the fix under `WARNINGS_AS_ERRORS=ON`, which is the configuration the defect made unbuildable. `test_unit_config_profiles` is the only `assert_throws` caller outside the framework, and it passes unchanged.

[CXXCBC-1003]: https://couchbasecloud.atlassian.net/browse/CXXCBC-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/framework/selftest.cxx
+++ b/test/framework/selftest.cxx
@@ -31,6 +31,7 @@
 #include <stdexcept>
 #include <string>
 #include <thread>
+#include <utility>
 
 namespace couchbase::test
 {
@@ -58,6 +59,24 @@ body_sleep()
   std::this_thread::sleep_for(std::chrono::milliseconds{ 200 });
 }
 
+// assert_throws asked for a base of the framework's own control-flow types. Both bodies exist to be
+// run through the runner rather than called directly: what is under test is the outcome the runner
+// reports, and a case body cannot observe its own verdict.
+void
+body_skip_inside_assert_throws()
+{
+  assert_throws<std::exception>([]() {
+    skip("intentional skip from inside a callable");
+  });
+}
+void
+body_failed_assertion_inside_assert_throws()
+{
+  assert_throws<std::exception>([]() {
+    assert_true(false, "intentional failure from inside a callable");
+  });
+}
+
 // A sink for the runner's progress output so the self-test stays quiet.
 auto
 run_quiet(const test_suite& suite, const std::set<std::string>& filter, bool real_cluster)
@@ -65,6 +84,21 @@ run_quiet(const test_suite& suite, const std::set<std::string>& filter, bool rea
 {
   std::ostringstream sink;
   return run(suite, filter, real_cluster, sink);
+}
+
+// The message an assertion produced, or empty if it did not fail. assert_throws cannot be used to
+// observe a test_assertion_failure: naming that type as the expected one leaves its own handler
+// unreachable, so the framework's own tests catch it directly.
+template<typename Fn>
+auto
+message_of(Fn&& fn) -> std::string
+{
+  try {
+    std::forward<Fn>(fn)();
+  } catch (const test_assertion_failure& e) {
+    return e.what();
+  }
+  return {};
 }
 
 // ── the self-test cases ──────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


avsej:
> The phrase "the case never ran" is inaccurate: `body_skip_inside_assert_throws` does execute, but the old handler swallows its `skip()` before `run_case` can classify it. Please describe this as the runner missing the skip and counting the executed case as passed, so the comment does not misstate the failure mode.

Correct, and fixed. The body runs to completion: the lambda throws, `catch (const Exc&)` claims it, `assert_throws` returns normally, the body returns normally, and the runner sees no exception at all. Nothing is skipped and nothing is unexecuted — what is lost is only the skip.

The comment now reads:

```
  // std::exception is a base of test_skip_exception, so an expected-exception handler matched ahead
  // of the control-flow ones claims the skip. The body still runs to completion; what is lost is
  // the skip itself -- assert_throws returns normally, nothing propagates, and the runner counts
  // the case as passed. A case that declared it does not apply then reads as one that verified
  // something.
```

Two further places stated the same thing wrongly and are corrected with it. The commit message said "A case that never ran and reports passed reads as coverage"; it now describes the body running on and the runner never seeing the skip. The comment on `assert_throws` itself called the outcome "a skipped case reported as a passing one", which has the same defect — the case was never skipped — and now says the case is counted as having verified something it had just declared it could not.

The ticket description and the pull request body were checked for the same wording and did not carry it.

Verification unchanged otherwise: `framework_selftest` is 22 cases at this commit and 26 at #1098, and the negative control still reports 4/4 mutations killed.

thejcfactor (approved):

PR #1131, opened 2026-09-17
CBD-6883: Correct BoringSSL version

udkyo:
The version listed in this repo does not match the SHA we're using, this is affecting Black Duck scans for products which depend on this repo:

```
% git describe --tags 2ff4b968a7e0cfee66d9f151cb95635b43dc1d5b
fips-20210429-4847-g2ff4b968a
```

(BlackDuck shortens this to `fips-20210429`)


yash-dodderi7 (approved):

avsej:
I would rather not change it here just because BlackDuck does not have that commit marked as version.

@udkyo is there any specific reason to change BoringSSL version?

I would rather merge this change instead https://github.com/couchbase/couchbase-cxx-client/pull/1122, BoringSSL does not use versioned releases.

udkyo:
I'm happy to go with AWS-LC, sure. And concede that I was 100% wrong about that tag!

